PR #72, opened 2015-02-02
Add hyphen to "WCAG-conforming".

lkmorlan:


lkmorlan:
I don't think the test failure has anything to do with my patch.

Issue #131, opened 2016-01-08
C7 - change needed in CSS code for example


awkawk:
C7: Using CSS to hide a portion of the link text contains the following CSS to visually hide an element:

a span {
    height: 1px;
    width: 1px;
    position: absolute;
    overflow: hidden;
    top: -10px;
}

Is there any reason to include the negative top value? If top, right, bottom, and left are not set, the element will stay in the rendered place. This would prevent any mess-ups with—for example—focus styles. A way around this would be to set position to relative on the parent element, but that means extra code that isn’t strictly necessary if top wasn’t set.

kwahlbin:
This is the CSS that we recommend that has worked across different platforms, browsers and AT

.visually-hidden
{
  position: absolute;
  clip: rect(1px 1px 1px 1px);  /* for Internet Explorer */
  clip: rect(1px, 1px, 1px, 1px);
  padding: 0;
  border: 0;
  height: 1px;
  width: 1px;
  overflow: hidden;
}

MichielBijl:
This opening post was [originally posted on the WAI IG mailing list](http://lists.w3.org/Archives/Public/w3c-wai-ig/2016JanMar/0048.html).

@kwahlbin that looks good to me. For which version(s) of IE is the space separated rect value?

awkawk:
I understand that the IE fix is for IE 6/7.  I think we can let it go if that is correct.

awkawk:
Can we make this change? If not, let's close this issue.

alastc:
I think it is worth an update, I did a PR in #630 

This [post from TPG](https://developer.paciellogroup.com/blog/2012/05/html5-accessibility-chops-hidden-and-aria-hidden/) was useful for seeing the test results of various methods (updated in 2016).

The question that struck me was: Is this still recommended, or would we point to a variation of [ARIA1](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA1)?

patrickhlauke:
> The question that struck me was: Is this still recommended, or would we point to a variation of [ARIA1](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA1)?

depends which browser/AT combos you're aiming to support. more specifically, if you're wanting to support (generally old) AT that doesn't have full ARIA support. I seem to remember (but shamefully, haven't got the ability to test just now) that even fairly recent screen magnifiers with screen reading functionality still struggle with some basic ARIA stuff...

alastc:
[...]

PR #132, opened 2016-01-08
Updates in response to Issue #131

awkawk:


wcag20/sources/techniques/css/C7.xml
```diff
@@ -30,8 +30,9 @@ a span {
 height: 1px;
 width: 1px;
 position: absolute;
+clip: rect(0 0 0 0);
+margin: -1px;
```
MichielBijl: What is the purpose of the negative margin?

mraccess77:
Shouldn't we also add an overflow:hidden; ?  also perhaps an !important after the position absolute?

MichielBijl:
My original proposal included overflow hidden. Important should be avoided at all cost, what if you want to enable the text/element on smaller/larger screens; don't want to litter your css with a bunch of important's. 

—Michiel

wcag20/sources/techniques/css/C7.xml
awkawk: This thread discusses it: https://github.com/h5bp/html5-boilerplate/issues/194  (in short, to avoid an errant single pixel of an item with a background color)

awkawk:
Jon, overflow:hidden was in there before and wasn't removed.

wcag20/sources/techniques/css/C7.xml
MichielBijl: That can be solved with background: none|transparent; right?

—Michiel

awkawk: I suppose so, I haven't seen that done or tested. One line of CSS either way?

PR #160, opened 2016-02-09
Issue159

jnurthen:
Removed the link to the ARIA Primer,
Switched the first example to reference the ARIA 1.1 APG and switched the 2nd example to also reference the ARIA 1.1 APG which has replaced the ARIA Primer.


jnurthen:
Need to make a few more changes

awkawk:
Changes are all in.

PR #175, opened 2016-04-07
Removed references to 3.2.1 from F52.

alastc:
Small changes to F52 to remove references to the AA level 3.2.1 SC.
Also, the reference to XHTML seems rather out-dated, is that ok to remove?


jnurthen:
@alastc I don't think you should remove the example. I would have simply changed the word XHTML to JavaScript - as this is what the code looks like to me. Indeed I believe both examples are labelled incorrectly as HTML and XHTML when they should both be labelled as JavaScript. One simply shows using the old fashioned onload attribute whereas the 2nd shows using event listeners.

alastc:
Are two examples needed? Perhaps it would be better to simply use the 'XHTML' one (labelled as JavaScript). The key feature is using the window load trigger, both do that and newer sounds better...

jnurthen:
I agree with this thought. 2 examples complicates things.

awkawk:
@alastc - this was approved with the assumption that you could remove the third file from the pull request that is dealing with a separate issue (about pt/px).  Can you take that file out of the pull?

awkawk:
The group discussed on June 7 and was not able to reach consensus.

alastc:
I take that means it is punted for now, is there anything I need to do to clean up, e.g. close or withdraw the request? (Not sure what the optiosn are, or what is appropriate.)

awkawk:
Closing without consensus

PR #177, opened 2016-04-15
Changes to Understanding 1.4.3

awkawk:
Reorganized a few paragraphs for clarity and added a sentence to
indicate the intent of 1.4.3 as discussed in Issue 157


awkawk:
The changes I made are to provide an indication of what was agreed on in Issue 157.  In addition, as I looked for the best place for the text I felt that the content was hard to read because there were paragraphs that seemed to be in the wrong sections.  Nothing major, and the paragraphs were moved intact.

Issue #121, opened 2015-10-06
failure for not identify required fields and input format before submit


goetsu:
I think we need a failure to cover the case where no information is given to identify required fields and input format before submit because for me that a clear violation of SC 3.3.2

goetsu:
example : 

## Description

This failure ensures that people with visual or cognitive disabilities will recognize required fields and understand what information need to  be provided before to submit the form. Required are frequently formatted in fixed, distinctive ways, and authors may feel that just providing visual formatting of the fields will be sufficient to identify them. However, this information must be be available programatically and visually before submitting the form.

## Examples

### Failure Example 1:
a form with visual information only
```html
<p>Field with * are mandatory</p>
<p>
<label for="name">Name</label>
<input type="text" id="name">*
</p>
```
### Failure Example 2 :
a form with programatically associated information only
```html
<p>
<label for="name">Name</label>
<input type="text" required id="name">
</p>
```
### Failure Example 3 : 
a form with name field required but without any information associated
```html

<p>
<label for="name">Name</label>
<input type="text" id="name">
</p>
```

## Tests 

### Procedure

- For each fields in the web page that represent a required field, check that this information is available in the accessible name programatically associated with the fields before submission of the form
- For earch field in the web page that represent a required field, check that this information is available with a visible text or cue positioned near the fields.

### Expected Results

If check #1 or check #2 are false, then this failure condition applies and the content fails this success criterion.

jnurthen:
Why is example #2 a failure?

awkawk:
I'd say that example 1 fails 1.3.1 but not 3.3.2, example 2 fails 3.3.2 but not 1.3.1, and example 3 fails both.

jnurthen:
Why does #2 fail? The default rendering of required on many browsers includes an indication that the field is required. In these cases it is fine for 3.3.2 IMO

awkawk:
Ah, right.  Assuming all of the necessary accessibility support then #2 would be fine for 3.3.2.

goetsu:
Ok then change replace required with aria-required in example 2 if you prefer

goetsu:
Or a class="required" and a JavaScript associated with that classname to make the field required.

My point with that failure proposal is to make more clear that not mentioning required status of fields in an accessible manner (visually and programmatically) PRIOR to submit is a failure.

jnurthen:
I'm fine with either of those so long as it is made clear in the failure example that there is not CSS indicating that the field is required based on that attribute.

[...]

PR #180, opened 2016-04-20
F94.xml

marcjohlic:
Adding a new Failure technique - F94 - per Issue #121.  
https://github.com/w3c/wcag/issues/121

### DIFF
--- a/wcag20/sources/techniques/failures/F94.xml
+++ b/wcag20/sources/techniques/failures/F94.xml
@@ -1,0 +1,100 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE technique
+  SYSTEM "../../xmlspec.dtd">
+<technique id="F94">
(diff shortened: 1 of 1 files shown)


jnurthen:
I disagree with this failure. There are times when fields are not known to be required until the data is submitted to the server. Any failure technique must allow for this case.

I agree that if the required indicator is available visually it must also be available programatically in some way (aria-required, required or as part of the label). However, this technique only seems to allow label which is not correct.

goetsu:
@jnurthen a note to indicate that case and an exception in the test procedure could be good for you ?
it's not my intent to allow only label as the correct solution maybe we can add correct example one using label and one using aria-describedby for example (any others ?). For sure title would be conforming but I don't think it's a good example to give

patrickhlauke:
as noted in #121 - at least for the case of using `aria-required`, or indeed just `required` - there's quite a bit of overlap with 4.1.2

awkawk:
Closing without consensus.

Issue #181, opened 2016-04-20
"Large text" - using px rather than pt as unit


patrickhlauke:
Something that's been bugging me for a while...

For on-screen/web content, `px` would be a far more natural unit of measure than `pt`. 

`pt` is (generally anyway) derived simply from the reference pixel (where then 1pt ~= 1.333px - see https://drafts.csswg.org/css-values/#absolute-lengths).

having "large text" (https://www.w3.org/TR/WCAG20/#larger-scaledef) defined in `pt` adds an extra layer of confusion for authors who would normally not actually use this unit of measurement in their CSS/HTML, and then may get very confused about whether or not this refers to actual physical `pt` as measured on the screen (which is something authors have no control over, as it will vary depending on computer/mobile phone/tablet, the OS, the browser, user settings, etc), and whether or not they need to take into account potential variances in dpi (again, not something they have any control over directly).

In short, suggest - in an update/errata to WCAG 2.0 - adding a note about the fact that this refers to CSS `pt` which are anchored on the reference `px` (which then further makes the assumption that the OS/browser/user has a reasonably close mapping of 1px to the idealised reference pixel as per the spec), or simply...using `px` instead (replacing `18pt` with `24px`, `14pt` with `18.5px`, etc)

At a stretch, at least complementing `18pt` with the equivalent `24px` - (e.g. "18pt (24px) ...")

/cc @yatil

WayneEDick:
The point is a physical measure. There is some disagreement on how much of an inch is occupied by a point. I use the Adobe definition that is 1/72 inch. The key is it is not relative. It is a size in inches.

That is the real problem with the pixel. Pixels cannot be used to measure font-size from the point of view of accessibility because the same measure 24px can be from 16 point to 22 point depending on the dpi. 16 point really isn't large print. 

The user does not care how many pixels it takes to make their critical print size, they just want that size because it facilitates reading. For example, to become readable at a 3:1 ratio rather than 4.5:1 requires a physical size in points, inches or centimeters.

I do agree that the point may be a bad measure because its definition is not set. The inch or centimeter would be better.

patrickhlauke:
> The point is a physical measure. There is some disagreement on how much of an inch is occupied by a point. I use the Adobe definition that is 1/72 inch. The key is it is not relative. It is a size in inches.

sorry wayne, but that's not (currently, and for the past few years) correct when talking about web content. from the spec

> For a CSS device, these dimensions are anchored either
>
[...]

PR #184, opened 2016-04-27
Issue181 - proposal for changed/expanded notes in "large scale (text)" definition

patrickhlauke:
Based on discussions on GitHub https://github.com/w3c/wcag/issues/181 and on mailing list https://lists.w3.org/Archives/Public/w3c-wai-gl/2016AprJun/0214.html

As the definition itself of "Large scale (text)" https://www.w3.org/TR/WCAG20/#larger-scaledef is normative, this proposed change leaves the "14pt bold / 18pt" text itself intact, and instead reorganises, changes and expands the following non-normative/informative notes to actually qualify what "points" mean in a far more precise way.

Note that even while trying to untangle the current notes, it was still not completely clear to me whether WCAG 2.0 actually means physical points as measured with a ruler on the screen, or not. As amply discussed already, if this IS the intention of the definition, than it would invalidate the whole premise of the normative definition, since there is NO WAY for an author to ensure any particular physical rendering size. It is all dependent on user agent and operating system defaults / settings, and to be of any use for actual authors it MUST be at least anchored on the assumption that the UA's base font size is appropriate, with "large scale" being relative to that. These rewritten notes try (as much as was possible without touching the normative part of the glossary definition) to convey this.


patrickhlauke:
I left the various steps that led from the current notes to the final rewrite here intact as separate commits, leaving the possibility for people to discuss individual changes rather than having to necessarily tackle the entire rewrite when commenting.

patrickhlauke:
as @goetsu rightly points out https://twitter.com/goetsu/status/725308370891247616, even with https://github.com/w3c/wcag/pull/184/commits/67c9230cc3ace1320afa9d0a82b79671a62dd94b there is still an element of variability here for auditors/authors: if they define font sizes with absolute units (be it `pt` or `px` or whatever) that won't guarantee that text is "large" on devices/UAs that do have a base font size that differs from the general `16px`. the only true way for an author to guarantee that their large sizes are actually rendered large relative to all users' UA base font size is to use relative sizing for fonts. this is not something that a note or glossary definition can mandate, and would likely be a new SC from the Low Vision TF (in short: mandating the use of relative, rather than absolute, font sizing generally). Once that is in place, the glossary definition would ideally be changed substantively to just mention proportions (e.g. that large text means 150% of the base font size, and as per the new SC this measure needs to be defined in relative terms to the base font size, without assumption of 16px).

In short, even this rewording still presents an element of uncertainty, but it does codify (in my mind anyway) current practice of how most (all?) authors and auditors interpret the "points" mentioned here (basically, take the points value, multiply by 1.333, get the pixel size, and use that as your guide). It has less uncertainty than the current "what is actually meant by 'point'?"

patrickhlauke:
And in case the remaining level of uncertainty above leads to "so why not define 'points' as meaning the actual physical points, wouldn't that remove the uncertainty altogether" - in theory yes, but as demonstrated, in practice authors have no way of actually guaranteeing/meeting any specific physical size rendering, so while mandating a physical point size elegantly circumvents having to qualify the sizing in relation to all these uncertainties, it simply sweeps these uncertainties under the carpet for authors to try and resolve (which, with current technology, they simply can't). So these notes at least put these problems of variability and dependency out in the open.

patrickhlauke:
With the latest commit (https://github.com/w3c/wcag/pull/184/commits/c404ae2835cb4fe5f0721c539b93f50ad4c608cc) I believe that even the uncertainty around UAs with non-16px default font size has been addressed now (if we consider, again, that the definition here is to do with visual size, rather than establishing visual hierarchies or readability)

patrickhlauke:
And to clarify where the remaining bit of uncertainty now lies: if as an author I try to do the right thing and only size my fonts using relative sizing (`em`, `%`, `rem`, etc), then I as author can't predict at what `px` size these will be presented on UAs that deviate from the `16px` default font size value; ditto for an auditor, who can only make a judgement call on whether something is large scale or not under the assumption of a fairly common `16px` base font size when fonts are defined using relative units.

To be clear: this uncertainty is already present in the current unaltered WCAG 2 wording, just that it's not actually tackled/addressed/spelled out.

patrickhlauke:
The last commit may be controversial https://github.com/w3c/wcag/pull/184/commits/975f098ace40c9f30c2ccc7907bbed1ec310f742 but pragmatically this removes any remaining uncertainty from an auditor/author point of view by explicitly stating that if only relative font sizes are being used, a base font size of `16px` has to be taken as the base from which to calculate whether something is `large scale`.

This is pragmatic, and most likely what authors actually do today. Indeed, most authors actually assume `16px` IS a standard default font size, and need to be told that this is NOT the case https://nicolas-hoizey.com/2016/03/people-don-t-change-the-default-16px-font-size-in-their-browser.html

And once again, this uncertainty/problem is already in the current WCAG 2 definition, just silently swept under the carpet.

patrickhlauke:
A certainly very valid comment sent to me separately by @alastc was that the notes are now very verbose, wondering if a large chunk of them should be moved to the "Understanding..." document rather than being kept here.

While I agree that the notes are verbose, I think there's a danger in sweeping the actual important explanation about "points" (which unfortunately does involve all the theory about CSS reference pixel, viewports, base font size in UAs, etc) into a separate document. This would require an author who wants to actually find out what the definition means to first click through from an SC (currently only the color contrast one references this, to my knowledge) to the glossary definition, and THEN separately checking the "Understanding..." document for the SC. Also, this wouldn't scale as soon as there are two or more SCs referencing "large scale (text)", unless we're happy to have repetition in the separate "Understanding..." sections.

alastc:
I just wonder if that explanation is needed: most people didn't realise there was a difference, so stating this solves the problem:

>  The ratio between sizes given in "points" and CSS pixels is approximately 1pt = 1.333px.

If they did know the difference, they don't need the explanation.

The other key bit needed is regarding the relative sizing, which can be accomplished with:

> In most user agents, the default text size is 16px (approximately 12pt). As such, 14pt and 18pt are roughly equivalent to 18.5px and 24px. If using relative units that equates to 1.2em and 1.5em / 120% and 150% of the default text size (assuming that the body text size has not been changed from the user agent's default).
> 
> Although some user agents may have a default text size that differs from the common 16px/12pt, this definition will assume a default text size of 16px when calculating the size of larger text.

Authors know what to do, testers know how to test... job done?

patrickhlauke:
> If they did know the difference, they don't need the explanation.

But what if they don't know the difference, but assumed (wrongly) that "points" means the physical points that they may know from print? And then think this is to do with measuring with a real-world ruler on a screen? (as evidenced in some of the discussions on list)

The proposed shortening of the whole relative sizing bit is seductively terse, though. Maybe I'm just too paranoid about removing the context/background for some of the decisions to include all this, to ensure no future ambiguity/questioning.

Anyway, open to tweak some more...let's see what others think :+1:

alastc:
I think it might be the working group that needs the explanation most - in order to understand we don't need much explanation??

Anyway, I put the cut down version here for comparison: 
https://github.com/alastc/wcag/commit/80ec984e3948f382de2d0bef69cfcdb57a7edf52

I don't know about you, but I could do with some fresh eyes on it from others...

patrickhlauke:
So what do i do now with all this? Just close it? really wish the WG had looked at / commented on this concrete PR, rather than the more general issue...

(and yes I still believe the better option would be to clarify the glossary definition of "large scale (text)" right here, rather than adding an extra layer of abstraction by then clarifying it further in the understanding docs, but ho hum)

joshueoconnor:
@patrickhlauke If you feel this is a separate issue, as such, and would like the group to look at it - thats totally fine. If not, it can be labelled as 'Deferred/WCAG.next'. Let me know what you would like to do.

patrickhlauke:
@joshueoconnor i feel that some of the individual changes as part of this large PR could be cherry-picked as errata/additions that do not change the overall meaning. namely:

- Clarify the "content is delivered" to include UA defaults https://github.com/w3c/wcag/pull/184/commits/fa7920204f69de5243c3592dfb723aee4fe1a07f
- Remove incorrect sentence https://github.com/w3c/wcag/pull/184/commits/c34de7dd8bd6fb81d3a0d62eeff73a919fe27dbe
- Expand the note about low vision users (at least in part) https://github.com/w3c/wcag/pull/184/commits/cbb6092332d25778d4e12c35e8b61eaba548ced2
- Add some actionable advice for authors https://github.com/w3c/wcag/pull/184/commits/c15ab1771b8bc61e5cef5bfea1e1e26ad6cf43bd

If, however, these are still felt to be off-limits for change, then...*shrug*..."Deferred/WCAG.next" it. Does the issue then get closed, or does it stay open?

joshueoconnor:
Thanks for those @patrickhlauke. The issue will stay open with the appropriate label. The commits can be surveyed by the group.

awkawk:
Actually, according to the process we defined, the issue gets closed and labeled with WCAG.next.  Closed or open, as long as it has the WCAG.next label then we will review it in due course, but we will close it to keep the issue list manageable.

joshueoconnor:
Apoplogies @patrickhlauke. This is going to be closed and deferred till WCAG.next.

PR #188, opened 2016-05-24
Adding px note to two understanding docs.

alastc:
Added a note about PT to PX conversion in understanding docs for 1.4.6 and 1.4.3. Also adjusted note on image of text to simplify.


alastc:
I'm guessing the bibref doesn't work? Not sure how it should work though.

patrickhlauke:
wonder if the note about CSS pt should be more explicit/walk the reader through how the value is reached (basically, this https://github.com/w3c/wcag/pull/184/commits/738095ca633e610f4424cf36a33aaea9a5677aee).

and no the bibref won't work unless you add it as well (see part of https://github.com/w3c/wcag/pull/184/commits/291f68c76771ce3de3198c759c453383360402b7)

awkawk:
@alastc - just use a loc element with an href to the section that you want to link to.

awkawk:
Approved by group. https://lists.w3.org/Archives/Public/w3c-wai-gl/2016AprJun/0670.html

Issue #191, opened 2016-05-31
F78 is intended for keyboard - should be in the test procedure


jnurthen:
The description of F78 States
"This describes a failure condition that occurs when the user agent's default visual indication of keyboard focus is turned off or rendered non-visible by other styling on the page without providing an author-supplied visual focus indicator."

The test procedure states.
### Procedure

1. Set the focus to all focusable elements on a page.
2. Check that the focus indicator is visible.

### Expected Results

#2 is true.

I propose changing to

### Procedure

1. Set the focus to all focusable elements on a page using the keyboard.
2. Check that the focus indicator is visible.

### Expected Results

#2 is true.

awkawk:
Changes submitted in Pull Request #192 were accepted, so closing.

PR #192, opened 2016-05-31
Change for Issue 191

jnurthen:
#191

### DIFF
--- a/wcag20/sources/techniques/failures/F78.xml
+++ b/wcag20/sources/techniques/failures/F78.xml
@@ -40,7 +40,7 @@
       <procedure>
          <olist>
             <item>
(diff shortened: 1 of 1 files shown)


joshueoconnor:
This is a change to  F78 is intended for keyboard - should be in the test procedure #191 - adding the text '[...]using the keyboard'.

@jnurthen Please provide a short description (as suggested here) to future pull requests.

awkawk:
F78 is a failure on 2.4.7 ("Any keyboard operable user interface has a mode of operation where the keyboard focus indicator is visible.") which is very much directed at keyboard access.  This clarification is regarded as editorial and we will accept this change as a result.

PR #196, opened 2016-06-28
Removing UA issues from H44 - Issue 190

alastc:
Removed the UA issues from both H44 and the UA notes. Also updated the description to include the implicit method.
Someone please check that this works, the UA-notes were not obvious in how they should be updated!


patrickhlauke:
trawling some of the really old PRs ... is this still needed/relevant or can it be closed?

fstrr:
@alastc 

1. the "explicitly" typo has been fixed in the [2.1 document](https://www.w3.org/WAI/WCAG21/Techniques/html/H44).
2. I can't find a current version of the User Agents Support Notes (just this [old page](https://www.w3.org/WAI/WCAG20/Techniques/ua-notes/html)), so I'm assuming that, after almost seven years, that doesn't need updating now.

Can this be closed?

alastc:
No UA notes anymore...

PR #214, opened 2016-07-30
H36 Add link to spec

momdo:


awkawk:
Will add link to HTML 5.1 when it reaches Recommendation.

momdo:
Yes, HTML 5.1 is currently Candidate Recommendation, but https://w3.org/TR/html has already pointed HTML 5.1 not HTML 5.0.

awkawk:
That is interesting and is probably a W3C process issue.  To keep in line with the W3C process we will link to it when HTML 5.1 reaches REC.

PR #217, opened 2016-08-05
Updated in response to Issue 216

awkawk:


awkawk:
Approved https://lists.w3.org/Archives/Public/w3c-wai-gl/2016JulSep/0580.html

PR #219, opened 2016-08-05
Removing H46

awkawk:
Noembed is not allowed in HTML5, and was not part of HTML.  If we are looking to use techniques to advocate for good practices, this probably isn't it.


awkawk:
Closing pull without merging due to WG agreement.

Issue #228, opened 2016-08-16
Add some actionable advice to "Large scale (text) > Note 1" in the glossary


patrickhlauke:
In https://www.w3.org/TR/WCAG20/#larger-scaledef Note 1 currently reads

> Note 1: Fonts with extraordinarily thin strokes or unusual features and characteristics that reduce the familiarity of their letter forms are harder to read, especially at lower contrast levels.

This does not provide any actual advice to authors (or auditors). I'd recommend a (slightly expanded) cherry-pick of https://github.com/patrickhlauke/wcag/commit/c15ab1771b8bc61e5cef5bfea1e1e26ad6cf43bd (which does not substantially change the meaning of the definition, nor the note, but closes the loop in terms of why it's actually mentioned)

joshueoconnor:
Thanks Patrick. That's one for the LVTF to review. @lauracarlson could you make sure this gets on the LVTF radar?

lauracarlson:
@joshueoconnor  will do.

joshueoconnor:
Thanks! @lauracarlson

mraccess77:
My initial thoughts: We could start by adding "advisory" and then go on to explain that just because large scale text is used doesn't mean the text will be readable or have the needed contrast for users with low vision etc.  That is one thing that Gregg V. has brought up before -- just because we allow large text to have lesser contrast doesn't mean the text will be readable by people with low vision.    Also the larger the text does not mean less contrast is always acceptable.  You could have text a foot high and if it has low contrast it may be unperceivable by some people with low vision.

alastc:
I've created a PR in this repo, #774.

It seems like a small update, but reasoanble advice for people.

Anything further would overlap with #665

alastc:
From the group's discussion: We won't accept the pull request, but we'll make a similar change to the understanding document.

PR #229, opened 2016-08-16
Add actionable advice to "Large scale (text) > Note 1" in the glossary

patrickhlauke:
Closes https://github.com/w3c/wcag/issues/228


joshueoconnor:
This looks editorial (and makes sense to me) but is it?

awkawk:
This should be discussed for 2.1.  There's no point in making this change in the 2.0 document.  As to whether it is editorial, I don't think that adding actionable advice can be editorial.  We will need to route this for group review, although it may be easy.

patrickhlauke:
Yes, this was intended for 2.1 (I've given up on trying to make any changes to 2.0 ;) ). And agree it's not editorial since it gives actual advice/suggestion

awkawk:
closing with the WCAG 2.1 label on it for future reference.

PR #254, opened 2016-11-12
Change for Issue 253

### LINKED ISSUE #253: Generate valid HTML5 with XSLT
According to <a href="https://www.w3.org/pubrules/doc/rules/?profile=WG-NOTE#valideHTML">Pubrules</a>, we need to generate valid HTML5.

Thus, at least, we should modify following points:
- Omitting XML Declaration
- Changing DOCTYPE
- Using `<abbr>` instead of `<acronym>` (especially in the copyright section)
- Fixing `<script>` tag (about fixup.js)

momdo:
Fix #253


patrickhlauke:
trawling some of the really old PRs ... is this still needed/relevant or can it be closed /cc @michael-n-cooper

fstrr:
This seems to relate to this [Quick Reference page](https://www.w3.org/WAI/WCAG20/quickref/20070521/Overview.php) which was a draft in 2007. I'm going to close this soon, but do wonder if the related content should also be deleted.

Issue #236, opened 2016-09-17
Luminance / Luma confusion


xi:
Some time ago filed a [bug report with Less](https://github.com/less/less.js/issues/2829) because I thought they may have confused the terms "luminance" and "luma". They use luma to calculate the color contrast although the [spec](https://www.w3.org/TR/WCAG20/#contrast-ratiodef) clearly states that you have to use "relative luminance".

We came to the conclusion that Less is actually right and the spec may be wrong. From an [article by Charles Poynton](http://www.poynton.com/PDFs/YUV_and_luminance_harmful.pdf) I understand that the term "luminance" has been used ambiguously in the past and the term "luma" has been specifically invented to allow a clear distinction. The wrong wording in the spec may therefore be a relic of that ambiguous past.

So am I correct in thinking that the spec actually refers to luma? If that is the case, how can it be fixed? IMHO there should at least be a note about the luminance/luma ambiguity in the ["Understanding WCAG 2.0" document](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html).

_I have already brought this up on the mailinglist some time ago, without much success_

moekraft:
hi

moekraft:
Hi Xi, I am researching. Thanks, Moe

moekraft:
Hi Xi, 

I have read through your materials and I do believe the WCAG definition of relative luminance needs clarification. The formula used in Notes 1 is correct, however, the conversion of the gamma-corrected R'G'B' value I believe needs some clarification. 

If you read through the formula it's proposing to covert the non-linear RGB values to linear RGB values, e.g. R'sRGB = 12.92 x RsRGB (gamma-corrected) and the formula is reversing that RsRGB = R'sRB/12.92 (linear). The condition checked R,G,B <=.03928 is coming straight from the sRGB spec on which this is based although this seems backwards to me since we are trying to solve for R,G,B. 

I disagree that the spec is trying to define luma while it is converting luma (R'G'B') to RGB by removing the gamma correctness. 

I will work on a proposed update to the definition to clarify. But need to pull in some experts to review. 

Thanks,
Moe

xi:
@moekraft Thanks for working on this! What is the current status? (ie can we close this issue?)

As far as I understand #256, you concluded that the wording in the spec is right and the wording in LESS is wrong. Could you please verify that and maybe summarize your findings so that I can convince the LESS authors to change their wording?

moekraft:
[...]

PR #255, opened 2016-11-14
Working branch for q1 2017

moekraft:
These updates are in response to Issue #236 which highlights the confusion over the definition of relative luminance. The value of relative luminance is computed from the non-linear sRGB values by reversing gamma-correction applied. This update adds a statement to the definition clarifying this process. Additionally, when representing non-linear (gamma-corrected) sRGB values, the prime symbol must be used. Note: the prime symbol is represented in HTML as &prime; or &#8242;. Collaborating with color expert, Lars Borg, Adobe, to confirm this clarification through Andrew Kirkpatrick's request.


moekraft:
Trying to generate a pull request for issue #236 only. Removing pull request from here. Will work from separate fork since unable to create a branch.

PR #256, opened 2016-11-14
Proposed updates for Issue #236 clarifying definition of relative luminance.

moekraft:
Proposed updates to the definition of relative luminance to clarify how
relative luminance is calculated from non-linear (gamma-corrected) RGB.
Additionally, anytime non-linear (gamma-corrected) RGB is referenced the
prime symbol must be used.

These updates are in response to issue #236 which identified confusion over the existing definition which is based of sRGB specification. 

Collaborating with Lars Borg, color expert from Adobe, on these updates. Would like to send Lars a link to the pull request to view changes to the definition.


moekraft:
The definition for relative luminance is in the Glossary of WCAG 2.0 which is normative. This section cannot be updated. Closing this pull request and will instead propose a note on the formula in the Understanding 1.4.3 document and related techniques.

PR #265, opened 2017-02-08
Update G18.xml

jnurthen:
Changed incorrect reference to 1.4.5 to 1.4.6


patrickhlauke:
trawling some of the really old PRs ... need to decide if this is still relevant (for updating 2.0 quickref) or if any errors there are just left as they are (either merge or close?)

fstrr:
fixed in the html file that replaced the xml one.

PR #276, opened 2017-04-30
Modify link texts

momdo:


momdo:
This PR is very old. WAI-ARIA Authoring Practices links have been fixed already.

PR #277, opened 2017-04-30
Fix W3C status on F59

momdo:
WAI-ARIA 1.0 is W3C Recommendation now.


momdo:
This PR is very old. I have remade as #2510 instead.

PR #279, opened 2017-04-30
Update HTTP/1.1 resource to RFC 7231

momdo:
HTTP/1.1 Semantics is defined in RFC 7231 now.


momdo:
This PR is very old... I will remake PR.

momdo:
Refined as #2511.

PR #309, opened 2017-12-15
Fixed JavaScript error in SCR30 working example

G0dwin:
The [SCR30 working example](https://www.w3.org/WAI/WCAG20/Techniques/working-examples/SCR30/expand-links) does not work, it was using invalid JavaScript:

```javascript
for each (link in links) {
    ....
```


fstrr:
this has been fixed elsewhere in the [2.1 version of the working example](https://www.w3.org/WAI/WCAG21/working-examples/script-expand-links/), so I'm going to close this.

PR #314, opened 2018-01-19
In G200, list G201 as a related technique

lkmorlan:
Both are about opening links in new windows. G201 already links to G200.

### DIFF
--- a/wcag20/sources/techniques/general/G200.xml
+++ b/wcag20/sources/techniques/general/G200.xml
@@ -58,5 +58,6 @@
    <related-techniques>
       <relatedtech idref="H83"/>
       <relatedtech idref="SCR24"/>
+      <relatedtech idref="G201"/>
    </related-techniques>
 </technique>


mbgower:
**Official task force response**
The WCAG 2 Task Force facilitators reviewed this PR.
We are closing this PR without implementing it for the following reasons:

1. The "Techniques and Failures" section in each success criterion's Understanding document contains a list of all techniques which are considered relevant (whether sufficient, failure or advisory).
2. Each Technique page refers back to the Understanding documents with which it is associated under its Applicability section near the top.
3. Although many techniques documents have a Related Techniques section, what is included there is inconsistent between different techniques, and is rarely an exhaustive list of related techniques.

Points 1 and 2 seem to provide sufficient connections between the Understanding and Techniques documents. We do not believe it is necessary to review or update the many-to-many relationships possible between techniques themselves in the Related Techniques section. We also regard trying to update such material a low-priority task.

We are in the process of creating a single source of truth for the mapping of techniques to success criteria, which will be used to populate the sections of the Understanding and Techniques documents covered in points 1 and 2. We believe this is sufficient.

mbgower:
The draft response went through a working group review, and has now been marked as the official response.

PR #373, opened 2018-06-19
Understanding non-text-contrast updates

alastc:
From the discussion on the call 19/06/2018, updating the examples.


DavidMacDonald:
Typo
" Also, the visual effects which are implemented to order to indicate state, ..."

should be "in order..."

"a text link that only differs from adjacent text using color and there is no other visual indication that the text is linked (e.g. the link underline is removed) needs to ensure that the link color meets the 3:1 luminosity contrast ratio relative to the non-linked text color in order to meet this Success Criterion."

This is already true in order to meet 1.4.1 color SC. It sounds like a new requirement the way it's presented here. I suggest, 

"a text link that only differs from adjacent text using color and there is no other visual indication that the text is linked (e.g. the link underline is removed) needs to ensure that the link color meets the 3:1 luminosity contrast ratio relative to the non-linked text color in order to meet the related Success Criteria 1.4.1."

"Default link text is in the scope of 1.4.3 Contrast (Minimum), and the text is sufficient to indicate the link."
Shouldn't it be 
"...the underline is sufficient to indicate the link."

"User Interface Components that are not available for user interaction (eg: a disabled control in HTML) are not required to meet this color contrast requirements in WCAG 2.1."

There is actually a bit of controversy around the term "color contrast" ... perhaps best to leave off "color". Also a typo ... should be [requirement (no "s")]

"...A one-size fits all solution has been very difficult to establish, a method of varying the presentation of disabled controls <add>or an icon for disabled controls</add> based on user preferences is anticipated as an advancement in future...."

I still don't see where we allow focus indicators that use browser defaults to pass. We say "Where links or buttons have been styled (including the focus indicator), these must meet the 3:1 contrast ratio."

How about an addition to the link bullet before it?

"Where the focus style of the user-agent is not adjusted <add> on interactive controls </add>, the default focus style is sufficient."

"The phone icon is a simple shape within the orange circle. The meaning can be understood from that icon alone, the background behind the circle is irrelevant. <add>The orange background and white icon have a contrast ration of 3.4:1 which passes."


"Most of them have good contrast except the green triangles."
The green line appears to be 2.2:1. I think this line is sending a mixed message. Is it ok to have most of the lines right and have a failing line  or does this example fail the SC. I would say its the later. I think we should remove the green line from the example or make it darker passing green.

"The information is available in another form, <add> such as a table that follows a graph, which becomes visible when a "Long Description" button is pressed.</add>

"An infographic can use text which meets the other criteria to minimize the number of graphical objects required for understanding. For example, using text with sufficient contrast to provide the values in a chart.<add> A long description would also be sufficient because then the infograph is not relied upon for understanding.

johnfoliot:
> I still don't see where we allow focus indicators that use browser defaults to pass. We say "Where links or buttons have been styled (including the focus indicator), these must meet the 3:1 contrast ratio."

This is what I have been saying as well. Additionally, we have a (non-normative) [Failure Technique](https://www.w3.org/TR/WCAG20-TECHS/F78.html) which also appears to include background colors, which states:

> Description
>This describes a failure condition that occurs when the user agent's default visual indication of keyboard focus is turned off **or rendered non-visible by other styling on the page without providing an author-supplied visual focus indicator.**

DavidMacDonald:
I've added a pull request against this branch here.
https://github.com/w3c/wcag/pull/383

PR #379, opened 2018-06-22
Pointer gestures amendment

detlevhfischer:


alastc:
Preview at: https://rawgit.com/w3c/wcag/9608b6de7db85e236c15269dbf2a0e9006a08f4f/understanding/21/pointer-gestures.html

alastc:
Just checking on process, and I think this was accepted here:
https://www.w3.org/2018/07/17-ag-minutes.html 

But need to add it to a weekly survey soon.

alastc:
Hmm, I think this might be historical, we've made parallel update to this understanding doc separately, I think we've covered the issues raised, if @awkawk & @detlevhfischer agree let's close this.

I can't reconcile this against the master branch, the live version is quite different.

detlevhfischer:
I'm OK with closing this issue.  (We still need to resolve the one brought up by Mike Gower though - whether a slider thumb drag falls within free-form drag and is theefore excluded, or whether it counts as gesture under this SC.)

alastc:
Yep, we should resolve Mike's point, but I can't see it as part of this pull request, so I'll close this PR.

PR #383, opened 2018-06-25
Minor updates for clarity

DavidMacDonald:
One edit I didn't do because it's in the graph image and the text that follows it.

"Most of them have good contrast except the green triangles."
The green line appears to be 2.2:1. I think this line is sending a mixed message. Is it ok to have most of the lines right and have a failing line  or does this example fail the SC. I would say its the later. I think we should remove the green line from the example or make it darker passing green.

### DIFF
--- a/understanding/21/non-text-contrast.html
+++ b/understanding/21/non-text-contrast.html
@@ -18,7 +18,7 @@
 			<section id="user-interface-components">
 				<h3>Active User Interface Components</h3>
 
-				<p>For active controls on the page, such as buttons and form fields, any visual information provided that is necessary for a user to identify that a control exists and how to operate it must have sufficient contrast with the adjacent background. Also, the visual effects which are implemented to order to indicate state, such as whether a component is selected or focused, must also ensure that the information used to identify the control in that state has a minimum 3:1 contrast ratio.</p>
+				<p>For active controls on the page, such as buttons and form fields, any visual information provided that is necessary for a user to identify that a control exists and how to operate it must have sufficient contrast with the adjacent background. Also, the visual effects which are implemented in  order to indicate state, such as whether a component is selected or focused, must also ensure that the information used to identify the control in that state has a minimum 3:1 contrast ratio.</p>
 				<figure>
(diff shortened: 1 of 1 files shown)


understanding/21/non-text-contrast.html
```diff
@@ -158,7 +158,7 @@ <h3>Graphical Objects</h3>
 					<tbody>
 						<tr>
 							<td><img src="img/contrast-phone.png" alt="An orange circle with a white telephone icon in the middle." width="40" /></td>
-							<td><p>The phone icon is a simple shape within the orange circle. The meaning can be understood from that icon alone, the background behind the circle is irrelevant.</p>
+							<td><p>The phone icon is a simple shape within the orange circle. The meaning can be understood from that icon alone, the background behind the circle is irrelevant. The orange background and the white icon have a contrast ration grater than 3:1, which passes.</p>
```
AutoSponge: change `contrast ration` to `contrast ratio`

understanding/21/non-text-contrast.html
```diff
@@ -34,7 +34,7 @@ <h3>Active User Interface Components</h3>
...
+				<p>The <a href="visual-audio-contrast-without-color">Use of Color</a> success criteria addresses changing <strong>only the color</strong> of an object (or text) without otherwise altering the object's form. If a controls indicator of state (e.g. button background) varies only by color this is acceptable if the luminosity contrast ratio between the colors of the states differ by at least 3:1, or if there is another indicator of state. For example, a text link that only differs from adjacent text using color and there is no other visual indication that the text is linked (e.g. the link underline is removed) needs to ensure that the link color meets the 3:1 luminosity contrast ratio relative to the non-linked text color in order to meet the related Success Criteria 1.4.1.</p>
```
AutoSponge: `controls` to `control's`

awkawk:
Or we could just make it more clear that this graphic would trigger a failure unless the free line's color was addressed to be >3:1.

AutoSponge:
If you change the color of the "background" 3:1 and the border changes by less than 3:1, can we call that a pass? It penalizes someone for a stylistic choice to add a border. I'd still prefer the border as an indication of hit area.

PR #385, opened 2018-06-28
Removing the fixed-text examples

alastc:
IBM noted that there is no language in here saying text must be able to be adjusted, however, the SC text assumes that is possible, and deals with the results.

### DIFF
--- a/understanding/21/text-spacing.html
+++ b/understanding/21/text-spacing.html
@@ -55,15 +55,6 @@
             <img src="https://www.w3.org/WAI/GL/low-vision-a11y-tf/wiki/images/1/11/Spacing_overlap_fail.png" alt="Heading text overlaps part of paragraph text." width="510" height="170" />            
          </figure>
       		</section>
-      		<section>
-         <h4 id="fixed">Fixed Text</h4>
-         <p>Text  fails the SC when it is fixed and  not able to be overridden to the Success Criterion's metrics. </p>
-         <p>Text that allows for overriding to the metrics passes. Line height must be able to adapt to 1.5 times the font size. Letter spacing must be able to adapt to 0.12 times the font size. Word spacing must be able to adapt to 0.16 times the font size. Spacing underneath paragraphs must be able to adapt to 2 times the font size.</p>
-         <figure>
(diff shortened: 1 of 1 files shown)


lauracarlson:
I think the key bit in the SC text are the words "by setting". If you can't set the metrics, it is a failure.

An example of this type of failure was [documented last summer](https://www.w3.org/WAI/GL/wiki/Results_of_Bookmarklet_Tests_for_Issue_78) in the Argos mega menu: 

Check the [Screenshot](https://www.w3.org/WAI/GL/wiki/File:Argos_nav_spacing_test.png).

lauracarlson:
I did some testing. I was a glitch with the previous bookmarklet.  As this was an issue with mega menus, I checked an some:

- https://www.columbia.com/
- https://cmp.callawaygolf.com/
- http://www.brasnthings.com/
- https://www.officedepot.com/
- https://www.adidas.com/us
- https://www.northerntool.com/
- https://www.walmart.com/
- https://www.moosejaw.com/
- https://www.quiksilver.com/
- https://www.staples.com/
- https://www.sears.com/
- https://www.houseoffraser.co.uk/

They all adapt to the metrics with Steve's new bookmarklet. They don't with your old bookmarklet.

As it was a testing tool glitch, I'm okay with removing Example 4 in the Understanding Doc.

Thanks!

mbgower:
@lauracarlson the full phrase is "no loss of content or functionality occurs by setting..." The point the developer was making is that if you set the change and nothing happens, that's not actually a loss of content. So the wording

>  Figure 4: Absolutely fixed text is a failure. Adaptable text passes.

 is not actually true since fixed text isn't a failure of the SC wording.

Our larger discussion we were actually trying to answer was "are there any _bad_ coding techiques (CSS or otherwise) we can search for that allow us to do an automated flag of a failure of this SC?" If you have insight, we'd definitely welcome it!!

lauracarlson:
Hi @mbgower ,

I'll keep looking, but when testing with Steve's new bookmarklet, so far I haven't found any failures like Example 4. 

And thank you for bringing this issue up. Much appreciated.

alastc:
If there's no objections to this now, shall we survey it and resolve in a meeting?

PR #389, opened 2018-06-28
Tech media queries sticky

alastc:
New technique for sticky headers / footers, you can preview [Sticky headers / footers](https://rawgit.com/w3c/wcag/tech-media-queries-sticky/techniques/css/media-queries-sticky.html), and there is an associated [example page](https://rawgit.com/w3c/wcag/example-css-sticky/working-examples/css-sticky/index.html).

### DIFF
--- a/techniques/css/media-queries-sticky.html
+++ b/techniques/css/media-queries-sticky.html
@@ -1,0 +1,101 @@
+<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml" lang="en">
+	<head>
+		<title>Using media queries to un-fix sticky headers / footers</title>
+		<link rel="stylesheet" type="text/css" href="https://rawgit.com/w3c/wcag21/master/css/sources.css"/>
+	</head>
+	<body>
+		<h1>Using media queries to un-fix sticky headers / footers</h1>
+		<section id="meta">
+			<h2>Metadata</h2>
+			<p id="id"></p>
+			<p id="technology">css</p>
+			<p id="type">sufficient</p>
+		</section>
+		<section id="applicability">
+			<h2>When to Use</h2>
+			<p>This technique is applicable to Cascading Style Sheet / HTML technologies.</p>
+		</section>
+		<section id="description">
+			<h2>Description</h2>
+			<p>The objective of this technique is to be able to present content without sticky headers and footers when the device is held in landscape mode. This is done by using orientation techniques that adapt to the portrait or landscape mode of the device.</p>
+      <p>Sticky regions always stay visible in the viewport while the other content will disappear underneath when scrolling. This often is no problem in portrait mode but may block a big portion of the screen in landscape mode because of the height of the regions and leave only a small part of the screen for users to profit from. This is an effective way to create designs that present well in portrait mode while when users prefer landscape mode because of different reading and zoom preferences the sticky regions will be disabled.</p>
+      <p>The basic principle of un-fix sticky headers / footers is to:</p>
+      <ol>
+        <li>Define the sticky regions using orientation properties or properties for specific viewport sizes and media queries, so they get fixed or un-fixed depending on the portrait or landscape position of the device;</li>
+      </ol>
+		</section>
+		<section id="examples">
+			<h2>Examples</h2>
+			<section class="example">
+				<h3>Example 1: Medium complex un-fix sticky headers / footers in HTML and CSS</h3>
+				<p>The following medium complex example uses HTML and CSS to un-fix sticky headers / footers. The sticky regions get un-fixed as the orientation is changed. When the orientation property matches the device orientation defined via media queries, regions which are sticky in portrait mode will un-fix in landscape mode or vice versa. This particular example uses the CSS "orientation: portrait" property while it is also possible to use specific height properties.</p>
+        <code>
+        &lt;!DOCTYPE html&gt;
+          &lt;html lang="en"&gt;
+            &lt;head&gt;
+              &lt;meta charset="UTF-8"&gt;
+              &lt;title&gt;Using media queries to un-fix sticky headers / footers&lt;/title&gt;
+              &lt;style&gt;
+
+              /* Sticky Styling */
+
+              @media screen and (orientation: portrait) {
+                header { position: sticky; top: 0; }
+                footer { position: sticky; bottom: 0; }
+              }
+
+              &lt;/style&gt;
+
+            &lt;/head&gt;
+
+            &lt;body class="grid"&gt;
+
+              &lt;header role="banner" class="grid-item"&gt;
+                ...
+              &lt;/header&gt;
+
+              &lt;main role="main" class="grid-item"&gt;        
+                ...
+              &lt;/main&gt;
+
+              &lt;aside role="complementary" class="grid-item"&gt;
+                ...
+              &lt;/aside&gt;
+
+              &lt;footer role="contentinfo" class="grid-item"&gt;
+                ...
+              &lt;/footer&gt;
+
+            &lt;/body&gt;
+          &lt;/html&gt;
+        </code>
+				<p>Working example: <a href="https://rawgit.com/w3c/wcag/example-css-sticky/working-examples/css-sticky/index.html">Using CSS Flexbox for Reflow</a></p>
+			</section>
+		</section>
+		<section id="tests">
+			<h2>Tests</h2>
+			<section class="test-procedure">
+				<h3>Procedure</h3>
+				<ol>
+          <li>Display content on a mobile device / user agent in portrait mode.</li>
+          <li>Change the orientation to landscape.</li>
+          <li>Check whether the sticky header and footer will be un-fixed.</li>
+        </ol>
+			</section>
+			<section class="test-results">
+				<h3>Expected Results</h3>
+				<ul>
+          <li>Check #3 is true.</li>
+        </ul>
+        <p>If this is a sufficient technique for a success criterion, failing this test procedure does not necessarily mean that the success criterion has not been satisfied in some other way, only that this technique has not been successfully implemented and can not be used to claim conformance.</p>
+			</section>
+		</section>
+		<section id="related">
+			<h2>Related Techniques</h2>
+			<ul>
+				<li>ID</li>
+			</ul>
+		</section>
+	</body>
+</html>
--- a/techniques/css/media-queries-sticky.html
+++ b/techniques/css/media-queries-sticky.html
@@ -55,18 +55,14 @@
 
     /* Sticky Styling */
 
-    header { min-height: 100px; }
+    header { min-height: 130px; }
 
     @media (min-height: 480px) {
       header {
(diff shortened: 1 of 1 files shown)


detlevhfischer:
I was slighly puzzled by the focus on landscape vs. portrait - should the point not be that headers / footers should be made unstuck if the screen is too small to allow enough space for seeing the scrolling content? We are also moving into dangerous territory as having situations where only one orientation might be usable is in conflict with 1.3.4 Orientation (AA).
If I look at the example page referenced, in any of the mobile screen sizes that can be selected in developer tools, there is hardly any (or no) space for the scrolling text.
I believe a better approach migh be a page with sticky header at, say, 1280px that becomes unstuck as we move towards narrower viewports (potentially by zooming) - hasn't Alastair presented such a technique somewhere, recently?
So this, IMO is not ready,

mbgower:
I was also a bit puzzled about which SC this was supposed to be a sufficient technique for. As written, it seems to be for Orientation, but the requirement for Orientation is not that content needs to be transformed; just that it not be locked.
So it appears this is maybe a better fit with Reflow?
I would expect the technique to include a section on what SCs it is applicable to. Perhaps if this was added, it would help assess the technique.

alastc:
Hi everyone,

And thanks Jake for tackling this. A few notes from having been involved in the development of the reflow SC:

- This technique was intended to be an adviosry (but very recommended) technique for supporting reflow.
- We never got to the point of being able to say exactly how big a sticky header was too big, which is why it isn't a failure/sufficient, but is advisory.
- I think it would be simpler to focus on the height (and/or width) rather than orientation. I can see why using the orientation might help define your approach, but there is a simpler way.

What I'd like to encourage with this is a progressive-enhancement approach, the main issues we've come across in testing are: 

1. Sticky headers and/or footers that leave desktop-zoomers with a narrow slit of content.
2. Pop-out or floating menus that assume a certain height is available, and prevent scrolling, thus prevent people from getting to all the options.

By progressive enhancement, I mean that the application of  'stickyness' or non-scrolling is enabled when there is sufficient space.

For example, if a header is 100px tall and we assume that taking less than 25% of the vertical height is reasonable, then it should only stick if the height is 400px or greater.

If the menu is 500px high, then it should only prevent scrolling when there is 500px vertical height available.

In this way it doesn't matter what device is being used, it is about the content.

Cheers,

-Alastair

jake-abma:
Hi @alastc, thx for the inspect and suggestions, after my vacation will do the changes and techniques. Cheers!

lauracarlson:
+1

jake-abma:
@detlevhfischer @alastc @mbgower / all,

NEW OBJECTIVE:

> The objective of this technique is to be able to present content without sticky headers and footers when the space of the viewport is limited. This may be because of a device held in landscape mode or by zooming on a dektop.

Changed the example already and wondering if this suits the needs more...

https://rawgit.com/w3c/wcag/example-css-sticky/working-examples/css-sticky/index.html

In short there are just a couple of MQs at different sizes so:

- You'll see sticky headers on a mobile in portrait but not landscape.
- You'll see sticky headers on a tablet in portrait but not landscape.
- Sticky will be there on desktop but when you'll start zooming they disappear, depending a bit on viewport size.

Please provide feedback before I start to re-write the technique as this may change if the example is not sufficient. It may not cover all size scenarios but I think we're on the save side here.

Cheers!
Jake

alastc:
Hi Jake,

I wonder if it's possible to simplify even further? I'd suggest switching it around to more of a progressive enhancement approach, so:

```
    header {
      height: 100px;
    }

    @media (**min**-height: 480px) {
      header {
        position: sticky;
        top: -50px;
      }
    }
```
So the objective is to only use the sticky header when there is sufficient vertical space.

Or does it need to be the max-height verison to get the same effect?

jake-abma:
Like this?

https://rawgit.com/w3c/wcag/example-css-sticky/working-examples/css-sticky/index.html

alastc:
Yep, I just think it's slightly clearer, and maps to progressive-enhancement more easily. Doesn't make much difference in this case, but a good principle across the board.

It should also help in the wording, 'have enough space to include a sticky element' rather than removing it when there isn't.

jake-abma:
Hi @alastc 

I can add another style for tablet landscape like:

> @media (min-width: 576px) and (max-height: 768px) {
>   header {
>     position: static;
>     top: auto;
>   }
> }

and

> @media (min-width: 576px) and (max-height: 768px) {
>   footer {
>     position: static;
>     top: auto;
>   }
> }

This way we give a hint / idea to automatically get rid of the sticky part when in landscape mode.
For our [sticky example](https://rawgit.com/w3c/wcag/example-css-sticky/working-examples/css-sticky/index.html) the space of the header and footer is not that high but in other cases this might be the case but makes the example just a bit more complex...

Do or Don't?

jake-abma:
Correction, made the following change:

> @media (min-device-width: 576px) and (max-device-width: 1024px) and (orientation: landscape) {
>   header {
>     position: static;
>   }
> }

and

> @media (min-device-width: 576px) and (max-device-width: 1024px) and (orientation: landscape)  {
>   footer {
>     position: static;
>   }
> }
> 

Hope it's not too controversial, any case / edge case objections for this case?

jake-abma:
UPDATED: technique for sticky headers / footers, you can preview [Sticky headers / footers](https://rawgit.com/w3c/wcag/tech-media-queries-sticky/techniques/css/media-queries-sticky.html), and there is an associated [example page](https://rawgit.com/w3c/wcag/example-css-sticky/working-examples/css-sticky/index.html).

Please Review...

StommePoes:
So, this says one can use viewport sizes (in px).  What would a developer do who wants a sticky on a small screen, but only if the user hasn't set their font size larger/has enough vertical space? Or is that not detectable if it's set by OS? Would measuring viewport height or media queries in em work more flexibly?

Also, does it have to be a CSS technique if developers were using, say, JS to measure the sizes of things to determing the % of screen area a sticky takes up?
mobile screenshot of the footer where the background height is set in px yet the text spills out because the user has enlarged their text size

patrickhlauke:
@StommePoes 

> Would measuring viewport height or media queries in em work more flexibly?

Yes, that'd make more sense i'd say

> does it have to be a CSS technique if developers were using, say, JS

techniques don't have to be exhaustive / you don't really NEED to have techniques for any possible way of doing things. i'd say nowadays, media queries are likely the tool of choice for this sort of thing, rather than JS. (you can also do CSS MQs in JS with `window.matchMedia`, so an unholy mix of the two https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia)

alastc:
> So, this says one can use viewport sizes (in px). What would a developer do who wants a sticky on a small screen, but only if the user hasn't set their font size larger/has enough vertical space?

All the units are relative to something, so it really depends on what the content of the container is, and how it is laid out.

If you set the contianer to 100px, then using a vertical height media query as a minimum makes sense.
If you set the height in EMs, then set the media query in EMs as well.

Whether text wraps is another matter, and not really the target of this technique. (Especially as setting the height in EM wouldn't prevent text-wrapping and increasing the height outside of the container. Making sure the background extends would solve that problem.)

> does it have to be a CSS technique if developers were using, say, JS

Extending what Patrick said, each technique should be focused and clear. We can have another technique that's very similar, but for JavaScript. Given that it would be adjusting style properties, I think it's natural to start with a CSS one.

jake-abma:
Changed "height" to "min-height", scaling if fine now.

Cheers!

WayneEDick:
+1 I really love this example.
Wayne

detlevhfischer:
Looks good to me. I cannot reproduce the situations where there was hardly anmy space for scollable content, not sure whether this is because the example changed or I am doing it differently now...

jake-abma:
Hi @detlevhfischer , in our case the content has enough space because the sticky header / footer are not that 'high', but when you'll have higher ones you'll end up with not enough space and only a narrow view left for the content. The technique shows what happens / sticky becoming UNsticked... but indeed the need in this example page might not be / seem that important. Would you say to have the header keep it's height at all times so the needs seems more clear?

detlevhfischer:
I'd say a header of substantial height make the need for get it unstuck more obvious - but the example works as it stands.

jake-abma:
adjusted the example to bigger headers and footers, will this be more clear?

[example page](https://rawgit.com/w3c/wcag/example-css-sticky/working-examples/css-sticky/index.html)

jake-abma:
https://www.w3.org/2018/10/09-ag-minutes.html#item08


- Removed: "Medium complex" and now it is: "Example 1: Un-fix sticky headers / footers in HTML and CSS"

- Adjusted the example to bigger headers and footers, will this be more clear? [example page](https://rawgit.com/w3c/wcag/example-css-sticky/working-examples/css-sticky/index.html)

- @detlevhfischer  comment not answered yet: "in user testing sticky header when user tabs backward focus can be behind sticky header. Should we point out those potential problems? Or is this a specific technique? I've checked this by adding some links and indeed the focus end up behind the sticky header. But this is indeed also not what this technique is all about...

- @awkawk  mentioned something about: "The basic principle of un-fix sticky headers / footers is to:", not sure what the exact issue is here anymore...

@alastc will you do the "It would be good to have a quick refinement on the text and check in with Jake to make sure we have not changed anything significant. "?

Cheers!

jake-abma:
https://rawgit.com/w3c/wcag/tech-media-queries-sticky/techniques/css/media-queries-sticky.html

changed CSS example to bigger header/footers / simplified

detlevhfischer:
My suggestion for s re-formulation to include of a warning about issues brought upon uses by sticky headers would be this:

Old:

>  Sticky regions always stay visible in the viewport while the other content will disappear underneath when scrolling.
>  This often is no problem on desktop, or in portrait mode for mobile devices, but may block a big portion of the screen in landscape mode or when zoomed because of the height of the sticky regions and leave only a small part of the screen for users to profit from.
>  This is an effective way to create designs that present well with enough available space while when users prefer different reading and zoom preferences, or landscape mode, the sticky regions will be disabled.

Proposed change:

>  Sticky regions always stay visible in the viewport while the other content will disappear underneath when scrolling.
>  On mobile devices, sticky headers may block a big portion of the screen, especially in landscape mode or when zoomed in, because of the height of the sticky region may leave only a small part of the screen for the display of page content.
>  Disabling -- or un-fixing -- sticky regions is effective way to allow enough available space when users prefer different reading and zoom preferences or when using landscape mode.

> _Note:_ Be aware that the use of sticky regions carries disadvantages for keyboard users and should be considered judiciously. The problem for keyboard users tabbing through a page with a fixed header is that once the page has started to scroll, tabbing backwards to reach interactive elements higher up on the page will often mean that the focus becomes invisible once it moves behind the sticky header. Users must then scroll up to be able to see the focus - something they may not necessarily be aware of. In the same way, the visible focus can disappear behind a sticky footer, so users would need to scroll down to be able to see their focus position, which is a major inconvenience.

jake-abma:
Great @detlevhfischer, changed!

PR #418, opened 2018-07-03
Adding xmlns declaration

alastc:
The [reflow understanding doc](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html) isn’t being processed (plain-html doc), only difference I can see is the html tag / xmlns.

Does anything else need to happen to solve this?

If I was confident in the publication process I'd just merge it in, but if it is that simple please merge and delete this branch.

Thanks,

-Alastair


alastc:
V simple change, will merge and see the effect (if any).

PR #424, opened 2018-07-10
added mousedown complexities

detlevhfischer:
I have added some details on the complexities of the down event which may or may not be triggered by mousedown and tocuhstart handlers.


understanding/21/pointer-cancellation.html
```diff
@@ -14,6 +14,8 @@ <h3>Up-Event activation or completion</h3>
...
       
         <p>Up-event activation refers to the activation of a target when the pointer is released. In a touchscreen interaction, when the finger touches a target, the up-event activation only occurs when the finger is lifted while still being within the target boundary. Similarly in mouse interaction, the up-event occurs when the mouse button is released while the cursor is still within the boundary of the initial target set when the mouse button was pressed.</p>
+      <p>In contrast, the down-event can occur as soon as the pointer is pressed inside a target. Whether this actually happens can depend on the operating system, the user agent, and the mode of input. For example, the mousedown event will lead to a down-event in desktop operating systems when using mouse input, but will be ignored when using touch input in the same setting. Mousedown is currently ignored in mobile operating systems, but the down-event may be triggered here by the touchstart event, which (as the name suggests) is triggered by touch, but ignored when the same control is activated via mouse or trackpad input.</p>
```
patrickhlauke: a few clarifications:

- it's not that `mousedown` is "ignored" on touch devices (nothing to do with "mobile operating systems" ... this isn't specific to the OS, but to the touchscreen technology). touch devices/browsers do fire compatibility mouse events (see https://w3c.github.io/touch-events/#mouse-events), which include `mousedown`; it's just that these are fired all in rapid sequence when the finger is lifted from the touchscreen (so in effect fired on the "up" event)

understanding/21/pointer-cancellation.html
```diff
@@ -14,6 +14,8 @@ <h3>Up-Event activation or completion</h3>
 			<p>The most accessible way to incorporate pointer cancellation is to make activation occur on the up-event.</p>
       
         <p>Up-event activation refers to the activation of a target when the pointer is released. In a touchscreen interaction, when the finger touches a target, the up-event activation only occurs when the finger is lifted while still being within the target boundary. Similarly in mouse interaction, the up-event occurs when the mouse button is released while the cursor is still within the boundary of the initial target set when the mouse button was pressed.</p>
```
patrickhlauke: Clarification:

- "the up-event activation only occurs when the finger is lifted while still being within the target boundary" is a bit misleading. Yes, the `click` event isn't fired if the finger went down onto the touchscreen, then moved outside of the boundary, and then went up. However, regardless of whether the finger is inside or outside of the boundary, the `touchend` event is fired at the original element. If an author was listening to `touchend`, then they'd be firing functionality regardless of where the finger was at the end when it was lifted - inside or outside the element won't matter then.

detlevhfischer:
Well, then mousedown is not ‘ignored’ but ‘queued until the up-event happens’ - the point I was getting at is that in some environments (touch) it can appear that there is no “down-event problem” but in other contexts, there will be. Is this an unnecessary point, or does it just need rephrasing?

patrickhlauke:
@detlevhfischer i'd say for clarity it's worth rephrasing. the point stands of course that if an author was listening to `mousedown`, then on touch input it would - by virtue of how touch events work - satisfy the "up event" clause. but of course, it would still fail for mouse use (on desktop, or even on an android mobile/tablet if the user attached a bluetooth mouse)...so not a solution.

understanding/21/pointer-cancellation.html
```diff
@@ -13,7 +13,9 @@ <h2>Intent</h2>
...
       
-        <p>Up-event activation refers to the activation of a target when the pointer is released. In a touchscreen interaction, when the finger touches a target, the up-event activation only occurs when the finger is lifted while still being within the target boundary. Similarly in mouse interaction, the up-event occurs when the mouse button is released while the cursor is still within the boundary of the initial target set when the mouse button was pressed.</p>
+        <p>Up-event activation refers to the activation of a target when the pointer is released. In a touchscreen interaction, when the finger touches a target, the up-event activation only occurs when the finger is lifted while still being within the target boundary. Similarly in mouse interaction, the up-event occurs when the mouse button is released while the cursor is still within the boundary of the initial target set when the mouse button was pressed. Note however that the touchend event, if defined by the author, is fired at the original target element even when the pointer is released outside the target boundary.</p>
```
alastc: The last sentence is a little convoluted, does it mean the same as this?
> Note that if the author defines a touchend event, that event is triggered on the original target element, even if the pointer is released outside the target boundary.

understanding/21/pointer-cancellation.html
```diff
@@ -13,7 +13,9 @@ <h2>Intent</h2>
...
+      <p>In contrast, the down-event can occur as soon as the pointer is pressed inside a target. Whether this actually happens will depend on the mode of input. For example, the mousedown event will lead to a down-event down-event when using mouse or trackpad input, but it will not immediately fire when using touch input in the same setting. Using touch input, it only fires at the time of the up event. On devices with touch screens, a down-event will be triggered when authors use the touchstart event, while this event will not be fired when the same control is activated via mouse or trackpad input.</p>
```
alastc: Again, quite complex, can it be simplified? If I’m understanding correctly, could we say something like:
> In contrast the down-events can occur as soon as the pointer is pressed inside a target. The mousedown and touchstart events can trigger immediately, there are some differences depending on the input device, but unless the content will only be used in one modality they should both be avoided.

understanding/21/pointer-cancellation.html
detlevhfischer: Yes, your sentence is better, it means the same thing.

understanding/21/pointer-cancellation.html
detlevhfischer: I think your sentence is a bit unclear. I am also not sure whether we should give authors the idea that web content may only be used on mobile devices (where apparently mousedown is queued to occur just before the mouseup touchend event). I think @patrickhlauke Patrick might be the best one to phrase this...

alastc: Maybe ‘modality’ isn’t the right word then, it doesn’t mean just on mobile, it means in a mode (touch or mouse).

How about:
> In contrast the down-events can occur as soon as the pointer is pressed inside a target. The mousedown and touchstart events can trigger immediately.  There are some differences depending on the input device, but they should both be avoided for general web content.

PR #431, opened 2018-07-12
Example aria role status

mbgower:
Submitting this pull request, along with the technique for aria-status-role, which I will cross-reference.

PR #432, opened 2018-07-12
Tech using role status

mbgower:
Technique for 4.1.3 Status Messages. Working examples for this technique are contained in
https://github.com/w3c/wcag/pull/431


mbgower:
here's the rawgit
https://rawgit.com/w3c/wcag/tech-using-role-status/techniques/aria/aria-status-role.html

jake-abma:
Hi Michael / Marc,

Nice work so far!

It may be a good idea to also add and example for a "snackbar" like structure: [https://material.io/design/components/snackbars.html](https://material.io/design/components/snackbars.html)

It's the messages appearing at the bottom of a screen for a few seconds, are very populair and designers / developers often ask what to add to the code in order for PwD to all get the information.

mraccess77:
If we do mention snackbar it's probably worth mentioning that authors need to consider other SC as well such as 2.2.1 and 2.2.2 for a11y purposes.

mbgower:
> It may be a good idea to also add and example for a "snackbar" like structure: https://material.io/design/components/snackbars.html

@jake-abma, I've added in a new example in the Understanding doc where I give a snackbar example. It's in a[ new branch](https://github.com/w3c/wcag/tree/status-messages
), which I'll pull request as soon as the techniques go through and I can confirm the correct targets for the technique urls. 

Do you have any input to that materials.io information? If so, you may want to alert them that they do employ three screen shots where there is a Dismiss action offered, which is inappropriate for a Snackbar according to the documentation. These screen shots are used on examples labelled 'caution' and 'don't' AND there is a specific example not to use Dismiss later, BUT the Dismiss actions are not themselves called out as the thing that is wrong in the examples, so I think it makes more sense to make the action appropriate in this context.

snackbar example showing some photos and the message 'Your photo has been archived. Dismiss'

mbgower:
> If we do mention snackbar it's probably worth mentioning that authors need to consider other SC as well such as 2.2.1 and 2.2.2 for a11y purposes.

@mraccess77 I've avoided that by not including it as an actual technique example but just an example in the Understanding doc. I agree there are other considerations for these snackbars, but I don't think this technique is the appropriate place to go into them in detail.

mbgower:
@awkawk i have made changes to a number parts of the technique based on input from the call:

- title
- description
- code snippet
- procedure

 I believe it is now ready for round 2.

awkawk:
Thanks Mike.

I don't think that the "(such as a shopping cart image with proper ALT text) " is needed as this is covered in the example. We shouldn't be teaching people about the concepts in a technique procedure, in my opinion.

awkawk:
Also wondering if the description would be more readable if a paragraph break was added before "The aria live region role..."

alastc:
+1 (to the current draft).

Andrew's points would help, and I just noticed that the section under "Updating the shopping cart status" says "A screen reader will announce "Shopping cart, six items".

From the text before that, I think it would say 1 instead of 6? 

But not critical issues.

alastc:
@michael-n-cooper  - I updated the understanding doc, and added the examples to this pull request.

PR #435, opened 2018-07-18
Tech using role log

mbgower:
new technique for 4.1.3 Status Messages
https://rawgit.com/w3c/wcag/tech-using-role-log/techniques/aria/aria-log-role.html

Links go to working examples which have a separate pull request https://github.com/w3c/wcag/pull/436

### DIFF
--- a/techniques/aria/aria-log-role.html
+++ b/techniques/aria/aria-log-role.html
@@ -32,36 +32,34 @@
 			<h2>Examples</h2>
 			<section class="example">
 				<h3>Updating the contents of a chat conversation</h3>
-				<p>Comments that users type into a chat input field are appended to the end of the chat history region. The region is marked with role of <code>log</code> so that new additions are announced by ATs. The <code>aria-relevant</code> attribute is set to be "additions" only so that removed or edited messages are not re-announced. When each new chat message appears, a screen reader announces its content.</p>
-				<div class="code">
-					<pre>
+				<section class="description">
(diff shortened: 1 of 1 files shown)


mbgower:
I've added in a new example in the Understanding doc where I give a snackbar example. It's in a[ new branch](https://github.com/w3c/wcag/tree/status-messages
), which I'll pull request as soon as the techniques go through and I can confirm the correct targets for the technique urls.

@jake-abma, do you have input into the materials.io pages? On that snackbar page, there are some "Dismiss" action buttons shown in a number of examples, even though Dismiss should not be implemented. "Dismiss" admittedly appears in "caution" and "don't" examples, but with one except, use of "Dismiss" is not part of the example text. 
snackbars example showing some photos and the message Your photo has been archived. Dismiss

mraccess77:
Some have even suggested that we should avoid using aria-relevant altogether.

https://medium.com/dev-channel/why-authors-should-avoid-aria-relevant-5d3164fab1e3

mbgower:
re: aria-relevant

I can take this out for now. I _do_ think there is a case to be made for using it when a 'waiting' message disappears, so I may tackle it when I get to that technique/use case -- although I agree the AT support is spotty right now.

mbgower:
> It appears the first example once the "send" button is sent gives "undefined" if it is hit more than once.

Disabled Send button to make it one-time in https://github.com/w3c/wcag/pull/436

alastc:
> Using role=log to improve AT support

Should the main heading talk about "improve AT support"?

How about: "Using role=log to identify sequential information updates"

alastc:
Thanks Mike, I haven't used atomic in anger yet. 

I just assumed that 'atomic' meant peice by peice, therefore 'true' would mean per-change, and false would mean the whole lot. I shall update my mental model ;-)

PR #437, opened 2018-07-20
Tech autocomplete

DavidMacDonald:
https://rawgit.com/w3c/wcag/tech-autocomplete/techniques/html/autocomplete.html


rachaelbradley:
This is looking good and I agree with Mike Gower's edits. Please consider the following suggestions. +1 to moving this forward once the edits are complete. 

1. Consider adding an example tag into the first sentence  "The technique works by adding the appropriate autocomplete token to each form field, such as autocomplete="given-name", to make the identified...
2. The code sample is missing, though it may be the way I'm looking at this. 
3. Consider rewording the test steps to the following:  

Step 1: The form field has a valid and well formed autocomplete value.
Step 2: The purpose of the form field indicated by the label corresponds with the autocomplete token on the input.

lauracarlson:
+1

alastc:
+1
I did a little tidy up on the example to it comes through as a code block, encoding the tags.

alastc:
@michael-n-cooper This one is approved and ready, do you need to check anything?

alastc:
@michael-n-cooper I've done a little tidy up on the description. No new / different info, just moving a paragraph up and adjusting the spacing.

alastc:
Ok, probably need to update the [readme](https://github.com/w3c/wcag#user-content-images-examples-cross-references-for-techniques) then, that says to use the full URL. Is it plain, or with a # at the start?

michael-n-cooper:
The readme says use the full URI to parts of WCAG *other* than SC and guidelines, which have Understanding pages so we should link to understanding for those. It's without a hash, and from an understanding page is just the ID of the page, the way you edited it. As I say, I can update the generator to look for this to ensure consistency and probably should, rendering that comment unnecessary.

PR #380, opened 2018-06-22
Technique - ensuring single pointer usage

alastc:
Initial creation of the technique, comment on this pull request, and [preview on rawgit.com](https://rawgit.com/w3c/wcag/tech-ensuring-single-pointer/techniques/general/ensuring-single-pointer.html).

PR #454, opened 2018-08-21
Tech ensuring single pointer deltev

detlevhfischer:
This Technique seems to make a few other proposed Techniques for 2.5.1 Pointer Gestures redundant, for example:

- Not relying on path-based gestures
https://github.com/w3c/wcag/blob/tech-not-path-gestures/techniques/general/not-path-gestures.html

- GXXX: Do not rely on multi-point gestures


detlevhfischer:
@alastc You are right - I will try to re-write it more narrowly for the slider case.

detlevhfischer:
yes,  I think #380 has the substance of what we were discussing here - I was thinking of drawing up a more specific Tech and had talks with Patrick whether he can supply an example but we haven't made progress on that yet. So yes, I think we can close this one.

PR #455, opened 2018-08-21
Tech visible labels match names

detlevhfischer:


detlevhfischer:
Changed rawgit link from wcag21 to wcag repo

alastc:
I think in the last call we had a discussion, and the outcome was: It is ok for there to be text before or after the visible text. 

Given that, would the main heading/title be: Ensure the accessible name includes the visible text.

The examples are fine (you don't have to cover every possible variation, just positive examples of passing).

The test-procedure would need to update step 2: The visible label is included in the accessible name of the control.

In fact, when you do that it makes step 1 redundent.

detlevhfischer:
Have implemented Alastair's suggestions and taken the liberty to change the test description to include a string match requirement: "The visible label is contained as exact string in the accessible name of the control".
Before we finalise this, I would still like us to verify that an accessible name that does not start with the label text but contains it verbatim can actually be activated by speech input technology - @KimPatch  , could you verify / comment?

brooksienoodlesoup:
Would the following inputs pass the test for this technique?

<label>Weight in lbs.<input type="text"></label>
<label>Weight in st.<input type="text"></label>

The label and the accessible name are one in the same using these implicitly labeled inputs.  But, will the abbreviations through the speech recognition software off, considering that people don't really verbally announce the abbreviated form, but instead, they verbalize the expanded forms of the terms?

alastc:
It would pass as it doesn't override the visible label at all. I don't know whether speech input would deal with the contractions appropriately or not. 

Either way, I think it would be a user-agent issue to solve. If voice output can read the first as "weight in pounds", then presumably voice _input_ could also translate that verbalisation.

mraccess77:
If the visual and programmatic match it seems like a pass.  If one uses abbreviations and the other doesn't such as with a span and aria-label then I would likely think it a problem especially if the programmatic name contained the abbreviation.

KimPatch:
@detlevhfischer @alastc The way this was originally worded, with the visible label coming first, was intentional, and is important for speech input users. If the label comes first, you can enable the first few words of the label for speech input. If the label can be anywhere in the command, however, you have to enable all the words of the command to make sure that speech users can say the label. This creates a whole bunch of hidden commands, which are a big problem for speech input users. Not specifying that the visible label should come first, at least as a best practice, would be counter to the first and second lines of the understanding.

awkawk:
This is one technique, and there are >1 possible techniques. Perhaps we should focus this on that best practice with "Ensure the accessible name starts with the full string of visible text".

We can still have another technique that is "Ensure the accessible name includes the visible text" if we want, and/or we can allude to the possibility of including the visible text but not at the start.

If the core technique that we want to promote is to front load the text in the accessible name (and it seems to be) then we should focus the technique on that as clearly as possible.

mraccess77:
@KimPatch can you help me understand the issue if there is extra text in front in the accessible name.  For example, if we have an input field that has a visible label of "TV Shows" and it has an accessible name of "Search TV Shows" are you saying when someone says "search" it would focus this field which might not be the intended field? thus creating a hidden command?  When you say "enable all of the words in the command" are you saying you would create a custom command using Dragon or rely on the default command set?  Do you turn off the "Require click" setting in Dragon?  Thank you for helping me understand the specific situation.

alastc:
There were a few comments from the survey to address, from the [minutes](https://www.w3.org/2018/11/27-ag-minutes.html#item02).

alastc:
Based on the comments, I've updated the technique.

[Preview of the technique](https://cdn.staticaly.com/gh/w3c/wcag/tech-visible-labels-match-names/techniques/general/visible-labels-match-names.html?x=1).

The changes were:
- Cut down the description a bit, moving some to the understanding doc.
- Added a Japanese example with the hidden text first, mostly for testing at this stage.
- Updated the procedure to not use the term 'exact'.

PR #456, opened 2018-08-21
Tech drag drop cancel

detlevhfischer:
Rawgit view:
http://htmlpreview.github.io/?https://github.com/w3c/wcag/blob/tech-drag-drop-cancel/techniques/general/G208.html


alastc:
Hi Detlev,

Presumably this is to support 2.5.2, particularly "Completion of the function is on the up-event, and a mechanism is available to abort the function before completion or to undo the function after completion;"

A few little points:

- I think the description should be first, I found the UA support notes a little confusing to start with, it made sense after I'd read the objective. It might even be a sub-section of the Description?
- There's an extra left-bracken in the Resources H2 closing tag.
- In the procedure, rather than step one I'd put it in a paragraph above the llist, and say: 

> "For content which is draggable check whether:"
> 1. The drag-and-drop action can be reversed by releasing the picked-up item outside a drop target
> 2. The drag-and-drop action can be reversed by dragging the picked-up item back to its old position.
> 3. The drag-and-drop action can be reversed after releasing the item over the drop target via a dialog or an undo control.
> 
> Expected Results
> Any of #2, #3, or #4 are true

I suspect someone will say "this isn't a technique, it is a test procedure", but I can't think how to convert it to be a more specific technique at the moment :-/

mbgower:
typo: us for is, first line

alastc:
@detlevhfischer I *think* this is a sufficient technique for pointer cancellation, but I have to admit it took a lot of thought to get there, is that right?
@michael-n-cooper another one ready, including an update to the technique listing in the understanding doc.

detlevhfischer:
The rawgit link abz

alastc:
This was approved on the call, not sure why Detlev closed it. I was just checking that I had targeted the correct SC.

mbgower:
> For content which is draggable check whether:

Can't review the whole file at the moment, so I'll just note that a concern I have with this phrase is that some things are draggable which I might not classify as drag and drop. For instance, a slider is draggable. However, I haven't seen anyone implement a cancel function on a slider. To me, that would be pretty intrusive (given the dragger is just a change of value and it is just a matter of repositioning it.

PR #459, opened 2018-08-23
chore: Run eclint fix for whitespace and indents

nschonni:
Ran https://github.com/jedmao/eclint to apply the checked-in whitespace settings in the EditorConifg

***
<a href="https://pr-preview.s3.amazonaws.com/nschonni/wcag/pull/459.html" title="Last updated on Aug 23, 2018, 4:04 AM GMT (aaf54cf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/459/8f54c66...nschonni:aaf54cf.html" title="Last updated on Aug 23, 2018, 4:04 AM GMT (aaf54cf)">Diff</a>


nschonni:
If you look at the whitespace ignored diff https://github.com/w3c/wcag/pull/459/files?utf8=%E2%9C%93&diff=unified&w=1 there are a few diffs for the first lines in files. I am guessing those might be stripped BOMs

alastc:
Need some input from @michael-n-cooper.

I'm assuming this is theoretically a good thing to do, but I'm unclear whether:
a) It is desirable in practice, and 
b) if it is going to cause conflicts with every other pull request in the queue (I suspect so).

nschonni:
> a) It is desirable in practice, and

I think so, but if the cleanup shouldn't be done, then maybe the EditorConfig should be removed till the repo is ready to use it.

> b) if it is going to cause conflicts with every other pull request in the queue (I suspect so).

I think only in the case where the PR is editing a line that had trailing whitespace, but I'm not sure if those are merge conflicts that git won't just figure out itself by merging back the branch to the PR

alastc:
Having merged one thing in, that's showing as a conflict now!

It's worth noting that we have many contributors, mostly on the github interface so whitespace is going to be all-over the place anyway.

I think merging in that many files is bound to cause conflicts, perhaps @michael-n-cooper could look at the config.

For this type of wide-scale change, it's better to raise an issue first (but thank you for contributing!)

nschonni:
GitHub does have native support for EditorConfig, but I'm not sure if their web editor does the full file trimming like and IDE does. I know it does doe the line ending modifications and indent levels based of of the settings

michael-n-cooper:
This changes too many files to be able to review and assure myself that it does not have adverse effects. While whitespace changes seem innocuous, I can think of a number of risks, and it would be unfeasible to do the amount of review needed to manage those.

While I understand the desire for "clean code", the workflow of the AG WG does not support use of some of the automated code maintenance tools that are suitable for other repositories. Rather than pull requests with massive changes, I suggest filing issues with suggested editorial procedures, such as "use eclint to normalize whitespace". There may be "stopping points" where the editors can apply queued-up fixes if the suggestions are on the radar, but we are not at one of those now.

PR #462, opened 2018-08-23
Incorporate changes to status message Understanding post-michael cooper

mbgower:
Went in and added missing information from prior Pull Request that is not in conflict with Michael Cooper's updated version with proper sectioning.
I also added in an additional section tag which I think was missing for one of the H4 sections.


alastc:
I resolved a couple of conflicts, the last was around including classes on the techniques, referring to the 4.1.2 understanding doc it appears there should be a class on each link for aria/general/CSS etc.

detlevhfischer:
+1 for a very clear and detailed text. I have the following suggestions:

> They are already surfaced by assistive technologies

'surfaced' is probably the same as the often used "exposed" but I guess many readers might struggle to understand that term. Maybe reword to something like:
"A Change of context interrupts the user by taking focus. Assistive technologies recognise such a change and alert the user to new content." 
This my be too simple or not considered accurate enough though, not sure.
 
> In this  case it would also be a courtesy to add offscreen text such as "in shopping cart" to the message. 

One could argue that it is not just a courtesy but rather a requirement for meeting 2.4.6 Labels and descriptions.
 
Homogenize  "off screen" and "offscreen" (occurs several times) to "off-screen"

PR #466, opened 2018-08-29
non-text-contrast: Fix HTML code issues that impact generation of Understanding as expected

michael-n-cooper:
This involved moving some content from "resources" to "examples", as it
was not structured the way resources for an Understanding should be.
Because this is heavy-editorial I'm submitting pull request for review.

### DIFF
--- a/understanding/21/non-text-contrast.html
+++ b/understanding/21/non-text-contrast.html
@@ -300,14 +300,14 @@
 		</section>
 		
 		<section id="examples">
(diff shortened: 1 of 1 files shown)


michael-n-cooper:
I missed the blank section, thanks for that. Adjusting the heading levels wasn't necessary as the generator overrides them based on section nesting (which is one reason proper sections are needed for *every* heading), but it's not harmful either. :) This should be good to merge if you're ok with it.

PR #467, opened 2018-08-29
status-messages: Fix HTML code issues that impact generation of Understanding as expected

michael-n-cooper:
This involved moving some content from "resources" to "examples", as it
was not structured the way resources for an Understanding should be.
Because this is heavy-editorial I'm submitting pull request for review.


alastc:
Resolved the conflicts in favour of this branch, not 100% that is correct, it removed an ID and added a note?

michael-n-cooper:
I made a minor fix; otherwise this still looks good wrt the changes I was making. I haven't looked closely that it still covers the changes you were merging in, but if you're satisfied on that then it should be good to go.

alastc:
I think the main changes were from @mbgower, but I read everything under the description, and the flow looks good to me, so I'll merge.

Issue #433, opened 2018-07-16
Failure technique F87 contradicts the Accessible name and desc algorithm


Wildebrew:
The [Accessible name and description algorithm](https://www.w3.org/TR/accname-1.1/) (v1.1) describes how content included using the CSS :before and :after pseudo classes should be included in the accessible name calculation for an element.
This actually works with most browser and screen reader combinations, see my [CSS pseudo classes for screen readers](https://codepen.io/Wildebrew/pen/yqNWXG) Code Pen example.
But [WCAG 2.0 Failure technique 87](https://www.w3.org/TR/WCAG20-TECHS/F87.html) forbids use of CSS content to communicate information.

Since CSS content is now a part of an element's accessible name, the WCAG spec can no longer outlaw its use, in my opinion.
 

I believe F87 pre dates the ANDC algorithm by a few years and the web techniques landscape has changed. In light of that I recommend updating this technique to say somethihg like "whereas CSS content can be used to communicate helpful information about an element it should not be used for essential information" altogether.
or to drop it.

awkawk:
Thanks @Wildebrew 

Arguments to keep it: 
1. The test procedure requires that one checks whether the information is read by AT so if the content is read then the failure won't apply.

Arguments to remove it:
1. AT support is much better now than when this technique was authored.
2. The technique ignores that CSS is a technology that can be relied on, so it isn't appropriate to require that the information is available when CSS is off.

Wildebrew:
I don't think we can consider IE11 (or MS Edge)  as the mainstay of
accessibility support anymore.
The browser landscape is changing, even for assistive technology users.
IE11 is no longer being updated and Edge is barely used (see the
WebAIM screen reader survey #7 for numbers).
From a screen reader user perspective I don't think lack of IE11
support is sufficient to mark something as a failure.

The low vision/custom CSS stylesheet argument I can definitely see the
point, but then the wording of F87 should be changed to explain that
the failure is primarily related to the problems of users relying on
custom style sheets.
I wonder if custom style sheets can be written that preserve the
content of pseudo elements. My CSS background is admittedly not strong
enough to provide the answer on that.


On 7/28/18, Jonathan Avila <notifications@github.com> wrote:
> Personal thoughts  -  I belove it is not supported by screen readers in IE.
>  Seems like pseudo elwmrns also complicates the situation for low vision
> users who use custom css fonts/color.
>
> --
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
[...]

PR #473, opened 2018-09-11
F87 - removing the with-AT clauses

alastc:
For issue #433

### DIFF
--- a/techniques/failures/F87.html
+++ b/techniques/failures/F87.html
@@ -46,7 +46,7 @@ q:after  { content: close-quote }</pre>
       <section class="procedure"><h3>Procedure</h3>
   <p>For each instance of content inserted through use of the :before and :after pseudo-elements and the content property:</p>
         <ol>
-            <li>Check that non-decorative information presented by the generated content is available when styles are overridden./li>
+            <li>Check that non-decorative information presented by the generated content is available when styles are overridden.</li>
          </ol>
       </section>
       <section class="results"><h3>Expected Results</h3>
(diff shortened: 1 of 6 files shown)


techniques/failures/F87.html
```diff
@@ -1,7 +1,7 @@
...
       <p>All technologies that support CSS.</p>
    </section><section id="description"><h2>Description</h2>
-      <p>The CSS :before and :after pseudo-elements specify the location of content before and after an element's document tree content. The content property, in conjunction with these pseudo-elements, specifies what is inserted. For users who need to customize or turn off style information in order to view content according to their needs, assistive technologies may not be able to access the information that is inserted using CSS. Therefore, it is a failure to use these properties to insert non-decorative content.</p>
+      <p>The CSS :before and :after pseudo-elements specify the location of content before and after an element's document tree content. The content property, in conjunction with these pseudo-elements, specifies what is inserted. For users who need to customize or turn off style information in order to view content according to their needs, they may not be able to access the information that is inserted using CSS. Therefore, it is a failure to use these properties to insert non-decorative content.</p>
```
awkawk: I'm wary of adding "or turn off" since in WCAG you can rely on a technology like CSS or Javascript and requiring support for users turning off styles isn't mandated. I think that if we strike the "or turn off" phrase we will be ok.

techniques/failures/F87.html
```diff
@@ -45,7 +45,7 @@
          <ol>
             <li>Examine all content inserted through use of the :before and :after pseudo-elements and the content property</li>
             <li>Verify that the content is decorative.</li>
-            <li>If the inserted content is not decorative, check that the information is provided to assistive technologies and is also available when CSS is turned off.</li>
+            <li>If the inserted content is not decorative, check that the information is available when the :before and :after styles are overriden, or when CSS is turned off.</li>
```
awkawk: I'm wondering how this evolves. What if the browsers allow stacking of styles so the "jim:" that precedes a line in the can also have a paragraph symbol or series of dots to indicate a heading if the user wants that? Would the "or when CSS is turned off" still make sense?  I get that today the turning off of CSS is the easy substitute for the actual test, but am not sure if it is the right thing to do.

mraccess77:
I'm ok with the change although I agree it's a nebulous area.

techniques/failures/F87.html
alastc: That phrase was there already, I just removed the "assistive technologies may not be able to access the information" bit as that is not true anymore.

detlevhfischer:
"overriden" should probably be "overridden"
In the light of good(?) SR support of CSS-generated content I am a bit wary of this as a general failure. Are there actual use cases where LV users completely replace author's styles (not just beef up particular elements) and thereby prevent CSS-generated content form working? I have personally so far not come across a user doing that (I would imagine many other things would just break so it seems not a workable approach for most common sites) - but I am ready to support this if there is enough evidence that thisis a real use case.

awkawk:
Per https://www.w3.org/2018/09/25-ag-minutes.html#item05

PR #479, opened 2018-09-14
Fix/meta viewport examples

jake-abma:
- meta name="viewport" added


alastc:
I'm surprised these examples worked without that(!?), what triggered you to notice?

alastc:
Ah, I didn't test on mobile, only desktop, that's why.

@michael-n-cooper this is a simple change to the working examples, any problem with merging this?

Issue #403, opened 2018-06-29
Second thoughts on pointer gestures and drag-and-drop


michael-n-cooper:
_From @detlevhfischer on May 23, 2018 10:13_

I have some (late) second thoughts on the new SC Pointer Gestures.
Take a look at the [Salesforce example of drag-and-drop that has purpose of laying out elements on a two-dimensional plane](https://salesforce-ux.github.io/dnd-a11y-patterns/#/canvas?_k=l04mef) (as in composing a diagram) referenced in [a recent accessible drag-and-drop article on Medium](https://medium.com/salesforce-ux/4-major-patterns-for-accessible-drag-and-drop-1d43f64ebf09) by [Jesse Hausler](https://twitter.com/jessehausler).
This drag-and-drop example has been made keyboard-accessible (with added instructions for desktop screen reader users via aria-live) and also allows dragging with a single pointer (including touch). Now, do we really mean there should be an extra mechanism to move the objects with single clicks / tabs? This seems to necessitate either

- four discrete buttons at each side of a moveable object, or
- a mechanism (e.g. single tap) whereby the object can be selected / "picked up" and on-screen controls (arrows) are then available to move the selected element in discrete steps

I fear we might have gone too far; I fear that adding extra functionality for draggable objects to afford single click/tap operation might decrease the usability / affordance overall, and may not be worth it (quite apart from adding a lot of complexity for developers). Even so, the current SC text seems to mandate exactly that, and would fail the example above. Since accommodations are possible, this does not seem to be a case for the 'essential' exception - but others may disagree.

Thoughts? 

_Copied from original issue: w3c/wcag21#937_

michael-n-cooper:
_From @patrickhlauke on May 23, 2018 10:20_

I'd note that the intention for the SC is to prohibit *path-based* gestures. i.e. where a user has to follow a specific path / make a specific gesture (e.g. a precisely timed and executed swipe or similar). Dragging/dropping is more of a freeform interaction...while the user is dragging, it doesn't matter if they stray off a specific path or not, what counts is only the start and end points. Maybe this should be clarified (assuming my take on this is correct).

michael-n-cooper:
_From @patrickhlauke on May 23, 2018 10:24_

[...]

PR #484, opened 2018-09-24
Pointer gestures understanding: implementing mbgover's comments

detlevhfischer:
Rawgit view: https://rawgit.com/w3c/wcag/understanding-pointer-gestures/understanding/21/pointer-gestures.html
Removed dragging in general (but left in 'dragging nof a slider thumb'), incorporated a variant of what Michael had suggested as a replacement text. Changed the last example (floor-planning app) away from the confusing drag-n-drop focus, and made it into a Kanban widget which allows horizontal swiping to move elements to adjacent silos.


alastc:
Based on [previous issue](https://github.com/w3c/wcag/issues/403#issuecomment-407862763), this change makes it clear 'drag and drop' is not caught by the pointer gestures SC.

@awkawk one for the survey.

PR #485, opened 2018-09-25
Scr29 example update

awkawk:


detlevhfischer:
> // Was the Enter key pressed?
>    if (key == 13) {
>    if (key == 13 || key == 32) {

should probably be updated to read 
Was the Enter or Space key pressed?

momdo:
According to MDN, [KeyboardEvent.keyCode](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) is deprecated. We need to use [KeyboardEvent.code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) instead.

PR #487, opened 2018-09-25
Updating the stylish / stylus reference.

alastc:


awkawk:
I'm not sure if it matters but the change makes the format for the entry differ from the other items in this list...

lauracarlson:
@awkawk wrote:

> I'm not sure if it matters but the change makes the format for the entry differ from the other items in this list...

How about reformatting to something such as:

*[Stylus Team](https://github.com/openstyles/stylus/graphs/contributors) (2012). [Stylus browser extension](https://github.com/openstyles/stylus/blob/master/README.md) (Firefox, Chrome, and Opera) (compatible with Userstyles.org material). 

@JasonBarnabe's  [initial commit was on April 15, 2012](https://github.com/openstyles/stylus/commit/b019038f5701b67d475a36275fcbe69ac5d6ee04).

alastc:
Yes, I was reading the code rather than rendered version, I've updated to Laura's suggestion.

alastc:
@awkawk this seems like a "merge and notify" editorial situation, or should it go to the survey?

Issue #476, opened 2018-09-13
1.4.3 Resource is not exist


jidaikobo-shibata:
Hi,

https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum
> (See The American Printing House for the Blind Guidelines for Large Printing and The Library of Congress Guidelines for Large Print under Resources)

This line says "see Resources". But, there is no resource about "The American Printing House for the Blind Guidelines for Large Printing and The Library of Congress Guidelines for Large Print" under the "Resources".

I think "Resources" should contain below.

APH — Guidelines for Print Document Design
https://www.aph.org/research/design-guidelines/

and also, link text of "Resources" are broken (there are two broken link).

thanks.

awkawk:
This link was removed before the March 2016 version was published (https://www.w3.org/TR/2016/NOTE-UNDERSTANDING-WCAG20-20160317/) because the old link was 404 and we couldn't find a new link.

The 
http://www.loc.gov/nls/reference/circulars/largeprint.html link was removed the version before that (link text was "National Library Service for the Blind and Physically Handicapped (NLS), The Library of Congress Guidelines for Large Print")

New link may be: 
https://www.loc.gov/nlsold/reference/guides/largeprint.html

Since we have the text in the description that references these links we should probably reinstate them.

mbgower:
> and also, link text of "Resources" are broken (there are two broken link).

The anchors are the same for both 2.0 and 2.1
https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html#visual-audio-contrast-contrast-resources-head
https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum#visual-audio-contrast-contrast-resources-head
(although the paths are obviously different) so since the actual resource is pulled in 2.1, the link breaks. But I question why these aren't pointing at the id for the heading ("resources") instead of the actual link content.

<code>a href="#visual-audio-contrast-contrast-resources-head">Resources</a</code>

As a basic convention, I would recommend it point here:
<code>a href="#resources">Resources</a</code>

I'm going to do a pull request that addresses this by adding in the new resource, but the target will point to the section heading. @michael-n-cooper I wonder if it would be an idea to do a global search and see if there are other instances where "resource" links should be pointing to the heading instead of the list item?

awkawk:
Yes, the linking to resources needs to be addressed in the generator - @michael-n-cooper is on that part, we should keep it separate from this Issue.

[...]

PR #489, opened 2018-09-26
updated anchors and supplied new link targets

mbgower:
In accordance with https://github.com/w3c/wcag/issues/476 and have made the link to Resources generic to the heading, and added in two new resources for APH and NLS.


awkawk:
+1 for me

Issue #481, opened 2018-09-14
Include details on user interface components in 1.4.3 & 1.4.6 Understanding


jnurthen:
1.4.11 includes 
https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html#inactive-controls

This same section should be included in the understanding documents for 1.4.3 and 1.4.6.
Alternatively the defintion for "Inactive user-interface components" could be placed in the glossary (but this would be a normative change)

awkawk:
Seems reasonable, some sections will need to be edited for that (e.g. the backward compatibility item that references 1.4.3).

mbgower:
Note that while I was investigating this, I noticed the use of "eg" in this section (my emphasis):
> User Interface Components that are not available for user interaction (**eg:** a disabled control in HTML) are not required to meet contrast requirements in WCAG 2.1. An inactive user interface component is visible but not currently operable. Example: A submit button at the bottom of a form that is visible but cannot be activated until all the required fields in the form are completed.

That is a deviation from the standard style for the Latin _exempli gratia_ abbreviation, which is "e.g.," as per Chicago style (which I think is what w3c follows; it happens in 28 of the Understanding docs). 
I am going to do a PR that not only adds in the newer section to those respective Understanding docs, but also updates the abbreviation. I'm also going to remove the hyphen in "in-active" and the terminal punctuation in the figure caption (another Chicago convention), and generally clean up any style inconsistencies in the source doc so we don't just copy over things that need to be changed later in multiple sources.

awkawk:
That's fine, but ideally the 100% editorial items would go in a separate pull request and Alastair or I would just merge it without further discussion (assuming only editorial changes)

mbgower:
I have a suggestion about how to proceed with this. The first paragraph and example in the section reads as follows:
> ### Inactive User Interface Components
> User Interface Components that are not available for user interaction (eg: a disabled control in HTML) are not required to meet contrast requirements in WCAG 2.1. An inactive user interface component is visible but not currently operable. Example: A submit button at the bottom of a form that is visible but cannot be activated until all the required fields in the form are completed.
> [image of Grey button with non-contrasting grey text.]
> An in-active button using default browser styles.

I am suggesting that " in WCAG 2.1" be removed from the first sentence. With my other changes, it would read as:

> ### Inactive User Interface Components
[...]

PR #491, opened 2018-09-27
editorial changes to existing 1.4.11 understanding doc

mbgower:
This commit is ancillary to the request in https://github.com/w3c/wcag/issues/481 to propogate the inactive control information to the two 2.0 Contrast SCs.
- It fixes up the inconsistent use of e.g., and i.e., ;
- removes terminating punctuation in figure captions that are sentence fragments (but for now leaves in place periods in those captions that form complete sentences; pending clarity from @michael-n-cooper );
- does trivial editorial changes (adding a missing possessive, removing or adding dashes as necessary);
- slightly alters a phrase for better readability

### DIFF
--- a/understanding/21/non-text-contrast.html
+++ b/understanding/21/non-text-contrast.html
@@ -22,7 +22,7 @@
 				<figure>
 					<!-- example from https://getbootstrap.com/docs/4.1/components/buttons/ -->
 					<img alt="Two buttons on a white background. The first has a blue border to indicate its boundary, white internal background and blue text. The second adds a thick blue-grey outer border to show focus." src="img/first-button-example.png" /> 
-					<figcaption>An active control showing the visual boundary by default (color of <code>#007bff</code> on white), and the focus indicator.</figcaption>
+					<figcaption>An active control showing the visual boundary by default (color of <code>#007bff</code> on white), and the focus indicator</figcaption>
 				</figure>
 				
 				<p>This success criteria does not require that controls have a visual boundary indicating the hit area, but if the visual indicator of the control is the only way to identify the control, then that indicator must have sufficient contrast. If text (or an icon) within a button is visible and there is no visual indication of the hit area, or the visual indication does not provide the only indication, then there is no contrast requirement beyond the text contrast (<a href="contrast-minimum">1.4.3 Contrast (Minimum)</a>) or icon contrast covered by the Graphical Objects part of this Success Criteria. Note that for people with cognitive disabilities it is recommended to delineate the boundary of controls to aid in the recognition of controls and therefore the completion of activities.</p>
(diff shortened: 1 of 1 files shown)


mbgower:
This PR is strictly editorial triviality, and should be able to be executed without validation by the group or a CFC process. It constitutes minor fixes to the Understanding doc noticed while I was working on a different issue.

alastc:
@michael-n-cooper (next week when back from holiday) Is this in keeping with our usual styles? Previous understanding docs haven't used figcaptions that I've noticed. I don't mind either way, just not sure what would be consisent.

understanding/21/non-text-contrast.html
```diff
@@ -31,14 +31,14 @@ <h3>Active User Interface Components</h3>
...
+				<p>The <a href="use-of-color">Use of Color</a> success criteria addresses changing <strong>only the color</strong> of an object (or text) without otherwise altering the object's form. If a control's indicator of state (e.g., button background) varies only by color this is acceptable if the luminosity contrast ratio between the colors of the states differ by at least 3:1, or if there is another indicator of state. For example, a text link that only differs from adjacent text using color where there is no other visual indication that the text is linked (i.e., the link underline is removed) needs to ensure that the link color meets the 3:1 luminosity contrast ratio relative to the non-linked text color in order to meet the related Success Criteria 1.4.1.</p>
```
awkawk: @mbgower this is the only one that I'm not 100% on. This feels like an example but maybe needs to be better clarified as an example?  How about:

(e.g., black paragraph text with blue links with the link underline removed)

awkawk:
@alastc If this is a different style that is a different change that can be discussed. We can accept this once @mbgower clarifies my one question since all of his changes are editorial.

understanding/21/non-text-contrast.html
mbgower: I assume you're talking about where I change the e.g. to i.e., ? (The only other change here is adding in a comma after the prior e.g.).
If you think the actual wording needs to be changed, that can be done, but I simply did a search for e.g., style, not for content. I changed this to i.e. (_id est_, 'in other words') because the _only_ way I could think of have no other indication of text being a link is to remove the underline. So my feeling was that this isn't an example, it's a rewording. Hence the i.e.

BTW, since we're talking about wording, I find this whole thing a little awkwardly phrased. I believe this would both make more sense AND be more accurate

> For example, if a text link only differs from adjacent text using color (because the browser-supplied persistent underline for the link has been removed by the author) the author needs to ensure that the link color meets the 3:1 luminosity contrast ratio relative to the non-linked text color in order to meet the related Success Criteria 1.4.1.

I also want to add in an additional note saying the underline has to appear on focus, but maybe that is just making the use case too messy when we're only trying to address...

You know what?! I think this link example is really not a good example here. "If in doubt, take it out". I would be inclined to strip the entire final sentence out and just end it at "...another indicator of state." Bringing in a link example is just making it too messy!

mbgower: It is, after all, supposed to be about NON-TEXT contrast.  Geez, one pulls a thread and the whole sweater begins to unravel!

awkawk: I would be happy to take the parenthetical out of the sentence. The sentence is already a "for example" so it should be simple enough without an example.

awkawk: The link example is good, in the "for example" - the contrast is about the non-text aspects of the link.

mbgower: @awkawk, 

Here's what I think we should do. Let's restrict the changes to the ones I made, which is almost completely concerned with swapping out e.g. for e.g.,  (and replacing one e.g. with an i.e.,)

I will then actually tackle a larger review of this Understanding doc in a separate issue https://github.com/w3c/wcag/issues/492. 

-- 
In relation to where this conversation has gone, I'd much rather have the example be about non-text. For instance, it could talk about inversing the foreground and background colors (as is the case with the 3rd blue-on-white button, where it becomes white on blue on focus) or thickening the border (as is the case with the first blue-on-white button that gets a wider blue border on focus).  And as I write that, I'm wondering if the figures should be numbered?

My concern with the example as it stands is that if there is no underline indicating link, then the color example is wholly about text, not non-text.  But, I suggest we just ignore that for now and get these 'easy to pass' editorial changes in so that  481 can get closed.

alastc:
My comment was just on the figcaption, and whether it finishes with a full stop or not, I couldn't think of other understanding docs with images to see.

understanding/21/non-text-contrast.html
awkawk: ok

awkawk:
My gut is that it would depend on whether the caption is a sentence or not.

PR #494, opened 2018-10-05
Tech fitting images viewport

jake-abma:
New technique for Using Fitting Images for Reflow, you can preview [Using CSS max-width and height to fit images](https://rawgit.com/w3c/wcag/tech-fitting-images-viewport/techniques/css/tech-fitting-images-viewport.html), and there is an associated [example page](https://rawgit.com/w3c/wcag/example-fitting-images-viewport/working-examples/css-fitting-images/index.html).

### DIFF
--- a/techniques/css/tech-fitting-images-viewport.html
+++ b/techniques/css/tech-fitting-images-viewport.html
@@ -33,7 +33,7 @@
     <section id="examples">
       <h2>Examples</h2>
       <section class="example">
-        <h3>Example 1: Simple fitting images in HTML and CSS</h3>
+        <h3>Example 1: Fitting images in HTML and CSS</h3>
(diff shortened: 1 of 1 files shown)


jake-abma:
Wondering if we should also mention a case to the use of width: 100% to not only allow an image to scale down if it has to, but also scale up than its original size if the container allows.

jake-abma:
For stock images I'm wondering what kind we would like to use... ?! For now I've used some WAI images to work with.

jake-abma:
And what about the use of the `srcset` attribute to provide multiple versions of images and have the browser do the work of figuring out which image fit best?

detlevhfischer:
I don't mind if any of these are used: http://oturn.net/photos/index.html

jake-abma:
- pictures added
- "text intended changed to "content intended"
- "height: auto" deleted as this will be set by default
- paragraph starting "Layouts..." changed

@alastc will you do a final text check or will this be enough?

PR #500, opened 2018-10-05
Example css sticky

jake-abma:


alastc:
Noting that this was required for #389, created a 404 on the techniques site without it.

PR #502, opened 2018-10-05
Added video style to additional.css

alastc:
Just a max-width to prevent it overflowing.


alastc:
I started picking apart the WAI video player, but it would take too long remove all the features that they use. This is a much simpler solution...

alastc:
Thanks. It resolves the video in Reflow expanding out of view. We were going to try the WAI player, but I'm not sure I'll ever have time for that :-/

PR #506, opened 2018-10-06
Tech failure content hover focus not hoverable

detlevhfischer:
View in Rawgit: https://rawgit.com/w3c/wcag/tech-failure-content-hover-focus-not-hoverable/techniques/failures/tech-failure-hover-content-hoverable.html

Drafted this fairly obvious failure for content on hover or focus. One point I was not sure about is whether it should exclusively focus on pointer operation or whether it is better to include both content shown on hover AND focus even if the failure condition then applies to not being able to move the pointer over the content, once displayed.
(Not sure why this PR also shows my changes to another Failure Technique in another branch - still struggling with Github...)


Ryladog:
+1 (LVTF)

** katie **

Issue #517, opened 2018-10-11
1.4.11 seems to include the ACTIVE state of links and buttons.


DavidMacDonald:
The SC seems to require **_every_** state to have sufficient contrast. Technically I would say that includes the ACTIVE state of clicking which is visible for a spit second while clicking.

https://www.w3.org/TR/WCAG21/#dfn-states

Do we need language in the understanding about transient states not being what was intended.

awkawk:
This was discussed as part of https://github.com/w3c/wcag/issues/157 and we addressed the bulk of the issue but this part was missed. I doubt that this comes up too often in reviews, but we should add something in the understanding document to clarify.  

It seems like since the SC speaks to "Visual information required to identify user interface components and states" that we would be able to make the case that the visual information for the active state isn't required to identify the state because the user is actively clicking on it.

mraccess77:
I agree a note would be useful.  When we discussed I believe we had agreed it was the case that the active state was excluded for the reason Andrew describes.

alastc:
Someone asked me if [all 16 states](http://blog.selfthinker.org/2011/07/10/trick-question-how-many-link-states-can-be-defined-in-css/)(!) of links needed to be contrasting with each other, so this could do with clarification!

patrickhlauke:
on the "with each other" part: as the SC talks about "adjacent colors", am i right in thinking that the SC itself doesn't require that there be sufficient contrast between different states of the same control, provided they're not adjacent? (example of adjacent colors would be, for instance, a horizontal row of rectangular buttons, with no gap between them, and one of them is "selected" with a darker background color - there the requirement would be that the selected background color, since it's adjacent to non-selected background color, have a sufficiently different color)

mraccess77:
It was my understanding that the states between a selected tab and non-selected tab would be covered here by the contrast requirement if there was no other mechanism to visually tell them apart.  However, separate states that aren't necessary to understand the purpose aren't covered like active state.  In addition, comparison between states like focus and hover state doesn't add any value so there is no need to have contrast differentiation between the different hover and focus states -- just each state by itself.

I think the question people should ask - if you removed the insufficient contrasting colors do you lose anything required to identify/provide understanding that you can't get another way.

patrickhlauke:
> It was my understanding that the states between a selected tab and non-selected tab would be covered here by the contrast requirement if there was no other mechanism to visually tell them apart

[...]

PR #530, opened 2018-11-07
changes to address #517

awkawk:
changes to the Understanding document to address the issue.

### DIFF
--- a/understanding/21/non-text-contrast.html
+++ b/understanding/21/non-text-contrast.html
@@ -18,8 +18,11 @@
 			<section id="user-interface-components">
 				<h3>Active User Interface Components</h3>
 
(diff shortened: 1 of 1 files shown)


alastc:
That solves the source issue, I wonder if we need to tackle the issues about contrasting between states.

I've been asked if link-states need to contrast with each other (and the background), or just against the background.

My reply was that it is against adjacent colors, so a link's contrast is not against itself, perhaps add something like:
"...Also, any visual effects implemented which are necessary to indicate state, such as whether a component is selected or focused, must also ensure that the information used to identify the control in that state has a minimum 3:1 contrast ratio. **Added:** The state indicators are not required to contrast with each other. For example, a text link which changes from blue to purple when in a visited state has to contrast with the background, not un-visited state."

awkawk:
@alastc Let's resolve the issue at hand and then deal with further changes as needed.

alastc:
Approved here: https://www.w3.org/2018/11/20-ag-minutes.html#item06

Issue #144, opened 2016-01-12
[LowVision] Techniques needed for Icon Fonts


joshueoconnor:
The use of Icon fonts is more prevalent, do we need techniques for Icon fonts? My sense is yes. They are semantically rich, whether in a Unicode like strings etc or whatever.

awkawk:
Preparing to mark as deferred due to inactivity and group resource constraints (https://www.w3.org/WAI/GL/wiki/Handling_Issues).

lauracarlson:
Hi Andrew and Josh,

I worked on some techniques in the Wiki last February.

* [Using a Decorative Icon Font](https://www.w3.org/WAI/GL/wiki/Using_a_Decorative_Icon_Font)
* [Providing an On-Screen Text Alternative for an Icon Font](https://www.w3.org/WAI/GL/wiki/Providing_an_On-Screen_Text_Alternative_for_an_Icon_Font)
* [Providing an Off-Screen Text Alternative for an Icon Font](https://www.w3.org/WAI/GL/wiki/Providing_an_Off-Screen_Text_Alternative_for_an_Icon_Font)

Thoughts?

Thanks,
Laura

joshueoconnor:
Good work @lauracarlson. I think they look really good, and just got missed somehow. I'd say they are ready for WG review @awkawk is this ok with you, and enough reason to not mark as deferred?

awkawk:
Yep, this qualifies as activity!

joshueoconnor:
Excellent, and many thanks to @lauracarlson for this - looks great.

joshueoconnor:
Thanks for this Laura - great stuff. Please let me know when the LVTF have reviewed it and we will bring it the wider group for approval.

patrickhlauke:
Possibly related...advice to actually not use icon fonts altogether: https://speakerdeck.com/ninjanails/death-to-icon-fonts / http://blog.cloudfour.com/seriously-dont-use-icon-fonts/

joshueoconnor:
Thanks @patrickhlauke . @lauracarlson  that may be useful.

lauracarlson:
Thank you very much @patrickhlauke. I added the links to the resource sections of the Wiki pages.

joshueoconnor:
Thanks @lauracarlson are these updated technique ready for review? 
    Using aria-hidden="true" on an icon font that AT should ignore
    Providing an On-Screen Text Alternative for an Icon Font

lauracarlson:
[...]

PR #531, opened 2018-11-08
Technique - setting role of img on icon fonts

alastc:
Latest Update by @jake-abma 

**!!! Seems like we have a "github htmlpreview" issue so example only properly working in Chrome !!!**

Technique: https://github.com/w3c/wcag/blob/tech-icon-font-img-role-v2/techniques/aria/icon-font-img-role.html

Preview: https://htmlpreview.github.io/?https://github.com/w3c/wcag/blob/tech-icon-font-img-role-v2/techniques/aria/icon-font-img-role.html

Examples: https://github.com/w3c/wcag/blob/example-icon-font-img-role/working-examples/aria-icon-font-img-role/index.html

Preview: https://htmlpreview.github.io/?https://github.com/w3c/wcag/blob/example-icon-font-img-role/working-examples/aria-icon-font-img-role/index.html => Preview Example

History

https://github.com/w3c/wcag/issues/144
https://github.com/w3c/wcag21/issues/296
https://github.com/w3c/wcag21/issues/297

--------------------

**OLD:** Technique originally provided by @lauracarlson, [preview here](https://htmlpreview.github.io/?https://github.com/w3c/wcag/blob/tech-icon-font-img-role/techniques/aria/icon-font-img-role.html).

Note: There are a couple of things to tidy up for the examples, which need to be added to the WCAG repository.

Question: Would it make sense to incorporate SVG & background images in this technique, or would it get too long?


detlevhfischer:
+1 - looks good

marcjohlic:
+1 to this technique

Small typo in Description:  "The point is to provide a technique to differentiate icons fonts from general font (text) usage."   

'icons' should be 'icon'

mraccess77:
+1

allanj-uaag:
+1

jake-abma:
There are certain issues with this technique that don't feel right and I suggest to at least make clear why we decided to provide such a solution and what the drawbacks are.

This technique only covers a "atomic element" case of icon fonts and not all icon font usages. A "atomic element" case here means if an element ONLY contains the icon font AND is not part of a combined graphic like multi coloured icon fonts or already semantically structured content.

At least some text should be changed like the test procedure and one or more notes added of what this technique exactly solves and what not. 

Sites using a combination of atomic element icon fonts, multi coloured icon fonts and icon fonts added to other elements containing more semantics than only the icon font may end up not being able to use the test procedure and the possibility only some icon fonts remain on the page but other will still be deleted as they are not targeted.

1.

> The objective of this technique is to show how to semantically identify an element that uses a font to create an icon.

This is not absolute as I can also use a font icon on a link with the :before where you do NOT want to change the semantics of the link and DO want to change the font style of the link text. In the following example the element uses a font to create an icon and this technique / procedure won't work. It will only work when a span is used inside the link, not directly on the link (which you'll see a lot). Not always will developers be able to add a atomic version...

```
<a href="">
(a:before => icon)
Text node
</a>
```

Lots of other examples are out there where font icons are shown in combination / attached at other structures.

2. 

> There are millions of multiple coloured font-icons out there where these solutions will not work. The multi coloured icons consist of more than 1 font-icon stacked on top of each other. Often these are used in a mix together with the simple one-off font-icons. Think of favourite icons where there are 5 next to each other, four full yellow, the last half yellow / half grey. The last one is a stacked one which would not be shown with the examples provided. The techniques shown will change the meaning of the star rating. (four instead of four-and-a-half).

See: https://github.com/w3c/wcag21/issues/297#issuecomment-305276347

> Semantics /should/ be ok as the sub images are hidden with aria-hidden so the tree shouldn't be altered. besides role="img" makes all children presentational anyway in the accessibility tree.

See: https://github.com/w3c/wcag21/issues/297#issuecomment-305306843

And

> ... I started by saying it would be hacky ;)

See: https://github.com/w3c/wcag21/issues/297#issuecomment-305306843

Besides the fact that if we already says it's kind of hacky, at least we must communicate it this way otherwise the gut feeling from developers might be like "the WCAG working group does hacky stuff not following the standards, we will also do so..."

Correct me if I'm wrong but at https://www.w3.org/TR/wai-aria/#img it says:

>  img (role) A container for a collection of elements that form an image.

It says elements (plural) not element (single) (or should the ARIA group change this?)

And

> In order for elements with a role of img be perceivable, authors MUST provide alternative text or a label determined by the accessible name calculation.

So do we want if to be perceivable by ALL people or only some? Example: I see some icons with value worth mentioning but it won't be read aloud...

At https://www.w3.org/TR/wai-aria/#dfn-perceivable it says: Perceivable: Presentable to users in ways they can sense. Adding aria-hidden="true" to an element with role="img" feels awkward.

This is also true for nested role="img" attributed, I do understand it is possible, and also that it hacks a hacky implementation for icons, but to provide it as good practice is another thing (without mentioning)

3.

The test procedure as is:

> Tests Procedure:
> 
> For each icon created with a font check that:
> 
> The element providing the icon font has role="img"

Is not correct as you do not want this in all use cases.

jake-abma:
Also in the third example I hear "link" followed by "link heading 2 email" when using my arrow keys (2 separate links...) NVDA / Firefox combination, can anyone double check? not sure why, din't investigate...

alastc:
Hi @jake-abma,

Some good points, I think we need to both refine the technique and provide some context.

The idea is that there people who override fonts (LV and COGA), and there needs to be a standard way to say 'don't override this particular font'. If it is identified as an image, it can be ignored, e.g. `*:not([role=img])`

What we found from testing was that for both this and screenreader use, we really do want the icon to be in it's own element, e.g. a `<span>`. I don't think that is coming through in the description, we want to encourage that it is in it's own element, not using `:before` or similar to embed it in another element.

So where it says:
> The key is for the author to semantically markup icon fonts with role="img"....

Add something like: "... which means using the font in a separate element, rather than using CSS pseudo selectors and `content` to embed the icon.

NB: This is a positive technique, not a failure. We are not precluding other methods.

> Besides the fact that if we already says it's kind of hacky, at least we must communicate it this way otherwise the gut feeling from developers might be like "the WCAG working group does hacky stuff not following the standards, we will also do so..."

In the scenario where it is an icon in an element, is it hacky? It doesn't seem so. An alternative (and preferred) technique would be to use an SVG, which would be an element. This is moving towards that approach.

The first two examples are straightforward, is it the combination of `role=img` and `aria-hidden` in the third that's the problem?

> It says elements (plural) not element (single) (or should the ARIA group change this?)

I read that as a 'one or more' thing, presumably role=img is building on the HTML img, so it doesn't have to be more than one element.

Perhaps we should state this technique is aimed at scenarios where you have one meaningful icon, rather than multiple?

jake-abma:
I see 3 use cases we need to tackle, where we want to end up with the same technical implementation as an advise, the stand-alone icon with role img, accessible name and when needed hidden.

I would even prefer have the 'wrong' way be shown first and than the right solution so it will be understood better, like:

### 1. Stand-alone (favourite)

**Instead of...**

```
<span class="icon icon-favourite"></span>

.icon-favourite:before {content: "\f005";}
```

**Do...**

```
<span class="icon icon-favourite" role="img" aria-label="favourite"></span>

.icon-favourite:before {content: "\f005";}
```

### 2. Stacked / two coloured icon (star rating)

This one is the hacky one with nested role="img", accessible names and aria-hidden.

**Instead of...**

```
<span class="icon-stacked icon-stacked-half">
	<span class="icon icon-favourite-bg"></span>
	<span class="icon icon-favourite-half-grey"></span>
</span>

.icon-favourite-bg:before {content: "[font-name-bg]";}
.icon-favourite-half-grey:before {content: "[font-name-half-grey]";}
```

**Do...**

```
<span class="icon-stacked icon-stacked-half" role="img" aria-label="Half star">
	<span class="icon icon-favourite-bg" role="img" aria-label="Star bg" aria-hidden="true"></span>
	<span class="icon icon-favourite-half-grey" role="img" aria-label="Star half grey" aria-hidden="true"></span>
</span>
```

### 3. Icons as part of another (semantic) element (the link to favourites as example)

**Instead of...**

```
<a href="" class="double-icon-link">Favourites</a>

.double-icon-link:before {content: "[font-name-favourite]"; left: 8px;}
.double-icon-link:after {content: "[font-name-chevron]"; right: 8px;}
```

**Do...**

```
<a href="" class="double-icon-link">
	<span class="icon icon-favourite" role="img" aria-label="favourite" aria-hidden="true"></span>
	<span class="icon icon-chevron" role="img" aria-label="chevron" aria-hidden="true"></span>
	Favourites
</a>

.icon-favourite:before {content: "[font-name-favourite]";}
.icon-chevron:after {content: "[font-name-chevron]";}

.double-icon-link .icon-favourite {left: 8px;}
.double-icon-link .icon-chevron {right: 8px;}
```

Cheers!

lauracarlson:
Hi @jake-abma,

If you want to change the examples [per your proposal](https://github.com/w3c/wcag/pull/531#issuecomment-445523547), that  is fine with me. 

Thanks!

jake-abma:
Hi @lauracarlson, will do soon, it's on my list

lauracarlson:
Thank you @jake-abma .

jake-abma:
@lauracarlson / @alastc think we're ready here for review, please check the links in the files as these don't seem to work properly via Github...

Technique: https://github.com/w3c/wcag/blob/tech-icon-font-img-role-v2/techniques/aria/icon-font-img-role.html

Preview: https://htmlpreview.github.io/?https://github.com/w3c/wcag/blob/tech-icon-font-img-role-v2/techniques/aria/icon-font-img-role.html

Examples: https://github.com/w3c/wcag/blob/example-icon-font-img-role/working-examples/aria-icon-font-img-role/index.html

Preview: https://htmlpreview.github.io/?https://github.com/w3c/wcag/blob/example-icon-font-img-role/working-examples/aria-icon-font-img-role/index.html => Preview Example

lauracarlson:
They don't render right for me either. Maybe @alastc  has a fix?

jake-abma:
Hi @alastc have you encountered this previously? Also got a 'no render' once in a while and the script also doesn't work while when I check Windows / Edge and Firefox via codepen with same code it does work.

jake-abma:
Hi @alastc could you please take a look at this one? It's ready for review for a couple of weeks now but only the examples have some issues with github htmlpreview

Thanks!
Jake

jake-abma:
Hi @alastc , please check this one, thx!

alastc:
Just following up, there were a couple of updates to make prior to publication/merging:
https://www.w3.org/2019/05/07-ag-minutes.html#item06 

I can see a separate branch (v2), is that a better one to update and merge?

jake-abma:
@alastc v2 is the one

jake-abma:
Technique: https://github.com/w3c/wcag/blob/tech-icon-font-img-role-v2/techniques/aria/icon-font-img-role.html

Preview: https://htmlpreview.github.io/?https://github.com/w3c/wcag/blob/tech-icon-font-img-role-v2/techniques/aria/icon-font-img-role.html

Examples: https://github.com/w3c/wcag/blob/example-icon-font-img-role/working-examples/aria-icon-font-img-role/index.html

Preview: https://htmlpreview.github.io/?https://github.com/w3c/wcag/blob/example-icon-font-img-role/working-examples/aria-icon-font-img-role/index.html => Preview Example

jake-abma:
+ Also got a 'no render' once in a while and the script also doesn't work while when I check Windows / Edge and Firefox via codepen with same code it does work.

So the https://htmlpreview.github.io seems to not work properly but the script does...

alastc:
Closing this one as the best branch is V2, will merge from there.

PR #533, opened 2018-11-16
Adding a basic css file for previewing techniques/understanding

alastc:
Includes a copy of the base and technique CSS files from live.

The additions I've made are at the end, things like adding the 'Note' text at the start of notes so people don't add it themselves.


alastc:
Also cleaned up the gitignore a bit, by the by.

alastc:
Ah, I misunderstood when you said 'put it at top level', no problem.

I was just a bit nervous about adjusting the .gitignore straight in master, but understood.

PR #534, opened 2018-11-20
Removed the link.

alastc:


alastc:
Approved here: https://www.w3.org/2018/11/20-ag-minutes.html#item05

PR #535, opened 2018-11-21
First draft technique changelog for review.

alastc:
We've discussed a changelog, how is this for a location and format for techniques?


alastc:
I'll merge now so it's easy to update if we approve more techniques next week, we can change the format or remove if @awkawk doesn't agree.

Issue #529, opened 2018-11-06
New Technique draft "Making content on focus or hover hoverable, dissmissible and permanent"


detlevhfischer:
I have written a draft Sufficient Technique "Making content on focus or hover hoverable, dissmissible and permanent".
This is the Github link: 
https://github.com/w3c/wcag/blob/tech-hover-focus-hoverable/techniques/general/tech-hover-focus-hoverable.html
This is a rendered view (via raw.githack.com):
https://raw.githack.com/w3c/wcag/tech-hover-focus-hoverable/techniques/general/tech-hover-focus-hoverable.html

There is now just one example for popups offering additional info (like a preview) for links. The JS is embarrassing and should be improved by someone better versed in scripting (would be nice if it was general enough to work for any pair of consistently named link / corresponding hover content on the page).

Another example which I haven't tackled yet would be similar, but the popup content would not be nested inside the trigger but be in a separate section below (think of footnotes listed at the end of a doc which can be displayed in a popup next to the footnote in a text when hovering over it).

detlevhfischer:
There does not seem to be a consistent way how the [practical example ](http://www.3needs.org/en/testing/code/hover-focus.html) is rendered by NVDA/Firefox: first the link plus the popup text was read of firstr kb focussin, then nothing is read, then, after back and forth, just the trigger. It seems inconsistent at best (and a bit unclear what the deisirable behaviour should be - I guess best might be the renderng of the popup with some delay (as with accDescription in some UAs), bur I am not sure. Is this an anti-pattern? How can it be improved?

alastc:
Replacing PR #450 450

mraccess77:
I made an example of the technique here https://labs.levelaccess.com/index.php/Additional_Content_on_Hover_or_Focus

detlevhfischer:
Thanks, @mraccess77  - just tied it out, when I focus the trigger with a mouse it brings up th epopup but I cannot get rid of it with ESC (Win 10, Firefox).

mraccess77:
The trigger has to be focused for escape to work.  It works for me in FF with Win 10 and Win 7.

detlevhfischer:
[...]

PR #543, opened 2018-11-27
Tech hover focus hoverable

detlevhfischer:
A [rendered version](https://raw.githack.com/w3c/wcag/tech-hover-focus-hoverable/techniques/client-side-script/tech-hover-focus-hoverable.html).


detlevhfischer:
This was the example provided by @mraccess77 in the original issue (now closed) https://github.com/w3c/wcag/issues/529 :

> I made an example of the technique here https://labs.levelaccess.com/index.php/Additional_Content_on_Hover_or_Focus

Just putting it here for reference. There were also responses in that issue as to its suitability and likely changes needed.

alastc:
+1

jake-abma:
+1

Small text fixes:

1. dissmissible  => dismissible

2.
> focus the trigger with the keyboard 

can we delete "with the keyboard" here, or add other methods here as the keyboard is not the only method to put focus on a element.

detlevhfischer:
@mraccess77 ,
Many thanks for improving the code - but I think the example still needs more work (I felt from the start that it was not easy-peasy).
The problem is if you focus the trigger with the keyboard and then use the mouse to move the cursor over the popup, the popup closes. So the two input modalities are not in sync.
Detlev

detlevhfischer:
@mraccess77  I guess strictly speaking your code example is fine since the SC does not require that content on (keyboard) focus is hoverable. However, arguably it would be nice if it the additional content brought up by keyboard focus would remain visible for pointer exploration (see the issue I have raised today: https://github.com/w3c/wcag/issues/582 ).

PR #545, opened 2018-12-03
Tech failure orientation lock

awkawk:
Rendered version: http://htmlpreview.github.io/?https://github.com/w3c/wcag/blob/tech-failure-orientation-lock/techniques/general/failure-orientation-lock.html


mbgower:
I feel like the test is missing one more test, which is whether the orientation changes as a device change of orientation occurs.

brooksienoodlesoup:
It seems to me that if a user is looking to the Failure Technique as a way determining whether or not content fails because of a locked orientation, it would be important to consider whether or not the orientation is essential.  So, my suggestion would be to add another step at the beginning of the procedure, such as "Ensure that one display orientation is not essential to the information or functionality of the content."  Then, adjust the expected results to make it so that content that has an essential orientation doesn't fail.

alastc:
On @brooksienoodlesoup point, the only example of a WCAG 2.0 failure for an SC with the 'essential' exception I could find is [F16](https://www.w3.org/WAI/WCAG21/Techniques/failures/F16), and apart from the extraordinary long title, that aspect isn't mentioned.

I think (as David said on the call) that you can't refer to a technique for something which doesn't apply. At the SC level, if locking orientation is essential, it doesn't fall-through to the techniques, you get the exception at the SC level.

Perhaps it would be appropriate for the title to be "Failure due to locking the orientation to landscape or portrait view where it is not essential", and add something to the description. 

E.g. the first line could be:
> The objective of this technique is to describe how restricting the view of content to a single orientation is a failure unless it is for an essential purpose.

awkawk:
@mbgower Does the SC require that that orientation changes at any time, or would page load be sufficient?

awkawk:
@alastc it should still be in the procedure. I updated the procedure and added a sentence to the description.

@brooksienoodlesoup I updated for essential, but also expanded "control" in the last check to remind people that there may be user agent, OS, or device settings that should be checked.

awkawk:
@Ryladog I was thinking about your wall-mounted display comment from the call minutes. In that situation we are really talking about a closed system anyway and I wonder if adding that to this may confuse people?

@detlev - the reason I split the procedure is because I didn't think that we were saying that content needed to be able to rotate on the fly to orientation changes, and as a result I wanted to have "open the content" for each separate.

jake-abma:
I'm just having difficulty with the text as the sentences don't seem to be logical.
Suggest to rethink and shape the sentences a bit to be a bit more clear.

For example: 

> The objective of this technique is to describe how restricting the view of content to a single orientation is a failure to allow content to be viewed in multiple orientations.

"is to describe" can be deleted as this is what we do in a technique document.

> ... to allow content to be  viewed...

Seems like another part of another sentence mixed with the previous part.

> When content is presented in this way, 

What way are we talking about here?

>  users must view the content in the orientation that the author selected.

Also here I need to read it multiple times to see what's meant.

> with landscape and portrait display orientations

Reads a bit difficult, shouldn't it be "with landscape and portrait orientation modes"

jake-abma:
Also the procedure and expected results can be rewriting so it 'flows' better

mbgower:
I question the need to have the exception information in here. Looking at existing Failure techniques that apply to SCs where there is an exception, such as https://www.w3.org/TR/WCAG20-TECHS/failures#F54
I don't see any language about the exception in the failure technique. It seems to me that one only wants the failure technique to test failure, not the exception.
Someone can argue that a particular implementation meets the exception language, but at that point whether or not it fails is immaterial; it doesn't need to meet the SC.

mbgower:
@awkawk in regard to your https://github.com/w3c/wcag/pull/545#issuecomment-444500218 I think that the language of the SC means that testing dynamically is the simplest way to confirm the Failure.
1. You confirm switches and OS settings to lock orientation are disabled
1. You open content in one orientation and confirm content appears oriented correctly
1. You rotate to the other and confirm content responds.

If 2 or 3 are not true, the failure applies.

awkawk:
Looks like it will need to be F97

awkawk:
Thanks Michael.

Issue #540, opened 2018-11-26
Clarify in understanding whether SC 1.4.11 Non-text contrast adjacent colors requires both sides of adjacent colors or only one side have sufficient contrast


mraccess77:
Take a dark blue button with white text.  When the button is focused a black border appears around the button.  The black border has sufficient contrast with the page background of white but does not have sufficient contrast with the button background color of dark blue.

Options
Fail: While outer side has sufficient contrast this materially help the user distinguish the focus state.
Pass:  outside border is sufficient.

alastc:
Does the suggested technique in the [other issue](https://github.com/w3c/wcag/issues/541#issuecomment-443174870) cover this?

alastc:
Hi @mraccess77, can we close this one?

mraccess77:
My question applies outside of focus.  Say the button has a border.  Does the border have to conform to both the outside and inside colors?  Button is dark gray.  Border is black and background of page is white.  No states are present.  The gray contrast with the white background but the gray border does not contrast with the button's black internal background.

alastc:
> Does the border have to conform to both the outside and inside colors?

The adjacent colors section of the [understanding doc](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html) was overhauled to answer this type of question, so the question is now: does that answer it for you?

mraccess77:
Yes.  It answers my question.

alastc:
Great, good to know the updates are helping with this. Just to wrap this up then, my _proposed_ response is:

<hr>

The updated Understanding document for 1.4.11 should now cover this question, the relevant bits for this question include:


What aspects of a control need contrast?
> visual information provided that is necessary for a user to identify that a control is present and how to operate it

Boundaries are not necessarily included:
> This success criteria does not require that controls have a visual boundary indicating the hit area, but if the visual indicator of the control is the only way to identify the control, then that indicator must have sufficient contrast.

And the several visual examples in the 'Adjacent colors' section (and the examples section) should help to indicate various methods.

Issue #400, opened 2018-06-29
SC 1.4.11 contrast on hover included or not 


michael-n-cooper:
_From @goetsu on May 11, 2018 17:0_

Hi,

current text inside understanding state : 
"Regardless of the whether or not there is a visible indication of the hit area for the button, the focus indicator for the component must have sufficient contrast when the component is focused."

I think hover indicator must also have a sufficient contrast otherwise people using mouse only may not see the change of state specially on button element where cursor isn't changing to pointer by default

_Copied from original issue: w3c/wcag21#913_

michael-n-cooper:
_From @goetsu on May 11, 2018 17:22_

Also in case like this one https://www.lvmh.fr/ (click on the search icon in the header) or https://knowbility.org/ where search input has no border and transparent background having a hover effect with suffisant contrast will definitly be helpfull for everyone specially low vision user to understand those are interactive area

michael-n-cooper:
_From @detlevhfischer on May 11, 2018 18:24_

As to mouse operation, while it is desirable to have good hover focus, the fact that the user places the visible cursor over the interactive element is IMHO a sufficient indication in terms of minimal SC requirements (for mouse operation).

>

michael-n-cooper:
_From @mraccess77 on May 14, 2018 14:8_

I believe the hover state needs to have sufficient contrast as well.  Just because a point is over something doesn't mean it's the item we want to activate.  If we have to move the mouse away to then get sufficient contrast that might mean panning the magnified view.

michael-n-cooper:
_From @goetsu on May 14, 2018 18:43_

Also as I already said in some cases (button element for example) cursor isn’t changing. So if user don’t see the change of state because contrast isn’t sufficient he will may consider the element as not interactive

michael-n-cooper:
_From @detlevhfischer on May 15, 2018 5:10_

I would support the requirement, I just think the current SC text and understanding text do not make it unambiguously clear that a hover state is required. The current master understanding text has:

> This success criteria does not require that controls have a visual indicator, but that if there is an indicator, it has sufficient contrast.

[...]

Issue #13, opened 2014-02-27
Technique F47 improvement


webcc:
In the file:

https://github.com/w3c/wcag/blob/master/wcag20/sources/failure-tech-src.xml

The technique F47 should make a reference to the own documentation of Mozilla, where the use of blink is deprecated and discouraged:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/blink

joshueoconnor:
Item merged on 22/9/2014

https://github.com/w3c/wcag/pull/24

PR #550, opened 2018-12-06
Non-text contrast updates

alastc:
This update to the understanding doc for Non-Text Contrast addresses several issues raised.

[Preview the rendered understanding doc](https://cdn.staticaly.com/gh/w3c/wcag/non-text-contrast-updates/understanding/21/non-text-contrast.html?update=2), or review the [diff](https://github.com/w3c/wcag/pull/550/files#diff-3860e6fe857f7805dda94be580f919cc).

The substantive updates are in the Intent section, from the second paragraph down to the table. They are based on the following assumptions:

From 1.4.11:

- Components must continue to have contrast in different states ("required to identify").
- Where a state indicates functionality of the component (e.g. a check in a checkbox), that aspect must have sufficient contrast ("required to identify … states"). However, it does not create a requirement to differentiate states by color internally (e.g. between default and hover, visited link, visited + hover etc.). NB: A better term than "functionality" would be useful there.

In combination with Focus Visible (2.4.7) & Use of Color (1.4.1)

- if the focus indicator relies on color alone (so all three SCs!), that must provide contrast compared to the non-focused state. (E.g. changing the background color of a button.)
- if the focus indicator adds/removes parts of the component, or changes the shape of the component, that change must have contrast. (e.g. adding thickness to a border that has contrast, or adding a separate outline.) However, it doesn't have to contrast with itself if it changes the shape.

Not in scope:

- Boundaries for links / buttons. 
- Default (untouched) focus indicators.
- Measurement is against adjacent colors of the component, not different states within the component (except when color-change-only is used for focus state).
- Hover / visited state (except that the component must continue to have contrast). Whilst desirable that a component would indicate hover state in a perceivable way, it is not required by 1.4.11.

This update should address these issues:

- SC 1.4.11 contrast on hover included or not #400
- Color contrast SCs still insufficient for LV and users with color deficiencies ([WCAG21#649](https://github.com/w3c/wcag21/issues/649))
- Clarify in understanding whether SC 1.4.11 Non-text contrast adjacent colors requires both sides of adjacent colors or only one side have sufficient contrast #540
- Does SC 1.4.11 require comparing focused and non-focused states #541
- And a [webaim discussion](https://webaim.org/discussion/mail_thread?thread=9054).


allanj-uaag:
In the first section, second bullet you say
"Where a state indicates functionality of the component (e.g. a check in a checkbox),"
the check is not the functionality of the component, it is the indicator of the state. perhaps...
"The state indicator (e.g. a check in a checkbox) must have sufficient contrast  ("required to identify … states")."

alastc:
Hi @allanj-uaag,

In the understanding doc that became the second paragraph:
> This Success Criterion does not require that all states are differentiated within the component. With the exception of focus indicators, links and buttons are not required to differentiate states such as hover. Controls which convey the value of an input, such as checkboxes, sliders, and radio buttons must meet the contrast requirement for those states. For example, the tick in a checkbox must meet the contrast requirement.

We need another way of saying 'state', without using 'state', as then it gets mixed up with hover, visited etc.

I wasn't totally happy with 'value', but couldn't think of a better term.

jared-w-smith:
Below are several recommended edits on the first few sections...

- I suggest consistency between "color" (35 instances) and "colour" (10 instances). I'll presume "color" is appropriate to align with WCAG.

- "For active controls on the page..." to "For active controls, those that are not inactive or disabled, on the page...". This stems from previous discussions where "active" in this context has been incorrectly interpreted to mean the "active" state, meaning the state it is in when being activated or clicked.

- "...must have sufficient contrast with the adjacent background." to "...must have a minimum 3:1 contrast ratio with the adjacent background." It seems odd to use "sufficient contrast" in this sentence, but 3:1 in the next sentence. This could be interpreted to mean that there are different contrast requirements for these - one just "sufficient" and the other 3:1.

- "With the exception of focus indicators..." to "With the exception of keyboard focus indicators...".

- "Controls which convey the value of an input..." to "Information within controls that convey the value or state of an input...". This better clarifies that it's not just the control that needs contrast, but also information that is being indicated by that control.

- The first example with two buttons has an incorrect caption, or is out of place and has incorrect alt text. I believe the caption should read "A button without a visual boundary, and the same button with a defined visual boundary." or similar. I think the suggestion that a focus indicator is suitable as a "visual boundary" (the topic of the previous paragraph) here would be confusing.

- "Focus indicator and Use of Color" to "Focus Indicator and Use of Color" (capital "I").

- "The Use of Color success criteria..." to "The Use of Color success criterion". This should probably link to the SC, not its supporting materials.

- The sentence that starts "If a control's indicator of state..." should probably be removed. This is adequately covered in the next paragraph. This sentence introduces "state", but the rest of the paragraph is an example of only Use of Color for links, so this adds a point of potential confusion.

- "Where links or buttons have been styled (including the focus indicator), these must meet the 3:1 contrast ratio." What does "these" refer to? If it's a text link, then this statement is not applicable to the text (assuming it's not "large" text). I think this should be scoped to "Focus indicators that have been styled by authors must meet the 3:1 contrast ratio."

- "Where a text-input has an indicator, such as a bottom border, that indicator it must meet 3:1 contrast ratio." to "Where a text input has a visual boundary, such as a bottom border, that indicator must meet the 3:1 contrast ratio." "Indicator" is a bit ambiguous and could be confused with "focus indicator", whereas "boundary" has already been introduced previously. Note the removal of the extraneous "it".

- "Where a text-input has an indicator, such as a bottom border, that indicator it must meet 3:1 contrast ratio." to "Where a text input has a visual boundary, such as a border, that boundary must meet the 3:1 contrast ratio." As above, plus removal of extraneous "bottom".

- "Adding a focus indicator is required, and must meet 3:1 contrast ratio." to "A focus indicator is required. If the focus indicator is styled by the author, it must meet the 3:1 contrast ratio." The removes the suggestion that authors must "add" a focus indicator (it's there by default) and properly scopes this to styled focus indicators (default ones are exempt).

- "Text input using a different in background colour to indicate the input, no border is required to indicate the input box." I'm not entirely sure what this means. I think the following is the intent: "Text inputs that have no border and are differentiated only by a background color must have a 3:1 contrast ratio to that background".

- "Where an author-created focus style is applied to a dark background, it should have 3:1 contrast with the surroundings and/or input." The focus style isn't applied to the background. I think the following conveys this concept: "When controls are differentiated only by a background color, focus indicators for those controls must have a 3:1 contrast ratio to their background and/or the control." Note the change from "should" to "must".

- I find the "Inactive User Interface Components" to be extremely lengthy and is mostly details that are irrelevant to authors, but mostly notes written for working group members. This could be condensed to the following:

> User Interface Components that are not available for user interaction (e.g., a disabled control in HTML) are not required to meet contrast requirements in WCAG 2.1. An inactive user interface component is visible, but not currently operable. An example is a submit button at the bottom of a form that is visible, but cannot be activated until all the required fields in the form are completed. Because inactive controls are inherently presented with low contrast, and thus may introduce confusion or difficulty, consideration should be given to the use of disabled controls.

mraccess77:
Thanks for the updates Jared and Alastair.   I have two additional things we should attempt to address.
* Explanation of what adjacent means.  Does it require all adjacent colors or just enough adjacent colors.  For example, does a border focus indicator need to contrast with the page background and button background or only one is sufficient.
* The selected and non-selected states of tabs and listbox items should be included along with focus and non-focused states since they are part of the same control and adjacent.  We should make this clear.

alastc:
Thank you @jared-w-smith, 

Re: color / colour, Doh! that's my muscle memory kicking in, color is the correct spelling here.

I've integrated almost all of that, a couple of notes on the the things not integrated:

> Focus indicator and Use of Color

SCs use title-case, and I realised I was trying to convey the related SCs, so I've updated that to "Focus Visible and Use of Color", matching the order of the content.

> Where a text-input has an indicator, such as a bottom border...

I would like to phrase it in a way that says: Whatever the visual indicator is, that's what should have contrast, so I've updated that to: "Where a text-input has a visual indicator that it is an input, such as a bottom border, that indicator must meet 3:1 contrast ratio."

> Text inputs that have no border and are differentiated only by a background color must have a 3:1 contrast ratio to the adjacent background

Now: "Text inputs that have no border and are differentiated only by a background color must have a 3:1 contrast ratio to the adjacent background."

> Where an author-created focus style is applied to a control on a dark background, it must have 3:1 contrast with the adjacent background

Now: "Where an author-created focus style is applied to a control on a dark background, it must have 3:1 contrast with the adjacent dark background."

> I find the "Inactive User Interface Components" to be extremely lengthy 

Yes, and I agree, but it is kind of an FAQ, we keep getting asked about that. I'd like to find somewhere else to put it before removing it.

(I'll respond to Jon's later, my train is pulling in.)

jared-w-smith:
Nice work on this! Here are a few thoughts:

A couple "colours" made their way in again. :-)

> For user interface components 'adjacent colors' means the colors adjacent to the component, rather than within the component. For example, if an input has a white internal background, dark border, and white external background the 'adjacent color' to the component would be the white external background.

This example isn't always accurate. This suggests that Text box 1 at https://webaim.org/temp/1-4-11examples.htm would fail in its default state. Because it has sufficient contrast to its border, this is sufficient.

Also consider Text box 2. In this case the border has insufficient contrast to the text box, but sufficient contrast to the background. So this would also pass.

Additionally, this statement would not be true for an input without a border at all, but on a different colored background. In this case the color within the input and the background color must sufficiently contrast.

So, sometimes we measure internal to border, sometimes we measure border to external, and sometimes we measure internal to external. As such, I think we need to be careful about defining this as "internal" vs. "external", but should instead focus on what is needed to identify the component.

Perhaps the following:

> For user interface components, 'adjacent colors' means the colors adjacent to whatever is needed to visually identify the component itself. For a text box with a border, the border is used to identify the component, so it must contrast with its adjacent color, either the text box background color or with its surrounding background color. For a text box without a border, the input background color is used to identify the component, so this must contrast with the background. The tick in a checkbox must contrast to the checkbox background color. 

You wrote:

> That style of outline must contrast with the background, but it is not required to contrast with the component.

As above, I don't believe this is always accurate. Consider button 12 at https://webaim.org/temp/1-4-11examples.htm The focus indicator in this case only has >3:1 contrast with the internal button color (and also >3:1 from its default state), but not the background.

Perhaps change to simply:
> That style of outline must contrast with its adjacent colors.

The SC requires that it be visible - considered to be 3:1 to adjacent colors (which isn't necessarily background/external) - so I think this is more accurate.

Wow, this is tough stuff to document!

alastc:
Ach, I consciously tried to avoid colour, but my muscle memory is strong... despite years of CSS.

> This suggests that Text box 1 at https://webaim.org/temp/1-4-11examples.htm would fail in its default state. Because it has sufficient contrast to its border, this is sufficient.

There is a concept that is easy to say but hard to convey: If you couldn't see the non-contrasting bit, would it still look like the thing it is intended to be?

Text box 1 is a good example, if you ignore the border (which doesn't contrast with the white background) it has still got plenty of contrast for the input.

I'll start with your suggested text, but I don't think it clearly covers the textbox 1 example either, we need to include something about ignoring a border if that isn't required to discern the control due to the background.

> Consider button 12 at https://webaim.org/temp/1-4-11examples.htm The focus indicator in this case only has >3:1 contrast with the internal button color (and also >3:1 from its default state), but not the background.

I consider that another example of 'changing the shape', it is conceptually the reverse of the 2nd button in the set of 4 just above "Use of Color and Focus Visible". Adding the grey border _inside_ the boundary of the button means the button changes shape as the grey _doesn't_ contrast with the white background. It effectively reduces the size of the dark internal background so changes shape.

If we use simply:
> That style of outline must contrast with its adjacent colors.

I think that opens up more questions: adjacent to internal and external? External to the component?

I'll keep thinking...

alastc:
Hi @jared-w-smith, and couple of updates to try and make things clearer from your last comments, [diff of just those](https://github.com/w3c/wcag/pull/550/commits/a8b016e370e33458df2c1ffdf1309ab771f6ba35), and the [preview](http://htmlpreview.github.io/?https://github.com/w3c/wcag/blob/non-text-contrast-updates/understanding/21/non-text-contrast.html) should have updated.

alastc:
Hi @mraccess77, I'm not sure how that would be clearer? If you look at the examples it matters whether you are trying to contrast with the internal and/or external colors. Otherwise adding a light grey outline around a dark control won't help.

jared-w-smith:
> If you look at the examples it matters whether you are trying to contrast with the internal and/or external colors.

Right. What we're suggesting is that it will be very difficult to define the conditions where one or the other is used. It would be better, I think, to allow the author to determine what is needed to discern the control (or its state), then ensure 3:1 contrast to is adjacent colors - whatever that may be (internal color, border, or external color)

> Otherwise adding a light grey outline around a dark control won't help.

Sure it will - if it has >3:1 contrast.

alastc:
> Sure it will - if it has >3:1 contrast.

With what though?  It won't (in reasonable cases) be able to contrast with both.

Sorry if I'm missing something, I'm just confused about 'any adjacent color' making it clearer.

Compare examples 4 & 5, and 12 & 13 from: https://alastairc.ac/tests/wcag21-examples/ntc-focus-styles.html 

How do you word that to pass 4 & 12, but fail 5 & 13?

My suggestion is that sufficient techniques include:
- Adding a border/outline that wasn't there in the default presentation, which can contrast with component and/or adjacent background. (examples 3, 6, 7, 9)
- Changing the shape of the component by adding/removing something, either:
  - Adding a border/outline that increases the size of the component (example 4), i.e. does not contrast with the component, but does with the adjacent (to the component) colors.
  - Adding a border/outline inside the component that shrinks it (example 12), i.e. does contrast with the component, not with the adjacent background.
- Changing a color of the component that contrasts with itself (example 10 passes, 11 fails).

jared-w-smith:
> > Consider button 12 at https://webaim.org/temp/1-4-11examples.htm The focus indicator in this case only has >3:1 contrast with the internal button color (and also >3:1 from its default state), but not the background.
> 
> I consider that another example of 'changing the shape', it is conceptually the reverse of the 2nd button in the set of 4 just above "Use of Color and Focus Visible". Adding the grey border _inside_ the boundary of the button means the button changes shape as the grey _doesn't_ contrast with the white background. It effectively reduces the size of the dark internal background so changes shape.

Please note that the border is present in the default state, but is the same as the internal color. It changes color on focus. I don't believe a border color change alone is 'changing the shape' - unless there is also a 3:1 contrast (as is the case with your examples). Otherwise, this is problematic...

Consider my new Button 13 at https://webaim.org/temp/1-4-11examples.htm It is nearly identical to Button 12, except that the border style is a bit thinner and the color on focus does not provide >3:1 to either the button color OR to the background OR to the default border color. If you consider the border color change alone to be "changing the shape", then this would pass, despite it being nearly impossible to determine the focus state.

I think that exceptions will always exist any time you try to dictate the specific the conditions for contrast between internal vs border vs external. As above, the key is to simply identify what is needed to differentiate the control or state (in this case the border change on focus) and ensure >3:1 to the adjacent colors (internal OR external would work), and, if it's color-reliant indicator of state, a >3:1 change from the default state.

alastc:
> Consider my new Button 13 at https://webaim.org/temp/1-4-11examples.htm It is nearly identical to Button 12, except that the border style is a bit thinner and the color on focus does not provide >3:1 to either the button color OR to the background OR to the default border color. If you consider the border color change alone to be "changing the shape", then this would pass, despite it being nearly impossible to determine the focus state.

That's very similar to the example 13 on my page. 

No, I'm not saying that passes. For it to be a change of shape, it must contrast with either the component or the background.

If you look at the sub-bullets from my last comment:

- Shrinking = inside the component change that conrtasts with itself. 
- Expanding = outside the component and contrasting with outside background.

> and ensure >3:1 to the adjacent colors (internal OR external would work),

I don't think internal or external does work, otherwise examples 5 & 13 would pass.

-Alastair

jared-w-smith:
> How do you word that to pass 4 & 12, but fail 5 & 13?

> I don't think internal or external does work, otherwise examples 5 & 13 would pass.

These would both pass if they had >3:1 contrast to the default state, as I've indicated is necessary. In #5, the #ccc border is <3:1 to the white that was at this location before focus. In #13, the dark grey on focus is <3:1 to the default black.

You've already indicated that color-reliant focus indicators (such as a button changing color) must have >3:1 from the default. Why would this be required of a button color change on focus, but not a button border color change on focus?

Yes, this can be constraining in certain designs, but this is a constraint the author has invoked by not having a standard focus outline (which, if customized, must also have >3:1 contrast to the background).

Or maybe we're still misunderstanding each other here.

alastc:
> is <3:1 to the white that was at this location before focus

Ah, ok, I see what you mean now, that’s another way of describing this aspect of contrast. What colour was **replaced**, and does that have contrast with what was there before?

I had started with a similar explanation when talking it through with a developer at gov.uk (check out their focus styles, there are some interesting examples).

They had struggled with that explanation, at least when considering their examples, and found the ‘change of shape’ concept easier to understand. 

I don’t have a preference, the effect is the same, I’ll see what impact that change would have on the doc tomorrow and try and get some other perspectives.

alastc:
Alternative explanation created in #557.

understanding/21/non-text-contrast.html
```diff
@@ -20,30 +20,37 @@ <h3>Active User Interface Components</h3>
...
+
+				<figure>
+						<img alt="Two buttons, the first has a blue outline and inner white background with blue text. The second has a blue background with white text, reversing the color scheme." src="img/button-example2.png" /> 
+						<figcaption>An active control with a visual boundary, and a focus (or other) state that is highlighted with a contrasting change of color</figcaption>
+				</figure>
+
+				<p>Another method to provide a visual focus indicator that requires contrast is changing the shape of the control. For example, if a button has a dark border the focus states could add a dark outline, effectively changing the shape of the component. This method does not require contrast with both the component and the background, because the shape of the component has changed.</p>
```
awkawk: perhaps "shape or size"?

understanding/21/non-text-contrast.html
```diff
@@ -20,30 +20,37 @@ <h3>Active User Interface Components</h3>
...
+
 				<figure>
-					<img alt="Two buttons, the first has a blue outline and inner white background with blue text. The second has a blue background with white text, reversing the color scheme." src="img/button-example2.png" /> 
-					<figcaption>An active control with a visual boundary, and a focus (or other) state that is highlighted with a contrasting change of color</figcaption>
+						<!-- example from https://getbootstrap.com/docs/4.1/components/buttons/ -->
+						<img alt="Two buttons on a white background. The first has a blue border to indicate its boundary, white internal background and blue text. The second adds a thick blue-grey outer border to show focus." src="img/first-button-example.png" /> 
+						<figcaption>An active control showing the visual boundary by default, and the focus indicator which changes the shape of the component</figcaption>
```
awkawk: in this case, it changes the apparent size of the component, not the shape.

understanding/21/non-text-contrast.html
```diff
@@ -20,30 +20,37 @@ <h3>Active User Interface Components</h3>
...
         
-				<p class="note">In cases where the state of a user interface component is apparent without relying on visual information, this Success Criterion does not require that visual information meet the 3:1 contrast ratio. For example, the Active state on a link does not rely on the visual information because the user is actively clicking on the component at the same time.</p>
+				<p>This Success Criterion does not require that all states are differentiated within the component. With the exception of focus indicators, links and buttons are not required to differentiate states such as hover. Controls which convey the value of an input, such as checkboxes, sliders, and radio buttons must meet the contrast requirement for those states. For example, the tick in a checkbox must meet the contrast requirement.</p>
```
awkawk: Perhaps "This Success Criterion does not require that all states are differentiated within the component - only where the visual information is required to identify the component or the state. "

understanding/21/non-text-contrast.html
```diff
@@ -20,30 +20,37 @@ <h3>Active User Interface Components</h3>
...
-					<figcaption>An active control with a visual boundary, and a focus (or other) state that is highlighted with a contrasting change of color</figcaption>
+						<!-- example from https://getbootstrap.com/docs/4.1/components/buttons/ -->
+						<img alt="Two buttons on a white background. The first has a blue border to indicate its boundary, white internal background and blue text. The second adds a thick blue-grey outer border to show focus." src="img/first-button-example.png" /> 
+						<figcaption>An active control showing the visual boundary by default, and the focus indicator which changes the shape of the component</figcaption>
 				</figure>
 
+
 				<section>
 					<h4>Active User Interface Component Examples</h4>
 					<p>For designing focus indicators, selection indicators and user interface components that need to be perceived clearly, the following are examples that have sufficient contrast.</p>
```
awkawk: We don't provide any examples of selection, seems like a gap

alastc:
Updated baseds on the [last call](https://www.w3.org/2018/12/18-ag-minutes.html#item03) and [survey](https://www.w3.org/2002/09/wbs/35422/technique-approvals3/results):

- Updated the text to show that contrast between states (including functional) is not required, but that functional ones are likely to be covered in future.
- Removed the focus indicator examples. They are still valid, but should be added as techniques under focus-visible.
- Added a "to be added" technique at the bottom of the focus-visible understanding doc.

jared-w-smith:
> Non-text information within controls that convey the value or state of an input, such as checkboxes, sliders, and radio buttons are not currently in the scope of this success criteria.

This sentence needs much clarification or should be removed. I believe what you're trying to convey is fully covered by the previous paragraph - that various states don't have to contrast between themselves. As written, this sentence directly conflicts with the SC itself and statements in both of the previous paragraphs. It essentially exempts both "User interface components" and "States" - generally rendering this critical part of the SC moot. I don't believe this is the working group's intent.

Similarly, the checkbox example should not have been removed. This example was to highlight that the check itself still requires 3:1 to the background. This is not merely a differentiation between states, but is a critical part of the user interface component itself.

While contrast differences between states are not covered, I still think it would be useful for this page to indicate that these _should_ have sufficient contrast for optimal usability and accessibility. A toggle button example would be very good to demonstrate this - the states of the toggle aren't required by WCAG to have 3:1 contrast between themselves (only to the adjacent colors), but sufficient contrast between the states is still necessary for the button to be usable.

alastc:
Hi Jared, 

A couple of bits of discussion to highlight:
https://lists.w3.org/Archives/Public/w3c-wai-gl/2018OctDec/0177.html

Essentially: If states such as hover do not have to differentiate within the component, then the same SC text can't logically require functional states (e.g. checkboxes) to differentiate either.

There is a proposed new SC in #559 to address that.

How about:
> Non-text information within controls that convey the value or state of an input, such as checkboxes, sliders, and radio buttons are not currently in the scope of this success criteria. These parts of a control should provide good contrast and this gap is likely to be addressed in a future version of the guidelines.

mraccess77:
We need to be careful here with the proposed wording  I believe the individual states are covered within the control-- just not compared to other states.  So for example, a checkbox checkmark has to have sufficient contrast when examined -- but not compared to the non-checked state.  So a selected state would need to sufficient contrast -- but not the selected state against the non-selected.  The comparison  bit unfortunately will have to wait until a future version based on how the SC was written.

mraccess77:
To be clear I am in agreement with Jared that.

alastc:
> But the user interface component itself must still have sufficient contrast to the adjacent colors (e.g., background) when in the hover state (or any other state - except inactive). 

Agreed.

> Similarly, a check in a checkbox must have sufficient contrast to the adjacent colors. Again, the check (or dot in a radio button or handle to a slider) isn't merely a state differentiator - it is part of (the most critical part of) the component itself, so must have 3:1 to the adjacent colors.

That's the bit we can't seem to square. From the SC text we don't differentiate states (i.e. hover and checked are treated the same, as states), and some of these states may not have adjacent colors. 

For example, a radio button could be a black circle on white. The selected state might fill in the black circle with a black background. That is clear, and the component still contrasts with it's adjacent colors, but the 'state' does not change from that point of view.

Another example from Jake was a nav has a selected state (brown bar at the top) and hover (grey bar at the top): https://lists.w3.org/Archives/Public/w3c-wai-gl/2018OctDec/0193.html 

> This note instead states that "Non-text information within controls" are not covered by this SC at all. 

**Within,** as in a check inside a checkbox, not the checkbox. The visual info to ID the component (in any state) must be maintained across states, but not inside the component.

The paragraph above is the blanket statement, then the note is about specific states that we would like to cover but (as per the paragraph above) are not currently.

How about: 
> Non-text information inside controls that convey a value such as checks in a checkbox are not currently in the scope of this success criteria. This gap is likely to be addressed in a future version of the guidelines.

awkawk:
I'm confused by the new language as well. 

I think that the first paragraph under "Active User Interface Components" is very clear, and indicates that visual information that is used to identify the control in a state must meet 3:1. For example, a checkbox on a white background which has a black border and a light pink check would not meet the criterion, but if the check was dark in color it would.

> For active controls that are not inactive or disabled, such as buttons and form fields: any visual information provided that is necessary for a user to identify that a control is present and how to operate it must have a minimum 3:1 contrast ratio with the adjacent colors. Also, any visual effects implemented which are necessary to indicate state, such as whether a component is selected or focused must also ensure that the information used to identify the control in that state has a minimum 3:1 contrast ratio.

How about then saying (and removing the two following paragraphs):
This Success Criterion does not require that changes in color that differentiate between states of an individual component meet the 3:1 contrast ratio. For example, a text hyperlink may be blue initially and red after it is visited. The blue and red colors need to meet the 3:1 contrast ratio against the link background, but the blue and red colors are not required to meet the 3:1 against each other. The Working Group intends to address the contrast between states in a future revision of the guidelines.

Does that help, or confuse things further?

awkawk:
I don't think I ignored visited in my example - it needs to meet 3:1. 

The pink check mark fails because you can't identify the state of the control based on the visual information since the check is integral to that understanding.

alastc:
But the SC doesn't differentiate hover / visited / checked, the color changes for the link do not work to differentiate default from visited.

Based on the SC text why is the check integral to understanding but the visited state is not?

mraccess77:
@alastc in this case the visited state has to have 3:1 just as the unvisited state has to have 3:1 but you don't need to have sufficient contrast between the visited and non-visited states just like you don't have to have sufficient contrast been the checked and unchecked states.

awkawk:
I think that my link example isn't great since that actually probably falls under 1.4.1 Use of Color. I'm not sure that people have failed links/visited links on that basis.

For a checkbox, the non-text elements would be the border and the check, and 1.4.11 requires that the visual information required to identify the component and required to identify the state has a 3:1 contrast ratio against adjacent colors. I think that we are agreed that adjacent means "next to" in space rather than in time, so if the check and the border are needed to identify the component and the state, then the check and the border need to have a 3:1 ratio against the background.

If the checkbox is done like the Material style (https://material.io/design/components/selection-controls.html) then in the unchecked state the dark gray border needs to meet 3:1 against the background, and in the checked state the edge of the fill defines the component and the lack of fill defines the check, so it is the purple:white contrast comparison.

If you have a star rating widget you might have one where the stars are the same color if filled or not filled (https://image.shutterstock.com/z/stock-vector--star-rating-vector-icon-flat-design-best-vector-star-rating-illustration-778439206.jpg) or with different colors (https://cdn2.vectorstock.com/i/1000x1000/38/16/5-star-rating-symbol-design-vector-21353816.jpg) - in this case I think that the all black one would pass but the other wouldn't because the not selected stars don't have good contrast against the navy colored background.

awkawk:
Here's a different version of my added paragraph that doesn't use text links:

This Success Criterion does not require that changes in color that differentiate between states of an individual component meet the 3:1 contrast ratio. For example, a rating component with five stars may have 3 blue and 2 gray filled stars to indicate a rating of three. The blue and gray colors each need to meet the 3:1 contrast ratio against the star's background, but the blue and gray colors are not required to meet the 3:1 against each other. The Working Group intends to investigate ways to address the contrast between states in a future revision of the guidelines.

jared-w-smith:
I agree with @awkawk's examples and summary, and his updated paragraph.

To take Andrew's star example (https://image.shutterstock.com/z/stock-vector--star-rating-vector-icon-flat-design-best-vector-star-rating-illustration-778439206.jpg) a step further, if the stars changed appearance (say to grey) on hover, the grey would not have to contrast with any other color of star, but only with the background white while in that hover state. It may thus be difficult to know it is hovered, but the star is still perceivable.

Another example would be a menu button that looks like a link, but presents a blue drop-down carat on hover to indicate that it's a drop-down menu button. The blue carat only appears in the hover state, but would still be covered by this SC and must be 3:1 to the background because it is necessary to discern the component in that particular state. If the carat changes color to red when the menu is in the open state, the red does not have to contrast with the blue, but the red does still have to contrast to the background.

This does leave out some use cases that could be impactful, such as differences between pressed and non-pressed states of toggle buttons. Or color changes for a round dot that changes from green to yellow to red to indicate number of remaining characters available in a tweet, for example. Each of the colors of these components must be 3:1 to their surrounding colors, but not between each other, which could cause usability issues. These issues could perhaps be addressed by the new proposed SC.

alastc:
@awkawk:
> visual information that is used to identify the control in a state must meet 3:1

If that is the case, how do we square that with *not* applying to hover/visited etc? Do you mean the change creates a 'change of form'? E.g. a text link changing color doesn't change it's form, but a check being added does change it's form/shape?

@jared:
> You are conflating "components that have states" and "states of components", and incorrectly throwing both out. The "components that have states" (the link or checkbox) is in scope, but differences between "states of components" (blue vs. red or checked vs. non-checked) are not.

Ok, I'm agreeing that components should maintain contrast through different states (e.g. visited link needs contrast, although that would be under text in most cases), but struggling to see how that works with the checkbox example.

> Because the check becomes part of the component in that state. And the component must have 3:1 to surrounding color. 

But the component (the whole checkbox does maintain it's contrast to the surrounding color, even with a pink on white check. 

Also, by that logic this ['Modern' radio button](https://www.npmjs.com/package/modern-radio-buttons) example would fail then, as the check doesn't contrast to 'adjacent' colours?
checkbox example with a solid filled centre

You are assuming that there are adjacent colours, and that isn't always the case.  How do we square that with the SC text?

@awkawk:
> If you have a star rating widget you might have one where the stars are the same color if filled or not filled, or with different colors - in this case I think that the all black one would pass but the other wouldn't because the not selected stars don't have good contrast against the navy colored background.

I agree that the solid color ones are good, but how do you measure the 'adjacent colour' for the state?
examples of star rating where the stars fill in

However this is an interesting example:
Star rating with a solid blue background that's part of the component, and filled with yellow or dark grey

The component (dark blue) has good contrast with it's surroundings, and it is conveying the same information as the other one, but it has 3 colours being used.

If you assume the dark-grey / blue doesn't contrast (2.8:1) and can't be differentiated, it is still conveying the same information as the solid example. I.e. empty or full (yellow). 

If the  grey and blue and yellow have to contrast with each other we are making it unfeasible, as it is a three-way contrast situation.

From the proposed paragraph:
>  The blue and gray colors each need to meet the 3:1 contrast ratio against the star's background, but the blue and gray colors are not required to meet the 3:1 against each other.

That would be relying on color then, probably not a good example? I also think that would lead to a worse situation as we'd be encouraging people not to use contrast to differentiate states.

I understand (and like) the aim of saying: If something switches colour (like text links) then the change isn't covered except to maintain contrast. But if the form changes (e.g. checked, or the menu indicator Jared outlined) then that is covered. 

However, I'm struggling to square that with the SC text (& definition of states).

awkawk:
> You are assuming that there are adjacent colours, and that isn't always the case. How do we square that with the SC text?

I think that this is the same as the thinking around gradients. The evaluator needs to ask what adjacencies are relevant to understand the graphic enough to identify the state. For a checkbox the contrasts inside the bounding box are relevant and need to be taken into account. 

For this:
image

The stars have sufficient contrast in both states, so they pass. If either of these were used as icons for some function (e.g. on a button to open some sort of rating function dialog) we would pass either, right?

For this:
image

This one is interesting because I could try to make an argument that the grey/green is relevant or irrelevant depending on the page background that they are placed on. The yellow is always compared to the green, and passes. The grey effectively disappears into the green, but since it is supposed to show and empty star or no star that can be ok as long as the user is able to determine that there is a set of rating stars to work with - and that is the problem, because the star shape that provides the clue as to the type of control lacks sufficient contrast.

The way that this control might need to be adjusted to conform is to include a yellow outline around the grey (empty) stars.

alastc:
> The stars have sufficient contrast in both states, so they pass. If either of these were used as icons for some function (e.g. on a button to open some sort of rating function dialog) we would pass either, right?

Of those options yes I agree, but if there is a (arguably superfluous) hover effect that added a mid grey (non-contrasting) state, would that cause a fail? I'm struggling with the scope, .

> This one is interesting because I could try to make an argument that the grey/green is relevant or irrelevant depending on the page background that they are placed on.

Indeed, the white background helps here. I generally see these graphics as as layers of light/dark contrast, so the dark border 'houses' the light yellow stars, and the grey is essentially irreverent.

Of these variations, it's the white one (3rd) I would fail based on lack of differentiation from the others:
variations on stars filled with grey/white/and outline

awkawk:
> Of these variations, it's the white one (3rd) I would fail based on lack of differentiation from the others:

But the differentiation with the others is what we are saying isn't covered, aren't we?

alastc:
So on the call we got to a point where we agreed (I think) that components maintain contrast, and we can require that graphical indicators (but not *changes in color*) within components need to contrast.

How about this as the content of the "Active user interface components" bit:

> For active controls that are not inactive or disabled, such as buttons and form fields: any visual information provided that is necessary for a user to identify that a control is present and how to operate it must have a minimum 3:1 contrast ratio with the adjacent colors. Also, any visual effects implemented which are necessary to indicate state, such as whether a component is selected or focused must also ensure that the information used to identify the control in that state has a minimum 3:1 contrast ratio.
>
> This Success Criterion does not require that changes in color that differentiate between states of an individual component meet the 3:1 contrast ratio when they do not appear next to each other. For example, there is not a new requirement that visited links contrast with the default color, or that mouse hover indicators contrast with the default state. However, the component must not loose contrast with the adjacent colors, and non-text indicators such as the check in a checkbox, or an arrow graphic indicating a menu is selected or open must have sufficient contrast to the adjacent colors.
>
> Note: Non-text information within controls that convey the value or state of an input, such as a 1-5 star indicator with a black outline for each star filled with either yellow (full) or white (empty) is likely to fail <a href="../20/use-of-color.html">Use of color</a>, rather than Non-text Contrast.

I'm going to update the 'notes' page of examples to double-check we're all happy with the outcome of that text for various examples, I'll add the ones above.

alastc:
I've added example 21-23 based on the discussions, and udpated the pass/fail 'decusion tree' here: https://alastairc.uk/tests/wcag21-examples/non-text-contrast.html#examples

alastc:
Here's the <a href="https://cdn.staticaly.com/gh/w3c/wcag/non-text-contrast-updates/understanding/21/non-text-contrast.html?x=7">version to be published soon</a>, any other comments/updates will go into the next version.

alastc:
Understood. Those examples have been there for a whilte by the way, not new in this update!

PR #319, opened 2018-05-31
List skip-link and landmark techniques as related to each other

lkmorlan:
Skip-links and landmarks have overlapping use cases. Their techniques pages should be shown as related to each other.

PR #552, opened 2018-12-07
List skip-link and landmark techniques as related to each other

michael-n-cooper:
This is a redo of #319 which was against the 2.0 XML sources. Now that
the related techniques have been ported, it is possible to make these
changes on the 2.1 HTML sources. @lkmorlan


lkmorlan:
Thanks

PR #554, opened 2018-12-10
Removing duplicate TOC H2 in understanding index

alastc:
To align with techniques, looks like the one in the index.html should be removed, there is one in the TOC file.


michael-n-cooper:
Closing in favor of a93895ddfcd1e9a5faf56dde19110f022a46b621 which fixes this a bit more solidly.

Issue #505, opened 2018-10-06
Broken Link in F89


kerstinp:
Just a small issue: the link "Accessible Name Calculation" in F89 (https://www.w3.org/TR/WCAG20-TECHS/F89.html) is broken and gives a 404.

alastc:
Thank you, doing a search in the code there were 5 files referencing that link, so PR #505 updates them.

awkawk:
Closing as addressed in #555

PR #555, opened 2018-12-10
Updating accname link references.

alastc:
From issue #505 I checked for other references to the broken accname link, and updated them to https://www.w3.org/TR/accname/

In ARIA7 there was a slight rewording in a paragraph, but not major.


alastc:
Accepted on [the call](https://www.w3.org/2019/01/08-ag-minutes.html#item07).

PR #557, opened 2018-12-11
Alternative method of explaining focus states.

alastc:
This is an alternative explanation for #550 for discussion, it is a change to PR #550.

[Preview](http://htmlpreview.github.io/?https://github.com/w3c/wcag/blob/26d63f851bb3b9023fc554d16acb125f4435e4c1/understanding/21/non-text-contrast.html), the changes are under the ‘states’ heading.

### DIFF
--- a/understanding/21/non-text-contrast.html
+++ b/understanding/21/non-text-contrast.html
@@ -52,13 +52,6 @@
 					<figcaption>A checkbox showing a contrasting state indicator for a check.</figcaption>
 				</figure>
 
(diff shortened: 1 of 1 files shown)


detlevhfischer:
One problem I see ist that in the text here, there is not a clear differentiation between the two kinds of states: those intrinsic ones that we have provisionally called „functional“ and that result from interaction (a checkbox gets checked), and those that convey operational affordances such as ‚focused‘.
I‘ll try to find out what terms have been used for these different meanings of state in the literature.

alastc:
Hi Detlev,

This particular PR/ammendment is only about focus styles, so I don't think it matters about the functional states aspect, as the SC in question is really focus-visible.

However, this comment probably applies to the other changes we've been emailing about on the list, and #559 looks like a good start.

-Alastair

PR #565, opened 2018-12-18
updates in response to MBgower comments

awkawk:


alastc:
I think these were integrated with the last NTC update (but conflicted so I didn't use it directly.)

Closing for cleanliness...

PR #568, opened 2018-12-19
Removed G139 from Situation B techniques for Status Messages Understanding doc

mbgower:
G139 is about Creating a mechanism that allows users to jump to errors. For pratical purposes it will almost invariably involve a change in focus (otherwise the user would need to hunt for the mechanism) or else will appear on page reload (following an attemtp at submitting a form). As such, it is not the most appropriate technique to cite in Status Messages, and I think the Understanding document is better without it.


mbgower:
@michael-n-cooper. I'm not sure how your techniques are cross-referenced with Understanding docs, but obviously with the removal of G193 from 4.1.3, the technique itself should no longer reference 4.1.3.

michael-n-cooper:
Understanding docs linking to techniques define the associations. Any mention in the technique (in the meta section) is stripped by the generator. There shouldn't be mentions in other parts of the technique in any case.

awkawk:
merged per 1/29/19 call and survey

PR #575, opened 2019-01-05
Fix typos in 'The relation of Reflow to... Resize Text'

mitchellevan:
Clarify that 1.4.4 does not require resizing text to 200% _larger than_ the default, but only to 200% _of_ the default.

### DIFF
--- a/understanding/21/reflow.html
+++ b/understanding/21/reflow.html
@@ -55,7 +55,7 @@
     	</section>
       <section>
  <h3>The relation of Reflow to the Success Criterion 1.4.4 Resize Text</h3>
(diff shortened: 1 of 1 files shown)


mraccess77:
@mitchellevan can you explain the difference between the two in your understanding?"Clarify that 1.4.4 does not require resizing text to 200% larger than the default, but only to 200% of the default."

mitchellevan:
Yeah, my comment is pretty confusing without the markdown. Here's what I
was trying to say. It's a grammatical issue. The number 11 is 10% "larger
than" 10. The number 11 is 110% "of" 10. Similarly, SC 1.4.4 does not
require text to be "200% larger" -- that would be 300% of the original size.

On Fri, Jan 4, 2019, 5:39 PM Jonathan Avila <notifications@github.com>
wrote:

> @mitchellevan <https://github.com/mitchellevan> can you explain the
> difference between the two in your understanding?"Clarify that 1.4.4 does
> not require resizing text to 200% larger than the default, but only to 200%
> of the default."
>
-- 

Mitchell Evan
mtchllvn@gmail.com
+1 (510) 375-6104

alastc:
It's a good point about the large/of the default, the only bit I wasn't sure about was:

> size of all text to at least 200% while simultaneously meeting the reflow requirement.

That makes it look like you have to have 200% text plus 400% reflow. Well, it doesn't say that per-se, but I can see people taking it that way and panicking about 600% text size.

However, the rest of the text is quite exact about it, so shouldn't be a problem.

PR #576, opened 2019-01-07
Technique - Boundary contrast

alastc:
A new technique to show how to create and test graphics with sufficient contrast between areas from Rachael.

[Preview the technique](https://cdn.staticaly.com/gh/w3c/wcag/tech-color-boundaries/techniques/general/boundary-color-contrast.html?v=4).


detlevhfischer:
Looks good - I think the test needs some work.
Currently,  #3 packs contrast mesurement of areas without border and with border into the same bullet point. The first two bullet points therefore seem slightly superfluous.
With regard to content, I wonder whether one can relax border-to-field contrast requirements when the adjacent field has more than 3:1 contrast. I.e. the second example, "Favouire candy flavors", would in my opinion also pass if the light blue (Chocolate) was a very dark blue with insufficient contrast to the black border as long as the contrast to both the segments Tart and Mint was above 3:1. (not sure whether this chimes wit hthe hard-one understanding, though, so I am ready to withdraw this commwent if all this has been discussed at length already).

rachaelbradley:
Below is a proposal for revised test text:

For each graphical object required for understanding, use a color contrast tool to:
1. Measure the contrast ratio of each color compared to the adjacent color(s) or to the surrounding border (if present).
2. Check that the contrast ratio of each is equal to or greater than 3:1.

jake-abma:
Change:

...with a 3:1 color contrast with each color.

to:

... with at least 3:1 color contrast with each color.

PR #584, opened 2019-01-11
Tech failure unmodifiable single key shortcut

detlevhfischer:
A preview is here: 
[Failure of Success Criterion 2.1.4 due to implementing character key shortcuts that cannot be turned off or remapped](https://raw.githack.com/w3c/wcag/tech-failure-unmodifiable-single-key-shortcut/techniques/failures/unmodifiable-single-key-shortcut.html)


detlevhfischer:
I believe it is still unclear whether in repeating the pressing of all character keys with SHIIFT held down would be necessary.

Issue #538, opened 2018-11-23
Clarify if modal dialog on page load is a failure of SC 3.2.1 + 3.2.5


JamesCatt:
I'm trying to understand if having a modal dialog open immediately on page load constitutes a failure of SC 3.2.1 and/or 3.2.5.

Both SCs list F52 (opening a new window on page load) as a failure. Although "window" is not specifically defined in F52, the code examples both demonstrate opening new browser windows via javascript (`window.open`). 

So it would seem that F52 does not include modal dialogs, but a modal _does_ seem to fit the definition of "change of context" since it involves a change of focus.

To muddy the waters even more, F52 is listed as a failure for SC 3.2.1 for some reason, even though it does not involve focusing a component (see #170 ). I understand this issue has been raised previously, and initially [there was consensus to remove F52 from 3.2.1](https://www.w3.org/2016/04/05-wai-wcag-minutes.html#item06), but then once the PR (#175 ) was submitted the [working group reversed course and decided against](https://www.w3.org/2016/06/07-wai-wcag-minutes.html#item06) the change.

According to posts in #170 it sounds like all of this was supposed to be addressed as part of 2,1, but it seems like that didn't happen.

detlevhfischer:
I have drafted a Working Group response (see mailing list).

JamesCatt:
@detlevhfischer Thanks! I looked the mailing list and saw in the meeting minutes from Jan 15 that you took this on, but I don't see the response you drafted in the list archives. Is this because it needs to be discussed by the group first?

(sorry I'm not familiar with the working group's process)

detlevhfischer:
Sorry, it might be that I accidentally sent this to the list from the wrong email address, which was rejected  - will check later. I wasn‘t quite sure myself whether the thing to do was to post this straight into the github issue, that‘s why I sent it to the list.


Sent from phone

alastc:
Hi Detlev, you did send it, it went to the AG list on Thursday at 6:14am GMT, but I'll copy it here as well:
<hr>

We apologize for not yet having addressed this issue in our work for WCAG 2.1 as we had intended according to issue https://github.com/w3c/wcag/issues/170.

[...]

Issue #567, opened 2018-12-18
Do user interface control labels count as part of the control for the purposes of Contrast Requirement exemptions for disabled controls?


tigt:
We’re implementing a design system at my job, and we’re unsure if dimming the `<label>` associated with a disabled `<input>` is a violation or not:

```html
<input id="w" type="checkbox"> <label for="w">Checkbox label</label>
```

```css
input:disabled,
input:disabled + label {
  opacity: 0.4; /* Or color:gray or whatever */
}
```

I looked at existing UI guidelines from Apple and Microsoft, but they don’t seem to mention this case.

Our team is unsure if the [WCAG Guidelines for SC 1.4.3](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html) are for or against, depending on if a label is considered “part of an inactive user interface component” or not.

jake-abma:
This issues was also raised on slack, we might consider writing a technique / specifically mention in the Understanding doc.

fuzzbomb:
I was the person who asked on Slack. It's publicly archived, my question can be found at https://gitter.im/w3c/a11ySlackers?at=5c3f7f7c32a8370605e0c8b4

I think it's good to be able to read what an input is for, even if it's not currently operable. So I'm in favour of saying labels aren't exempt.

Disabled buttons  are an interesting case, because the disabled control and the text label are the _same element_. I'm not sure what to think about that one.

mraccess77:
The following is a proposed response that has not yet been discussed or accepted:

Thank you for submitting this question.  It is an important topic.
The exemption for SC 1.4.3 states:
"Text or images of text that are part of an inactive user interface component ...  have no contrast requirement."

The definition of component states:
"a part of the content that is perceived by users as a single control for a distinct function Note 1: Multiple user interface components may be implemented as a single programmatic element. Components here is not tied to programming techniques, but rather to what the user perceives as separate controls."

We understand the user need for inactive controls and related text to have sufficient contrast -- this challenge is complicated because sufficient contrast can be confusing for other user groups who might mistake the component for being active.  We have previously identified this as an area to consider when working on newer versions of accessibility standards.  People have proposed ideas such as an icon or other visual indicator that would be ubiquitous to the current dimming of a control to indicate it's inactive state.

Based on the current exemption for "inactive" content in SC 1.4.3 the label text is part of the user interface component and thus is not required to have a contrast minimum.

alastc:
[...]

PR #590, opened 2019-01-22
Removing F52 from on-focus

alastc:
Related to issue #538, this is a very small change to simply remove F52 from SC 3.2.1.

### DIFF
--- a/understanding/20/on-focus.html
+++ b/understanding/20/on-focus.html
@@ -173,14 +173,6 @@
             
             <li>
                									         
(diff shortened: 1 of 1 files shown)


detlevhfischer:
Hi Alastair, 
I think you meant to point to issue https://github.com/w3c/wcag/issues/538 not #567 (Do user interface control labels count as part of the control for the purposes of Contrast Requirement exemptions for disabled controls?)

alastc:
Oops, yes, thank you!

PR #600, opened 2019-01-29
Cleanup of reflow techniques.

alastc:
- Transforming the pre failure technique into a positive sufficient one.
- Removing the 'reflowing simple data tables', there is disagreement that is possible without creating other accessibility issues.
- Modifying the working of the mechanism to 'switch to mobile' technique.

[Preview update](https://cdn.staticaly.com/gh/w3c/wcag/reflow-technique-list/understanding/21/reflow.html?x=3#techniques) (under techniques)


patrickhlauke:
@alastc could you check if this is still relevant/needed? certainly the parts where you're sorting out the `@@` placeholder links has now been addressed already elsewhere

Issue #582, opened 2019-01-09
1.4.13 Content on Hover or Focus - hoverable requirement, clarification of point "persistent"


detlevhfischer:
When transient content is brought up via hover, we offer as one option the use of the keyboard to close it (press ESC), so we mix input modes here.
However, when content is brought up via keyboard focus, we do not seem to require that it is pointer-hoverable since the SC text says 

> "If pointer hover can trigger the additional content, then the pointer can be moved over the additional content". 

Is that the correct reading? I.e. if something like a drop-down menu is brought up by tabbing to it, there is no requirement that the user can move the pointer over it and interact with it?

Finally, the SC text says 

> The additional content remains visible until the hover or focus trigger is removed(...). 

I **think** the intention was to say "until the hover or keyboard focus on the trigger is removed". If so, this is not clear, so the text would be a candidate for errata. If however the formulation is deliberate, I don't understand it (the triggering element will rarely be completely removed ).

alastc:
First point: Jon will take to the LV task force, seems unlikely to be a serious issues (compared the complexity of the other way around), but need to check.

Second point: Someone needs to think about the nitty-gritty of what the errata would be for the second point.

mraccess77:
I think there is a case where the user might bring something up with the keyboard and the item doesn't fit in the magnified area.  The user may want to use the mouse to pan to the content -- but moving the mouse causes it to go away.  The user then must move the point over the hover trigger then it get it to come up again but this causes the user to have to move out of the magnified area.

mbgower:
@mraccess77 I'd like to pinpoint the issue here, and flag the potential danger I anticipate trying to adjust to address your concern.

First, let's talk about what results from the existing language requirements:

1. When I am navigating by keyboard, new information appears on focus. I can dismiss it by keyboard.
1. Depending on my level of magnification, all some or none (with high magnification) of this information appears in the viewport. 
1. Need to confirm, but I believe should users need to see content outside the viewport, in most situations they can use a cursor key (e.g., right arrow or specialized combo) to reposition the viewport to see the additional information. Since they are arrowing, the focus of the object isn't affected.

If users wishes to adjust the viewport by mouse, depending on how their mouse tracking is set (see below) when they touch the mouse, it either reorients the mouse to the current viewport or makes the current location of the pointer the new point of regard in the viewport.

[...]

PR #603, opened 2019-01-31
Update content-on-hover-or-focus.html

mbgower:
As per the discussion in https://github.com/w3c/wcag/issues/582, this erratum is intended to address potential misreading of the normative text for 1.4.13 Content on Hover or Focus in the final Persistent bullet. The key objective was to replace the word "trigger" so that users would not interpret that the trigger itself must be removed to meet the criterion.

The revised version reads:
> The additional content remains visible until the user dismisses it, its information is no longer valid, or the hover or focus is removed.

### DIFF
--- a/guidelines/sc/21/content-on-hover-or-focus.html
+++ b/guidelines/sc/21/content-on-hover-or-focus.html
@@ -17,7 +17,7 @@
    <dd>If pointer hover can trigger the additional content, then the pointer can be moved  over the additional content without the additional content disappearing;</dd>
 
     <dt>Persistent</dt>
-    <dd>The additional content remains visible until the hover or focus trigger is removed, the user dismisses it, or its information is no longer valid.</dd>
+    <dd>The additional content remains visible until the user dismisses it, its information is no longer valid, or the hover and focus is removed.</dd>
 
   </dl>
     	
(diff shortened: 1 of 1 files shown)


awkawk:
Why change the order? Can we just remove "trigger" so the change is as minimal as possible?

mbgower:
As per comments, I can live with that, but it _does_ read somewhat awkwardly, in terms of what "it" refers to.
> The additional content remains visible until the hover or focus is removed, the user dismisses it, or its information is no longer valid.

If you think it will be easier to pass as an erratum, I can live with it and I do not anticipate objections from those involved in the discussion (which is summarized [here](https://github.com/w3c/wcag/issues/582#issuecomment-457234323).

As Steve mentions, I also think this could be clarified in the Understanding document without issuing an erratum, should that prove the path of least resistance.

mbgower:
PS normally when I create a local branch, my client offers me a chance to push it. I didn't get it this time. Not sure if that is due to permissions? but I assume that is what is causing the "This branch has not been deployed" message

alastc:
Don't worry about the deployment messages, there's some build-test type stuff happening but it seems a bit flakey, the errors are generally due to time-outs. That it is showing the 'merge pull request' (to me at least) is a good sign.

mbgower:
As per https://github.com/w3c/wcag/issues/582#issuecomment-462393140, I am recommending this PR be withdrawn and, with @detlevhfischer's agreement, the original issue be closed.

alastc:
Reviewing potential errata, given there are voices that would insist on it being an errata for 2.1, I'll close as requested.

Issue #165, opened 2016-03-01
Test #1 of H39 for table caption needs re-wording


spanchang:
The intent for Technique H39 (table caption) begins with " The objective of this technique is to programmatically associate captions for data tables where captions are provided in the presentation". 
Yet, the first test says:
"1. Check that the table includes a caption element."

This leads to the interpretation that all data tables need a caption  ... disregarding the "where captions are provided in the presentation. " 
So the test needs to be re-worded:
1. Check that the table includes a caption element where captions are provided in the presentation.

And although H39 and H73 (summary attribute) say that caption text should not be the same as summary text, there is no test in either technique to address that. Maybe there should be a test that checks for that. When both are identical, obviously one of them  (most likely the summary), is wrong. It may not  function as an effective summary to aid table comprehension / navigation. 
Thanks,
Sailesh

DavidMacDonald:
Proposed response:
It is correct that Caption elements on tables are not necessary to meet WCAG. Therefore, we will adopt a variation of your wording for check 1.
"Where a caption is visually provided in the presentation, check that the table includes a caption element ."
It is not a failure of WCAG to have an identical Summary and Caption, although it is a best practice that they are different. We don't see a significant problem in the wild with this, and <summary> has been deprecated in HTML5, so we will not do further techniques on it.

MikeElledge:
+1, with addition of "check that the table **code also** includes a caption element" for clarity.

spanchang:
Mike,
The initial proposal already addresses  your concern:
"Check that the table includes a caption element where captions are
provided in the presentation".
I am not sure why  this is not accepted as is.

Besides about table summary:
How is  a table caption that is identical to the table summary
attribute not problematic?
Imagine for a moment, the summary  text is visible. Will the table
then ever pass UI design QA if the caption is the same as the summary?
Why should screen reader users be subjected to the same text  repeadedly?
Or is it the attibude: be thankful you have access to that content at
least once; do not complain if it is repeated?
Or is it a presumption that non-sighted users are nincompoop  and need
text repeated to them?
If the developer fixes it just once all screen reader users will
perpetually enjoy a good experience. No?
It is something that can be  detected by an automated testing tool and
not even require manual verification.
It is immaterial that the summary attribute  is not valid in HTML5:H73
is a valid technique. Does it  cause an SC 4.1.1 failure?
By the way, not using summary attribute   from HTML5 quality of code
point of view may be only a best practice from a code markup
perspective.
[...]

PR #604, opened 2019-02-04
Changes to address Issue #165

awkawk:


awkawk:
Accepted on feb 12, 2019 call

PR #606, opened 2019-02-07
corrections to Character key shorts and Reflow

mbgower:
The errata requests for these two SCs are as follows:
For Character Key Shortcuts, there are three changes.  The most important change is the replacement of "characters" by "keys" in the first bullet.  Simply, modifier keys are not characters. This change does not alter the interpretation but merely addresses the inappropriate use of a  word that could potentially cause confusion.

The second suggestion is more open to discussion, and concerns replacing "use" with the word "include". This is a suggested edit to address the literal meaning of the text. The purpose of the  mechanism is not to replace a character key with a non-printable key, but to create a keyboard shortcut combination -- e.g., instead of A use Ctrl+A. "Include" better asserts this notion.

Finally the parenthetical information has been updated to match what I believe is the acceptable style for e.g., (including a comma) and eliminating the "etc" which is superfluous in the context of an example. There is no supposition that the example is an exhaustive list, therefore the etc is unnecessary. This is a trivial change, only offered here with the assumption that if the text is being altered, such slight style questions may as well be addressed.

Reflow's changes are primarily concerned with number agreement. "content...are designed" becomes "content...is designed"

The first note has also been changed, but here rather than make "content which require" be "content which requires" we have added in "parts of" so that it reads "parts of content which require". This solution mirrors language in the SC exception.

A trivial  style change has also been added for two occurrences -- the comma after e.g.,

### DIFF
--- a/guidelines/sc/21/character-key-shortcuts.html
+++ b/guidelines/sc/21/character-key-shortcuts.html
@@ -14,7 +14,7 @@
     <dd>A <a>mechanism</a> is available to turn the shortcut off;</dd>
       						
    <dt>Remap</dt>
-     <dd>A mechanism is available to remap the shortcut to use one or more non-printable keyboard characters (e.g. Ctrl, Alt, etc);</dd>
+     <dd>A mechanism is available to remap the shortcut to include one or more non-printable keyboard keys (e.g., Ctrl, Alt);</dd>
 
     <dt>Active only on focus</dt>
     <dd>The keyboard shortcut for a <a>user interface component</a> is only active when that component has focus.</dd>
--- a/guidelines/sc/21/reflow.html
+++ b/guidelines/sc/21/reflow.html
@@ -15,7 +15,7 @@
   
    <p class="note">320 CSS pixels is equivalent to a starting viewport width of 1280 CSS pixels wide at 400% zoom. For web content which is designed to scroll horizontally (e.g., with vertical text), 256 CSS pixels is equivalent to a starting viewport height of 1024px at 400% zoom.</p>
    
-   <p class="note">Examples of parts of content which require two-dimensional layout are images, maps, diagrams, 
+   <p class="note">Examples of content which requires two-dimensional layout are images, maps, diagrams, 
      video, games, presentations, data tables, and interfaces where it is necessary to keep toolbars in view while
      manipulating content.</p>
    				
--- a/guidelines/sc/21/reflow.html
+++ b/guidelines/sc/21/reflow.html
@@ -13,9 +13,9 @@
    </ul>
    <p>Except for parts of the content which require two-dimensional layout for usage or meaning.</p>
   
-   <p class="note">320 CSS pixels is equivalent to a starting viewport width of 1280 CSS pixels wide at 400% zoom. For web content which are designed to scroll horizontally (e.g. with vertical text), the 256 CSS pixels is equivalent to a starting viewport height of 1024px at 400% zoom.</p>
+   <p class="note">320 CSS pixels is equivalent to a starting viewport width of 1280 CSS pixels wide at 400% zoom. For web content which is designed to scroll horizontally (e.g., with vertical text), 256 CSS pixels is equivalent to a starting viewport height of 1024px at 400% zoom.</p>
    
-   <p class="note">Examples of content which require two-dimensional layout are images, maps, diagrams, 
+   <p class="note">Examples of parts of content which require two-dimensional layout are images, maps, diagrams, 
      video, games, presentations, data tables, and interfaces where it is necessary to keep toolbars in view while
      manipulating content.</p>
    				


detlevhfischer:
Hi @mbgower - Read through it, looks good to me!

guidelines/sc/21/reflow.html
```diff
@@ -13,9 +13,9 @@ <h4>Reflow</h4>
    </ul>
    <p>Except for parts of the content which require two-dimensional layout for usage or meaning.</p>
   
-   <p class="note">320 CSS pixels is equivalent to a starting viewport width of 1280 CSS pixels wide at 400% zoom. For web content which are designed to scroll horizontally (e.g. with vertical text), the 256 CSS pixels is equivalent to a starting viewport height of 1024px at 400% zoom.</p>
+   <p class="note">320 CSS pixels is equivalent to a starting viewport width of 1280 CSS pixels wide at 400% zoom. For web content which is designed to scroll horizontally (e.g., with vertical text), 256 CSS pixels is equivalent to a starting viewport height of 1024px at 400% zoom.</p>
    
-   <p class="note">Examples of content which require two-dimensional layout are images, maps, diagrams, 
+   <p class="note">Examples of parts of content which require two-dimensional layout are images, maps, diagrams,
```
awkawk: I don't think that this is necessary. In WCAG content can be a piece of the page (see the "change of context" def for an example).

guidelines/sc/21/reflow.html
```diff
@@ -13,9 +13,9 @@ <h4>Reflow</h4>
    </ul>
    <p>Except for parts of the content which require two-dimensional layout for usage or meaning.</p>
   
-   <p class="note">320 CSS pixels is equivalent to a starting viewport width of 1280 CSS pixels wide at 400% zoom. For web content which are designed to scroll horizontally (e.g. with vertical text), the 256 CSS pixels is equivalent to a starting viewport height of 1024px at 400% zoom.</p>
+   <p class="note">320 CSS pixels is equivalent to a starting viewport width of 1280 CSS pixels wide at 400% zoom. For web content which is designed to scroll horizontally (e.g., with vertical text), 256 CSS pixels is equivalent to a starting viewport height of 1024px at 400% zoom.</p>
```
awkawk: Yikes.

guidelines/sc/21/character-key-shortcuts.html
```diff
@@ -14,7 +14,7 @@ <h4>Character Key Shortcuts</h4>
     <dd>A <a>mechanism</a> is available to turn the shortcut off;</dd>
       						
    <dt>Remap</dt>
-     <dd>A mechanism is available to remap the shortcut to use one or more non-printable keyboard characters (e.g. Ctrl, Alt, etc);</dd>
+     <dd>A mechanism is available to remap the shortcut to include one or more non-printable keyboard keys (e.g., Ctrl, Alt);</dd>
```
awkawk: Agree on the e.g./etc.

I think that use/include an be explained in understanding and probably needs to be clarified whether include or use is the term used/included in the SC text.

guidelines/sc/21/reflow.html
alastc: Oh dear, I wonder if that got changed from "web pages which are designed" and not updated the is/are.

alastc:
My thoughts: 

> one or more non-printable keyboard **characters/keys** (e.g., Ctrl, Alt)

"keyboard keys" sounds a bit odd, would just "keys" work? I haven't found anyone mentioning this as confusing, but if it's changed, just 'keys' would probably do it.

> remap the shortcut to **use/include** one or more

"Use" makes the most sense to me, if we changed it I would suggest "add" instead. I don't remember the ins & outs of this one from the time, but was there a reason it couldn't be this?
"A mechanism is available to remap the shortcut to add one or more modifer keys (e.g., Ctrl, Alt);"

> Examples of **parts of** content which require two-dimensional layout are images

I don't think that is necessary, 'content' conveys the meaning and the addition makes it less clear to me.

awkawk:
>"keyboard keys" sounds a bit odd, would just "keys" work? I haven't found anyone mentioning this as confusing, but if it's changed, just 'keys' would probably do it.

I vote for "keyboard keys". A keyboard is a board of keys, so it is normal to say keyboard keys even if it seems odd.

>"Use" makes the most sense to me, if we changed it I would suggest "add" instead. I don't remember the ins & outs of this one from the time, but was there a reason it couldn't be this?
"A mechanism is available to remap the shortcut to add one or more modifer keys (e.g., Ctrl, Alt);"

"add" makes it seem like a shortcut of ONLY non-printing character keys might not be ok, but it is.

> I don't think that is necessary, 'content' conveys the meaning and the addition makes it less clear to me. I realise that it says 'parts of the content' above, it's hard to explain, but that was added specifically to convey that that you can have pages with content, and a _part of_ that page/content can scroll in both directions.

I made this comment on the code in the pull request, so we agree.

mbgower:
Just to clarify, in Reflow, for "parts" the issue is that there is not agreement as it is written right now. "content which require" is not grammatical. If "parts of" is not added in, than it needs to become "content which requires"

mbgower:
@awkawk is there a reason this isn't part of this week's survey?

guidelines/sc/21/reflow.html
mbgower: If you don't put in "parts of" then it needs to be "requires" > "Examples of content which requires..."

awkawk: I'd change the verb to "requires"

awkawk:
@mbgower Forgot. Just added and will notify the 3 people who took the survey already.

guidelines/sc/21/reflow.html
marcjohlic: I prefer the change to "requires" as well.
```suggestion
   <p class="note">Examples of content which requires two-dimensional layout are images, maps, diagrams, 
```

awkawk:
Accepted by WG on 2/12/2019

mbgower:
For whatever reason, these changes were not incorporated, so I'm re-addressing.  I believe the Reflow ones were adopted. So for Shortcuts...

Existing:

> A mechanism is available to remap the shortcut to use one or more non-printable keyboard characters (e.g. Ctrl, Alt, etc);

I definitely prefer either "include" or "add" instead of "use". An author can't have a shortcut that is just Ctrl; it's a modifier key. So it seems to me saying the shortcut "uses" continues to imply something other than we want. I can live with "add", but it just seems a little clunky. Not going to die on this one, but...

In the _keyboard characters_  versus _keyboard keys_ versus _keys_ discussion, it seems like we have agreement it shouldn't be characters. I kinda get Alastair's keyboard key quibble, and I think we have clarifty we're talking about keys (e.g., Character Key Shortcuts), so I'm leaning to "keys".

We have agreement that the ", etc" can go. 

So in Mike's world, we end up with:

> A mechanism is available to remap the shortcut to include one or more non-printable keys (e.g., Ctrl, Alt)

mbgower:
@awkawk looks like there's a travis build failure, which is why this didn't come across?

alastc:
Travis build is something only Michael pays attention to I think, I started digging in once but it is generally noise, and doesn't prevent a merge. (Conflicts would, but not travis.)

mbgower:
@awkawk and @michael-n-cooper I'm assuming that because of the travis error, this never happened. There's actually a very odd history. I see that Andrew committed one of the changes for the second Note on Reflow on Feb 12,
 but it has never shown up

>   <p class="note">Examples of content which requires two-dimensional layout are images, maps..

Tthe other grammar fudge in Reflow has never had any controversy, and yet wasn't committed. 
>    <p class="note">320 CSS pixels is equivalent to a starting viewport width of 1280 CSS pixels wide at 400% zoom. For web content which are designed to scroll horizontally (e.g. with vertical text), the 256 CSS pixels is equivalent to a starting viewport height of 1024px at 400% zoom.</p>
becomes
>   <p class="note">320 CSS pixels is equivalent to a starting viewport width of 1280 CSS pixels wide at 400% zoom. For web content which is designed to scroll horizontally (e.g., with vertical text), 256 CSS pixels is equivalent to a starting viewport height of 1024px at 400% zoom.</p>

mbgower:
@awkawk I see you did the merge, but I did want to point out that one of the modifications that I thought had been pulled out was not (changing "use" to "include"). With this merge it now reads "include". I feel like this is my mess, so suggest I make a new PR for this and include the Character key shortcut line change in the same one. Does that work?

awkawk:
@mbgower It was "use" before and that was one of the suggested changes. I don't care which way that word goes, seems very editorial. You can make a PR if you want, but I'll sleep fine if you don't. :)

Also, these changes will be in the editor's draft but won't be in the public WCAG 2.1 until we republish with errata.

PR #607, opened 2019-02-07
Correct links to the WAI-ARIA spec.

alastc:
A collegue pointed out a 404 link from the [ARIA 11](https://www.w3.org/TR/WCAG20-TECHS/ARIA11.html) technique's procedure, which is also in the [WCAG 2.1 version of ARIA11](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA11).

Whilst correcting that, I noticed there's more, and there's a pattern. The links from the techniques go to URIs like:
https://www.w3.org/TR/wai-aria/roles#landmark_roles
The correct link is:
https://www.w3.org/TR/wai-aria/#landmark_roles

I.e. there is a sub-page specified before the within-page locator.

These pages don't exist, so the link is a 404.

From a file-search I found 12 techniques & failures in the  2.1 versions to fix.
Also, some of the URIs go to the 'latest' version, e.g. https://www.w3.org/TR/wai-aria/roles#heading
but the link text says "version 1.0". So I tried to align the link text to the URI, taking off the version number.

Requesting review from Michael in case there is a reason those links were 404 and a different change is needed.


alastc:
Also to note: Many people are still finding the WCAG 2.0 techniques, we really need to de-reference those now...

awkawk:
Looks like the 1.0 spec had a roles/ page and that has been removed in 1.1. Seems like your fix covers it.

PR #612, opened 2019-02-09
Fixes to WCAG 2.1 Understanding 2.4.6 and 3.3.2

patrickhlauke:
Closes https://github.com/w3c/wcag/issues/610

(note I wasn't sure whether to make the fixes on these files, which are in the 2.0 folder, or whether or not I should have created new instances of these in the 2.1 folder)

PR #614, opened 2019-02-11
Label in name understanding doc update, large

mbgower:


awkawk:
@mbgower The only thing that I find to be a problem is the "close proximity" bit - this is too subjective and I think will be seen as an attempt to redefine "label". The Label in Name SC speaks to "User Interface components with labels" and there is nothing in 2.0 or 2.1 that specifies distance from the control for a label to "count".  I think that the group should not make this clarification without a full understanding of the implications. (lines 18-27 the the lines in question).

understanding/21/label-in-name.html
```diff
@@ -9,23 +9,60 @@
...
-			<p>This issue is even more important for speech input and text-to-speech users who also have cognitive challenges. It’s an extra
-			cognitive load for a speech input user to remember and say a speech command that is different from the visible label they see on a
-			control. It’s also an extra cognitive load for a text-to-speech user to absorb and understand speech output that does not
-			match the visible label.</p>
+			<p>The intent of this Success Criterion (SC) is to make text that is <em>visually</em> associated with a component also associated with the component <em>programmatically</em>. This helps ensure that people with disabilities can rely on visible labels as a means to interact with the components.</p>
```
patrickhlauke: To be picky, the first sentence makes it sound to me like it has to be the same piece of text in the markup that is associated. However, it can be duplicated as well (i.e. the visual text itself, and then an alternative that duplicates that particular text, e.g. `First name <input ... aria-label="First name">`(

patrickhlauke:
> @mbgower The only thing that I find to be a problem is the "close proximity" bit - this is too subjective and I think will be seen as an attempt to redefine "label". The Label in Name SC speaks to "User Interface components with labels" and there is nothing in 2.0 or 2.1 that specifies distance from the control for a label to "count". I think that the group should not make this clarification without a full understanding of the implications. (lines 18-27 the the lines in question).

having just proposed removing some "close proximity" wording from 3.3.2 here https://github.com/w3c/wcag/pull/612/ i agree that it's not a concept that's specific enough, nor is it defined anywhere.

mbgower:
The notion of close proximity already exists:
> Field labels located in close proximity to the associated field assist users of screen magnifiers because the field and label are more likely to visible within the magnified area of the page.

(https://www.w3.org/TR/UNDERSTANDING-WCAG20/minimize-error-cues.html)

> Labels for most fields are positioned immediately before the field, that is, for left-to-right languages, either to the left of the field or above it, and for right-to-left languages, to the right of the field or above it. Labels for radio buttons and checkboxes are positioned after the field.

(https://www.w3.org/WAI/GL/2016/WD-WCAG20-TECHS-20160105/G162)

I have concerns -- I don't think it is too much to describe them as grave concerns -- that if we do not restrict the guidance to "near proximity", this SC is going to be interpreted in a way that is going to worsen the experience for both speech input users and screen reader users. I anticipate a lot of well-meaning people adding instructions, group labels, headings, input mask information, etc., to the accessible name which will increase both accidental triggering for speech input users and verbosity for screen reader users.

I'd rather deal with the very few exceptions in a separate technique (as was done with http://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/H65 for Labels or Instructions.) If you feel we can work in such qualifications into the Understanding doc, I'm happy to review the language. 

I'm going to try putting in some qualifiers into the section to see if I can make you happy.

awkawk:
> Field labels located in close proximity to the associated field assist users of screen magnifiers because the field and label are more likely to visible within the magnified area of the page.

This is just an example - it isn't saying that this is the only way.

> (https://www.w3.org/WAI/GL/2016/WD-WCAG20-TECHS-20160105/G162)

This is a technique that is saying to use the commonly expected locations, which are close to the control. It is just saying that this is an acceptable way, not that it is THE way.

> I have concerns -- I don't think it is too much to describe them as grave concerns -- that if we do not restrict the guidance to "near proximity", this SC is going to be interpreted in a way that is going to worsen the experience for both speech input users and screen reader users. I anticipate a lot of well-meaning people adding instructions, group labels, headings, input mask information, etc., to the accessible name which will increase both accidental triggering for speech input users and verbosity for screen reader users.

In the example that we've discussed, I think that is exactly what you are doing as a result of your close proximity language!

I'll read it again and make some suggestions/pull request.

patrickhlauke:
(for info, i searched through the rest of understanding, and found another use of "proximity" in https://www.w3.org/WAI/WCAG21/Understanding/status-messages - apart from that there's some old Silverlight techniques that use the term)

mbgower:
@patrickhlauke The language about proximity has been in place for a decade. I'm not sure that referring to your newly suggested edit to remove it is how citing precedence would work :)

understanding/21/label-in-name.html
mbgower: If you have a wording suggestion that solves that, I'll be happy, but I find trying to be clear about text vs text string, etc., has resulted in it being more confusing (or at least a lot harder to parse). Yes, one could just recreate the visible string in an aria-label; however at some point automated rules are going to start failing the multiple variations which would probably be okay with speech recognition -- e.g.,  "First Name:" and "firstname" -- and it is just so much easier for everyone if aria-label is left out where possible.

patrickhlauke:
> @patrickhlauke The language about proximity has been in place for a decade. I'm not sure that referring to your newly suggested edit to remove it is how citing precedence would work :)

i wasn't claiming any kind of authority. just pointing out that IF the group doesn't mind removal of the proximity language since it's not clearly defined anywhere, it may not be the best time to reintroduce that somewhere else instead. by all means, let's keep it... *shrug*

mraccess77:
Proximity is very important for users with low vision.  We are considering an SC for WCAG 2.2 to attempt to address these needs.  Based on the response here it sounds like a clear SC would be better than a technique that is not rooted in any WCAG 2.0 requirement.

understanding/21/label-in-name.html
patrickhlauke: I would tend not to handwave away the ways in which it's perfectly valid to pass this SC (e.g. using `aria-label`). So starting with something like

```
The intent of this Success Criterion (SC) is to make sure that text that is <em>visually</em> associated with a component also forms part of the component's accessible name.
```

and then adding more explanations, references, examples could be better?

Examples could be along the lines of (grabbing something I threw together not so long ago for an unofficial explainer):

```
<a href="..." aria-label="Please read more about WCAG 2.1">Read more</a> would pass, as the “Read more” string (disregarding capitalization) is present in the aria-label

<a href="..." aria-label="Read about WCAG 2.1">Read more</a> would fail, as the string is not present

<a href="..." aria-label="Read some more about WCAG 2.1">Read more</a> would fail, as the aria-label does not contain the actual string “read more” (the word “some” is between “Read” and “more”)
```

mbgower:
okay, I"ve made changes to this PR that incorporate much of the feedback from yesterday's survey and @awkawk PR version. They show up in the commits further up this thread.

understanding/21/label-in-name.html
awkawk: How about:
The intent of this Success Criterion (SC) is to make the accessible name include (or match) the text used for the visual label for a component. This helps ensure that people with disabilities can rely on visible labels as a means to interact with the components.

understanding/21/label-in-name.html
```diff
@@ -9,23 +9,61 @@
...
+			<p>Mismatches between visible labels and programmatic names for controls are even more of an issue for speech-input and text-to-speech users who also have cognitive challenges. Mismatches create an extra cognitive load for speech-input users, who must remember to say a speech command that is different from the visible label they see on a control. It also creates extra cognitive load for a text-to-speech user to absorb and understand speech output that does not match the visible label.</p>
+			<h3>Confining the label to adjacent text</h3>
```
awkawk: This was the one from the call where we suggested "Identifying label text for components"

understanding/21/label-in-name.html
```diff
@@ -9,23 +9,61 @@
...
+<li>immediately to the right of checkboxes and radio buttons</li>
+<li>inside buttons and tabs or immediately below icons serving as buttons</li>
+</ul>
+The rationale for some of these conventions is explained in <a href="https://www.w3.org/WAI/GL/2016/WD-WCAG20-TECHS-20160105/G162">G162: Positioning labels to maximize predictability of relationships</a>.
+</p>
+<p>It is important to bias towards treating only the adjacent text as a label because liberal interpretations of what constitutes a text label can jeopardize the value of this Success Criteria by lessening predictability. Isolating the label to the single string in close proximity to the component makes it easier for developers, testers and end users to identify the label targeted for evaluation in this SC. Predictable interpretation of labeling allows users of speech recognition technologies to interact with the element via its conventionally positioned label, and allows users of screen reading technologies to enjoy consistency between the nearby visible label and the announced name of the component.</p>
```
awkawk: "Criterion"

understanding/21/label-in-name.html
```diff
@@ -9,23 +9,61 @@
...
+<label for="firstname">First Name: </label><input id="firstname" type="text" name="firstname"><br />
+Likewise, "Next…" visibly shown on a button could have "Next" as the accessible name. When in doubt, where a meaningful visible label exists, match the string exactly for the accessible name.<br />
+<input type="submit" value="Next..."></p>
+
+<h4>Mathematical expressions and formulae</h4>
+<p>Mathematical expressions can be used as labels, for example "11x3=33" and "A>B" convey meaning. The label should not be overwritten in the accessible name, and substitutions of words where a formula is used should be avoided, such as making the name "eleven times three equals thirty-three" or "A is greater than B". Doing so would mean that a user who said "eleven multiplied by 3 is equivalent to thirty-three" would be unlikely to match. Further, converting a mathematical formula label into an accessible name that is a spelled-out equivalent may create issues for translation. The name should match the label's formula text.<br />
```
awkawk: Just a note to remind that this was the section Bruce had concerns about and needed a sentence.

mbgower: Bruce has signed off on my changes.

understanding/21/label-in-name.html
mbgower: Okay, I had modified but hadn't quite matched your words. Now updated.

mbgower:
@michael-n-cooper I have now resolved all of the outstanding comments, and we approved this for adoption in our last call. Can you please merge? Note that I _will_ be updating techniques, but I think it is good to get this new material in front of people. I will make a separate PR which will cover changes to the cross-referenced techniques.

PR #615, opened 2019-02-11
updates to address AWK's concerns and placeholder

awkawk:
@mbgower I'm going to keep working on this.


understanding/21/label-in-name.html
```diff
@@ -9,22 +9,21 @@
 		<h1>Understanding Label in Name</h1>
 		<section id="intent">
 			<h2>Intent</h2>
-			<p>The intent of this Success Criterion (SC) is to make text that is <em>visually</em> associated with a component also associated with the component <em>programmatically</em>. This helps ensure that people with disabilities can rely on visible labels as a means to interact with the components.</p>
+			<p>The intent of this Success Criterion (SC) is to make text that is <em>visually</em> associated with a component the same as text used to represent the component <em>programmatically</em>. This helps ensure that people with disabilities can rely on visible labels as a means to interact with the components.</p>
```
mbgower: I'm now finding the text biasing towards the programmatic text. We want the visual text to guide the programmatic text, not the other way around. How about
"...with a component also reflected in the text used to represent..."

understanding/21/label-in-name.html
```diff
@@ -9,22 +9,21 @@
...
+<p>When labels use a single string of text in close proximity to the component, developers, testers, and end users find it easy to identify the label targeted for evaluation in this SC. Obvious labeling allows users of speech recognition technologies to interact with the element via its conventionally positioned label, and allows users of screen reading technologies to enjoy consistency between the visible label and the announced name of the component.</p>
```
mbgower: maybe slight change for readability
"If labels use a single string of text in close proximity to the component, then developers..."
or better
"If labels are interpreted as a single..."

Concern here is that it now looks like we're offering advice on labelling, whereas before it was advice on how to interpret the label. Do you see the difference? The former is design advice, the latter is developer/tester interpretation guidance, which is I believe what this SC is about. We're not trying to alter how designers label stuff. Just ensure what they've done is reflected programmatically for the betterment of speech users.

awkawk:
No longer needed due to separate updates

Issue #635, opened 2019-02-21
"at least" should be "at most" in WCAG 2.1 SC 1.4.12


jrpool:
"at least" occurs 30 times on https://www.w3.org/TR/WCAG21/. It means what it says in 26 of those cases, but it means the opposite in the 4 cases where it occurs in SC 1.4.12.

For example, if the user sets line spacing to 1.3 times the font size, no loss of content or functionality should occur, but 1.3 is not at least 1.5. Conversely, if the user sets line spacing to 50 times the font size, content creators should not be required to protect against loss of content or functionality, although 50 is at least 1.5.

More colloquially, "at least" in these cases could be changed to "up to", rather than "at most".

awkawk:
I see your point, but from a testing perspective, setting the line spacing to 1.3 wouldn't allow you to make a conformance claim because you need to verify that the content isn't lost when you set it to 1.5 or greater. If you set the paragraph spacing to at least 1.5, including 2, 9 or 40 and no loss occurs then you can.

patrickhlauke:
changing it to "up to" could work, however. for testing, you'd test the extreme (1.5) - or, if you've got the time/inclination, all values leading up to 1.5.

jrpool:
Thanks for your thinking about this. It is not felicitous that we are in doubt about whether a word in a success criterion means what it says or means the opposite of that; or whether it modifies one or another word in a sentence. Clarity calls for a reformulation. @awkawk is correct that a test of 1.3 line spacing wouldn't permit a conformance claim. But a test of 40 line spacing wouldn't either. Neither test proves what the SC intends (at least I believe it intends) to require, namely that every line spacing from 1 to 1.5 is lossless and fully functional. The SC could be reformulated to say that, so we all understand it.

mraccess77:
On a related note SC 1.4.4 resize text says "up to 200%".   Many of us read that SC to require increments between the default size and 200% of the default size although practically speaking that would be cumbersome to test.

jrpool:
Yes, that's a good indication of the intent. While testing every increment might be expensive, testing all possible conditions is usually expensive, so we sample. But the SC serves as notice of what a legitimate complaint could report. Here I think we want the public to understand that a user who sets a line height of 1.4 and loses content has a basis for complaint, and a user who sets a line height of 1.6 and loses contant has no basis for complaint. I'd say the SC as formulated does not make that clear.

alastc:
[...]

PR #637, opened 2019-02-24
Editorial change to normative text of 1.4.12: at least > up to

patrickhlauke:
In line with the discussion at https://github.com/w3c/wcag/issues/635 this is an editorial (non-substantive) change to the normative language of SC 1.4.12 which better clarifies the intent of the
SC. Using "up to" (the same way that SC 1.4.4 uses "up to 200%" as pointed out by @mraccess77 ) better conveys that testing should ideally be done for any values from the default up to those
defined in the normative text. It also makes it clearer that values that go beyond those defined normatively that result in loss of content or functionality do NOT result in a failure of the SC.

Closes https://github.com/w3c/wcag/issues/635

### DIFF
--- a/guidelines/sc/21/text-spacing.html
+++ b/guidelines/sc/21/text-spacing.html
@@ -9,10 +9,10 @@
    <p>In content implemented using markup languages that support the following <a>text</a> <a>style properties</a>, no loss of content or functionality occurs by setting all of the following and by changing no other style property:</p>
    
    <ul>
-     <li>Line height (line spacing) to at least 1.5 times the font size;</li>
-     <li>Spacing following paragraphs to at least 2 times the font size;</li>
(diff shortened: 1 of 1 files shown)


detlevhfischer:
+1  
I think this would be an improvement, making the intention clear.

jake-abma:
+1 as well

patrickhlauke:
mulling this over further...the SC itself may still not help in the case of fonts that have a very tall x-height or similar (as even text with the same "font size" can appear wildly different/taller/shorter depending on the actual typeface itself). but i guess this is a good enough initial compromise, without having to go into very specific metrics calculations on a per-font basis for testers...

jake-abma:
Is there a conclusions for this one?

seems a critical part to say we demand a bandwidth of (example) "between 0.5 and 1.5" OR "from 1.5 and higher"

Issue #636, opened 2019-02-21
Do we need to clarify inactive user interface component


DavidMacDonald:
I did a Twitter poll asking A11y pros to say whether in a tabbed interface the unselected items could be interpreted as inactive and therefore greyed out. There was and 80/20 spit on results.
https://twitter.com/davidmacd/status/1093231012262293504

The question is. Should we allow an inactive user interface be a control that a user clicks on to become active. I would say "no, its not inactive, its unselected".

mraccess77:
I agree -- it's not inactive -- its unselected.   Treating unselected as inactive and ignoring contrast minimums is a very concerning issue and mis-application of the term.

awkawk:
I would say that the unselected tab is part of an ACTIVE UI Component - the Tab Bar - so it needs to meet the ratio.

awkawk:
Following up on this - it seems that 82% of people got this right. As such I don't think that addressing this is a huge priority, but if someone wants to suggest a sentence or two to speak to inactive components the WG can review/accept it.

patrickhlauke:
a small note/clarification in the understanding doc that "inactive user interface component" is to be understood as a disabled/non-actionable component?

DavidMacDonald:
We should probably explain that somewhere.

jake-abma:
Agree when adding / comparing with "disabled":

@patrickhlauke 
> a small note/clarification in the understanding doc that "inactive user interface component" is to be understood as a disabled/non-actionable component

@awkawk 
Being part of a Tar Bar though is not by definition "meet the ratio" as Tab Bars can contain inactive / disabled tabs. (Example: the current tab, two possible clickable tabs (options you already have) and 3 inactive / disabled tabs (options you do not yet have but are available when...)

So the following is not fully correct:

> I would say that the unselected tab is part of an ACTIVE UI Component - the Tab Bar - so it needs to meet the ratio.

awkawk:
@jake-abma good point.

awkawk:
@DavidMacDonald ok, you brought it up - want to suggest some text? :)

patrickhlauke:
made a first verbose stab at it https://github.com/w3c/wcag/pull/638 ... feel free to rip it apart/take it as a starting point 👍

alastc:
It's a pretty big addition for the glossary, and the 'inactive' part isn't included in the definition term. would it be better in the understanding document(s)? It is only used for the contrast ones.

patrickhlauke:
[...]

PR #638, opened 2019-02-24
Add clarification of "inactive" to user interface component glossary definition

patrickhlauke:
Admittedly very verbose, but clarifies the subtle distinction between an inactive and a disabled control. The first sentences were cribbed from the 1.4.11 understanding document
https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html

Opted to make the clarification directly in the glossary, rather than cramming it into both the 1.4.3 and 1.4.6 understanding documents.

[it wasn't clear if this was the right document to make the edits against, as there are additional .xml files in the `/sources` folder that seem to contain glossary definition bits]

Closes https://github.com/w3c/wcag/issues/636

### DIFF
--- a/guidelines/terms/20/user-interface-component.html
+++ b/guidelines/terms/20/user-interface-component.html
@@ -17,6 +17,8 @@
    <p class="note">What is meant by "component" or "user interface component" here is also sometimes
       called "user interface element".
    </p>
+
+   <p class="note">User interface components are "inactive" when they are not available for user interaction (e.g., a disabled control in HTML). An inactive user interface component is visible but not currently operable. An example would be a submit button at the bottom of a form that is visible but cannot be activated until all the required fields in the form are completed. This should not be confused with user interface controls which are not currently "selected". For example, in a tabbed interface, the tabs which colloquially would be referred to as not being "active" are <em>not</em> "inactive user interface components" - as long as they aren't disabled completely, they can be interacted with by the user (to switch tabs in the tabbed interface), and therefore aren't "inactive". More accurately, these tabs are merely "not selected".</p>
    
    <p class="example">An applet has a "control" that can be used to move through content by line or page
       or random access. Since each of these would need to have a name and be settable independently,


michael-n-cooper:
My initial reaction to this proposal is to suggest that, instead of adding yet another note to the definition of "user interface component", create a new term for "inactive user interface component", put most of the content from this proposal into the definition there, and then reference it from a much shorter note in the "user interface component" definition (and ideally from elsewhere, such as wherever "inactive" precedes existing links to "user interface component" or "inactive component" is used). The reason for this suggestion is the note reads to me as a pretty big tangent and I think glossary terms should have more crisp definitions.

This is just a thought for consideration though, not a formal input, so I'm putting this as an ordinary comment rather than a review.

awkawk:
Proposed WG response: This is a normative change so we can't do this until WCAG 2.2, so we are labeling it for WCAG 2.2 to be considered.

patrickhlauke:
feels to me it's on the border between normative and editorial change, as it doesn't really change anything, just clarify the intention that was already there but just not clear? not too bothered though either way...

awkawk:
@patrickhlauke The glossary is normative, so additions to it or changes to it are normative changes. Sorry, clearly on one side of the border! :)

alastc:
The working agreed with the response, I've added the WCAG 2.2 label, but given that we're so close to WCAG 2.2 I'll leave it open.

MelSumner:
May I recommend adding readonly along with disabled, and perhaps clarifying the difference between the two?

mraccess77:
In my opinion read only should be considered active as the user may need to read the text to copy it.   I agree clarification of active is needed as Apple grays out inactive alarms on the Apple watch -- yet these alarms are interactive and can be set -- they just are not currently set.     What is meant by inactive in this criterion is unavailable for interaction.     

Longer term people with low vision need to be able to read disabled content as well as that content often provides a clue as to other options that are available or that will be exposed based on some user action.  Thus, long term we need to figure out a better way to communicated disabled state without relying on contrast, e.g. an icon or pattern, etc.

patrickhlauke:
superseded by https://github.com/w3c/wcag/pull/1154

PR #639, opened 2019-02-26
Edits for PR614 for review

awkawk:


mbgower:
I did a heavier edit on my PR using input in the survey. I'll revisit each of your changes and see if it is now covered.

awkawk:
@mbgower I had other edits that were in a branch that for some reason didn't wind up in my pull request to your branch. I'll sort those out.

In general, I'm still concerned that your edits still treat "close proximity" as a normative concept and I don't think that we can do that. Some of my missing edits help make this more clear. I'm perfectly fine with suggesting that as a best practice, but since the definition of close proximity is subjective we can't use it to define the bounds of the SC.

awkawk:
No longer needed as #614 was updated.

PR #649, opened 2019-03-05
general techniques for label in name

mbgower:
techniques, some working examples, images


mbgower:
@MakotoUeki can you please verify that I have accurately captured what I think was intended by the Japanese example in G208? It's been over 30 years since I took Japanese :)

mbgower:
@awkawk if you have cycles please look at what I've done to these two techniques. I still ahve to add in a couple more examples to G210 covering stacked labels and icons, but I think/hope this approach is becoming comprehensible. It may make sense to split G210 into 2 separate techniques, but I'm seeing a few challenges.

detlevhfischer:
Looks good to me.

mbgower:
@awkawk the 2 techniques and supporting material in this pull request are now ready for group consideration.

MakotoUeki:
> @MakotoUeki can you please verify that I have accurately captured what I think was intended by the Japanese example in G208? It's been over 30 years since I took Japanese :)

@mbgower I've never known you took Japanese :)

I just found this. So sorry about that. And I don't think I caught up with this issue. 

I'm looking at the current version of G208 at:
https://www.w3.org/WAI/WCAG21/Techniques/general/G208
It looks strange. Text within span element is visible in the "Rendered example". And it is not needed to be invisible.

Could you please tell me what I should check here?

mbgower:
@MakotoUeki 
First, let me emphasize that I did not create this working example. I'm merely trying to adjust what was previously created to make it be understandable and correct.
You are correct that the katakana is appearing on the button. That is because the

MakotoUeki:
@mbgower 
> When you say the katakana is not needed, do you mean that the kanji is properly pronounced regardless? maybe this is just a bad example that should be retired?

The katakana "メールを" means "mail"(objject) and the kanji "送信する" means "send"(verb). The katakana is not the reading of the kanji. 

I'm confused because we don't hide "メールを" in general. So this looks weird to me. We will have "メールを送信する" (send email) or "送信する" (send) as the button text. Either will be fine. There is no need to hide a part of the button text in the first place. I'm not sure if I got the intent though.

Does the Example 5 make sense in English?

mbgower:
@MakotoUeki 
Ah, okay. I can help provide context, and your response has given me clarity on what I think was intended.
This technique is covering ensuring that the visible label is included in the accessible name.

Example 5 and 6 are the same scenario, in different languages. Example 5 gives the scenario in English, where the undisplayed object ("email") follows the displayed label ("send") to provide more context. In English, the non-displayed word "email" grammatically makes sense following the label and thus meets the recommendation to put the visible label first. 
I think they wished to demonstrate that the recommendation to put the visible word first in the accessible name does not work for some languages, and so added the Japanese example, where the invisible noun (mail) is placed in front of the visible verb ("send"), which makes sense grammatically in Japanese, and thus serves as an example that the recommendation to have the visible text at the start of the label is not always appropriate.
My feeling is we should just remove the Japanese example, because to me it raises more questions than answers.

MakotoUeki:
> and so added the Japanese example, where the invisible noun (mail) is placed in front of the visible verb ("send"), which makes sense grammatically in Japanese,

Yes, it makes sense on that point. But we don't hide text in the first place. If we want to have "mail"(object), we add it to the button text without hiding. I don't see any cases where we should hide "mail". That's why I'm confused with this example.

I agree with you on removing the Japanese example from G208. Or I'd like to contact the person who wrote this example and confirm the intent.

patrickhlauke:
> I don't see any cases where we should hide "mail". That's why I'm confused with this example.

It's not so much "should", but "could". If designers/developers do this sort of thing to make the visual appearance of the button smaller, this example aims to show that it can be done this way and it passes the label in name SC.

mbgower:
> If designers/developers do this sort of thing to make the visual appearance of the button smaller, this example aims to show that it can be done this way and it passes the label in name SC.

Agreed, and the last prior example in English demonstrates that too. I've pulled the Japanese example.

mbgower:
@michael-n-cooper This PR is now ready to go. I have removed the aria technique that was created out of the second general technique and put in in another branch, pending my reserach. So this  PR includes massive G208 changes, the new G211 technique and 5 examples (although only 3 of them are referenced here; I'll tidy that up in the PR for the aria branch when I do it).

PR #659, opened 2019-03-12
Update reflow.html

mbgower:
Modified section on browsers in mobile operation systems


understanding/21/reflow.html
```diff
@@ -39,11 +39,6 @@ <h3>Content exceptions for reflow</h3>
...
       <p>Interfaces which provide toolbars to edit content need to show both the content and the toolbar in the viewport. Depending on the number of toolbar buttons, the toolbar may need to scroll in the direction of text (e.g., horizontally in a vertically scrolling page). This Success Criterion therefore does not apply to interfaces which provide toolbars.</p>
     	</section>
-       <section>
-      <h3>Browsers in mobile operating systems</h3>
-      <p>The layout method of most common browsers in mobile operating systems does not offer reflow in the same way as on desktop browsers. In these browsers, content can be magnified in different ways, for example, by using a pinch gesture to scale up content, or a double tap on a particular column to make it fill the viewport width. This means that currently, mobile browsers usually exhibit horizontal scrolling regardless of what an author does.</p>
-      <p>In principle, mobile user agents can equally offer reflow when users zoom into content, as evidenced by the Dolphin browser for Android. The lack of reflow support in mobile operating systems can therefore be regarded as a user agent support issue.</p>
```
patrickhlauke: explicit mention of Dolphin is accurate, but feels a oddly specific. also, it seems at odds saying "in principle" and then citing an example of it in practice, unless it was changed to "In principle, all mobile user agents". Maybe reword along the lines of

```Some mobile user agents do offer reflow as an option, but this is far from common. The lack of reflow support in browsers on mobile operating systems can therefore..."

File: understanding/21/reflow.html [hunk 2]
```diff
@@ -39,11 +39,6 @@ <h3>Content exceptions for reflow</h3>
 <p>Complex data tables have a two-dimensional relationship between the headings and data cells. This relationship is essential to convey the content. This Success Criterion therefore does not apply to data tables.</p>
       <p>Interfaces which provide toolbars to edit content need to show both the content and the toolbar in the viewport. Depending on the number of toolbar buttons, the toolbar may need to scroll in the direction of text (e.g., horizontally in a vertically scrolling page). This Success Criterion therefore does not apply to interfaces which provide toolbars.</p>
     	</section>
-       <section>
-      <h3>Browsers in mobile operating systems</h3>
-      <p>The layout method of most common browsers in mobile operating systems does not offer reflow in the same way as on desktop browsers. In these browsers, content can be magnified in different ways, for example, by using a pinch gesture to scale up content, or a double tap on a particular column to make it fill the viewport width. This means that currently, mobile browsers usually exhibit horizontal scrolling regardless of what an author does.</p>
```
patrickhlauke: "browsers in mobile operating systems" perhaps better "browsers *on* mobile operating systems"

mbgower: I have no issue changing that, although it is pre-existing language used in both the current heading and the text.

understanding/21/reflow.html
mbgower: The Dolphin language was pre-existing. I'll see if I can rejig

mbgower: Okay, I made minor edits. please check

understanding/21/reflow.html
mbgower: Made the changes. Please check and resolve if it looks good.

understanding/21/reflow.html
patrickhlauke: I'd still tend closer towards my proposal (but that's of course not to say that's any better...happy to hear other opinions - just feels a bit odd singling out Dolphin). One thing I would say though, which is perhaps less controversial: I'd definitely change "The lack of magnified reflow support in mobile operating systems" to "The lack of reflow support *in browsers* on mobile operating systems" (as it's not the OS itself that lacks support, but the specific UA)

patrickhlauke: (also i noticed only now that my previous comment was badly formatted in markdown, which meant my actual code sample wasn't showing properly ... fixed now)

understanding/21/reflow.html
detlevhfischer: I don't mind removing the reference to the Dolphin browser to make this more future-proof.

> Mobile user agents _can_ in principle offer reflow when users zoom into content [Dolphin reference deleted]. The lack of magnified reflow support in mobile operating systems can therefore be regarded as a user agent support issue.</p>

understanding/21/reflow.html
mbgower: Added in your "in browsers" language. On the Dolphin langauge, I can live either way. I left it as is since it was pre-existing.

Issue #657, opened 2019-03-12
Technique G134: Outdated Links


cstrobbe:
Editorial issue: several links for technique G134 are outdated:

* [Validating an entire site](http://www.standards-schmandards.com/?2005/04/10/18-massvalidate) is now at http://www.standards-schmandards.com/2005/massvalidate/. 
* The [HTML Validation WIDGet](http://www.htmlhelp.com/tools/widgets/#valid) still "exists" but the latest version was for Internet Explorer 6 and probably not compatible with current versions of Internet Explorer or Edge.
* The articles [NetBeans: Working with XML, Part 1](http://www.onjava.com/pub/a/onjava/excerpt/netbeans_2/index1.html) and [NetBeans: Working with XML, Part 2](http://www.onjava.com/pub/a/onjava/excerpt/netbeans_2/index2.html) no longer exist.
* [XML Nanny](http://xmlnanny.com/) no longer hosts an XML or XHTML validator.

awkawk:
Changes in #678 accepted by WG April 9, 2019

PR #678, opened 2019-03-29
fixes per #657

awkawk:


awkawk:
Accepted by WG April 9, 2019

PR #681, opened 2019-04-03
Non-text contast - Adjacent colors and example clarifications.

alastc:
Updating the [explanation of Adjacent colors based on this email thread](https://lists.w3.org/Archives/Public/w3c-wai-gl/2019JanMar/0071.html).

Also:
* Added a couple of clarifications on the example descriptions from @mbgower.
* Removed the text-based techniques from the techniques section, as they aren't specifically for non-text contrast. However, a note has been added in the 'required for understanding' section cross-referencing text contrast.

[Preview on statically](https://cdn.staticaly.com/gh/w3c/wcag/e357f39c2bbd8d287da375a095b723aa669f17d8/understanding/21/non-text-contrast.html).


Myndex:
Hi Alastair,

The terms "brightness" and "luminance" are not interchangeable. Brightness is a perceptual _quality_, luminance is a measurable _quantity_. As such, I suggest removing the parenthetical (luminance) at the end of the sentence.

I'd like to mention that drop shadow, outlines & borders are often not just ornamental, but used to enhance perceptual contrast. In that regard they should indeed be used as part of a contrast assessment. However the equations listed in the W3C understanding section on contrast are not useful for such assessment. 

I'm not certain that "subsumed" is correct, more like _"while multiple colors used in borders and drop shadows can have an effect on total perceived contrast, there is not a practical way to define this programatically. As such the designer should be cautious and use appropriate judgement to ensure adequate contrast."_ 

An "Image Appearance Model" such as ICAM might be able to do this programmatically, but otherwise requires judgment on the part of the designer. This is related to a separate issue I recently opened for discussion. xref #695

alastc:
> Brightness is a perceptual quality, luminance is a measurable quantity.

I trust that is correct, but luminance is not a well known term. Brightness is a well known term, and the closest I know of to describe how someone would perceive luminance. 

We use a luminance measure, but what (common) term would you use to convey that? 

We could say something like this, but it seems unnecessarily wordy:
> color closest in brightness (measured by luminance)

> I'm not certain that "subsumed" is correct, more like ...

That misses the point of the section, it is trying to say that you can ignore certain colors for the calculation. That should make it easier to measure. If you look at the [rendered version](https://cdn.staticaly.com/gh/w3c/wcag/e357f39c2bbd8d287da375a095b723aa669f17d8/understanding/21/non-text-contrast.html) I think that's clear.

Myndex:
> > Brightness is a perceptual quality, luminance is a measurable quantity.
> 
> I trust that is correct, but luminance is not a well known term. Brightness is a well known term, and the closest I know of to describe how someone would perceive luminance.
> 
> We use a luminance measure, but what (common) term would you use to convey that?

Hi @alastc  I see, you want to relate to the measured term used elsewhere. Then I think you could say:

..brightness (perceived luminance).

PR #685, opened 2019-04-05
Remove empty/unnecessary example section from understanding 2.5.5

patrickhlauke:
> Screenshot of the empty example section as rendered in the understanding document - just an empty coloured block


detlevhfischer:
+1

PR #688, opened 2019-04-05
Sort out figure labels

michael-n-cooper:
Require id on figure, automatically put "Figure #: " at beginning of label. Will require cleanup of labels.


awkawk:
@michael-n-cooper I think that this is ready.

PR #693, opened 2019-04-12
Update motion-actuation.html

mbgower:
Changes to improve understanding and readability. The only one which may need scrutiny is my replacement of "indirect" with "incidental"

### DIFF
--- a/understanding/21/motion-actuation.html
+++ b/understanding/21/motion-actuation.html
@@ -23,7 +23,7 @@
 				<h3>Benefits</h3>
 				
 				<ul>
-					<li>This Success Criterion helps people who may be unable to perform particular motions (such as tilting, shaking, or gesturing) because the device may be mounted or users may be physically unable to perform the necessary movement. This success criterion ensures that users can still operate all functionality by other means such as touch or assistive technologies.</li>
+					<li>This Success Criterion helps people who may be unable to perform particular motions (such as tilting, shaking, or gesturing) because the device may be mounted or users may be physically unable to perform the necessary movement. This success criterion ensures that users can still operate all functionality by other means such as touch or via assistive technologies.</li>
   <li>Other users will benefit in situations where they are unable to move their devices.</li>
 				</ul>
 			</section>
--- a/understanding/21/motion-actuation.html
+++ b/understanding/21/motion-actuation.html
@@ -12,19 +12,19 @@
 		</h1>
 		<section id="intent">
 			<h2>Intent of this Success Criterion</h2>
-			<p>The intent of this success criterion is to ensure that functions that are triggered by moving a device (for example, shaking or tilting) or by gesturing towards the device (so that sensors like a camera can pick up and interpret the gesturing), can also be operated by more conventional user interface components, unless the motion is essential for the function or not using motions or gestures would invalidate the activity.</p>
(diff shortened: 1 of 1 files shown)


detlevhfischer:
Hi @mbgower you have removed, in examples, "A user can choose an application setting which turns off Shake to Undo features." That may be fine - I am not quite sure about the implications. Does it mean that content would fail that does not offer another way of undoing actions (even though shake to undo is a system level feature)? Say, you accidentally delete input in a textarea, shake the device, input is restored. Without 'shake to undo' there may be no explicit way to undo that deletion. Would that already be a failure of 2.5.4? I guess not, but maybe we should be clearer still about this aspect (e.g., scoping  the SC to motion implemented by content, excluding functionality that is implemented on UA or OS levels).

patrickhlauke:
> Does it mean that content would fail that does not offer another way of undoing actions

i'd say yes, that's the point of 2.5.4? "Functionality that can be operated by device motion or user motion can also be operated by user interface components "

i'd also tend to fail under 2.1.1 in this case.

detlevhfischer:
@patrickhlauke The ability to undo by shaking may just be considered a OS-supplied bonus (I believe there is no equivalent for Android users?) and users in general may not have that option to undo. So taking your line, would that not mandate authors to replicate OS-provided functionality across platforms inside their content to cater for any case where the OS functionality is not available? That seems untenable.

patrickhlauke:
ah, i thought you meant actual functionality implemented by the page author. the iOS typing and shake-to-undo is a functionality of the user agent. as that's not a functionality of the *web content* itself, then it's exempted

detlevhfischer:
It actually says application setting, not OS setting, so it may have been clear enough. Still I wonder why Mike thought it better to remove this bit.

understanding/21/motion-actuation.html
```diff
@@ -12,31 +12,32 @@ <h1><strong>Motion Actuation</strong>
...
         
       
       <section id="benefits">
 				<h3>Benefits</h3>
 				
 				<ul>
-					<li>This Success Criterion helps people who may be unable to perform particular motions (such as tilting, shaking, or gesturing) because the device may be mounted or users may be physically unable to perform the necessary movement. This success criterion ensures that users can still operate all functionality by other means (e.g., touch or voice input).</li>
-  <li>Other users may benefit in situations where they are unable to hold and move the device because their hands are occupied with some other activity.</li>
+					<li>This Success Criterion helps people who may be unable to perform particular motions (such as tilting, shaking, or gesturing) because the device may be mounted or users may be physically unable to perform the necessary movement. This success criterion ensures that users can still operate all functionality by other means such as touch or assistive technologies.</li>
```
awkawk: maybe add "via" before "assistive technologies? Reads oddly to me without.

patrickhlauke: agree with @awkawk

mbgower: I agree. Incorporating

mbgower:
@detlevhfischer 
> Hi @mbgower you have removed, in examples, "A user can choose an application setting which turns off Shake to Undo features." 

I removed this because it is not an author solution. If a device does not offer such an application setting, the app would still fail. If the author implements shake to undo, it is the author's job to offer a means of disabling.

PR #694, opened 2019-04-12
Update status-messages.html

mbgower:
Modifications for clarity. Adjusted wording of advisory techniques


understanding/21/status-messages.html
```diff
@@ -17,7 +17,7 @@ <h2>Intent of this Success Criterion</h2>
...
 	</ol>
-<p>Information can be added to pages which does not meet the definition of a status message. For example, the list of results obtained from a search are not considered a status update and thus are not covered by this Success Criterion. However, brief text messages displayed about the completion or status of the search, such as "Searching...", "18 results returned" or "No results returned" would be status updates if they do not take focus. Examples of status messages are given in the section titled <a href="#status-examples">Status Message Examples</a> below.</p>
+<p>Information can be added to pages which does not meet the definition of a status message. For example, the list of results obtained from a search are not considered a status update and thus are not covered by this Success Criterion. However, brief text messages displayed <em>about</em> the completion or status of the search, such as "Searching...", "18 results returned" or "No results returned" would be status updates if they do not take focus. Examples of status messages are given in the section titled <a href="#status-examples">Status Message Examples</a> below.</p>
```
awkawk: Might be clearer changing "if they do not take focus" to "if they do not cause a change in context by taking focus"

understanding/21/status-messages.html
```diff
@@ -163,8 +160,8 @@ <h3>Advisory</h3>
 		<li>Using aria-live regions to support <a href="https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html">1.4.13 Content on Hover or Focus</a> (future link)</li>
```
awkawk: I need a reminder why this 1.4.13 technique that doesn't exist yet is important for 4.1.3.

understanding/21/status-messages.html
mbgower: Note that only change in language here was  emphasis on "about". Fine if you want to make that change as long as you don't think it changes readability.

understanding/21/status-messages.html
mbgower: Because by having the content that is exposed by focus read out as a live region, the user does not need to do any action to hear the info, such as putting their focus on it. In some cases the exposed content may qualify as status messages, and live region is a nice technique. But since not all content exposed with hover/focus will be a status message, it is advisory.

understanding/21/status-messages.html
```diff
@@ -89,20 +89,17 @@ <h3 id="excepted-examples">Examples of Changes That Are Not Status Messages</h3>
...
-      <ul><li>A form field asks “How many children do you have?” and when the user responds with the number 5, five 'child name' fields are added to the form. Offscreen text may announce “5 form fields added”</li>
-<li>	A list of 15 products are on a page with a “See More” button at the bottom. When the user clicks the button, offscreen text announces, “15 products added to the page”</li></ul>
-        <p>Several Advisory Techniques are incorporated into this Understanding document.</p>
+<p>Live regions and alerts can be usefully applied in many situations where a change of content takes place which does not constitute a status message, as defined in this Success Criterion. However, there is a risk of making an application too "chatty" for a screen reader user. User testing should be carried out to ensure the appropriate level of feedback is acheived. The Advisory Techniques provide examples of how alerts or live regions can enhance the user experience.</p>
```
patrickhlauke: "level of feedback is acheived" > "level of feedback is achieved"

patrickhlauke:
Wondering if this may also be an opportunity to clarify even further in the understanding document that this SC is NOT about requiring status messages where there currently aren't any...something along the lines of

```
Note that this criterion is only concerned with ensuring that where status messages are displayed, they are correctly / programmatically identified in a way that allows assistive technologies to present / announce them to the user. This success criterion does <em>not</em> mandate the use of status messages (either visible to all sighted users, or off-screen / visually hidden but announced by assistive technologies) if the content doesn't currently implement them.
```

mbgower:
Addressed Patrick's typo and comment. @awkawk I think this PR may have been overlooked?

patrickhlauke:
LGTM

alastc:
@mbgower yep, sorry, added the 'ready for survey' label so it goes into teh survey queue.

In the meantime, in the interest of spreading this knowledge around, the build is failing.

The thing to do in that scenario is:

- Select the details link next to the failling message (the top one in this case).
- Select the link "the build" in the middle of the page.
- If you see the message "The program 'ant' is currently not installed.", it means you need to merge master into this branch. That adds a file for ant to use.
- If you don't see the ant message, search for the term "fatal", and that should provide some information on the build issue. Usually a missing end tag or similar HTML markup issue.

mbgower:
Thanks, @alastc . in case anyone is looking at this, 
> you need to merge master into this branch

is accomplished in the client by choosing 'Update from master' from the Branch menu in the GUI client

PR #701, opened 2019-04-25
In 1.4.11 Understanding text, adds field boundary exception when contrasty placeholder present 

detlevhfischer:
In section "Boundaries" the example used is a button that does not necessarily need a contrasty outline in text acts as indicator and has  4.5:1 or better contrast. The same is probably true for placeholder (contrast above 4.5:1) before a text field gets focused. I have just added that case (we had questions around that).


awkawk:
Perhaps it would be better to say "there is no visual indication of the complete hit area" (adding "complete". There is an indication of the hit area in that the text provides it, just not the full extent.

Issue #708, opened 2019-04-29
Key terms have word "new" preceding definition, 2.5.2: Pointer Cancellation


jake-abma:
Key terms have word "new" preceding definition

see: https://www.w3.org/WAI/WCAG21/Understanding/pointer-cancellation.html#key-terms

> down-event
> new
> platform event that occurs when the trigger stimulus of a pointer is depressed

> single pointer
> new
> pointer input that operates with one point of contact with the screen, including single taps and clicks, double-taps and clicks, long presses, and path-based gestures

> up-event
> new
> platform event that occurs when the trigger stimulus of a pointer is released

mbgower:
I note that in [Reflow](https://www.w3.org/WAI/WCAG21/Understanding/reflow#key-terms), the "New" is capitalized. 
These Key Terms sections are not actually in the Understanding documents, but generated. @michael-n-cooper, I have made the change under /terms/21.
I note that the dfn for CSS pixel (from Reflow) contains a data-lt attrbute missing from the up-event and down-event. Since other of the new ones also have this, and the -lt attribute seems to be a form of plurality indicator, I didn't touch it here, but wanted to note.
https://github.com/w3c/wcag/pull/711

michael-n-cooper:
I'm not sure but assuming the request here is to suppress the generation of "new" from key terms, which is a hold-over from the working draft stage. I'll put that on my list.

jake-abma:
At 2.5.1: Pointer Gestures we have another one:

https://www.w3.org/WAI/WCAG21/Understanding/pointer-gestures.html#key-terms

> **single pointer**
> new
> pointer input that operates with one point of contact with the screen, including single taps and clicks, double-taps and clicks, long presses, and path-based gestures

patrickhlauke:
@jake-abma a "point" on that key term...noting that a mouse is also a "pointer" and has no "contact with the screen" as such (only in an abstract sense)

jake-abma:
@patrickhlauke thanks, I know, I use Wacom pen, most of the time don't touch anything. But this one is about the word "new", not about the definition :-)

patrickhlauke:
yeah sorry, i knee-jerked there. filed a proper separate issue here https://github.com/w3c/wcag/issues/749

PR #711, opened 2019-04-29
Pointer cancellation jake suggestions

### LINKED ISSUE #707: 2.5.2: Pointer Cancellation mentions failure for  2.5.3, editorial issue
A failure reference for techniques mentions 2.5.3, think that needs to be 2.5.2

https://www.w3.org/WAI/WCAG21/Understanding/pointer-cancellation.html#techniques

> @@ Failure of SC 2.5.3 due to activating a button on initial touch location rather than the final touch location.

mbgower:
addressed editorial raised by Jake

Closes #707


detlevhfischer:
+1 looks fine to me

jake-abma:
Hi @mbgower shouldn't we need to remove the word new altogether?

jake-abma:
see also: https://github.com/w3c/wcag/issues/708#issuecomment-488067248

alastc:
Hi @jake-abma,

The 'new' is from the build-generator, that's in Michael's (very long) to-do list, and shouldn't affect the other aspects.

alastc:
In fact, the technique updates were mostly already done in master, so I'm going to merge this one, it's purely editorial and will be over-ridden by whatever MichaelC does with the generator.

Issue #522, opened 2018-10-18
2.5.1 Pointer Gestures: clarity on path-based language in Understanding document


mbgower:
2.5.1 says actions must be achieved "without a path-based gesture". I took this to mean the same thing as in the 2.1.1 Keyboard SC , whose language gives an exception "where the underlying function requires input that depends on the path of the user's movement and not just the endpoints."

That seems to be the same concept as "path-based" in 2.5.1. However, 2.5.1's Understanding document gives the following for disallowed path-based gestures (my emphasis):
>  Examples include swiping (which relies on the direction of movement) the **dragging of a slider thumb**, or gestures which trace a prescribed path, as in the drawing of a specific shape. 

The slider thumb seems to contradict the interpetation in the pre-existing 2.1.1 Understanding document, which provides the following :
> Drawing straight lines, regular geometric shapes, re-sizing windows and **dragging objects to a location (when the path to that location is not relevant)** do not require path dependent input.

With a slider, I can normally drag in any direction I want, and the slider just interprets my movement as it relates to a pre-determined axis. So I can take _any_ path I want to arrive at the slider position (end point) I desire. 

With this in mind, I would suggest that operating a slider by dragging should be given the same pass as drag-and-drop:
> Note that free-form drag and drop actions are not considered path-based gestures for the purposes of this Success Criterion

detlevhfischer:
Hi Mike,
following your argument, based on the 2.1.1 Understanding doc, that "dragging objects to a location (when the path to that location is not relevant) do not require path dependent input" would also apply to sliders, there would be no need to provide single tap/click operation alternatives to sliders at all, which (I hope) is not your intended meaning. I would argue that sliders that move a thumb on a defined track with endpoints but also intermediate stops indicating some value between max and min are an entirely different beast compared to free-form drag and drop of elements from location A to B. The fact that once you pick up the thumb on pointer-down, you do not have to follow the **exact** straight line of the track does not make this a free-form gesture to me. 
[...]

PR #714, opened 2019-04-30
Update pointer-gestures understanding doc

mbgower:
Addresses resolution from group discussion for https://github.com/w3c/wcag/issues/522

I removed most (but not every) dragging example, and added in language about the desirability of alternatives.
I also made some minor editorial changes for readability which are not intended to alter in any way the meaning of the Understanding document.


patrickhlauke:
I'll admit that I'm torn on this issue myself. There are certainly bad/naive implementations of sliders that do not offer all the good affordances such as allowing users to stray off the slider track, and essentially require users to very precisely follow the slider's track/path very closely. I would count these (bad) implementations as failures under this SC unless they provided alternatives (such as the ability to tap/click anywhere on the track to jump the slider button directly to that position/value, or extra arrows/controls to move the slider up/down or left/right). so discounting/out-of-scoping sliders altogether here feels wrong to me.

A rather broken example of such a bad mouse-driven slider can be found here https://codepen.io/patrickhlauke/pen/yMpbXa (and yes, it has many other issues, such as not being AT or kbd friendly, but just to demonstrate the problem of slider thumbs not "capturing"/"locking onto" the user's pointer when clicked) - compare to this other example (as broken as the previous one, but at least it follows the mouse's X position even after the user moves the mouse outside of the slider) https://codepen.io/patrickhlauke/pen/eZgqRo

detlevhfischer:
@patrickhlauke I think the issue of Mike's proposed edit is if sliders would be considered out of scope, which would exempt a significant group of interface objects from being operable by single point activation. Whether this or that slider implemetation would be in scope due to being or not being constrained vertically seems a different point. From the perspective of single-point activation (click or tap at one point) they both fail since you can't click / tap the groove to reposition the thumb (and there are no alternative inputs for incementing/decrementing).

patrickhlauke:
as @detlevhfischer asked me to clarify (since my usual word salad may have buried the lede): when I said "I would count these (bad) implementations as failures under this SC" I meant, in a roundabout way "I would not remove/exempt sliders from this SC outright, as there are slider implementations that *can* be problematic for users in ways that this SC aims to actually cover".

or shorter: no, i wouldn't exempt sliders.

alastc:
It's a good point about swipe-to-reveal, it is somewhere between a slider and a gesture.

Can we draw the line just prior to full drag and drop? I.e. drag 'n drop has specific start and end points, and the line between them does not matter.

Whereas a slider / gesture does not have a specific end point.

I still think the rational that drag 'n drop is too difficult to do (with all the other criteria to account for as well) is the best one.

I've done a PR (#725) to take Mike's readability updates, keep sliders, and add an example for short/long swipe.

Feel free to update that, I'll pop this whole question back in the survey for next week, with two clear options.

jake-abma:
just some 2 cents, we've implemented sliders using drag-and-drop from a technical perspective...

mbgower:
>  From a user perspective and in terms of the needs of motor-impaired users, there is no clear conceptual separation between swiping and dragging 

Gotta disagree with this to some degree. One swipes in a direction. Therefore, it's directional and so must be path-based. The simplest example I can find is in the Yahoo mail app. I go to an item and put my finger on it. Each direction I pull causes a different action. (up/down, scroll inbox; right, delete message; left, mark as read).  I can't turn one action into another; it's triggered by direction.

That's clearly path-based, to me.

I'm happy to see you incorporate that into a proposed new rewrite of Understanding and get traction on that once there is a consistent, easily understood distinction.

> There are complex swipe / drag gestures where the functionality triggered can be dependent on the length of the path 

If it's a complex swipe/drag that has dependencies, then obviously it wouldn't qualify in the exception. Add in language about that in the doc.
I don't think time is called out in the document. Since we don't define 'complex' I think we have the ability to introduce time into the discusion in the Understanding document. To me, if the speed of gesture alters it's interpretation (I guess we'll need limits for that) then it is a clear indication of a complex gesture.

patrickhlauke:
perhaps muddying the waters further, "swipe" at least to me also suggest more of a fast motion, while dragging is more measured/slow. though you then also get interfaces/components where you can either slowly drag or "throw" things (with momentum etc).

wondering if the distinction in general is more about "you can do things by doing simple taps/clicks, without the need to move a finger on the screen/keep mouse pressed and move it" (though our original intent was, i think, that movement itself was fine, just not the need for precise/following a particular path type movement...but maybe that isn't useful enough to disambiguate situations?)

detlevhfischer:
The question has been raised before - and more recently by @mbgower - if the definition of "path-based" that we intend to add needs to be in line with the exception for path-dependent input in SC 2.1.1 Keyboard:

>[Everything is keyboard-operable] "...except where the underlying function requires input that depends on the path of the user's movement and not just the endpoints."

> "Note 1: This exception relates to the underlying function, not the input technique. For example, if using handwriting to enter text, the input technique (handwriting) requires path-dependent input but the underlying function (text input) does not."

The path-dependent definition in 2.1.1 was, as I see it, motivated by functions (drawing, signing) that require the production of a particular shape - something that cannot be achieved with a keyboard.  ‚Path’ is thus a component of such a shape.  Its nature is not input-operational but **descriptive** (for want of a better term) - what counts is the resulting form that may be taken as-is or pattern-matched / converted. 

In contrast, the purpose of the pointer input paths *we* are dealing with here is **operational** - moving something represented on screen, whether content slider, menu, or control slider thumb. That is why I feel it is appropriate to depart from the 2.1.1 exclusion wording for path-dependent. I consider it reasonable (and importantly, addressing demonstrated user needs) if we now come to define ‘path-based’ as separate from 2.1.1, as operational pointer path, which would cover both swiping and dragging. 

If memory serves, Mobily A11y TF's core idea of what became SC2.5.1 Pointer Gestures has been: Offer an single point operable alternative to any pointer path on screen -- for example, to enable users who simply CANNOT swipe or drag, e.g. who position the mouse with a voice-operated grid in speech AT and depend on single point activation. With that logic, free-form drag-n-drop **should** really be in, but I accept the argument that this would set the bar for developers very high **and** introduce complexities/issues for other users. This is clearly not the case for elements where movement is constrained - adding single point operation can be as trivial as making a slider groove respond to touch.

detlevhfischer:
I did not intend to work on this branch and rename it - I hope that can be reverted. I have created a separate PR https://github.com/w3c/wcag/pull/735/ for a proposed rewording of the understanding text that would clarify the original Mobile a11y TF intent to require a single point activation alternative for all pointer input actions that require users to draw a path (be it swipe or drag).

jake-abma:
@patrickhlauke wrote:

> though our original intent was, i think, that movement itself was fine, just not the need for precise/following a particular path type movement...but maybe that isn't useful enough to disambiguate situations?

Is that so?

When I read the part of the SC:

> can be operated with a single pointer **without a path-based gesture**

I thought exactly the opposite.

So included or no?

alastc:
Ach, just found this comment not posted... will discuss on the call in a moment anyway.

Hi @detlevhfischer,

You wrote (to MAFT):
> I believe the original intent of the SC that became 2.5.1 Pointer Gestures was to require a single point operable alternative for all path-based interactions, including dragging. My feeling is that the issue is important because excluding control sliders from 2.5.1 via the understanding doc would affect many users with severe motor impairments. 

Presumably that was adjusted because looking at drag&drop made that infeasible, and now we've zoomed in a level into the Mandelbrot picture, and are noticing different types of sliders & components.

Given the separate 2.1.1 SC, I just want to check the motor-impairments aspect, as I'm not convinced that a well-done slider should be in scope on that basis.

For example, if we took the [first color-picker slider I found on the web](https://www.cssscript.com/demo/rgb-color-picker-slider/), and made the 'thumbs' larger, I _think_ sliding would be more effective than tapping. Being able to hold it down, experiment, get it right, then let go seems like a better interaction.

In usability testing with someone with quite a severe motor impairment they general tapped (with their wrist-knuckle), but had to use a joystick for certain interactions (including typing). In that scenario they'd be using a keyboard-like interface. For someone with a less severe impairment I think the un-restricted slider would be easier. It's hard to argue that tapping a tiny slider-target is harder than tapping a thumb-target and then being able to move freely. Unless there's something I'm missing?

Also, there are other types of widgets that fall into this little in-between space, such as [resizable boxes](https://jqueryui.com/resizable/) and [sort-able lists](https://jqueryui.com/sortable/). It is difficult to see how that could be adjusted to a tap-only interface.

I do agree that using the path-based concept from 2.1.1 isn't quite the same. Perhaps we call this one 'path based gestures' instead?

mbgower:
@detlevhfischer  was quoted as saying to the MATF:
> I believe the original intent of the SC that became 2.5.1 Pointer Gestures was to require a single point operable alternative for all path-based interactions, including dragging. 

I was a bit surprised to see this since I clearly showed in an email to Detlev over a week ago that:
1. while swiping is mentioned in the original description, "drag" was never in the [original wording](https://github.com/w3c/wcag21/issues/61) provided by the MATF
1. while "path" is mentioned once, the phrase is "perform specific gestures (e.g. draw a complex path with their finger on a touchscreen) in a precise and timely manner". That seems a lot more restrictive than Detlev posits.
1. dragging as an in-scope concept is never mentioned in any meeting or issue I could find until it was introduced [a year later by Detlev](https://github.com/w3c/wcag21/issues/196#issuecomment-292242824).
1. [Greg V](https://github.com/w3c/wcag21/issues/356#issuecomment-341544162) and others have questioned the wisdom of extending the scope in the way being advocated
1.  there was a voted-on [official response to an issue](https://github.com/w3c/wcag21/issues/478) raised against Pointer Gestures clearly stating that path-based, as used in keyboard, was the basis of the SC wording (and that it would be turned into a normative definition, which didn't happen).

Obviously the SC was further altered after this point, but I think it's crucial we understand that this discussion has been visited and voted on multiple times (including just a few weeks ago, which triggered this PR to align the text with what was agreed on). I'm somewhat frustrated both that there are multiple discussions going on in many forums _and_ that valid points which are raised in one place are ignored in the other discussions (like this one).

BTW, for the record I have never advocated sliders being out of scope. If someone creates a slider that depends for operation on path of movement, multitouch or (if we ever bring it in scope) timing variations, that would fail the SC.  I objected to the example of dragging a slider being in scope for what I hope are now obvious reasons to do with the fact the path of the user's movement in no way affects a 'standard' slider. All that matters is the end points. Pretty much the exact same distinction used in keyboard.

patrickhlauke:
>  I objected to the example of dragging a slider being in scope for what I hope are now obvious reasons to do with the fact the path of the user's movement in no way affects a 'standard' slider

I believe the issue is more when the slider *isn't* standard (i.e.  a badly custom built one), and fails to allow users to stray from the track, or doesn't allow users to just tap anywhere on the track, or otherwise interact with it other than forcing the user to carefully drag the slider button along the slider track ... and whether those cases should fail this SC?

detlevhfischer:
Am 23.05.2019 um 18:47 schrieb Mike Gower:
>
> @detlevhfischer <https://github.com/detlevhfischer> was quoted as 
> saying to the MATF:
>
>     I believe the original intent of the SC that became 2.5.1 Pointer
>     Gestures was to require a single point operable alternative for
>     all path-based interactions, including dragging.
>
> I was a bit surprised to see this since I clearly showed in an email 
> to Detlev over a week ago that:
>
>  1. while swiping is mentioned in the original description, "drag" was
>     never in the original wording
>     <https://github.com/w3c/wcag21/issues/61> provided by the MATF
>
IN MATF we talked about 'complex gestures' and the MATF call seemed to 
confirm today that assuming this included dragging was not just my own 
interpretation, but TF consensus at that time.
>
>  1. while "path" is mentioned once, the phrase is "perform specific
>     gestures (e.g. draw a complex path with their finger on a
>     touchscreen) in a precise and timely manner". That seems a lot
>     more restrictive than Detlev posits.
>
Admittedly that old MATF wording is a bit fuzzy and was superseded when 
the SC was taken up by the WG,  but the fact is that in order to drag 
something, you have to (be able to) perform a path on screen (move your 
finger across it).
>
>  1. dragging as an in-scope concept is never mentioned in any meeting
>     or issue I could find until it was introduced a year later by
>     Detlev
>     <https://github.com/w3c/wcag21/issues/196#issuecomment-292242824>.
>  2. Greg V
>     <https://github.com/w3c/wcag21/issues/356#issuecomment-341544162>
>     and others have questioned the wisdom of extending the scope in
>     the way being advocated
>  3. there was a voted-on official response to an issue
>     <https://github.com/w3c/wcag21/issues/478> raised against Pointer
>     Gestures clearly stating that path-based, as used in keyboard, was
>     the basis of the SC wording (and that it would be turned into a
>     normative definition, which didn't happen).
>
As I have said before, I find the reference to start and end points in 
the 2.1.1 exception wording awkward when dealing with operational 
pointer gestures. Moving a content slider by swiping or dragging (both 
gestures usually work) does not envisage an endpoint (dragging control 
sliders might, but end point is recursively dependent on users' reaction 
to feedback of what changes, requiring fine motor control to adjust. I 
call that 'complex').
>
> Obviously the SC was further altered after this point, but I think 
> it's crucial we understand that this discussion has been visited and 
> voted on multiple times (including just a few weeks ago, which 
> triggered this PR to align the text with what was agreed on). I'm 
> somewhat frustrated both that there are multiple discussions going on 
> in many forums /and/ that valid points which are raised in one place 
> are ignored in the other discussions (like this one).
>
Mike, I tried to report the back story and your position in the call, so 
I certainly do not consciously ignore it - I just seek clarification. I 
fully understand your frustration and I will let it rest if the WG 
decides in the upcoming survey to exclude dragging. This is not a 
personal crusade - I am motivated by the feedback I have received from 
users with severe motor impairments who have advocated to have dragging 
in scope, and feel encouraged by other voices in MATF. I believe that 
the normative language of 2.5.1 allows for the scope I advocate, and it 
is up to us to decide whether we want this. For many months, the 2.5.1 
Understanding text has now included dragging in its description and its 
examples, and this will have shaped the interpretation of the community 
of the scope of 2.5.1. So narrowing it now needs a very good reason. 
Firming up the interpreation as it stands may require tweaking the 
proposed definition of path-based"User input that depends on the path of 
the user's movement and not just the endpoints" although I don't read it 
in a way that necessarily excludes dragging.

alastc:
Trying to keep discussions available in one place, I sent this to the list:
<hr>

Hi everyone,
 
Further to my point about needing to defining path-based gestures, I’ve been compiling [different examples and their attributes in this spreadsheet](https://docs.google.com/spreadsheets/d/1NGxqbwbcaG8dwSs5H2eTRy0gDELTdSuAi1ad0le4fDU/edit?usp=sharing), and [video examples in this folder](https://1drv.ms/u/s!AqcLQeCk6CjwiJpy0DUfE49R2LPiYg?e=mRI6w8). The IDs from the spreadsheet map to the videos.
 
One thing this exercise taught me was how few ‘gesture’ interactions do actually lock down the direction, undermining my previous approach. Things like Mail.app/Gmail gestures do NOT restrict the direction after the start of the movement, making it difficult to differentiate from (open) sliders.
 
In summary: It is really hard to see how to reasonably scope-out drag and drop (which we’ve agreed) without also scoping out sliders (and sortable lists).
 
The closest I can see is:

> A path-based gesture is an interaction where the user:
> - Engages a pointer with a defined starting point in the content (down event),
> - carries out a directional movement or movements before disengaging the pointer (up event),
> - where the end point is separate from the starting point and not a pre-defined area.
 
I’m not happy with the last bullet, but I can’t think of another way of putting it. The “separate” bit is to prevent people interpreting that as scoping-out strict sliders where you have to release at a particular point.
 
An alternative would be a last bullet of something like:
> - where the end point does not define a value.
 
That would scope out sliders, drag & drop, sortable lists, and possibly multi-value gestures.
 
Given the attributes identified in the spreadsheet, can anyone see another option?

alastc:
For those on interested in the follow up to the call today, I've made [an update](https://github.com/w3c/wcag/pull/714/commits/750acec0c2506f999c98e2eae1a2762f99f6f210) to the top of the understanding doc, [preview here](https://raw.githack.com/w3c/wcag/PointerGesturesUnderstandingDragChanges/understanding/21/pointer-gestures.html).

I haven't worked it all the way through, but I thought I'd get that starting point done to see if the description of path-based gestures holds up.

In the examples given near the top I included "some sliders and carousels", which I'll expand on in the examples tomorrow. Part of that is about whether we need to account for certain mobile gestures being adjusted by the user-agent.

I.e. if you select a slider and drag-down, it generally scrolls down without activating the slider. However, some sliders (or sometimes, it's hard to tell) it will both scroll _**and**_ activate your use of the slider, so sideways movement also moves the thumb. To me that doesn't qualify as a "gesture" as defined.

The [MDN 'range' example](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range) generally activates on my iPad regardless of direction.

If someone (e.g. pointer master @patrickhlauke) knows how browsers deal with that in general, that would be appreciated.

Other examples are linked from the [spreadsheet](https://docs.google.com/spreadsheets/d/1NGxqbwbcaG8dwSs5H2eTRy0gDELTdSuAi1ad0le4fDU/edit#gid=0).

understanding/21/pointer-gestures.html
```diff
@@ -3,23 +3,39 @@
...
+			<p>The intent of this Success Criterion is to ensure that content can be controlled with a range of pointing devices, abilities, and assistive technologies. Some people  cannot perform gestures in a precise manner, or they may use a specialized or adapted input device such as a head pointer, eye-gaze system, or speech-controlled mouse emulator. Some pointing methods lack the capability or accuracy to perform multipoint or path-based gestures.</p>
+
+			<p>A <strong>path-based gesture</strong> involves an interaction where the user engages a pointer (starting point), carries out a movement that goes through at least one mid-point before disengaging the pointer (end point). The mid-point defines the path as a gesture.</p>
```
awkawk: How about replacing the last sentence with "The mid-point defines the gesture as requiring a specific path, even if the path is not defined for the complete path." This seems backwards as it is.

understanding/21/pointer-gestures.html
```diff
@@ -3,23 +3,39 @@
...
+
+			<p>Authors must ensure that their content can be operated without multipoint or path-based gestures. Multipoint or path-based gestures can be used so long as the functionality can also be operated by another method, such as a tap, click, double tap, double click, long press, and click &amp; hold. </p>
+
+			<p>This Success Criterion applies to author-created gestures, as opposed to gestures defined on the level of operating system or user agent. Examples for gestures provided at the operating system level would be swiping down to see system notifications, and gestures for built-in assistive technologies (AT) to focus or activate content, or to call up AT menus. Examples of user-agent-implemented gestures would be horizontal swiping implemented by browsers for navigating within the page history, or vertical swiping to scroll page content.</p>
```
awkawk: I don't think that this is exactly right.  I'd say:

This Success Criterion applies to gestures in the author-provided content, as opposed to gestures defined or use by the operating system, user agent, or assistive technology. Examples for gestures provided at the operating system level would be swiping down to see system notifications, and gestures for built-in assistive technologies (AT) to focus or activate content, or to call up AT menus. Examples of user-agent-implemented gestures would be horizontal swiping implemented by browsers for navigating within the page history, or vertical swiping to scroll page content.

understanding/21/pointer-gestures.html
```diff
@@ -3,23 +3,39 @@
...
+
 			<p>While some operating systems may provide ways to define "macros" to replace complex gestures, content authors cannot rely on such a capability because it is not pervasive on all touch-enabled platforms. Moreover, this may work for standard gestures that a user can predefine, but may not work for other author-defined gestures.</p>
```
awkawk: Maybe "supported" or "available" instead of "pervasive"?

understanding/21/pointer-gestures.html
```diff
@@ -3,23 +3,39 @@
...
-			<p>An exception is made for functionality that is inherently and necessarily based on complex paths or multipoint gestures. For example, entering one's signature may be inherently path-based (although acknowledging something or confirming one's identity need not be).</p>
+
+      <p>An exception is made for functionality that is inherently and necessarily based on complex paths or multipoint gestures. For example, entering one's signature may be inherently path-based (although acknowledging something or confirming one's identity need not be).</p>
+
+      <p>Gestures that involve dragging in any direction are not in scope for this SC, however, such gestures do require fine motor control. Authors are encouraged to provide non-dragging methods, for instance, a drag and drop operation could also be achieved by selecting an item (with a tap or keyboard interaction) and then selecting its destination as a second step.</p>
```
awkawk: how about:

Gestures that involve dragging in any direction are not in scope for this SC because only the start and end points matter in a dragging operation. However, such gestures do require fine motor control.

understanding/21/pointer-gestures.html
```diff
@@ -32,11 +48,10 @@ <h3>Benefits</h3>
 		<section id="examples">
 			<h2>Examples</h2>
 				<ul>
-        <li><p>A web site includes a map view that supports both the pinch gesture to zoom into the map content and drag gestures to move the visible area. User interface controls offer the operation via [+] and [-] buttons to zoom in and out, and arrow buttons to pan stepwise in all directions.</p></li>
+        <li><p>A web site includes a map view that supports the pinch gesture to zoom into the map content. User interface controls offer the operation via [+] and [-] buttons to zoom in and out.</p></li>
```
awkawk: We can probably remove " via [+] and [-]"

awkawk: Just think about how that will read with a screen reader...

understanding/21/pointer-gestures.html
alastc: Refining slightly to: The mid-point defines the gesture as requiring a specific path, even if the complete path is not defined.

understanding/21/pointer-gestures.html
alastc: Integrated. I'd been working on that in parallel but along similar lines.

understanding/21/pointer-gestures.html
alastc: done.

alastc:
Comments generally integrated, but I had over-hauled the middle section in parallel, so some didn't apply any more.

I haven't made any significant changes to the examples/benefits, I don't think it needed it.

understanding/21/pointer-gestures.html
awkawk: yep, I meant to end with "gesture" but you fixed a different way.

understanding/21/pointer-gestures.html
awkawk: Still applies. Just trying to simplify language, even if a little amount.

alastc: Which is odd, as the word 'pervasive' isn't in the file anymore!?

alastc:
Hi @mraccess77,

You asked in another issue something I'd rather answer here:
> in your update you say a swipe gesture is a path based gesture (which I agree with) -- but I'm trying to understand how it matches the intermediate point criteria. In some samples I have seen on the Internet people are simply looking for a change in direction from the touch start location to the last touch move location (which is calculated on touch end). Maybe the platform creators do this differently than hacky devs -- but it feels like swipes are really just about the offsets of start and ending points with some tolerance for deviance of 30-60 pixels. So swipes aren't really about the start and end points but the delta within tolerances.

I think that still fits the definition, if you have to swipe in a particular direction (e.g. left), going far enough in that direction to be recognized as a swipe: So that is the mid-point. 

You can carry on with the movement, and some might require you stay within a particular path, but many don't. In fact, you can often (even in some hacky web ones I've tried), go back to the start and cancel the gesture, if the X-coordinate hasn't changed enough.

However, the fact you have to go in a direction means there is a 'mid-point', the alternative would be that you could go in a different direction and still trigger it (e.g. swipe up).

detlevhfischer:
"the relative direction of the gesture at some intermediate point" sounds rather vague and possibly misleading. The direction is critical at the *onset* of the gesture, to get it 'off the ground', as it were.

detlevhfischer:
Ah, @alastc you committed to a change while I was fiffling with anither wording. This is what I tried:

"A **path-based gesture** involves an interaction where not just the endpoints matter - its direction (at least initially, after beginning the gesture) also affects its meaning. The user engages a pointer (starting point), carries out a movement that goes through at least one mid-point before disengaging the pointer (end point). The mid-point defines the gesture as requiring a specific path, even if the complete path is not defined.

detlevhfischer:
@alastc I think your wording now also works, but it feels quite redundant.

alastc:
Hi Detlev,

We tried the 'direction' approach previously (that was my thought as well), and ran into trouble. I'd rather not re-run the dicsussion again, so I'll merge it when the build passes.

detlevhfischer:
@alastc  OK, I am not insisting. Maybe I was not part of that discussion - at any rate,  I do not remember why anyone objected to direction / directionality (that may just be my bad memory). But:  If we aim to make our understanding texts easy to understand, it seems a no-brainer to use a more straightforward wording so people won't scratch their heads. I would imagine anyone at EOWG agreeing to that.

alastc:
It wasn't so much objecting to the concept, as when you work the wording through it didn't include everything we wanted to due to the various implementations of swiping. I can't now remember the exact details, but from the [matrix spreadsheet](https://docs.google.com/spreadsheets/d/1NGxqbwbcaG8dwSs5H2eTRy0gDELTdSuAi1ad0le4fDU/edit#gid=0) and the testing of different implementations, talking about an initial direction didn't cover everything.

If someone can say the same thing more simply, great, I struggled. But let's update this one now to be correct (as per the discussions) and apply another pass for improving readability later.

NB: Pretty much every addition for clarity I have made, including those from other people, has caused other people to say that change is not clear, so there is a lot of subjectivity going on!

jake-abma:
Hi @alastc this one goes into master / live soon? Read through it yesterday and may need some small fixes... rawgit version not active anymore but saw some things to mention. Didn't make notes so can go through it when live if wished for.

I do remember we had two of the same words following each other and also I have a comment about the opening sentence:

> The intent of this Success Criterion is to ensure that content can be controlled with a range of pointing devices, abilities, and assistive technologies.

I don't think nr. 1 and 3 are part of the intent, "content can be controlled with a range of pointing devices" and "content can be controlled with a range assistive technologies".

Only abilities should be part, correct?

Cheers!

Issue #404, opened 2018-06-29
Understanding Success Criterion 2.4.1: Bypass Blocks


michael-n-cooper:
_From @devarshipant on May 23, 2018 17:55_

First bullet under [Benefits ](https://w3c.github.io/wcag21/understanding/bypass-blocks.html) does not sound like a **Benefit** on satisfying this criterion:

> When this Success Criterion is not satisfied, it may be difficult for people with some disabilities to reach the main content of a Web page quickly and easily.

Compare it to second bullet, where it's usage is more appropriate.

> Screen reader users who visit several pages on the same site can avoid having to hear all heading graphics and dozens of navigation links on every page before the main content is spoken.

_Copied from original issue: w3c/wcag21#939_

michael-n-cooper:
_From @awkawk on May 25, 2018 13:19_

Thank you for the comment. Can you suggest improved language for the first bullet? When I review it, it seems that a positive restatement of this first bullet is what the 2nd and 3rd bullets do. Perhaps we should just remove the first bullet?

michael-n-cooper:
_From @devarshipant on May 25, 2018 15:32_

What do you think of combining the first and second bullets to something like:

> A mechanism to bypass repetitive blocks of content allows screen reader and keyboard only users who visit several pages on the same site to reach content with fewer keystrokes.

michael-n-cooper:
_From @awkawk on May 25, 2018 15:47_

The three bullets are:
-When this Success Criterion is not satisfied, it may be difficult for people with some disabilities to reach the main content of a Web page quickly and easily.
- Screen reader users who visit several pages on the same site can avoid having to hear all heading graphics and dozens of navigation links on every page before the main content is spoken.
- People who use only the keyboard or a keyboard interface can reach content with fewer keystrokes. Otherwise, they might have to make dozens of keystrokes before reaching a link in the main content area. This can take a long time and may cause severe physical pain for some users.

Your suggestion makes the combined bullet overlap with the original 3rd bullet. What to do with that one?

michael-n-cooper:
_From @devarshipant on May 25, 2018 16:50_

Can we replace the first 3 bullets with the revised combined bullet?

Thanks,
Devarshi

alastc:
Pull request created, ready for survey.

alastc:
PR #721 was merged, closing.

fstrr:
This looks to have been resolved, so am closing.

PR #721, opened 2019-05-07
Bypass blocks issue 404

alastc:
I addressed the issue in #404 and added a common example. 

Adding the example needs AG agreement, but hopefully a simple one.


detlevhfischer:
+1 Looks good to me

jake-abma:
+1

Issue #416, opened 2018-06-29
Understanding Success Criterion 1.3.4: Orientation


michael-n-cooper:
_From @devarshipant on June 11, 2018 16:45_

https://www.w3.org/WAI/WCAG21/Understanding/orientation.html

Please refer to the following paragraphs:

> "The goal of this Success Criterion is that authors should never restrict content's orientation, thus ensuring that it always match the device display orientation."

1. Isn't "Content's Orientation" similar to "system-level orientation" setting? 

2. It is not clear what orientation settings can be controlled by content authors? I see Content, System-level, and Device-level Display orientation settings in use.

3. Should it be mentioned that restricting a display setting is fine if that specific display orientation is essential?

> " ...current orientation determined by any device senors."

senors should be sensors.

_Copied from original issue: w3c/wcag21#963_

michael-n-cooper:
_From @NIKHILSAIREDDY on June 21, 2018 5:2_

Then what about the gaming softwares

alastc:
I think thigs have changed anyway, the doc now starts:
> The intent of this Success Criterion is to ensure that content displays in the orientation (portrait or landscape) preferred by the user. Some websites and applications automatically set and restrict the screen to a particular display orientation and expect that users will respond by rotating their device to match, but this can create problems.

I think that clarifies 1 & 2, and the SC text itself has examples of 'essential':
> Examples where a particular display orientation may be essential are a bank check, a piano application, slides for a projector or television, or virtual reality content where binary display orientation is not applicable. 

@devarshipant does that resolve your comments/issues/questions?

StefanSTS:
The term "binary" display orientation sounds a bit strange to me. For an easier understanding I would prefer something like: neither landscape nor portrait. Since WCAG is all about making things easy to understand.

awkawk:
Suggest a modification that may address the issue but in a slightly different way:
Examples where a particular display orientation may be essential are a bank check, a piano application, slides for a projector or television, or virtual reality content where content is not restricted to simple landscape or portrait display orientation.

StefanSTS:
That sounds much better. 
Though I think there are actually simple orientations for virtual reality too, if you wear VR glasses. It might not be in a 3D cinema. But that distinction might just be to much to explain in this example.
Maybe: ... where content is not - necessarily - restricted to simple ...

mbgower:
BTW, isn't this normative language?

awkawk:
[...]

PR #724, opened 2019-05-09
Updates for issue 416, inc normative note

alastc:
Updates to address #416


alastc:
Noting for @michael-n-cooper that this update includes a small errata for the orientation SC.

michael-n-cooper:
Erratum added in https://www.w3.org/WAI/WCAG21/errata/

PR #725, opened 2019-05-10
Updates based on #714 discussion

alastc:
Taking @mbgower's readability updates, but still including sliders as a form of gesture. Adding an example for email swipe / long-swipe.


detlevhfischer:
I am very happy about this PR.
What threw me a bit was the sentence "A user may find it difficult or impossible to accomplish these if they type and point with a single finger or stick"  - I wasn't sure why we are talking about 'type' here (as we are dealing with a pointer context) and wondered whether it was 'tap' that was meant. Is 'stick' the joystick scenario? Other tech that might be mentioned here is sip-and-puff and laser pointers.

alastc:
Hi Detlev,

That wasn't one of the changes, that's there now. I assume it means 'type' as in type-in characters using the onscreen keyboard, with one finger/stick. 

Would this be clearer?
> A user may find it difficult or impossible to accomplish these if **they can only use** a single finger or stick, in addition to any of the causes listed above.

detlevhfischer:
Yes, I realize that it wasn't a change, and it isn't crucial, I just thought the mention of 'type' was a bit odd in the context of this pointer-oriented SC.  I still think we may mention more mono-pointer scenarios (mouse pointer, joystick, sip-and-puff mouse, laser pointer) to show it is a wide-ranging problem across input devices.

alastc:
Hi Detlev,

The text just above this section mentions quite a few examples, and I added sip-and-puff mouse & laser pointer.

alastc:
Closing, going back to the branch in #714

Issue #706, opened 2019-04-29
iOS and Android not part of examples for WCAG, should be deleted 2.5.2: Pointer Cancellation


jake-abma:
There are possible techniques suggested for 2.5.2: Pointer Cancellation:

https://www.w3.org/WAI/WCAG21/Understanding/pointer-cancellation.html#techniques

> @@ Activating a control using the up-Event in HTML, iOS and Android.

As the scope of WCAG is not native and we're not providing examples for native / iOS / Android we should not suggest we will add techniques for them.

mbgower:
I'll make a PR for this and the other issue you mentioned.

mbgower:
https://github.com/w3c/wcag/pull/711

jake-abma:
see it's fixed

PR #726, opened 2019-05-10
Technique: Native controls for up event

alastc:
New technique from @DavidMacDonald 

[Preview of the technique](https://raw.githack.com/w3c/wcag/tech-up-event/techniques/general/G211.html).

Also noting @jake-abma's comment in #706


detlevhfischer:
Copied from survey response:
Pointer-up event: Maybe we should use the term "up-event" here since it is the term in the 2.1 glossary. Readers might benefit from an explanation (in brackets) of what technical events are grouped as “pointer up” event i.e., change to "This can be accomplished most directly by relying on the pointer's up-event (for example, the onclick or mouseup event)."
If we include pointerup and touchend we might want to add a warning that pointer-specific up-events may not work across all devices (Patrick??).

Test procedure: The intro "Find all clickable controls with actions that are irreversible" seems a bit odd. Activating a link can mostly be reversed (bbrowser back button) but would, in my understanding, violate the SC if the link is triggered on the down-event. So testing for all controls in this way to first check whether the action is irreversible is not a particularly efficient testing approach. It seems easier (in a mouse-based testing scenario) to activate the down-event on each control, then move the pointer outside the target and release the mouse. (BTW we refer to a down-events bookmarklet https://testen.bitv-test.de/bookmarklets.html#downevent which unfortunately isn't terribly useful, but I would be keen to receive comments or hints for improvement).

patrickhlauke:
i'll add a quickfire comment first of all that just saying native elements misses out one thing...if i use a `<button>` element, it doesn't do anything on its own (exception: it's part of a `<form>`, and does NOT have `type="button"`, in which case I seem to remember it's treated as a submit button).

i, as a dev, need to add my event listeners there to do anything. and crucially here then, we should recommend that authors just rely on input-agnostic `click` handlers (which are fired for keyboard, mouse, touch, stylus, etc, - click is essentially fired in all possible scenarios - and where the decision WHEN to fire the event - on the up or on the down - is left to the UA and is generally the up event). but if authors HAVE to use input-specific event handlers (perhaps they're doing different things for touch versus mouse versus keyboard ?), then they SHOULD listen to `mouseup` (if they're doing something mouse-specific), `touchend` (if they're doing something specific to touchscreens) or `pointerup` (in browsers that support pointer events, and pointer events are fired for any kind of pointer, so could be mouse/touch/stylus) depending which event model / input devices they want to support.

alastc:
Hi @DavidMacDonald,

Comments from the survey: https://www.w3.org/2002/09/wbs/35422/Tech_Und_survey/results#xq11
Plenty of details in there, the only one I wouldn't implement is the test procedure covering all scenarios. It doesn't need to do that, but the irreversible bit is odd.

alastc:
For paper-trail, this was accepted as amended: https://www.w3.org/2019/05/21-ag-minutes.html#item06

techniques/general/G211.html
```diff
@@ -0,0 +1,70 @@
...
+		<section id="examples">
+			<h2>Examples</h2>
+			<section class="example">
+				<h3>Using an native onclick event in JavaScript</h3>
+				<p>In JavaScript native onclick events are triggered on the up-event by default.</p>
+			</section>
+      		<section class="example">
+				<h3>Using a native button or link in HTML</h3>
+				<p>In HTML native <code>&lt;button&gt;</code> or <code>&lt;a href ....&gt;</code> onclick events are triggered on the up-event by default.</p>
+				<p class="note"> The <code>&lt;button&gt;</code> element doesn't do anything on its own (Unless its part of a <code>&lt;form&gt;</code>, and does NOT have <code>&lt;type="button"&gt;</code>, in which case it's  submits the form). Authors should rely on input-agnostic click handlers. <p>
```
patrickhlauke: typo and needs lowercase "(Unless its part" > "(unless it's part"; typo "in which case it's  submit the form" > "in which case it submits the form"

alastc: Updated, thanks.

PR #732, opened 2019-05-14
Label in Name Understanindg edits for consistency

mbgower:
Editorial corrections to match conventions and keep all technique wording consistent. Added one advisory technique which is already referenced in the Understanding doc

### DIFF
--- a/techniques/failures/F96.html
+++ b/techniques/failures/F96.html
@@ -1,11 +1,11 @@
 <!DOCTYPE html>
 <html xmlns="http://www.w3.org/1999/xhtml" lang="en">
 	<head>
-		<title>Failure of Success Criterion 2.5.3 due to &quot;accessible name&quot; not containing the visible label text</title>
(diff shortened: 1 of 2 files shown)


mbgower:
@michael-n-cooper and @awkawk . This PR should be completed in tandem with #649 to properly publish Label in Name with the 2 techniques approved by the WG last week. I have gone through and made sure there is consistency between the techniques and the Understanding document titles (in use of the term 'accessible name').
I also put in the boiler plate which seemed to be missing in the Understanding document -- the paragraphs that follow each heading -- and matched the headings of existing techniques.
Finally, Michael, I thought I'd point out that the currently published version is missing a heading although it IS in the Master version: the "Identifying label text for components" subheading in the Intent section. Not sure what happened there, but I assume when this is pulled it will all be perfect :)

alastc:
Hi @mbgower,

It's worth having a look at the XSL templates sometime, you'll find the boilerplate stuff is in there already which is why the live version already has it.

PR #733, opened 2019-05-14
Motion actuation errata from issue 727

### LINKED ISSUE #727: Motion actuation (2.5.4) question on exception
Is the ‘supported interface’ exception of Motion Actuation supposed to apply to disabling motion actuation?

To snip the [SC text](https://www.w3.org/TR/WCAG21/#motion-actuation) to this bit:
> ...responding to the motion can be disabled to prevent accidental actuation, except when:
> The motion is used to operate functionality through an accessibility supported interface;

What would an accessibility supported interface for disabling motion actuation be? Presumably that's an operating system thing? In which case wouldn't that come under accessibility supported?

Or is it (in this scenario) simply that the getting to and applying the setting is accessibility supported?

I think an little explanation of that in the understanding would be helpful.

alastc:
Adjusting the exception and little tweak to the understanding doc, based on 727.

Closes #727


understanding/21/motion-actuation.html
```diff
@@ -13,10 +13,10 @@ <h1><strong>Motion Actuation</strong>
...
+			<p>In addition, some users may accidentally activate sensors due to tremors or other motor impairments. The user must have the ability to turn off motion actuation to prevent such accidental triggering of functions. Applications may be able to meet this requirement by using an <a href="../conformance.html#accessibility-support">accessibility supported</a> operating system settings which allows the user to disable motion detection.</p>
```
patrickhlauke: Not convinced that the "accessibility supported" thing is relevant to this part, in reference to OS settings. it's outside of the scope of WCAG to determine whether or not the OS setting is accessibility supported or not, surely?

understanding/21/motion-actuation.html
```diff
@@ -13,10 +13,10 @@ <h1><strong>Motion Actuation</strong>
...
+			<p>Devices often have sensors that can act as inputs, such as accelerometer and gyroscope sensors on a phone or tablet device. These sensors can allow the user to control something by simply changing the orientation or moving the device in particular ways. In other situations, web content can interpret user gestures via the camera or other sensors to actuate functions. For example, shaking the device might issue an "Undo" command, or a gentle hand wave might be used to move forward or backward in a sequence of pages. Some users with disabilities are not able to operate these device sensors (either not at all, or not precisely enough) because the device is on a fixed mount (perhaps a wheelchair) or due to motor impairments. Functionality must be implemented in a way that other or additional means of activation are available.</p>
```
patrickhlauke: This is probably more the part where the concept of having an alternative "accessibility supported" interface is more relevant (e.g. where the OS or UA offers an alternative way of triggering/simulating the "sensor"/"motion" data, so the author doesn't perhaps necessarily need to provide their own controls)

patrickhlauke:
Think the understanding doc needs a more thorough look, as I feel it still talks about the "accessibility supported" aspect in the wrong place. but the change to the normative wording of the SC looks good to me per the other conversation.

understanding/21/motion-actuation.html
alastc: Wouldn't an OS setting like reduce-motion fit into this? I'm not saying that querry is useful here, but it is an example of an OS system setting that could be used in a similar way. If there were a similar query to prevent the detection of motion, that could work.

mbgower: > Wouldn't an OS setting like reduce-motion fit into this?

As noted in the original issue, we have not normally allowed OS-level features to serve as a workaround for author-implemented content. For example, authors are not allowed to point to iOS or Android touch assist features as a means of overcoming the requirements of Pointer Gestures. The exception I can think of to this is the user agent zoom feature to meet Resize, which is specifically called out in that SC as allowed.

PR #734, opened 2019-05-15
Added href value

svinkle:
This adds a missing `href` value to the "2.2.2 Pause, Stop, Hide" link.


alastc:
Hi @svinkle,

Thanks, good spot. The target should be the other understanding doc, it also goes through a build process so isn't always the obvious link!

I've updated this straight into master, it will go live soon, thanks!

PR #753, opened 2019-05-22
Tech icon font img role v3

alastc:
Taking over from the #531 PR.


alastc:
Hi Jake,

I re-created the branch as the original one (that I created!) was too old. I think I got everything (technique, 1 working example page, font folder).

Do you know where the fonts came from? (I.e. are the licensing issues or a license file for that needed?)

jake-abma:
Hi @alastc I created the fonts myself with https://icomoon.io/

jake-abma:
**What is the license for IcoMoon icons/fonts?**
There is a license link for each icon pack in IcoMoon's library.
If you are using your own vectors, there are no license requirements for using the IcoMoon app.

https://icomoon.io/#faq

alastc:
Ah, so you created it rather than using one with a license, perfect. :-)

PR #754, opened 2019-05-23
Change on Request update for issue #738

### LINKED ISSUE #738: 3.2.5: Change on Request Understanding update
https://www.w3.org/WAI/WCAG21/Understanding/change-on-request.html#benefits

The first benefit for 3.2.5 is:

Benefits

> - Individuals who are unable to detect changes of context or may not realize that the context has changed are less likely to become disoriented while navigating a site. For example:
>            - individuals who are blind or have low vision may have difficulty knowing when a visual context change has occurred, such as a new window popping up. In this case, warning users of context changes in advance minimizes confusion when the user discovers that the back button no longer behaves as expected.

The text "are less likely to become disoriented" must be "are more likely to become disoriented"

alastc:
Closes #738


jake-abma:
Hi @alastc 

> Individuals who are unable to detect changes of context, or may not realize that the context has changed if they did not initiate the change are less likely to become disoriented while navigating a site.

This one has exactly the same problem, you tell here "people unable to detect the change are less likely to become disoriented" and "people not realizing that the context has changed  are less likely to become disoriented" While it is more likely (if yout not checking the SC normative text at the top.

The part "less likely disoriented if they make the change themselves" is the key here

alastc:
Hi Jake, I'm sorry, I'm not getting what you mean, could you provide a text update to that line?

jake-abma:
SC: "Changes of context are initiated only by user request or a mechanism is available to turn off such changes."

One Benefit:

It was: (and stand alone may be misinterpreted)

> Individuals who are unable to detect changes of context or may not realize that the context has changed are less likely to become disoriented while navigating a site. For example:

Now it is: (and stand alone may **still** be misinterpreted)

> Individuals who are unable to detect changes of context, or may not realize that the context has changed if they did not initiate the change are less likely to become disoriented while navigating a site.

Make it:

> Detecting or realising the context has changed after a request initiated by the user will make it less likely for the user to become disoriented while navigating a site.

understanding/20/change-on-request.html
```diff
@@ -56,8 +56,8 @@ <h2>Benefits of Change on Request</h2>
          
          <li>
             
-            <p>Individuals who are unable to detect changes of context or may not realize that the
-               context has changed are less likely to become disoriented while navigating a site.
+            <p>Individuals who are unable to detect changes of context, or may not realize that the
+               context has changed if they did not initiate the change are less likely to become disoriented while navigating a site.
```
patrickhlauke, line 60:
```suggestion
            <p>Individuals who are unable to detect changes of context, or may not realize that the
               context has changed if they did not initiate the change, are less likely to become disoriented while navigating a site.
```

patrickhlauke: I *think* the above suggestion would gets to what @jake-abma was saying? (though admittedly, I wasn't 100% sure about what he was getting at, myself)

patrickhlauke:
@alastc suggest that instead of leaving this to linger, we proceed with it? I believe the suggested rewording (from my suggestion) may be sufficient to address potential confusion.

understanding/20/change-on-request.html
mbgower: I think just adding a second 'who' clarifies
```suggestion
            <p>Individuals who are unable to detect changes of context or who may not realize that the context has changed are less likely to become disoriented while navigating a site.
```

patrickhlauke: but both those sentences say the same thing, unless i'm missing something? "people who are unable to detect change, or people who are unable to detect change if they didn't initiate it" ...

patrickhlauke:
merged my suggested simplification/rewording. if that was too rash, please back it out again

bruce-usab (approved):

understanding/20/change-on-request.html
```diff
@@ -56,8 +56,7 @@ <h2>Benefits of Change on Request</h2>
          
          <li>
             
-            <p>Individuals who are unable to detect changes of context or may not realize that the
-               context has changed are less likely to become disoriented while navigating a site.
+            <p>Individuals who may not realize that the context has changed if they did not initiate the change themselves are less likely to become disoriented while navigating a site.
```
mbgower, line 59: The original issue was triggered because the benefit statement was interpretted as a problem statement. By adding the 'solution' to the benefit statement, it becomes clear it is in the form of a [problem: solution] phrase
 ```suggestion
            <p>Individuals who are unable to detect changes of context or who may not realize that the context has changed are less likely to become disoriented if they initiate the change themselves.
```

patrickhlauke: either one works for me (though i still think the first part says the same thing twice... "are unable to detect changes of context" OR "may not realize the context has changed" seems the same, with perhaps subtle shades of severity only)

patrickhlauke:
need to check why the netlify preview failed there...some extra conflict somewhere?

bruce-usab:

PR #760, opened 2019-06-01
Tech providing single point control sliders

alastc:
Detlev's technique, just creating a PR.
[Rendered version](https://raw.githack.com/w3c/wcag/tech-providing-single-point-control-sliders/techniques/general/ensuring-single-pointer.html)


detlevhfischer:
The text above example 1 needs an update. It now says:

> A custom control slider built on top of a native slider (input type range) allows users to swipe left and right or drag the slider thumb to change the value of the slider. The slider track allows tapping or clicking to set the slider thumb to the activated position. 

We should add: "Buttons to the left and right of the slider also allow single point activiation for incremental changes to the value", or similar.

alastc:
Hi @detlevhfischer, I've updated that with a few extra words. In general I've tried to avoid 'single point activation' as that could include swipes/gestures.

detlevhfischer:
Hi @alastc - I think writing 

> The slider track allows tapping or clicking to set the slider thumb to the activated position, as do buttons on either side.

is somewhatr mnisleading clickign the track sets the positino arbitrarliy while activating the buttons increments / decrements, so this should be more explicit.
As to "single point activation" - I understand that 'single pointer' can be a swipe, but "single point activation" surely means tapping / clicking etc at ONE position? That was my understanding (I may have even coined that term to be precise that it is one, and just one, point on screen).

alastc:
OK, more explicit now. 

Single point "activation" doesn't make it specific to a postition, at least to me. Maybe I've been reading it as "single point**er** activation", but I'd rather avoid it and say tap/click/press or whatever. That's something I've (tried to) remove from the understanding doc.

detlevhfischer:
I think the current title of the Technique ("Providing simple pointer activation for all pointer operations") does not describe very clearly what the technique is about. This is about providing single point activation alternatives for the dragging of a control slider, ad that is what authors looking for practical implementations are likely to look for. Why not say so?  I find "all pointer operations" so generic as to be quite confusing. It's like saying "providing a cup for the hydration of living matter" instead of "Providing a cup for drinking".

alastc:
> it means activation at a single point on the display 

Ok, by I had to think hard about that (in comparison to "single pointer" which we use and define elsewhere). If you are thinking about the point behing your finger (which I was), it does not differentiate from a drag/gesture. 

Can we not be more basic? I.e. Tap or click.

detlevhfischer:
> Can we not be more basic? I.e. Tap or click.
Yes, sure!

techniques/general/ensuring-single-pointer.html
```diff
@@ -0,0 +1,58 @@
...
+			<p id="type">sufficient</p>
+		</section>
+		<section id="applicability">
+			<h2>When to use</h2>
+			<h3>Applicability</h3>
+			<p>Any technology that supports pointer input (e.g. supporting any or all of the following: mouse pointer, touch on touch screen or trackpad, stylus input, or laser pointer input).</p>
+		</section>
+		<section id="description">
+			<h2>Description</h2>		
+			<p>The objective of this technique is to ensure that users who have difficulties performing path-based gestures can operate a content slider with a single pointer (e.g., a single tap on a touch screen or a single mouseclick). A content slider contains chunks of content in a row. Usually several chunks of content are hidden, only one chunk is visible at any time. Left and right horizontal swiping over the visible part of the slider brings adjacent hidden chunks of content into view. Providing controls (for example, arrow buttons) as alternative means of input allows advancing the slider with single pointer input.</p>
```
patrickhlauke: "mouseclick" > "mouse click" ?

patrickhlauke: "are hidden, only one chunk" > "are hidden, and only one chunk"

techniques/general/providing-single-point-control-slider.html
```diff
@@ -19,37 +19,34 @@ <h3>Applicability</h3>
...
 				<img src="img/input-range.png" alt="A horizontal line with a round button halfway along the line, labelled volume." />
-				<figcaption>Screenshot of an example slider showing the thumb control in the middle of the track.</figcaption>
+				<figcaption>Screenshot of an example control slider showing the thumb in the middle of the track.</figcaption>
 			</figure>
 			
-			<p>Note: Excluded from scope for this technique are alternatives for the cancellation action achieved by a user when moving the pointer from the target area after a down-event has been initiated. This form of cancellation is well-established and the intent of this technique is not to force an alternative method of cancellation, although other methods for Undo may be feasible in many circumstances.</p> 
+			<p>It allows a user to set a value in a certain range, e.g. setting the volume, changing the hue value of a color, putting in the amount of money needed in a loan calculator, or picking a sum to be donated to a charity. A slider that requires path-base gestures would use swiping left or right to change the value, or dragging the thumb of the slider in a specific direction to the change the value.</p>
```
patrickhlauke: "path base gestures" > "path based gestures"

patrickhlauke: "to the change the value." > "to change the value."

techniques/general/providing-single-point-control-slider.html
```diff
@@ -19,37 +19,34 @@ <h3>Applicability</h3>
...
+
+        	<p>A simple fallback for activation without a path-based gesture is to make the control slider track clickable. This way, a value can be specified using single tap or click on the track.</p>
```
patrickhlauke: "using single tap or click" > "using a single tap or click"

techniques/general/providing-single-point-control-slider.html
```diff
@@ -0,0 +1,66 @@
...
+     		<p>Providing controls (e.g., arrow buttons) as alternative also allows incrementing or decrementing the value with single pointer input. This allows for a more fine-grained setting of the value.</p>
+		
+      		<p class="note">On touch screen devices, author-supplied path-based gestures usually do not work when OS level assistive technologies (AT) like a built-in screenreader are turned on. This is because AT generally consumes path-based gestures so they would not reach the authored content. When custom controls are built on top of native controls, however, these may also be operable with AT gestures such as vertical swiping to change the value (see example 1).</p>
```
patrickhlauke: "AT generally consumes" > "AT generally consume" ("AT" being plural?)

awkawk: I think of AT = Assistive Technology (singular)

patrickhlauke: > I think of AT = Assistive Technology (singular)

but in the sentence just preceding it it's been set as plural `OS level assistive technologies (AT)`. should both be the same either way, i'd say.

techniques/general/ensuring-single-pointer.html
alastc: Done

techniques/general/providing-single-point-control-slider.html
alastc: Updated to be consistent (singular).

alastc:
Accepted today: https://www.w3.org/2019/07/16-ag-minutes.html#item08

PR #763, opened 2019-06-04
Revert "general techniques for label in name"

awkawk:
Reverts w3c/wcag#649


awkawk:
Need to fix tech numbers

PR #769, opened 2019-06-05
Remove WAI glossary text on Charter

slhenry:
a completed / published "WAI Glossary" doesn't exist


slhenry:
"Internal Working Draft". Valiant effort! And EOWG worked on Glossary, too, years ago -- just other things took higher priority.

I didn't mean to bring up an touchy issue -- just correct the draft charter. :-)

PR #774, opened 2019-06-10
Adding suggestion from 228

### LINKED ISSUE #228: Add some actionable advice to "Large scale (text) > Note 1" in the glossary

alastc:
Originally from https://github.com/patrickhlauke/wcag/commit/c15ab1771b8bc61e5cef5bfea1e1e26ad6cf43bd

Michael, I assume this HTML file is what's used for the defintitions?

Closes #228

### DIFF
--- a/guidelines/terms/20/large-scale.html
+++ b/guidelines/terms/20/large-scale.html
@@ -8,7 +8,7 @@
    
    <p class="note">Fonts with extraordinarily thin strokes or unusual features and characteristics that
       reduce the familiarity of their letter forms are harder to read, especially at lower
-      contrast levels.
+      contrast levels. Authors may choose to exceed the recommended point sizes / relative size differences from the base font size in order to compensate for this.
    </p>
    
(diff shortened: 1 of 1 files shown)


patrickhlauke:
(added the "Closes..." bit in the description, which then autocloses the issue when this gets merged)

alastc:
Ah, didn't know that. @awkawk did you? If not, that will make surveys & calls easier! :-)

PR #775, opened 2019-06-10
Updating 3rd example in H80

### LINKED ISSUE #233: H80 Example 3
Example 3 is imo not a good example for H80. The "more here"-link is part of a paragraph which offers (also) information about the purpose of the link, which would be H78. I think a more concise example is needed to show the difference between H78 and H80.

alastc:
[Preview](https://raw.githack.com/w3c/wcag/1405a037ca0483af7e423fc61fef74341cbdc525/techniques/html/H80.html) and scroll to the newspaper example.

Closes #233


patrickhlauke:
depends how the script adds the link. might need to clarify that it adds it add the end, and not as part of the content/paragraph of the card.

patrickhlauke:
yup, much better :)

PR #767, opened 2019-06-05
Fixes for Understanding 1.3.3 Sensory Characteristics

patrickhlauke:
- expand/refocus the intent and benefits to be more explicitly about *instructions* (removing the vague implication that controls/content themselves shouldn't use shape/colour/etc alone)
- swap out first example, which was a failure of 1.1.1 not 1.3.3, with @awkawk's suggested example

Closes https://github.com/w3c/wcag/issues/750

[edit: since F26 was modified - see https://github.com/w3c/wcag/pull/767#issuecomment-516050840 - leaving these last two points out]
> - remove reference to failure technique F26, which is at best a 1.1.1 failure and not a 1.3.3 failure
> - remove F26 from techniques altogether
>
> Closes #258 

*Note to self: this seems to be the last active/open PR still on my old WCAG fork - once this is merged or thrown out, the fork can be removed...*

PR #776, opened 2019-06-10
F26 update to address 258

### LINKED ISSUE #258: Failure F26 links to the wrong SC?
From a reading of SC 1.3.3 I gather that it does not relate to images but to instructional text that
refer to other components identifying them by their shape or other
visual attribute.
In my opinion, F26: "Failure of Success Criterion 1.3.3 due to using a graphical symbol alone to convey information" does not seem to  have anything to do with 1.3.3. It is a failure of SC 1.1.1 because the symbol has no text alternative.

Thanks,
Sailesh

alastc:
Closes #258


patrickhlauke:
@alastc note that https://github.com/w3c/wcag/pull/767 actually deletes F26 (i had forgotten about that other issue #258)

alastc:
Gah, that's what happens when you start from the bottom. Ok, this is a fallback if that one doesn't go through then.

alastc:
So, somewhat confusingly this update was approved (to close #258), so I'll merge it in.

It may well be that the whole technique is removed later, but in the meantime there is one less PR & issue to deal with :-)

Issue #264, opened 2017-02-07
Tweak glossary definition for "images of text"


patrickhlauke:
(prompted by https://lists.w3.org/Archives/Public/w3c-wai-gl/2017JanMar/0642.html)

https://www.w3.org/TR/WCAG21/#dfn-images-of-text currently reads

> text that has been rendered in a non-text form (e.g., an image) in order to achieve a particular visual effect

This definition arguably limits the applicability of 1.4.3 to only apply to images of text where the intent of the author was to achieve a particular visual effect - what about images of text where the author had some other reason to use an image of text? Without actually trying to apply any particular effect, maybe the author simply thought it was the best thing to do? Or their tool/development environment only allowed images to be used, rather than actual HTML text?

I'd tweak the definition here to make the second part an example of intent, rather than seeming to imply it's the only intent that matters.

> text that has been rendered in a non-text form (e.g., an image); often, this is done in order to achieve a particular visual effect

patrickhlauke:
further the note

> Note: This does not include text that is part of a picture that contains significant other visual content.

seems like a get-out clause. what if the text is essential to understand the rest of the picture? or if it presents a separate piece of information that the other parts of the picture? why would it then be exempted, if a user's inability to properly perceive it will result in the user not getting the information/content it conveys (in the case of contrast)?

DavidMacDonald:
We should also say something in the understanding about our opinion on text that is in a timed media. Reasons that I haven't been watching videos to fail text contrast under 1.4.3 is because:

1) It would be a significant requirement, but it text in animations and movies is not mentioned in the understanding, which limits itself to "pixels" and static images.
2) The video was not likely rendered in non-text form for the purpose of achieving a particular effect on the text. It was probably made for a different reason.
3) I don't remember any discussion of text contrast in videos during the creation of WCAG 1.4.3.

jasonjgw:
+1 to expanding the scope of what is covered by the definition, as proposed. The purpose of using the image of text has nothing to do with determining whether there is a need for proper contrast or whether it should be provided to the user.

alastc:
I created an update for this, as suggested by Patrick.

It doesn't address the time-based media (which would be better done in the understanding doc), but would apply as an errata...

alastc:
@patrickhlauke there was a good reason for the note, I've updated it in the PR: 
> This does not include **incidental** <a>text</a> that is part of a picture that contains significant other visual content, **such as a picture which includes street signs**.

[...]

PR #777, opened 2019-06-10
Update from issue 264

alastc:
#264

### DIFF
--- a/guidelines/terms/20/image-of-text.html
+++ b/guidelines/terms/20/image-of-text.html
@@ -1,11 +1,8 @@
 <dt><dfn data-lt="images of text">image of text</dfn></dt>
 <dd>
    
-   <p>text that has been rendered in a non-text form (e.g., an image) in order to achieve
+   <p>text that has been rendered in a non-text form (e.g., a bitmap image). Often this is done in order to achieve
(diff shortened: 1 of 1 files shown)


alastc:
Updated from the dicsussion on Aug 13th. The note did have a good purpose - excluding incidental text.

However, I think it is worth updating as I read that as including things like info graphics. I.e. if you included a graph then the labels would be excluded, not the intent!

alastc:
To clarify from the last call, this (editorial?) change is to close a small & unintended loophole.

From the SC text for 1.4.5:
> If the technologies being used can achieve the visual presentation, text is used to convey information rather than images of text.

Images of text are defined as:
> text that has been rendered in a non-text form (e.g., an image) **in order** to achieve a particular visual effect 

Therefore if an image is **_not_** trying to achieve a particular visual effect, it is not an image of text... very odd!

E.g. this would not qualify as an 'image of text' under the definition:
A regular bit of text saying All checks have passed

The other change is that the note appeared to exclude things like info graphics:
> This does not include text that is part of a picture that contains significant other visual content. 

Infographics were not a big thing in 2006-8, but are common now. Minor update to reflect the intent as discussed on the last call.

PR #790, opened 2019-06-18
Tech failure doorslam

jake-abma:
Doorslam Technique for orientation.

[Preview](https://raw.githack.com/w3c/wcag/tech-failure-doorslam/techniques/css/tech-ask-reorient-device.html)
(The example is linked from the preview of the technique.)

PR #791, opened 2019-06-18
Example doorslam

jake-abma:
See: https://github.com/w3c/wcag/pull/790


alastc:
I've added the working example to the technique branch.

PR #793, opened 2019-06-19
Tech no reliance on path or multipoint

alastc:
Creating the PR from Detlev


mbgower:
@detlevhfischer is there an issue associated with this?

mbgower:
@detlevhfischer I realize this is more than 5 years old. I would also appreciate you indicating if you believe this PR is covered by https://www.w3.org/WAI/WCAG22/Techniques/general/G215

If you believe so, please close.

patrickhlauke:
@detlevhfischer having reviewed this PR and the now existing [Technique G215: Providing controls to achieve the same result as path based or multipoint gestures](https://www.w3.org/WAI/WCAG22/Techniques/general/G215) I *think* the latter covers the same ground as this proposed technique. I'm inclined to therefore close this. If there are aspects that you think are not sufficiently covered in G215 that you'd like to see there, can I suggest making a fresh PR against that technique?

detlevhfischer:
@patrickhlauke @mbgower The test in G215 needs work, I think. Currently it is limited to testing content sliders ("For content sliders that respond to path-based gestures...") but there can be other content relying on complex gestures (path-based or multipoint) which are not covered. So it should say instead something like:
> For any functionality that is operated by a multipoint or path-based gesture and does not fall under the essential exception:

Also, one could consider using the wider set of examples (possibly modified) I had in my technique draft, such as controls for panning a map where zooming out and zooming in to another location by tapping / clicking on aother place in the map is not supported. Another case could be a different swipe gesture: swipe left to delete (or reveal a delete button), swipe right to archive (or reveal an archive button). The latter could exemplify a way of meeting the SC by tapping on the item to go to an item detail view where you then activate the delete and archive buttons (arguably not quite equivalent in terms of effort, but I guess we would call that a PASS?) So I guess instead of saying "check that single pointer operable controls exist over or adjacent to the content slider" it should something more general, perhaps like
> "Check that for the functionality implemented by the gesture there exists an alternative that can be operated by a simple pointer input (or a sequence of simple pointer inputs)."

Apart from improving the test, I agree that G215 already does the job in principle, and do not mind if you close my PR.

patrickhlauke:
filed new issue/follow-up here https://github.com/w3c/wcag/issues/4892

PR #794, opened 2019-06-19
Tech control orientation

awkawk:
Pull request for General technique for orientation

Preview: http://htmlpreview.github.io/?https://github.com/w3c/wcag/blob/tech-control-orientation/techniques/general/control-orientation.html


patrickhlauke:
One aspect that may be confusing here is that in general (at least for phone/tablet use), users engage orientation locks at the OS level when they want to lock things into a certain orientation (i.e. in portrait mode, they lock it into that mode before lying on their side). But I gather you mean here is that it's the content itself that is doing the locking, counteracting what the normal behaviour would be - which would be a failure on its own - but is offering a custom control to then also allow orientation switching - making it ok again in light of the SC?

Maybe see if it can be emphasised that by default the content would adapt to landscape mode, but that the developer has explicitly counteracted this / locked the orientation, to cater for the "reading on their side" use case?

techniques/general/control-orientation.html
```diff
@@ -0,0 +1,53 @@
...
+			<p>When the orientation of the page is locked, provide a button to allow a user to change the orientation.</p>
+		</section>
+		<section id="description">
+			<h2>Description</h2>
+			<p>The objective of this technique is to allow users to access content whether their device is oriented in portrait or landscape view. A content provider may expect that most users will view content using a specific orientation or may expect that a user will want to maintain the original view within the device, and as a result the provider prevents the content from rotating in response to device information. By providing a control to allow the user to rotate the content a user who uses their device in a fixed orientation, such as a user who cannot hold the device and has it mounted attached to a wheelchair or bed, will be able to view the content in a comfortable manner.</p>
+		</section>
+		<section id="examples">
+			<h2>Examples</h2>
+			<section class="example">
+				<h3>A e-Reader application</h3>
```
patrickhlauke: A e-Reader > An e-Reader

(as in the next para)

patrickhlauke:
also...is the example of an e-Reader application appropriate for web content guidelines? the e-Reader would be like a user agent of sorts. maybe at least e-Reader *web* application?

awkawk:
@patrickhlauke sure - web app

awkawk:
@patrickhlauke It may seem a bit backward relative to what we would hope people do, but here's the logic:
1) The SC exists because people do restrict content to one viewing orientation or another.
2) Some people may feel that there is a good reason to do this.
3) This technique provides a way for them to still conform.

Perhaps what we should add is a note that says:
Many users will expect to use OS-level settings to prevent the orientation changing, so it is usually preferable to satisfy this Success Criterion by not restricting the content's ability to adapt to changes in orientation and allowing users to use the OS orientation lock when desired.

patrickhlauke:
On second reading, having worked out the intent, it actually sounded ok. But yes, having a note along those lines could help.

techniques/general/control-orientation.html
```diff
@@ -0,0 +1,54 @@
...
+			<p id="id">ID: @@G210</p>
+			<p id="technology">Technology: General</p>
+			<p id="type">Type: Technique</p>
+		</section>
+		<section id="applicability">
+			<p>When the orientation of the page is locked, provide a button to allow a user to change the orientation.</p>
+		</section>
+		<section id="description">
+			<h2>Description</h2>
+			<p>The objective of this technique is to allow users to access content whether their device is oriented in portrait or landscape view. A content provider may expect that most users will view content using a specific orientation or may expect that a user will want to maintain the original view within the device, and as a result the provider prevents the content from rotating in response to device information. By providing a control to allow the user to rotate the content a user who uses their device in a fixed orientation, such as a user who cannot hold the device and has it mounted attached to a wheelchair or bed, will be able to view the content in a comfortable manner.</p>
```
alastc: Readability suggestion: The objective of this technique is to allow users to access content whether their device is oriented in portrait or landscape view. A content provider may expect that most users will view content using a specific orientation or may expect that a user will want to maintain the original view within the device. As a result the provider then prevents the content from rotating. By providing a control to allow the user to rotate the content, someone who needs to use a particular orientation will be able to view the content in a comfortable manner. For example, a person who cannot hold the device and has it mounted attached to a wheelchair or bed.

mbgower:
Created #821 to address some readability issues

alastc:
Accepted today:
https://www.w3.org/2019/07/16-ag-minutes.html#item06

PR #806, opened 2019-06-28
Initial draft to address syc-media #796

### LINKED ISSUE #796: Remove exception phrasing from 1.2.1, 1.2.2, and 1.2.3
There is inconsistent phrasing used within the SC of Guideline 1.2.  This should be corrected.

SC 1.2.1, 1.2.2, and 1.2.3 all include the condition:  "except when the media is a media alternative for text and is clearly labeled as such".

This conditional is _not_ used in other SC with similar structure (namely 1.2.5, 1.2.6, 1.2.7, and 1.2.8).

The conditional is not necessary (in the SC) because the [normative definition for synchronized media](http://www.w3.org/TR/WCAG20/#synchronizedmediadef) already includes that qualification.
> audio or video synchronized with anotherformat for presenting information and/or with time-based interactive components, **unless the media is a media alternative for text that is clearly labeled as such** 

In other words, media alternatives for text are _not_ synchronized media.

Proposal:  In 2.2, we should remove the exception phrasing from 1.2.1, 1.2.2, and 1.2.3.

alastc:
Only applied to the AA guideline so far, could apply to the AAA guidelines depending on discussion.

Closes #796.

### DIFF
--- a/guidelines/sc/20/audio-description-prerecorded.html
+++ b/guidelines/sc/20/audio-description-prerecorded.html
@@ -5,7 +5,7 @@
    <p class="conformance-level">AA</p>
    
    <p><a>Audio description</a> is provided for all <a>prerecorded</a>
-      <a>video</a> content in <a>synchronized media</a>.
+      <a>video</a> content in <a>synchronized media</a>, except when the media is a <a>media alternative for text</a> and is clearly labeled as such.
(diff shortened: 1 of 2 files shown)


alastc:
Closing as didn't get traction.

Issue #297, opened 2017-07-24
Clarification on 3.1.2 Language of Parts


goodwitch:
So, I need clarification on [3.1.2 Language of Parts ](https://www.w3.org/TR/WCAG20/#meaning-other-lang-id)

Is it a failure if 
- (an html web site has alternative language links to versions of the page in other languages (e.g., Deutsch, Français, Nederlands, Castellano, etc.)) AND
- (the text of each link is the name of the language, in that language) AND
- (the language of each link is indicated via a lang attribute) AND
- (the text of each language link does not have a language attribute to indicate the change in language?

Why am I asking this question?  Because I've been calling this as a failure...but one of my colleagues just pointed out that the normative part of this SC says "passage or phrase", so they tell me, that since these language links are single words, they are not required to be marked.  

My brain says...sometimes a phrase can be a single word.  And I've found evidence of this in linguistics.  

If you agree that this is a failure, would it be worthwhile to add this as a failure technique for 3.1.2 in WCAG 2.0?

Thanks for considering this.

Ryladog:
My interpretation has always been that single word links to other language pages that appear in the secondary language is in fact a failure.

 

​​​​​* katie *

 

Katie Haritos-Shea 
Principal ICT Accessibility Architect (WCAG/Section 508/ADA/AODA)

 

Cell: 703-371-5545 |  <mailto:ryladog@gmail.com> ryladog@gmail.com | Oakton, VA |  <http://www.linkedin.com/in/katieharitosshea/> LinkedIn Profile | Office: 703-371-5545 |  <https://twitter.com/Ryladog> @ryladog

NOTE: The content of this email should be construed to always be an expression of my own personal independent opinion, unless I identify that I am speaking on behalf of Knowbility, as their AC Rep at the W3C - and - that my personal email never expresses the opinion of my employer, Deque Systems.

srinivasuchakravarthula:
I agree. 

Regards,
Srinivasu Chakravarthula
+91-9900810881
Sent on my phone. Excuse typos, if any.

kerstinp:
@goodwitch, all,

this example is mentioned in the Understanding of SC 3.1.2 (Example 2). So it seems that a single word can be seen as a "phrase". But because there is no definition of "phrase" in the glossary it's not very lucky when just the informative understanding document indicates this. That's why I filed an issue for WCAG 2.1 that a "definition of phrase in the glossary is needed": https://github.com/w3c/wcag21/issues/287.

[...]

PR #808, opened 2019-06-28
Add clarifications to understanding 3.1.2

patrickhlauke:
Closes https://github.com/w3c/wcag/issues/297


awkawk:
@patrickhlauke thanks for the suggestion. This strikes me as not correcting anything that is _wrong_ so as we struggle to finish off techniques and work on 2.2 SC's this (and items others propose that are of the same sort) are likely to take a back seat.

patrickhlauke:
ping @alastc (while i'm going through my PR left hanging open for ages)

patrickhlauke:
Ping @alastc

alastc:
@patrickhlauke was this closed due to delay, or because it wasn't right/helpful?

I'm going back through old issues now, this appears to be a solution to a couple of issues.

patrickhlauke:
@alastc I think I closed it in a fit of "let me just cull things that have been hanging for 2+ years and have seen no action". Reopened.

mbgower (approved):
useful change that provides some clarity

understanding/20/language-of-parts.html
```diff
@@ -13,7 +13,7 @@ <h2>Intent of Language of Parts</h2>
       
       
       <p>The intent of this Success Criterion is to ensure that user agents can correctly present
-         content written in multiple languages. This makes it possible for user agents and
+         words, phrases or passages written in multiple languages. This makes it possible for user agents and
```
alastc, line 16:
```suggestion
         words, phrases, or passages written in multiple languages. This makes it possible for user agents and
```

understanding/20/language-of-parts.html
```diff
@@ -13,7 +13,7 @@ <h2>Intent of Language of Parts</h2>
       
       
       <p>The intent of this Success Criterion is to ensure that user agents can correctly present
-         content written in multiple languages. This makes it possible for user agents and
+         words, phrases, or passages written in multiple languages. This makes it possible for user agents and
```
dbjorge, line 16: Agree with @WilcoFiers 's point in today's meeting that it's problematic for this to be more strict than the normative wording of the SC, which is just "passage or phrase" rather than "words, phrases, or passages". I think this wording should match the normative SC wording.

understanding/20/language-of-parts.html
```diff
@@ -136,6 +136,13 @@ <h2>Examples of Language of Parts</h2>
             <p>An HTML Web page includes links to versions of the page in other languages (e.g.,
                <span lang="de">Deutsch</span>, <span lang="fr">Français</span>, <span lang="nl">Nederlands</span>, <span lang="ca">Catalan</span>, etc.). The text of each link is the name of the language, in that language. The language of each link is indicated via a <code class="att">lang</code> attribute.
             </p>
+<pre><code><ul>
+  <li><a href="..." lang="de">Deutsch</a></li>
+  <li><a href="..." lang="it">Italiano</a></li>
+  <li><a href="..." lang="fr">Français</a></li>
+  ...
+  <li><a href="..." lang="zh-hant">繁體中文</a></li>
+</ul></code></pre>
```
WilcoFiers, line 145:
```suggestion
<pre><code>&lt;ul&gt;
  &lt;li&gt;&lt;a href="..." lang="de"&gt;Deutsch&lt;/a&gt;&lt;/li&gt;
  &lt;li&gt;&lt;a href="..." lang="it"&gt;Italiano&lt;/a&gt;&lt;/li&gt;
  &lt;li&gt;&lt;a href="..." lang="fr"&gt;Français&lt;/a&gt;&lt;/li&gt;
  ...
  &lt;li&gt;&lt;a href="..." lang="zh-hant"&gt;繁體中文&lt;/a&gt;&lt;/li&gt;
&lt;/ul&gt;</code></pre>
```

patrickhlauke:
oops, good catch @WilcoFiers ... clearly i spend too much time doing markdown these days and forget about escaping/encoding angle brackets

understanding/20/language-of-parts.html
mbgower: To me, the idea of "words" being applicable is part of the normative text

> except for proper names, technical terms, words of indeterminate language, and words or phrases that have become part of the vernacular of the immediately surrounding text.

As well, there is substantial existing wording in the Understanding document about words, especially the following sentence:
> Because this is so common in some languages, single words should be considered part of the language of the surrounding text unless it is clear that a change in language was intended. 

 These things, together, make it clear to me that replacing "content" with "words, phrases, or passages"  in the _non-normative_ text intro is accurate, more clear, and supported by the existing normative and non-normative language.

WilcoFiers: Adding "words" to "passages and phrases" is not supported by the normative text IMO. Mike, that you need to point to the exception says everything to me. If the intent was for this to include every individual word, the success criterion would have said so explicitly. 

There are good reasons not to be this strict. It's really hard to decide consistently what words are part of jargon or idioms. This is a problem in places that tend to be multilingual like The Netherlands and Denmark. Most people here speak English, and because of our exposure to English media there are English words that are more familiar to people than their native counterparts.

Switching back and forth between speech synthesizers in the middle of a sentence can make that sentence far more difficult to understand than having a word mispronounced slightly. If your screen reader has a similar-sounding voice for both languages, that's less problematic than if it uses two completely different voices. It also matters how close the two languages are. Having an German word pronounced with a Dutch voice makes almost no difference.

IMO, individual words in a sentence should not be required to have a lang attribute. Doing so is likely to hinder the comprehension of that text, it's difficult to evaluate consistently, and it's a significant burden on content authors to boot.

WilcoFiers (dismissed):

understanding/20/language-of-parts.html
mbgower: "there are English words that are more familiar to people than their native counterparts." Then those meet the exception.

alastc, line 16:
```suggestion
          phrases, passages and some words written in multiple languages. This makes it possible for user agents and
```

alastc: As discussed in the meeting, a [phrase can be a word](https://en.wikipedia.org/wiki/Phrase), but many people haven't understood that. That doesn't mean any word, but some. [Anyway, approved in the meeting](https://www.w3.org/2023/05/23-ag-minutes#item10).

patrickhlauke:
of course the "in some cases" is very handwavy. considering that the understanding should aim to clarify, rather than handwave, things...I suggest some more future work on explaining this part...

cstrobbe:
@alastc Does this pull request also close #1174?

patrickhlauke:
@cstrobbe it probably does, in a slightly handwavy way

patrickhlauke:
also x-ref https://github.com/w3c/wcag21/issues/287 (though that's in an archived repo)

patrickhlauke:
"words (unless part of the vernacular / loanwords)" I guess

PR #818, opened 2019-07-08
JS error in script-expand-links working example, issue 817

### LINKED ISSUE #817: JavaScript error on working example (script-expand-links)
When trying to see the effects of the [script-expand-links working example](https://www.w3.org/WAI/WCAG21/working-examples/script-expand-links/), I got a JS error in my console that prevented it from working (on Win FF, Chrome, and Edge).

It's choking on line 16 (`for each (link in links)`) since it seems that `for each...in` [has been deprecated](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for_each...in). I tried a fix trying to be new school (`for (const link of links)`, I think) but it choked on Edge, then tried `for (const link of Array.from(links))` but that still choked on IE11, so I'll be by shortly with a pull request using old-school `for`.

maxstarkenburg:
Should fix #817 

As noted in the issue, I only had luck in all the browsers I tried using a regular `for` loop, and as noted in the commit message, I didn't actually test the code in the techniques file (just copied and pasted the same diff from the working example).

### DIFF
--- a/techniques/client-side-script/SCR30.html
+++ b/techniques/client-side-script/SCR30.html
@@ -23,9 +23,10 @@ function doExpand() {
 	var links = document.links;
 	
 	for (var i=0; i<links.length; i++) {
-		var cn = links[i].className;
+		var link = links[i];
+		var cn = link.className;
 		if (linkContext[cn]) {
-			span = links[i].appendChild(document.createElement("span"));
+			span = link.appendChild(document.createElement("span"));
 			span.setAttribute("class", "linkexpansion");
 			span.appendChild(document.createTextNode(linkContext[cn]));
 		}
--- a/techniques/client-side-script/SCR30.html
+++ b/techniques/client-side-script/SCR30.html
@@ -22,7 +22,7 @@ var linkContext = {
 function doExpand() {
 	var links = document.links;
 	
-	for (var i=0; i<links.length; i++) {
+	for (var i=0; i&lt;links.length; i++) {
 		var link = links[i];
 		var cn = link.className;
 		if (linkContext[cn]) {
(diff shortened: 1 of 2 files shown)


techniques/client-side-script/SCR30.html
```diff
@@ -22,10 +22,10 @@
 function doExpand() {
 	var links = document.links;
 	
-	for (link of links) {
```
patrickhlauke: could probably just change this to a `links.forEach(link) {` without touching the rest of the code

working-examples/script-expand-links/index.html
```diff
@@ -13,10 +13,10 @@
 function doExpand() {
 	var links = document.links;
 	
-	for each (link in links) {
```
patrickhlauke: ditto

techniques/client-side-script/SCR30.html
patrickhlauke: actually, just tried, and seems `links` isn't a proper array but a collection...so yeah, just ignore. better to just be explicit with a for loop...

maxstarkenburg: Yeah, I had no luck w/ `forEach` either, but I did just push another commit that hopefully makes the overall change a bit more terse.

alastc:
The [preview](https://raw.githack.com/maxstarkenburg/wcag/issue817-expand-links-example-broken/working-examples/script-expand-links/index.html) works for me...

alastc:
Thank you for solving that @maxstarkenburg. As it's solving a bug and not changing the meaning of the technique I can just merge this in. (Other types of changes need to go through more process.)

PR #821, opened 2019-07-12
Update control-orientation.html

mbgower:
Rewrite for readability and clarity

### DIFF
--- a/techniques/general/control-orientation.html
+++ b/techniques/general/control-orientation.html
@@ -1,11 +1,12 @@
 <!DOCTYPE html>
 <html xmlns="http://www.w3.org/1999/xhtml" lang="en">
 	<head>
-		<title>Using a control to allow override of content orientation restrictions</title>
+		<title>Providing a control to allow a user to change the orientation when it is locked by the content</title>
 		<link rel="stylesheet" type="text/css" href="https://rawgit.com/w3c/wcag21/master/css/sources.css"/>
 	</head>
 	<body>
-		<h1>Using a control to allow override of content orientation restrictions</h1>
+		<h1>Providing a control to allow a user to change the orientation when it is locked by the content</h1>
+		
 		<section id="meta">
 			<h2>Metadata</h2>
 			<p id="id">ID: @@G210</p>
--- a/techniques/general/control-orientation.html
+++ b/techniques/general/control-orientation.html
@@ -25,7 +25,7 @@
 			<h2>Examples</h2>
 			<section class="example">
 				<h3>A portrait-only e-Reader application with a rotate button</h3>
-        <p>Users of an e-Reader web application sometimes read when laying down on their side. In this situation, the device is viewed by the user in portrait mode even though the device is in the landscape orientation with regard to gravity. The author locks the content into the portrait orientation to allow for this reading behavior; however, the author also provides a button to allow users who have devices mounted in a landscape orientation to achieve this desired view.</p>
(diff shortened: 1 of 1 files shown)


techniques/general/control-orientation.html
```diff
@@ -1,30 +1,31 @@
...
+			<p>The objective of this technique is to allow users to override author orientation decisions and view content as they prefer. Authors may design content for a preferred orientation and expect that people will use this orientation when viewing their content, or the author may decide that a user wants to maintain the orientation with which they initiated viewing. As a result the author may prevent the content from rotating. However, this fails the Orientation Success Criterion, since the author has restricted the view and operation to one orientation.</p>
+			<p>By providing a prevalent mechanism to allow the user to rotate restricted content, the author meets the Orientation requirement. For example, a person who cannot hold the device and has it mounted to a wheelchair or bed will be able to re-orient the content to match their display's physical orientation.</p>
```
awkawk: Not sure about "a prevalent mechanism" here. We essentially want to say "By doing nothing to interfere with the operating system's ability to rotate content, the author meets the Orientation requirement."

techniques/general/control-orientation.html
```diff
@@ -1,30 +1,31 @@
...
+      <p class="note">Many users will expect to use OS-level settings to preserve a desired orientation. As a result, instead of restricting orientation at the content level, and then allowing a user to override that restriction, it is usually best to satisfy this Success Criterion by allowing users to enable	 the OS orientation lock when desired.</p>
     </section>
 		<section id="examples">
 			<h2>Examples</h2>
 			<section class="example">
-				<h3>An e-Reader application</h3>
-        <p>Users of an e-Reader web application sometimes read when laying down on their side. In this situation, the device is viewed by the user in portrait mode even though the device is in the landscape orientation with regard to gravity. The developer locks the content into the portrait orientation to allow for this reading behavior, but also provides a control to allow users who have devices mounted in a specific orientation to achieve the desired viewing orientation.</p>
+				<h3>A portrait-only e-Reader application with a rotate button</h3>
```
awkawk: Not going to fuss about this but prefer the simpler version from before

awkawk: meaning "Not going to fuss *any more beyond this*..." :)

awkawk:
I'm good in general, just a couple small comments.

techniques/general/control-orientation.html
mbgower: In regard to the "prevalent mechanism" I figured we'd end up with some discussion about this. Fine taking it out since it hasn't previously been discussed, but to me this is very similar to an existing requirement (can't think what one it is right now) where an alternative version of something is available and it must be prevalent. In the case of orientation, if someone had to dig through 3 layers of submenu while trying to read the screen sideways in order to rotate the page, I think most folks would say that is really not accomplishing the SC.

techniques/general/control-orientation.html
mbgower: I agree entirely, but this whole technique appears to be about how the author, having not 'done nothing' can also still meet the SC. It's why I was concerned with the technique in the first place. In regard to the "prevalent mechanism" I figured we'd end up with some discussion about this. Fine taking it out since it hasn't previously been discussed, but to me this is very similar to an existing requirement (can't think what one it is right now) where an alternative version of something is available and it must be prevalent. In the case of orientation, if someone had to dig through 3 layers of submenu while trying to read the screen sideways in order to rotate the page, I think most folks would say that is really not accomplishing the SC.

techniques/general/control-orientation.html
mbgower: BTW, I couldn't help thinking when I was reviewing this that it would be pretty odd to have an e-reader application that was web-based. I'm also not going to 'fuss anymore beyond this' either, but I did wish we could have come up with a more web-ish scenario.

mbgower:
made it 'cannot hold a device'

mbgower:
@alastc I've made the changes based on today's comments. This is good to merge.

PR #830, opened 2019-07-23
G131 

### LINKED ISSUE #339: Problem with Technique G131
_From @awkawk on July 19, 2017 12:49_

Summary of Issue: Use of the term "required" is ambiguous in Step 2 of G131 Test Procedure
Comment (Including rationale for any proposed change):
In reviewing the Sufficient Techniques for WCAG 2.0 SC 3.3.2, it's unclear to me what is intended by the use of the word "required" in step 2 of the Test Procedure for Technique G131: Providing descriptive labels.

G131 Test Procedure Step 2 currently reads: "Check that any required label is present."

In this context, was "required label" intended to mean:
A) a visible text (or text symbol or icon with text equivalent) indicator that user input in the field is mandatory/required OR
B) the type of user interface control/component requires a visible label separate from the control itself OR
C) both A and B?

Rationale for Option A: Not all UI components require a separate label. A toggle <button> would not need/require a separate label (e.g., <label> element, title attribute or aria labeling technique) since the authored contents between the starting <button> element and closing </button> element can serve as the visible label for the component. In contrast, 2 input text fields would require a visible label to convey to the user what input they should to include in one text field versus the other text field.

Rationale for Option B: The introduction to the G131 technique has a sentence that indicates a required field indicator MAY be included in the label but nothing explicitly states that the label for UI components MUST distinguish whether input in a field is required or optional.

Proposed Change:
Change G131's Test Procedure to clarify the intent of Step 2 and possibly add a fourth Step related to including a cue in the visible label to distinguish the fields for which user input is mandatory vs. optional.

One possible revision follows below:
Procedure

For each interface component in the content:
1. Identify the purpose of the interface component.
2. Check that a label [linked to WCAG 2.0 definition for "label"] is visible for the component.
3. Check that the label makes the component's purpose clear.
4. If the user interface includes both mandatory and optional input, check that the visible label includes text or a text symbol that allows users to identify the components that require user input.

_Copied from original issue: w3c/wcag21#307_

alastc:
Update to the procedure.

Closes #339

### DIFF
--- a/techniques/general/G131.html
+++ b/techniques/general/G131.html
@@ -24,16 +24,15 @@
       </section>
    </section><section id="tests"><h2>Tests</h2>
       <section class="procedure"><h3>Procedure</h3>
-         <p>For each interface component in the content:</p>
+         <p>For each interface component with a label:</p>
          <ol>
             <li>Identify the purpose of the interface component.</li>
-            <li>Check that any required label is present.</li>
             <li>Check that each label makes the component's purpose clear.</li>
          </ol>
       </section>
       <section class="results"><h3>Expected Results</h3>
          <ul>
-            <li>Checks #2 and #3 are true.</li>
+            <li>Checks #2 is true.</li>
          </ul>
       </section>
    </section><section id="related"><h2>Related Techniques</h2><ul>


techniques/general/G131.html
```diff
@@ -24,16 +24,15 @@ <h3>A form with required fields</h3>
       </section>
    </section><section id="tests"><h2>Tests</h2>
       <section class="procedure"><h3>Procedure</h3>
-         <p>For each interface component in the content:</p>
```
mraccess77: While this change makes sense for 2.4.6 this technique applies to SC 3.3.2 as well and this technique 131 is the foundation for other compound techniques (see the AND clause in understanding SC 3.3.2) and thus I think it is necessary for us to not assume a label is present and that the check needs to test to make sure a label is present.  So I don't think we should change it.

mbgower: @mraccess77 there is a note specifically about this already existing in the technique: "Note: This technique must be combined with other techniques to meet SC 3.3.2. See Understanding SC 3.3.2 for details."

G131 is about making existing labels clear, not about requiring a label to be present for every UI component. Confusion over the existing wording in the procedure is what initiated the issue. Remember that G131 can apply to things besides inputs -- things that are not required to have labels according to WCAG. Where a label in on an input, it is going to be covered by 3.3.2.

Issue #833, opened 2019-07-25
Bookmarklet for testing character key shortcuts  (SC 2.1.4)


detlevhfischer:
I'd be happy to get feedback about a bookmarklet that we are using to test 2.4.1. 
I have created a test page to demonstrate how it works:
http://3needs.org/en/testing/code/kb-shortcuts.html

patrickhlauke:
sorry, VERY late comment on this (as i'm digging through the mouldy parts of my github notification emails folder) ...

might be worth clarifying in the description page there when you say

> When the bookmarklet script reaches 'p' and 'P', an alert pauses the execution of the script. In real word cases, the script should now be checked to see whether any unexpected / disruptive functions are triggered.

that the alerts that pause the execution are dummy code in the test page itself - i.e. it's not the bookmarklet that noticed something was caught/intercepted by the page, but it's the page reacting to the synthetic key press.

in practice, this bookmarklet would trigger any functionality in fast succession - so if it wasn't an execution-breaking script in the test page, it would trigger the stuff and then carrying on triggering more stuff? i.e. if i fire off the bookmarklet on a youtube video page, it'll trigger all sorts of things one after the other really quickly

patrickhlauke:
randomly...can this issue be closed? (I just stumbled across it mainly because I was looking for the URL of the bookmarklet :) )

detlevhfischer:
Well, if there was more discussion to come, I'd like to keep this open. Just today I came across a page where a tester had identified two single key keyboard shortcuts that activate behaviour on the page under test https://www.demokratie-leben.de/das-programm/ueber-demokratie-leben
the m key (mutes video) and the f key (switches to fullscreen view). At least the latter intercepts screen reader commands (NVDA: next form field). The bookmarklet seems unable to detect this, it doesn't flag m and f keys as violating 2.1.4...

patrickhlauke:
this has been open since July 2019, so not sure what more feedback you're expecting ;)

might be worth putting the bookmarklet somewhere like github and allowing people to submit issues/PRs there?

JAWS-test:
@detlevhfischer 

In your example f and m only work as shortcuts within the video. I assume that a video is an interface component and therefore the exception is valid:

> Active only on focus:    The keyboard shortcut for a user interface component is only active when that component has focus.

detlevhfischer:
@JAWS-test Good point. BTW the bookmarklet does catch the first key (f) when the video running before the bookmarklet is triggered.

PR #815, opened 2019-07-06
Failure technique for 2.5.2

patrickhlauke:
Admittedly wordy (as per my usual) and a bit contrived ... take this as a first stab that can be further tweaked and optimised.

[Preview](https://raw.githack.com/w3c/wcag/patrickhlauke-issue372/techniques/failures/failure-downevent-not-upevent-no-mitigating-circumstances.html).

Closes #372

PR #836, opened 2019-07-25
Create shortcut-mechanism.html

mbgower:
First draft of sufficient technique.

[Preview](https://raw.githack.com/w3c/wcag/keyboardShortcutSufficientTechnique/techniques/general/shortcut-mechanism.html).


detlevhfischer:
Here is a rendered view:
https://raw.githack.com/w3c/wcag/baa2cd3a31f7c6187335bcda27e4fe334d9625e3/techniques/general/shortcut-mechanism.html

patrickhlauke:
does the test procedure need an initial step to first determine if the page/site actually uses any character key shortcuts in the first place, and maybe even a follow-up step to check if they're single character ones?

and if so, could it then also benefit from a note about being able to at least narrow this down by checking JS for presence of `keydown`/`keyup`/`keypress` listeners? (similar to the note i added to the 2.5.2 failure technique https://github.com/w3c/wcag/pull/815 )

detlevhfischer:
@patrickhlauke   This bookmarklet takes another route, checks if anything happens when character keys are pressed (by script)  - but it may have flaws:
http://3needs.org/en/testing/code/kb-shortcuts.html - comment if you like in issue #833

mbgower:
@patrickhlauke [asked](https://github.com/w3c/wcag/pull/836#issuecomment-515446079):
> does the test procedure need an initial step to first determine if the page/site actually uses any character key shortcuts in the first place, and maybe even a follow-up step to check if they're single character ones?

My feeling would be it doesn't. There are sufficient technique procedures where the assumption is that the user is aware of what they're doing.  For instance  [G14: Ensuring that information conveyed by color differences is also available in text](https://www.w3.org/WAI/WCAG21/Techniques/general/G14) has a procedure that begins:
> For each item where a color difference is used to convey information:

That seems pretty similar to 'For each character key shortcut..." I don't feel like a success criterion needs to be iron-clad in the same way as a failure. Here the user is just trying to know how they confirm they've properly handled their shortcuts.  I'm not going to die on this hill, but more verbage seems unnecessary to me.

> and if so, could it then also benefit from a note about being able to at least narrow this down by checking JS for presence of keydown/keyup/keypress listeners?

Remember, this isn't a test to locate shortcuts. It's a technique to be able to turn them off. We've got content somewhat similiar to what you mention in the failure technique, which is I think where it makes more sense. I'll also mention that feedback I had from some rules writers was that trying to work off those keydown/up/press events to find character key shortcuts would lead to flagging many 'false' PVs.

mraccess77:
While this is a technique about turning them off (or changing so they are not single key)-- sending keys might be the only way to know a shortcut exists.   In my opinion the first step would be to check the documentation for single key short cuts, then check settings to see if they can be turned off or changed.  Third would to be verify that actually works and fourth to test for undocumented shortcuts.

mbgower:
If we are going to go to this level in the procedure, I think we're going to have to add in additional information in the body of the technique. So I will add a section that talks about how to determine shortcuts.
I continue to feel like folks are asking for more information here than is in a lot of existing techniques. I'll review a bunch and see if I can find a sweet spot.

mbgower:
Many of the script techniques (which have many parallels to this) have either a similar set up to what I've proposed (e.g., "For pages that dynamically create new content") or pretty broad set ups (e.g., "Find all interactive functionality")
- https://www.w3.org/WAI/WCAG21/Techniques/client-side-script/SCR1
- https://www.w3.org/WAI/WCAG21/Techniques/client-side-script/SCR2 (which is a much broader case, with very little verbage around the establishing scenario)
- https://www.w3.org/WAI/WCAG21/Techniques/client-side-script/SCR16
- https://www.w3.org/WAI/WCAG21/Techniques/client-side-script/SCR18
- https://www.w3.org/WAI/WCAG21/Techniques/client-side-script/SCR20
- https://www.w3.org/WAI/WCAG21/Techniques/client-side-script/SCR21

I'm not saying there aren't more stringent ones, but...

mbgower:
I've pushed some changes to address these comments.

alastc:
Accepted in the call, just needs prepping for publication (changing filename, updating understanding doc).

PR #845, opened 2019-08-02
Motion actuation failure

rachaelbradley:
This failure technique for motion actuation is ready for the AG to review.

### DIFF
--- a/techniques/general/GXX_motionactuationfailure.html
+++ b/techniques/general/GXX_motionactuationfailure.html
@@ -1,0 +1,71 @@
+<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml" lang="en">
+	<head>
+		<title>Failure due to inability to deactivate motion actuation</title>
+		<link rel="stylesheet" type="text/css" href="../../css/editors.css"/>
+	</head>
+	<body>
+		<h1>Failure due to inability to deactivate motion actuation</h1>
+		<section id="meta">
+			<h2>Metadata</h2>
+			<p id="id"></p>
+			<p id="technology"></p>
+			<p id="type"></p>
+		</section>
+		<section id="applicability">
+			<h2>Applicability</h2>
+			<p>Use this technique on web pages that detect device or user motion such as shaking or tilting and use these motions as a means of input.</p>
+			<p>This technique does not relate to movement of users through space as registered by geolocation sensors or beacons, or events observed by the device other than intentional gesturing by the user. It also does not cover indirect motion associated with operating a keyboard, pointer, or assistive technology. </p>
+			<p>This technique relates to Success Criterion 2.5.4: Motion Actuation (Failure).</p<
+		</section>
+		<section id="description">
+			<h2>Description</h2>
+			<p>This describes the failure condition that results when motion actuation can not be deactivated.  People who may accidentally activate sensors due to tremors or other motor impairments need the ability to turn off motion actuation to prevent such accidental triggering of functions.  </p>
+			
+		</section>
+		<section id="examples">
+			<h2>Examples</h2>
+	
+			<section class="example">
+				<h3>Motion Activated Slider</h3>
+				<p>Description</p>
+				<code>Code sample</code>
+				<p class="working-example"><a href="../../working-examples/motion-actuation-failure/">Working example of a motion activated slider that can not be deactivated</a></p>
+			</section>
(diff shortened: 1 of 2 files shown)


techniques/general/GXX_motionactuationfailure.html
```diff
@@ -0,0 +1,60 @@
...
+				<h3>Motion Activated Slider</h3>
+				<p>A slider which uses tilting motion to increase or decrease the value of an input, with no mechanism to deactivate the motion detection.</p>
+				<p class="working-example"><a href="../../working-examples/device-motion-sensor-input-failure/">Working example of a motion activated slider that can not be deactivated</a></p>
+			</section>
+		</section>
+		<section id="tests">
+			<h2>Tests</h2>
+			<section class="test-procedure">
+				<h3>Procedure</h3>
+				<p>For each function that is triggered by a motion sensor:</p>
```
alastc: @awkawk I updated the procedure to account for the exceptions in the SC text, if it impacts what you were going to look at for the other failures?

techniques/general/GXX_motionactuationfailure.html
```diff
@@ -0,0 +1,60 @@
...
+				<ol>
+					<li>Check if the use of a motion sensor is essential.</li>
+					<li>Check if there is an accessibility supported interface to trigger the function.</li>
```
alastc: @patrickhlauke I'm not entirely sure I understand what an accessibility interface for this would be, does the procedure for motion actuation make sense? (Mainly item 2)

techniques/general/GXX_motionactuationfailure.html
```diff
@@ -14,58 +14,48 @@ <h2>Metadata</h2>
...
-      <p>This technique relates to Success Criterion 2.5.4: Motion Actuation (Failure).</p>
+			<p>Content using technologies that can support the detection of device or user motion such as shaking or tilting and use these motions as a means of input.</p>
 		</section>
 		<section id="description">
 			<h2>Description</h2>
-			<p>This describes the failure condition that results when motion actuation cannot be deactivated.  People who may accidentally activate sensors due to tremors or other motor impairments need the ability to turn off motion actuation to prevent such accidental triggering of functions.  </p>
+			<p>This describes the failure condition that results when motion actuation can not be deactivated.  People who may accidentally activate sensors due to tremors or other motor impairments need the ability to turn off motion actuation to prevent such accidental triggering of functions.  </p>
```
patrickhlauke: wondering it the "can not be deactivated" needs some extra clarification that the deactivation/option needs to be part of the content, and that deactivating it at OS level doesn't count? (or does it?)

alastc: Do we need to specify? Isn't there a (future) possibility that the user could disable at the OS level which the website can pick up on?

patrickhlauke: in future perhaps, but for now it raises the usual questions of "if the content doesn't do anything, but on iOS or whatever the user can do it at OS level...is that a pass?"

patrickhlauke: (though then of course the usual answer is "not unless ALL user agents/OS provide this functionality", most likely)

alastc: That does depend on 'accessibility supported' and your conformance claim, but as a failure technique it just needs to say 'Check if there is a user setting which disables the motion detection'.

patrickhlauke: so if there's an OS/UA level setting to disable motion controls, and i say i'm only conformant on that particular OS/UA, then it's a pass without me doing anything as i satisfy #3 ?

alastc: I think that would be appropriate, yes. Of course, you'd need another mechanism to trigger the function, but that's under other SCs (2.1.1 I assume).

patrickhlauke: it's probably something more for understanding to clarify...will chew this over when i'm not poolside on hols ;)

techniques/general/GXX_motionactuationfailure.html
patrickhlauke: some OSs offer some accessibility options that can fake motion (mapped to a button or similar instead), i believe. might be that sort of thing this is referring/alluding to?

PR #853, opened 2019-08-09
Status message error

mbgower:
New technique and working example for failing 4.1.3 Status Messages

### DIFF
--- a/techniques/failures/413 Failure.html
+++ b/techniques/failures/413 Failure.html
@@ -22,7 +22,7 @@
     
 		<section id="description">
 			<h2>Description</h2>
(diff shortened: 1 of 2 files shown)


detlevhfischer:
@DavidMacDonald I think the example is useful as it points to a very frequent issue - developers looking at 4.1.3 to see what it's all about will be helped by it.

@mbgower I think this looks good. 
The only question that came to my mind is whether the Failure check should cover situations where an appropiate role is present but the result is not rendered due to the way it has been implemented (e.g. inserting the element with `role=status` and its content at the same time - I believe the element should already be present before the content is inserted or updated for it to be announced). This may be considered technical detail - but today, I would fail cases which with common UA/AT combos fail to expose the status to screen readers due to the way it is implemented. Including this may be perceived as being to tech-dependent though.

PR #858, opened 2019-08-13
Fix links to CSS3 values in 1.4.3 and 1.4.6

jfhector:
#### The problem this is fixing

The Understanding documents of both [SC 1.4.3](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html) and [SC 1.4.6](https://www.w3.org/WAI/WCAG21/Understanding/contrast-enhanced.html) have a note explaining that point sizes are based on the CSS point sizes defined in CSS3 Values.

'CSS3 Values' is wrapped into an anchor tag, but in both documents clicking on the link leads nowhere, because the `href` value of these anchor tags are empty strings.

#### Changes introduced

This Pull Request sets the `href` attribute of both links to `https://www.w3.org/TR/css-values-3/`.

### DIFF
--- a/understanding/20/contrast-enhanced.html
+++ b/understanding/20/contrast-enhanced.html
@@ -48,7 +48,7 @@
          <p>The point size should be obtained from the user agent, or calculated based on font
             metrics as the user agent does when evaluating this success criterion. Point sizes
             are based on the CSS pt size as defined in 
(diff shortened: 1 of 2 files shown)


alastc:
Thanks, that was intended to point to the CSS3 pixel reference, so I targeted that anchor in the page.

PR #860, opened 2019-08-13
Text spacing update

alastc:
Updates from Jim's doc,[ preview on rawgit](https://raw.githack.com/w3c/wcag/text-spacing-understanding-update/understanding/21/text-spacing.html), but viewing the files tab is probably easiest to see the changes.

### DIFF
--- a/understanding/21/text-spacing.html
+++ b/understanding/21/text-spacing.html
@@ -9,11 +9,12 @@
       <h1>Understanding Text Spacing</h1>
       <section id="intent">
          <h2>Intent</h2>
-         <p>The intent of this Success Criterion (SC) is to ensure that people can override text spacing to improve their reading experience. Each of the requirements stipulated in the SC's four bullets helps ensure text styling can be adapted by the user to suit their needs.</p>
-         <p>This SC focuses on the ability to increase spacing between lines, words, letters, and paragraphs. Any combination of these may assist a user with effectively reading text. As well, ensuring users can override author settings for spacing also signficantly increases the likelihood other style preferences can be set by the user. For example, a user may need to change to a wider font family than the author has set in order to effectively read text. </p>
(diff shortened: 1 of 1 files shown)


understanding/21/text-spacing.html
```diff
@@ -9,11 +9,12 @@
...
-         <p>This SC focuses on the ability to increase spacing between lines, words, letters, and paragraphs. Any combination of these may assist a user with effectively reading text. As well, ensuring users can override author settings for spacing also signficantly increases the likelihood other style preferences can be set by the user. For example, a user may need to change to a wider font family than the author has set in order to effectively read text. </p>
+         <p>The intent of this Success Criterion (SC) is to ensure that people can override author specified text spacing to improve their reading experience. Each of the requirements stipulated in the SC's four bullets helps ensure text styling can be adapted by the user to suit their needs.</p>
+         <p>The specified metrics set a minimum baseline. The values in between the author's metrics and the metrics specified in this SC should not have loss of content or functionality.</p>
```
patrickhlauke: At first reading, it sounds like the second sentence (about in-between values) follows on directly from the fist (mentioning the minimum baseline). But that's not really the case. Maybe the second sentence needs an "also" somewhere to make it clear this is an additional intention/requirement.

Also, later on under author responsibility the same first sentence is repeated, but there's a different follow-on. That looks odd/confusing. I'd say maybe remove this paragraph here, and add the second sentence from here to the one in author responsibility, to keep it all in one place - as it's the author's responsibility to also ensure that the in-between metrics work.

WayneEDick:
Please ignore this if someone else mentioned it. I cannot read all
these emails.

The sentence: "For example, a user may need to change to a wider font
family than the author has set in order to effectively read text." is wrong.

The SC does not enable font family change. However if the default
letter spacing
for a font family is to narrow, the user can increase it.

Best, Wayne

PR #880, opened 2019-09-02
Tech failure path based gesture (2.5.1)

detlevhfischer:
Hi,
here is the promised Failure of 2.5.1, better late than never. Hope it vcan be dealt with in time to be out od the way for TPAC.
Rendered version: https://github.com/w3c/wcag/compare/master...tech-failure-path-based-gesture


alastc:
I made a little udpate to move things around a little, should be the same meaning but it simplifies the procedure a bit.

detlevhfischer:
@alastc thanks, good changes!

PR #881, opened 2019-09-03
Tech fail text spacing clipping

awkawk:


awkawk:
Editors need to assign number and edit understanding doc before merging

PR #889, opened 2019-09-06
Create F_1_3_5.html

awkawk:


detlevhfischer:
When it is not possible to clearly associate fields to the user as in the airline booking example where on one page, several sets of fields are presented without any indication which set is for which user, I would argue that both approaches - marking up **all** with autocomplete, or marking up **none** - would be valid. I think the only safe failure might be associating clearly wrong autocomplete values (e.g. "name" for an email input field).

PR #910, opened 2019-09-28
Tweaks/additions to Understanding Pointer Gestures

patrickhlauke:
Two proposed changes to https://www.w3.org/WAI/WCAG21/Understanding/pointer-gestures.html (first one discussed with @alastc yesterday)

* Add example of directional-only pointer gesture
> While it's implied/is a subset of the existing first gesture (where points 1, 2 and 3 have to be in a straight line), having an explicit example here that calls these types of gestures out will make it more
immediately obvious that these are also covered. As this is  the simplest subset of the existing first figure, this was also made the first example (so the complexity builds up from the 1-to-2 straight line
example, to the 1-2-3 example).
>
> Some added `<strong>` and `<em>` added to make it more immediately clear when something IS or ISN'T path-based/free-form.

* Replace scope/scoped with "apply"
> talking about scope/scoped SC is rather obscure/technical language (that word is used only one more time, in the WCAG2ICT document, but nowhere else in WCAG). this makes it a bit more comprehensible to
laypeople.

[Preview of pointer gesture understanding](https://raw.githack.com/w3c/wcag/patrickhlauke-understanding-pointer-gestures-addition/understanding/21/pointer-gestures.html)


patrickhlauke:
ping @alastc (while i'm going through my PR left hanging open for ages)

patrickhlauke:
Ping @alastc

detlevhfischer:
Added a [rendering o the revised Pointer Gestures Understandig doc with the new graphics](https://raw.githack.com/w3c/wcag/7cf4bd1de4b8b38b6288a404e2a3d940864ab361/understanding/21/pointer-gestures.html).

patrickhlauke:
thanks @detlevhfischer ... have to admit, as this was such a long time ago, I can't even remember the exact context of this PR now, but I *think* it related to discussions we had at the time

alastc:
Sorry, almost through the 2.2 CFCs, then I can turn my attention to other issues/PRs.

patrickhlauke:
@alastc this is probably one of my oldest PRs on hold (september 2019)...any chance this could be looked at?

detlevhfischer:
Rendered version:
https://raw.githack.com/w3c/wcag/7cf4bd1de4b8b38b6288a404e2a3d940864ab361/understanding/21/pointer-gestures.html

patrickhlauke:
/me goes back to look at this PR from 4 years ago to see what it was even about...

understanding/21/pointer-gestures.html
```diff
@@ -12,24 +12,34 @@ <h1><strong>Pointer Gestures</strong><br />Understanding SC 2.5.1</h1>
...
 
-			<p>A <strong>path-based gesture</strong> involves an interaction where not just the endpoints matter. If going  through an intermediate point (usually near the start of the gesture) also affects its meaning then it is a path-based gesture. The user engages a pointer (starting point), carries out a movement that goes through at least one intermediate-point before disengaging the pointer (end point). The intermediate point defines the gesture as requiring a specific path, even if the complete path is not defined.</p>
+			<p>A <strong>path-based gesture</strong> involves an interaction where not just the endpoints matter, but how the pointer moves between these points.</p>
+			
+			<p>If the gesture is only recognised if the user moves in a (mostly) straight line from the start point to the end point, it is a <strong>path-based gesture</strong>.</p>
```
alastc, line 17:
```suggestion
			<p>If the gesture is only recognised if the user moves in a (mostly) straight line from the start point to the end point, it is an example of a <strong>path-based gesture</strong>.</p>
```

understanding/21/pointer-gestures.html
```diff
@@ -50,7 +60,7 @@ <h3>Challenges for people with disabilities</h3>
 
 			<p>An exception is made for functionality that is inherently and necessarily based on complex paths or multipoint gestures. For example, entering your signature may be inherently path-based (although acknowledging something or confirming your identity need not be).</p>
 
-			<p>Gestures that involve dragging in any direction are not in scope for this SC because only the start and end points matter in a dragging operation. However, such gestures do require fine motor control. Authors are encouraged to provide non-dragging methods, for instance, a drag and drop operation could also be achieved by selecting an item (with a tap or keyboard interaction) and then selecting its destination as a second step.</p>
+			<p>This SC <em>does not apply</em> to gestures that involve dragging in any direction because only the start and end points matter in a dragging operation. However, such gestures do require fine motor control. Authors are encouraged to provide non-dragging methods, for instance, a drag and drop operation could also be achieved by selecting an item (with a tap or keyboard interaction) and then selecting its destination as a second step.</p>
```
WilcoFiers, line 63:
```suggestion
			<p>This success criterion <em>does not apply</em> to gestures that involve dragging in any direction because only the start and end points matter in a dragging operation. However, such gestures do require fine motor control. Authors are encouraged to provide non-dragging methods, for instance, a drag and drop operation could also be achieved by selecting an item (with a tap or keyboard interaction) and then selecting its destination as a second step.</p>
```

patrickhlauke: Note that we want to capitalise these, so `Success Criterion"

understanding/21/pointer-gestures.html
```diff
@@ -12,24 +12,34 @@ <h1><strong>Pointer Gestures</strong><br />Understanding SC 2.5.1</h1>
...
+			
+			<p>If the gesture is only recognised if the user moves in a (mostly) straight line from the start point to the end point, it is a <strong>path-based gesture</strong>.</p>
 
 			<figure id="figure-path-based-gesture-1">
-				<img src="img/path-based-gesture-1.png" width="400" alt="Hand showing a starting touch, 1. Moving through a second point, 2. Going to one of several points,3." />
-				<figcaption>A path-based gesture involves starting a pointer movement that goes through at least one intermediate point before the end-point. The end-point may be a continuation, or allow for various movements.</figcaption>
+				<img src="img/path-based-gesture-1.png" width="400" alt="Hand showing a starting touch, 1. Moving in a straight line to a second point, 2." />
+				<figcaption>A <strong>path-based gesture</strong> where pointer movement is only allowed in a straight line from the start-point to the end-point. If the user strays from the straight directional path, the gesture is not recognised, has no effect, or is aborted.</figcaption>
```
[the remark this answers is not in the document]
patrickhlauke: the image doesn't claim to cover all path-based gestures, which is why it's captioned as "a path-based gesture where..." (i.e. it's only an example)

understanding/21/pointer-gestures.html
```diff
@@ -12,24 +12,34 @@ <h1><strong>Pointer Gestures</strong><br />Understanding SC 2.5.1</h1>
...
+			</figure>
+			
+			<p>If going through an intermediate point (usually near the start of the gesture) affects its meaning, then it is a <strong>path-based gesture</strong>. The user engages a pointer (starting point), carries out a movement that goes through at least one intermediate-point before disengaging the pointer (end point). The intermediate point defines the gesture as requiring a specific path, even if the complete path is not defined.</p>
```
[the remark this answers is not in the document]
patrickhlauke: as this has now been in limbo for over 4 years ... can we merge this, and then *separately* work on a complete rework of the definition / examples?

(and yes, i agree this whole concept has been weirdly fudged/bent to try and cover certain scenarios that the original meaning of "path-based" didn't actually allow/cover, and it's ugly and confusing overall)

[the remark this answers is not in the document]
detlevhfischer: I am happy to see this closed.

understanding/21/pointer-gestures.html
```diff
@@ -12,24 +12,34 @@ <h1><strong>Pointer Gestures</strong><br />Understanding SC 2.5.1</h1>
...
+			
+			<p>If the gesture is only recognised if the user moves in a (mostly) straight line from the start point to the end point, it is an example of a <strong>path-based gesture</strong>.</p>
 
 			<figure id="figure-path-based-gesture-1">
-				<img src="img/path-based-gesture-1.png" width="400" alt="Hand showing a starting touch, 1. Moving through a second point, 2. Going to one of several points,3." />
-				<figcaption>A path-based gesture involves starting a pointer movement that goes through at least one intermediate point before the end-point. The end-point may be a continuation, or allow for various movements.</figcaption>
+				<img src="img/path-based-gesture-1.png" width="400" alt="Hand showing a starting touch, 1. Moving in a straight line to a second point, 2." />
+				<figcaption>A <strong>path-based gesture</strong> where pointer movement is only allowed in a straight line from the start-point to the end-point. If the user strays from the straight directional path, the gesture is not recognised, has no effect, or is aborted.</figcaption>
```
detlevhfischer, line 21: The trouble is this still does not differentiate clearly between dragging movements and pointer gestures - because sliders on mobile may also require (initial) directional movement.
These can be moved in both ways so that andicates another differentiator - directionality. As said elsewhere, I think there are benefits in moving aways form the "intermediate point" definition because I have seen it cause lots of confusion. My alternative suggestion has been to define it in another way (DRAFT): 
"A pathbased gesture requires a particular direction of movement of the pointer across the display. It may in addtion require a particular speed or length of path for attached functions to be triggered"

That would differentiate better from axis-constrained dragging movements, because in these, their can be two directions of movement (as in an image slider).

Not that it matters that much, because 2.5.1 and 2.5.7 now both require a single pointer alternative.

detlevhfischer (dismissed):
See inline comment.

understanding/21/pointer-gestures.html
alastc: If a slider can be moved both ways, I think that's an example of two path-based-gestures (PBG), one for the first direction, another for the second direction.
Overall, we could consider the slider as either a PBG or dragging depending on the implementation, that's why it doesn't make a good example. 
I think this example description is ok though, because most sliders wouldn't fit into the "If the user strays from the straight directional path, the gesture is not recognised, has no effect, or is aborted."

Regarding the alternative description:
> A pathbased gesture requires a particular direction of movement of the pointer across the display.

Didn't we have examples where, so long as you started in a direction, it didn't require you to finish in that direction? That's why we used the intermediate point.
Also, what about multi-direction gestuers, e.g. up and left.

alastc:
Noting that there are some comments above, but the commenter is happy to merge this and come back with a different update.

alastc (approved):

detlevhfischer (approved):
+1 to be able to close this and move on and tackle again. As my comments to the PR make clear I still think that the "intermediate point" definition causes a lot of confusion, particularly since behaviour differs depending on the OS/UA wher the gesture (or movement) is carried out.

Issue #898, opened 2019-09-12
Technique G13 - allows surveys to advance on input with notice, even radio buttons?


jhausler1266:
Question about g13, https://www.w3.org/WAI/WCAG21/Techniques/general/G13

I'm working with a team building a survey creation tool. We pointed to SC 3.3.2 and said they can't advance on input, especially since they use radio buttons. They highlighted this text in G13:

`A 50 question online survey displays one question at a time. Instructions appear at the beginning of the survey that explain that users will be taken to the next question of the survey upon selecting an answer to each question.`

Even with advance notice, surely selecting a radio button from a group cannot advance a survey. Keyboard users would have to go back and forth as many times as it too to get to the final item in the group.

JAWS-test:
I find 3.3.2 very unfortunate formulated, as it allows the exception that it would be ok if the user is informed in advance. For harmless examples cited by the Understanding, it could be indeed sufficient for the user to be informed in advance. In reality, many examples can be seen in which a web application is not or very difficult to use if a context change occurs during operation. What use is a hint in advance if there is a context change during operation that leads you unintentionally to a new page or places the focus at the top of the page? My question would be whether such cases would then be regarded as violations of 2.1.1 or 2.4.3.

alastc:
In the scenario where a series of radio buttons change context on-change (i.e. as soon as you move the selection), I don't think you could fail that under 2.1.1 or 2.4.3:
- It is possible, if not easy. You can shift-tab back and select the next option until you get the right one.
- The order is fine, it is the auto-movement that's the problem.

I agree that scenario is not good from a keyboard accessibility point of view, but I think G13 is correct.

We could add a note to the effect that there are still potential problems with this approach.

Personally, if I could then I'd make the developers use it keyboard-only in sprint demos, that usually sorts out this sort of issue!

mraccess77:
I consider focus moving on selection of a radio button a WCAG failure although moving focus on blur is not.

alastc:
@mraccess77 under 3.3.2? That's where the issue started though, if there's an instruction about it, then it is not a failure of that.

mraccess77:
@alastc - right - I'd fail under 3.3.2 without an instruction.  Just wanted to make sure there wasn't some disagree with that!

JAWS-test:
@alastc 

>   It is possible, if not easy. You can shift-tab back and select the next option until you get the right one.
 
[...]

PR #912, opened 2019-10-04
Update glossary contrast ratio

JAWS-test:
* 1.4.11 links to term "contrast ratio" so limitation to text contrast is wrong
* Note 3: Also foreground color should be mentioned
* Note 4: First sentence is superfluous


JAWS-test:
I wanted to create a new PR for the update of https://www.w3.org/WAI/WCAG21/Understanding/on-input (to solve https://github.com/w3c/wcag/issues/898), but unfortunately my changes were inserted into the old PR and I don't know how to remove it here nor how to create a new PR that only contains the current change. Is there a tutorial?

Issue #917, opened 2019-10-10
1.4.11: please list color values with examples


fstorr:
It would be really useful if colors could be listed next to their examples. Because the examples are images rather than "live" code, checking them to help understand the document needs a color picker. Color pickers are at the mercy of properly-calibrated screens, and depending where I'm working (I work in different places with different screens), I get different results. 

For example, the description for Figure 4A states that the contrast ratio is 5.6:1, but I'm currently getting a result of 5.99:1 using two different contrast-checking tools. Having the colors listed would also help clarify whether the borders of radio buttons in Figure 5 have, or don't have, a 3:1 contrast. I'm seeing 2.55:1 at the moment and don't know if that's accurate.

Thanks

Myndex:
Hi @fstorr

If you're using Digital Color Meter to sample screen pixels on your monitor, that takes the monitor profile to account. As a result, you may get different values and were actually used in the CSS. 

If the example you're looking at is a live webpage graphic, that all you need to do is highlight the element, and under development tools in your browser inspect that element to see the CSS values actually used. 

If on the other hand the example element is a JPEG or a PNG, you can load that into Gimp or Photoshop which will allow you to sample colors using the colorspace that the image was created with (which should be sRGB).

When sampling colors on your monitor, you can temporarily set the monitor profile to a generic sRGB profile, and that should allow you to use Digital Color Meter to get correct values.

Some other color meters, like Classic Color Meter in the Mac app store, have additional settings that can help provide additional color information.

fstorr:
@Myndex 

Thanks for the reply. I know how to audit colors and how to use dev tools / web inspectors to get CSS values. I also know that using a Digital Color Meter isn't accurate because it relies on monitor calibration—that's the crux of my problem and why I'm suggesting clarifying the Understanding document. What I'm asking for is to add the color values to any color-related example in an Understanding document.

For example, the radio-button example in 1.4.11 would be much more helpful if the the ```<figcaption>``` element said something similar to: "A series of radio buttons each with a border color of ```#a2a2a1``` [if that's the actual color]. The first radio button shows the default state, the second (completely filled with a color of ```#797979```),  and third (completely filled with the same color at the border) show the radio button selected and filled with a color that contrasts with the color adjacent to the component. The fourth example (partially filled with a color of ```#797979```) shows the state indicator contrasting with the component colors."

Side note: there's a typo—"contasts"—in the current Understanding document.

alastc:
[...]

Issue #903, opened 2019-09-19
Structure of 1.4.11 Non-Text-Contrast Understanding inconsistent with existing formatting


mbgower:
The techniques in the Non-Text Contrast Understanding document are grouped under H4 headings. This seems to be a departure from other Understanding docs.
Following existing conventions, these should likely be given Situations (A, B, C, etc) and organized that way. See 1.1.1 for an example, where the situations have the H4s and the technique groupings are simply bold.
Low priority, IMO, but thought I'd capture.

alastc:
Closing having updated NTC in #931

Issue #896, opened 2019-09-11
Wanted to confirm my understanding as SC 1.4.11 applies to hover


mraccess77:
Based on reading the SC and the associated github comments -- only the parts needed for identification have to meet contrast requirements for the hovered state.

* A unchecked checkbox when hovered has a border < 3:1 to adjacent colors - FAIL
* A checked checkbox when hovered has a border < 3:1 but the checkmark > 3:1 - PASS
* A button with text when hovered has a border < 3:1 to adjacent colors - PASS - borders are not required to identify a button according to this working group

There continues to be misunderstanding in the community around the hovered and focused states and this SC.  For example, this SC does not require a focus ring to meet 3:1, etc.

patrickhlauke:
i would say the "hover" aspect here is a bit of a red herring, and that generally the point is:

- the SC doesn't require that the contrast between any of the colors chosen for individual states need to be 3:1 amongs each other (i.e. it doesn't say something like "when a component is hovered, or focused, or selected, its color must contrast 3:1 against its normal default look" or something). only exception: if there are controls that butt up against each other, with no gap, making the focused or hovered or selected color *adjacent* to another component's default color - but technically, even a single pixel gap makes them non-adjacent.

- what the SC does say is that if the non-text element (the outline, or border, or extra icon, or whatever) is essential to identify a component itself, or its state, then that must have a 3:1 contrast - and this applies to all possible states of a component.

> For example, this SC does not require a focus ring to meet 3:1

the understanding doc does explicitly call this out though "In combination with 2.4.7 Focus Visible, the visual focus indicator for a component must have sufficient contrast against the adjacent background when the component is focused". and, from the normative language of the SC, having a focus ring IS required to "identify [...] state".

what this SC doesn't cover, in my view anyway, is a scenario where focus indication is simply a change of color and luminosity - say all regular controls are blue, but the focused control uses a different color (but otherwise no other visual indication) - as long as focused and non-focused controls aren't adjacent, there's no prescription that says anything about how contrasty the focused/unfocused colors need to be between them, only that taken individually (if they're required to identify that something IS a control) they're contrasty enough against the background of the page. (though it may be argued that this would also violate "color alone", but i do recall some fairly hairsplitting discussion about changes in both hue/saturation/lightness counting or not counting as a "color alone" case, I forget which one was argued in the end).

JAWS-test:
[...]

Issue #713, opened 2019-04-29
Notifications - UI component contrast understanding 1.4.11


petewoodhouse-rg:
I'm not sure how to treat notifications regarding Non-text contrast guidelines

Does the complete notification count a single component and requires 3:1 contrast against its background?

Or as long as the dismiss cross (x) has enough contrast is this the only actionable element that needs to conform?

I'd recommend the complete notification because if not there would be no way of linking the dismiss cross(x) to the notification

This is a screenshot of a blog post I'm now writing (Probably shouldn't be posting an image with text here though!):

<img width="951" alt="Screen Shot 2019-04-29 at 23 13 36" src="https://user-images.githubusercontent.com/9609126/56930522-d4f5cb80-6ad4-11e9-96a2-7304f1bbb484.png">

patrickhlauke:
My take would be that the first case is not a failure, as you can still see the notification (albeit not its boundaries, perhaps) regardless of the lack of contrast / not being able to perceive its extent.

petewoodhouse-rg:
### **From WCAG**
Active User Interface Components: Visual information provided that is necessary for a user to identify that a control is present and **how to operate it** must have a minimum 3:1 contrast ratio with the adjacent colors.

'if a graphic is needed to **understand** the content or **functionality**'

Boundaries: if the visual indicator of the control is the **only way to identify the control**, then that indicator must have sufficient contrast. 

### **My interpretation**
To **understand functionality** the dismiss cross (Control) needs to be linked to the notification to know **how to operate it**.

The **only way to identify the control** (The dismiss cross) is linked to the notification is if it had a contrasting boundary. 

### **Example**
<img width="958" alt="Screen Shot 2019-04-30 at 07 51 38" src="https://user-images.githubusercontent.com/9609126/56944861-f7133c00-6b1c-11e9-991e-5c5763817a17.png">

The example above shows that without the boundary and with other components nearby the control is not identifiable, and the functionality is not understandable - especially as all notifications are not dismissable.

Myndex:
IMO the real fail is that the dismiss cross is **too small** and too far away from the descriptive text and not as obvious the key clickable element.

But as to border, the border of the second version is _TO THIN_, if you are using it to claim a passible contrast, it's a fail.

[...]

PR #931, opened 2019-11-08
Non text contrast updates

### LINKED ISSUE #551: Understanding SC 1.4.11 Active User Interface Components button examples text contrast
Issue with the two button examples - text contrast.
As noted in the examples, the visual boundary has "color of #007bff on white" 
which is 3:1 contrast. 

Observed: The text uses the same color as outline (and the inverse in the second example)

Expected: Text /background should have contrast ratio 4.5 to 1 per  SC 1.4.3 AA  

Example 1:
screen shot 2018-12-06 at 11 40 48 am
 Example 2:
screen shot 2018-12-06 at 11 41 58 am

alastc:
A bumper update to the non-text contrast understanding document to tackle 5 issues.

[Preview of understanding doc](https://raw.githack.com/w3c/wcag/ntc-updates-2019-10/understanding/21/non-text-contrast.html).

<h3>Changes</h3>

#551 - Section on disabled elements is a bit of an explainer, shouldn't be in understanding doc, cut out the explanation.

#713 - Adding some similar examples as those noted in the article.

#896 - Adding examples, use this thread for inspiration.

#903 - Re-structure headings in the understanding doc.

#917 - Add CSS color values to the graphical examples.

Closes #551 #713 #896 #903 #917


alastc:
@awkawk did the [last change](https://github.com/w3c/wcag/pull/931/commits/05053de0458f0f690898c85712531193a52e0b3c) make the header thing better?

awkawk:
@alastc Well.... Maybe if the situations were:
Situation A: Color is used to identify user interface components or used to identify user interface component states
Situation B: Color is required to understand graphical content (no colon - this one is the only one that has one now)
Situation C - fold into A as it is about a control and is really a conforming alternate version technique

PR #974, opened 2019-11-26
Additional content pre FPWD

sajkaj:
a set of additions and edits from multiple sources not previously pushed for
AGWG review. These include content from the Silver research work.


alastc:
For future reference we need to be careful about XML style validation. Although you can miss `<li>` closing tags in HTML5, the build process treats things as XML, so it's best to keep it valid.

I did an indentation & validation cleanup in this [quick change](https://github.com/w3c/wcag/commit/4455441caa7bf5df44c6779571a9c650d005188a) (in master), which looks like a lot, but I didn't change any text, just the indentation and closing tags.

PR #981, opened 2019-12-03
Update index.html

joecron:
Minor edits and cleanup for consistency.


sajkaj:
PR applied. You should probably now sync just to keep the trackers happy, though it hasn’t yet shown up for me. As is often the case, we should give it up to half an hour.

PR #997, opened 2019-12-15
Fixed incorrect ids in aria-labelledby-workshop-booking-timetable example

jfhector:
#### The issue that this PR is solving

In the [aria-labelledby-workshop-booking-timetable working example](https://www.w3.org/WAI/WCAG21/working-examples/aria-labelledby-workshop-booking-timetable/):
* the 'attend' checkboxes in column 2 are named using the headers of column 3;
* the 'attend' checkboxes in column 3 are named using the headers of column 2.

For example, when reading the checkbox to attend the "The Mesozoic era overview", screen readers say that that the session happens between 9 to 12am, but the table shows that it happens between 2 to 5pm.

That mismatch is also present in the code example in [ARIA Technique 9](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA9#example-3-a-conference-workshop-booking-table).

#### Changes introduced by this PR

* The `aria-labelledby` attributes for the 'attend' checkboxes in columns 2 and 3 now point to the correct headers.
* The `id`s of the headers have been renamed to be more explicit, so that errors like this are easier to spot in the future.
* The code example in [ARIA Technique 9](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA9#example-3-a-conference-workshop-booking-table) is updated to reflect those changes.

#### Testing

I've tested the updated markup using macOS VoiceOver.

#### Possible further improvements

I believe that the table would be more usable if the 'attend' checkboxes had the days (i.e. 'Thursday' or 'Friday') included in their accessible names. At the moment, the name of the 'attend' checkboxes only mentions the times of the day, but not the day.

If you think that this would be a worthwhile improvement, I'm happy to submit another PR with these changes, following the same [Using aria-labelledby to concatenate a label from several text nodes](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA9#example-3-a-conference-workshop-booking-table) technique.

### DIFF
--- a/techniques/aria/ARIA9.html
+++ b/techniques/aria/ARIA9.html
@@ -70,10 +70,10 @@
 &lt;/tr&gt;
 
 &lt;tr&gt;
-	&lt;th scope="col" id="am1"&gt;9 to 12 AM&lt;/th&gt;
-	&lt;th scope="col" id="pm1"&gt;2 to 5 PM&lt;/th&gt;
-	&lt;th scope="col" id="am2"&gt;9 to 12 AM&lt;/th&gt;
-	&lt;th scope="col" id="pm2"&gt;2 to 5 PM&lt;/th&gt;
+	&lt;th scope="col" id="day1_am"&gt;9 to 12 AM&lt;/th&gt;
+	&lt;th scope="col" id="day1_pm"&gt;2 to 5 PM&lt;/th&gt;
+	&lt;th scope="col" id="day2_am"&gt;9 to 12 AM&lt;/th&gt;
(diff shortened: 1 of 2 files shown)


fstrr:
Looking at the [ARIA9 technique](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA9.html), and the [working example](https://www.w3.org/WAI/WCAG22/working-examples/aria-labelledby-workshop-booking-timetable/), this looks like it has been completed elsewhere.

I'm going to close this, but if it needs re-opening, then please do.

Issue #1007, opened 2020-01-12
F87 and custom-styled form controls


JamesCatt:
It's a common practice to visually hide checkbox/radio buttons and use pseudo elements to provide custom-styled alternatives (for example, see https://www.w3schools.com/howto/howto_css_custom_checkbox.asp)

I'm curious what the WG thinks of this practice and whether or not it should be considered a failure of F87. If yes, then the test for F87 may need some rewording.

On the one hand, it does seem to go against the spirit of F87, as the technique involves using pseudo elements to communicate essential information (i.e., the presence and state of a checkbox or radio button). If a user has a custom stylesheet, this information could be lost.

On the other hand, if a user is indeed using a custom stylesheet with pseudo elements for checkboxes/radio buttons, then it may not actually matter that the author-supplied styling is lost—it would simply be replaced by the user styling.

Either way, the current wording of the test scenario would actually allow this technique to pass because it relies on disabling CSS altogether—in which case the browser default checkbox/radio button would be visible and no information lost to the user.

patrickhlauke:
I would say this practice is fine, and doesn't clash with the intent of F87 (since F87, from my reading, is intended to prevent situations where the CSS pseudo elements are the ONLY place where this additional content is defined/presented, whereas the custom checkbox/radio scenario is a progressive enhancement to what is already defined in the markup). Yes, custom stylesheet users may be affected if their custom stylesheets don't cover these situations...but that is up to the users and their custom stylesheets. And yes, if users disable stylesheets altogether, then they'll just see the regular unstyled checks/radios.

JAWS-test:
If I understand it correctly, it is as follows:
* earlier (when F87 was developed) CSS content was banned because a) screen readers didn't output it and b) pages without CSS should work. Both are no longer relevant today, because a) CSS content is part of the Accessible Name according to the specification and b) it is no longer assumed that CSS is disabled.
* Today, the issue is that users use their own user styles and overwrite the existing CSS, e.g. with a different font. 

Unfortunately, F87 was not properly adapted to the new situation and the examples contradict the mentioned intention. You get a better impression with [ARIA24](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA24.html). According to ARIA24 it would be okay if the CSS graphics were marked with role=img, because then the user styles do not exchange the graphics. Of course, this assumes that all users of user styles know this. To what extent this is actually the case is of course very questionable...

JAWS-test:
[...]

PR #1016, opened 2020-01-26
Update Working Example ARIA24

JAWS-test2:
Related to: #1007

I have changed the following:

*    frame (shadow) around the stars, so that contrast is sufficient
*    contrast between yellow and gray in the star increased, thus contrast greater than 3:1
*    all font icons marked with aria-hidden=true, and aria-label removed
*    alternative text of the icons in the source code as text (class=sr-only), so that the text is also visible without CSS (see F87)

If these changes are accepted, I would also change the ARIA24 file to update the source code examples


patrickhlauke:
ARIA24 wants to illustrate how adding `role="img"` and `aria-label="..."` works as a way to both "signal" that a container is actually using a font icon, and how to then also provide a programmatic text alternative for it. This proposed PR instead removes the `aria-label` and adds `.sr-only` additional containers directly after them. I believe this goes way beyond (in a very idiosyncratic way) what ARIA24 tried to actually show. (incidentally, if you're doing `aria-hidden="true"` on the spans with the font icon, you can probably then just save yourself adding the `role="img"` in the first place ... in essence, this rewrite guts the whole purpose of ARIA24 and simply suggests hiding the font icon from AT and having a visually-hidden container after it). To me, that goes beyond a small tweak, and should be its own technique or something ("Using visually-hidden content...")

The one positive aspect (if the current examples do fail) is the strengthening of the colour contrast for the stars. This could be tackled as a standalone separate PR.

patrickhlauke:
As for the original https://github.com/w3c/wcag/issues/1007, it's a lengthy conversation that is worth re-examining, because I don't believe even this PR would have "fixed" the question there (I'm removing the "Fixes..." part of the original description from this for that reason, to avoid having #1007 auto-close if this is closed)

bruce-usab:
Discussed on Backlog call 3/7.  Lori agreed to review.

mraccess77:
I agree with Patrick that any PR should not change the intent of ARIA24 which is to highlight how role img can be used to mark icon fonts for low vision users to avoid font replacement.

ljoakley:
This has morphed into a total rewrite rather than an update. Thus technique adds hidden text, and it is addressing more than original issue. It is using HTML plus CSS rather than ARIA. PR no longer closes #1007 and F87 is being revisited. We would like to close this and open another PR for "contrast between yellow and gray in the star increased, thus contrast greater than 3:1".

patrickhlauke:
Discussed in today's backlog meeting. TF agrees to close this PR. Made a separate PR to address the contrast issue in the working example https://github.com/w3c/wcag/pull/4293

PR #1017, opened 2020-01-26
Update Understanding On Input (SC 3.2.2)

JAWS-test2:
Fixes https://github.com/w3c/wcag/issues/898


understanding/20/on-input.html
```diff
@@ -42,7 +42,22 @@ <h2>Intent of On Input</h2>
          
       </div>
       
-      
+      <div class="note">
+
+         <p>If the user interface component is not fully operable with the keyboard due to the <a class="glossary">change of context</a>, 
+		    this would be a violation of SC 2.1.1.
```
stevefaulkner, line 48: is "violation" the appropriate term?

understanding/20/on-input.html
```diff
@@ -42,7 +42,22 @@ <h2>Intent of On Input</h2>
...
+         </p>
+
+      </div>
+
+      <div class="note">
+
+         <p>A <a class="glossary">change of context</a> should only be made if this makes further operation of the application easier. 
+		 If the change of context makes the operation difficult for impaired people, it is recommended to avoid the context change.
```
stevefaulkner, line 56: is "impaired people" an appropriate term?

stevefaulkner (changes requested):

understanding/20/on-input.html
patrickhlauke: should be "failure", yeah

understanding/20/on-input.html
patrickhlauke: yes that's a rather unfortunate turn of phrase

bruce-usab:
Discussed on backlog call, 2/21.

patrickhlauke:
I reviewed this PR for the WCAG 2.x group. I'll admit that while it makes sense, I'm not sure it's necessary. The first change - mentioning that things still need to be keyboard-accessible - seems an unnecessary cross-reference (as the same hold true for all other SCs). The same is true for the additional note
`If the change of context makes the operation difficult for disabled people, it is recommended to avoid the context change.`

I can see how this seems to directly address the problem brought up in https://github.com/w3c/wcag/issues/898 ... but it seems overly specific for something that, in essence, is a case that currently falls between the cracks ... "nominally passes the word of the SC, but it a dreadful user experience"

mbgower:
I believe several people have indicated this is a failure of 2.1.1.
I agree that the wording of On Input suggests that if you tell someone it's going to happen in advance, you pass On Input, but you'd still fail 2.1.1

I'd like to see if we have agreement that the problem described in the original issue -- someone triggers an on input change in a radio button group, it advances a survey -- would be considered a failure.

mbgower:
Resolving this through #898

PR #1018, opened 2020-01-26
Update term "contrast ratio" for SC 1.4.11

JAWS-test2:
Contrast requirements also apply to non-text content in accordance with SC 1.4.11. Therefore "text" has been replaced by "content".

If no colors are defined, the notes for foreground and background apply.


Myndex:
## Objections:

1) Font smoothing and anti-aliasing applies primarily to fonts, so adding _"non-text"_ to that paragraph is not particularly helpful. 

**ALSO:** that entire paragraph should be deleted IMO, because authors can and unfortunately DO use things like _"webkit-font-smoothing: antialias"_ which destroys the contrast. The use of webkit smoothing is a particular problem today, especially with WP templates, and authod imposed font smoothing definitely SHOULD be discussed.

2) Adding _"If no  foreground color is specified, then black is assumed."_ is incorrect. Text is NOT to be assumed to be black. Common defaults are black, blue for link, and red/purple for visited link.

But this is another paragraph that makes an assumption and should be deleted or revised. White should not be assumed to be the background. The default background is TRANSPARENT. So if you have text on a DIV with no BG color set, that div will be transparent and the color thus whatever random object it is over.

**AND** 

Also this paragraph is in conflict with the later paragraph that says _"It is a failure if no background color is specified when the text color is specified, because the user's default background color is unknown and cannot color is specified, because the user's default background color is unknown and cannot be evaluated"_

😳

Best Regards,

Andy

JAWS-test:
I understand some of your objections. But I am not sure if you also have objections to the [current version of the WCAG at this point](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html#dfn-contrast-ratio). If so, perhaps you can make a suggestion on how it could be improved

alastc:
@patrickhlauke will review. It might help to minimize the changes as this is normative text.

bruce-usab:
Those [six notes attached to the definition are unchanged from 2.0](https://www.w3.org/TR/WCAG20/#contrast-ratiodef).  I agree it was a missed opportunity not to update the notes when 1.4.11 was added.

There is also the long standing [F24 Failure ... due to specifying foreground colors without specifying background colors or vice versa](https://www.w3.org/WAI/WCAG22/Techniques/failures/).

bruce-usab (changes requested):
I am okay _only_ with the word substitution for note 2 (changing text to content).

IMHO, it is better for notes 3/4/5 to remain unchanged and specific to text.

guidelines/terms/20/contrast-ratio.html
```diff
@@ -15,20 +15,20 @@
    
    <p class="note">Contrast ratios can range from 1 to 21 (commonly written 1:1 to 21:1).</p>
    
-   <p class="note">Because authors do not have control over user settings as to how text is rendered
-      (for example font smoothing or anti-aliasing), the contrast ratio for text can be
+   <p class="note">Because authors do not have control over user settings as to how text and non-text content is rendered
+      (for example font smoothing or anti-aliasing), the contrast ratio can be
       evaluated with anti-aliasing turned off.
```
patrickhlauke, line 20: while the sentiment/idea is valid - that it's not just text but also non-text content, it's not possible to turn off anti-aliaising for things like SVG or CSS lines etc. And even for text, it's not always possible - depending on OS or user agent - to turn it off even for text rendering.

patrickhlauke: instead of just "with anti-aliasing turned off", this may need to be expanded to something like 
```suggestion
      evaluated with anti-aliasing turned off or ignored.
```

patrickhlauke (changes requested):

guidelines/terms/20/contrast-ratio.html
```diff
@@ -15,20 +15,20 @@
...
    </p>
    
    <p class="note">For the purpose of Success Criteria 1.4.3 and 1.4.6, contrast is measured with respect
-      to the specified background over which the text is rendered in normal usage. If no
-      background color is specified, then white is assumed.
+      to the specified background over which the content is rendered in normal usage. If no
+      background color is specified, then white is assumed. If no
+      foreground color is specified, then black is assumed.
```
patrickhlauke, line 26: as 1.4.3 and 1.4.6 *are* specifically about text, there's no sense here changing things to be more generic

TestPartners:
The phrase "the contrast ratio can be evaluated with anti-aliasing turned off" is ambiguous because it allows the measurement to be made using with the author-specified colours or the actual rendered colours. Is this ambiguity intentional? I think we should mandate one method or the other, unless there is a good reason not to.

Mandating the use of the author-specified colours would ensure consistency in measurements even if it doesn't reflect the user experience. I am inclined to favour this approach.

Allowing the measurement of the rendered colours introduces inconsistency and potentially allows testers to choose which colours to pick to get the result they want, which may be either a pass or fail. Furthermore, anti-aliased text often changes colour as the zoom level is changed.

patrickhlauke:
@TestPartners so long story short, you're saying the definition should change "can" to "should"?

guidelines/terms/20/contrast-ratio.html
```diff
@@ -15,20 +15,20 @@
    
    <p class="note">Contrast ratios can range from 1 to 21 (commonly written 1:1 to 21:1).</p>
    
-   <p class="note">Because authors do not have control over user settings as to how text is rendered
-      (for example font smoothing or anti-aliasing), the contrast ratio for text can be
+   <p class="note">Because authors do not have control over user settings as to how text and non-text content is rendered
+      (for example font smoothing or anti-aliasing), the contrast ratio can be
```
patrickhlauke, line 19:
```suggestion
      (for example font smoothing or anti-aliasing), the contrast ratio should be
```

patrickhlauke:
also, more generally, agree that the second and third note could probably do with being killed off altogether (about assumptions to be made about default foreground/background colours, and then saying that it's a failure if background isn't defined ... because it's a throwaway that doesn't actually go into sufficient detail, and makes it sound like *every time* you define foreground you *must* define background on that same element ,,, which is incorrect, since it's fine as long as at the very least the root/ancestor has things defined. also, why say in a definition that "it fails", as this is not in an SC but just defining the term itself)

TestPartners:
> @TestPartners so long story short, you're saying the definition should change "can" to "should"?
> 
> Or something more substantitve? note that understanding docs already mention that you should refer to author-defined values (from a recent-ish change that caused no end of arguing back and forth at the time)

Yes, that's what I'm saying, nothing more.

The Understanding document for SC 1.4.3 currently says "when evaluating this Success Criterion, refer to the foreground and background colors obtained from the user agent, or the underlying markup and stylesheets, rather than the text as presented on screen." I interpret this as MUST rather than CAN or SHOULD, and I think the same wording should be used for SC 1.4.11 to avoid ambiguity.

patrickhlauke:
> and I think the same wording should be used for SC 1.4.11 to avoid ambiguity.

note that the PR here is purely about the definition of "contrast ratio". changes to 1.4.11 understanding should be tackled separately (i have a vague feeling i may have filed something along those lines as a PR already somewhere)

patrickhlauke:
> Yes, that's what I'm saying, nothing more.

proposed here https://github.com/w3c/wcag/pull/1018#discussion_r2073718888

bruce-usab:
Discussed on backlog call 5/16.  Group agrees it's a good catch that, with 1.4.11, glossy term is not just about text.  Unfortunately, there are other issues with the notes.  Thank you Patrick for your willingness take another stab, keeping this PR to just a minimum change.

patrickhlauke:
superseded by https://github.com/w3c/wcag/pull/4400

PR #1019, opened 2020-01-26
Outdated link in ARIA19

### LINKED ISSUE #893: Outdated link in ARIA19
https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA19

> There are also a number of special case live region roles which can be used instead of applying live region properties directly. 

links to https://www.w3.org/WAI/PF/aria-practices/#chobet. 
It should link to https://www.w3.org/TR/wai-aria-practices-1.1/

The only problem is that the page does not currently contain any information about Live Regions. But this should be changed (https://github.com/w3c/aria-practices/issues/78).

JAWS-test2:
Fixes #893


kfranqueiro:
Thanks for catching this, and sorry this PR went unreviewed for so long. It appears this was eventually covered in this change: https://github.com/w3c/wcag/pull/2458/files#diff-dc090a06bab2f85ac7dc8db2cf7b231b93cdfed985e1467251b6146d2643f391

PR #1020, opened 2020-01-26
Understanding Use of Color (SC 1.4.1)

### LINKED ISSUE #871: Does G183 only apply to "links and controls" with additional requirements or in general?
If I understood @patrickhlauke correctly (#857), the WCAG SCs are normative and all other documents (Understanding, Failure etc.) are not normative. This means that the latter can explain the SCs, but cannot extend, restrict or even contradict them. SC 1.4.1 just says: "Color is not used as the only visual means..." - and there's not a single hint of a possible exception.
But then there is this https://www.w3.org/WAI/WCAG21/Techniques/general/G183 which restricts 1.4.1. How can that be? 
Furthermore, 1.4.11 refers to this G183 and uses it for the examples in Figure 6 and 7. But:
* which contrast ratio is now sufficient according to 1.4.11, if 3:1 is sufficient for links?
* and why should a contrast difference be sufficient for the stars of Figure 6 and 7, when G183 clearly states: "it can be used if additional visual confirmation is available when a user points or tabs to the link". And how does the "additional visual confirmation" look like for the stars? The text also does not say that the stars are interactive elements. It could also be just an illustration of the rating - and then there would certainly be no additional effects as with the links.

I'm not against the contents of G183, but
* in WCAG 3.0, the desired exceptions should already be considered when writing the SCs
* and G183 should already now be revised to apply to other elements, if referred to in 1.4.11. Otherwise, the reference in 1.4.11 should be removed.

See also: https://github.com/w3c/wcag/issues/201

JAWS-test2:
Fixes #871

### DIFF
--- a/understanding/20/use-of-color.html
+++ b/understanding/20/use-of-color.html
@@ -45,6 +45,17 @@
          
       </div>
       
+      <div class="note">
+         
+         <p>If the contrast ratio between two colors is greater than 3:1, this is considered as an 
+			other visual mean and this success criterion is met. However, it is recommended to use 
+			other visual means instead of or in addition to contrast differences to meet this sucess criterion.
+			When using the High Contrast Mode or when adapting the CSS to user needs, color differences 
+			are sometimes no longer visible. If only color with sufficient contrasts is used, this criterion 
+			is fulfilled, but there may be a violation of <a href="info-and-relationships">1.3.1: Info and Relationships</a>.
+         </p>
+         
+      </div>
       
    </section>
    <section id="benefits">


Myndex:
## Objection

The proposed paragraph is out of context and not a useful addition to this SC. The proposed paragraph appears to be discussing luminance contrast initially, but 1.4.1 is discussing how chroma (hue/sat) should not be relied on for coding or distinguishing elements.

The proposed paragraph does not add anything to the SC or discussion, it is confusing, and the "what ifs" presented don't provide a useful insight or motivation to insert this.

Best regards,

Andy

JAWS-test:
Either there is the 3:1 exception at SC 1.4.1, in which case it should also be in the Understanding If it does not exist, all associated technologies should be removed.

JAWS-test:
This PR can be closed if PR https://github.com/w3c/wcag/pull/1500 is successful

alastc:
PR #1500 was merged, but thank you for the PR, it helps to add another voice to the issue.

PR #1021, opened 2020-01-26
Update F103 to remove progressbar as it's not an implicit live region

### LINKED ISSUE #892: Why is progressbar mentioned in F103?
https://www.w3.org/WAI/WCAG21/Techniques/failures/F103

Progressbar in F103 is listed as a live region, but it is not a live region: https://www.w3.org/TR/wai-aria-1.1/#progressbar

Even worse: a progressbar can never be turned into a live region, because the change of the value of the progressbar does not affect any output in a live region (according to the specification of live regions)

See also: https://github.com/w3c/aria-practices/issues/1165

JAWS-test2:
Fixes #892


bruce-usab (approved):

bruce-usab:
Discussed on TF call 11/8.  Patrick volunteered to double check, but is Ready for approval.

PR #1022, opened 2020-01-27
Update Understanding "focus visible" (SC 2.4.7)

### LINKED ISSUE #1001: 2.4.7 as it applies to static content and programmatic focus
Apologies if this has been answered elsewhere already. I'm struggling to find an agreed-upon answer.

There are some situations where authors decide to send focus to static content via JS (think `tabindex="-1"` and `.focus()`. By default, browsers render a visual focus outline around these static elements when they receive focus. However, many authors remove this visual focus indicator for static elements and leave them for interactive elements. I find that authors do this because elements with `tabindex="-1"` receive focus on click and programmatically.

Some examples include:

1. When submitting a form, focus is sent to the container of an error message above the form.
2. In a single page application, a routing event automatically focuses the heading at the top of new content. I put together an [example of this to illustrate SPA link routing](https://codepen.io/mfairchild365/pen/NWPpoxw).

The normative text of 2.4.7 states
> Any keyboard operable user interface has a mode of operation where the keyboard focus indicator is visible.

The informative text of the [understanding document](https://www.w3.org/WAI/WCAG21/Understanding/focus-visible.html) suggests that this is to help "anyone who relies on the keyboard", and users with "attention limitations, short term memory limitations, or limitations in executive processes" 1. understand **exactly where focus is** on the page **at all times**, so that, 2. they can **predict what will happen** when they issue a keyboard command.

Do these situations require a visible focus indicator to meet 2.4.7?

One read is 'yes', since, without it,  1. users may not know exactly where focus is at all times,  and 2. users may not be able to predict what will happen when they issue a keyboard command.

Another read is 'no' since the static element itself is not keyboard operable in of itself.

JAWS-test2:
Add extra notes about single focusable elements, and initial state of focus on page opening etc.

Fixes #1001


patrickhlauke:
Reformatted the additions into the note to account for the changed structure since this PR was first filed, to clear the merge conflict

understanding/20/focus-visible.html
```diff
@@ -25,11 +25,13 @@ <h2>Intent of Focus Visible</h2>
...
 
       <div class="note">
          <p>There may be situations where mouse/pointer users could also benefit from having a visible focus indicator, even though they did not set focus to an element using the keyboard. As a best practice, consider still providing an explicit focus indicator for these cases.</p>
+        
+        <p>If there is only one keyboard focusable control on the screen, the success criterion would be met because the visual design presents only one keyboard focusable item. The success criterion does not require the focus to be visible intially, e.g. after opening a new page, when displaying a pop-up or after activating in-page links. However, it is recommended to show the focus in these cases as well.</p>
```
gundulaniemann, line 34: How is the user supposed to work when the focus is not shown initially? How does a screen reader user perceive what happened to the focus?
If the focus is not visible initially, the user doesn't see where the focus is and might initiate an action that is not wanted. Also, having to interact to see the focus reduces efficiency considerable and distracts from the task all the time.

mbgower: I don't think I agree with this addition (and btw, there's a typo in "initially"):
> The success criterion does not require the focus to be visible intially

There is a difference between a page that does not have an element in focus initially, and one where there is a focusable element that is not properly shown as active. I don't think this nuance is clear, nor how to address in respect to Focus Visible. I'm also not sure that the _number_ of focusable controls on the screen is relevant to that. 

I do not think this PR sufficiently captures or addresses the initial question in #1001. 

I think these are good questions to provide guidance on:
- can an inoperable element on a page take focus?
- if so, should it have a focus indicator?
- if so, _must_ it have a focus indicator to meet Focus Visible (and not doing so is a failure of 2.4.7)?
- conversely, is there a time when inoperable elements receiving focus constitutes a failure of WCAG?

I think more detail is required, and I don't think this PR is sufficient as written. I'm okay to keep the scope of this change contained, but I think it should clearly answer at least one of the questions that seems to exist in #1001 discussion.

patrickhlauke (changes requested):
I don't believe this fully answers/closes #1001 

A more thorough answer may be https://github.com/w3c/wcag/pull/1643

Issue #825, opened 2019-07-17
Outdated links in Understanding SC 2.4.2 (Page Titled)


cstrobbe:
Two links in [Understanding SC 2.4.2 (Page Titled)](https://www.w3.org/WAI/WCAG21/Understanding/page-titled.html) are outdated:

* The article "Writing Better Web Page Titles" on SocialPatterns.com is no longer available; there is a newer one with a very similar title, [Writing Better Page Titles for SEO](https://socialpatterns.com/writing-better-titles/), dated 5 February 2017. I don't know to what extent the content is similar to the old article. There are several alternative articles and blog posts that are less focused on SEO, e.g. [Page titles – accessibility for web writers](https://4syllables.com.au/articles/accessibility-writers-titles/) (July 2011) and [How to write better page titles](https://www.nomensa.com/blog/2013/how-to-write-better-page-titles/) (June 2013).
* The URL http://redish.net/images/stories/PDF/interactions.html does not work either, but a PDF version of the article is available at https://redish.net/wp-content/uploads/Theorfanos_Redish_InteractionsPaperAuthorsVer.pdf.

cstrobbe:
I am closing this issue **in protest against the [W3C's allowing overlay vendors to set up a community group](https://adrianroselli.com/2022/09/accessibility-at-the-edge-w3c-cg-is-an-overlay-smoke-screen.html)**. Overlay vendors have been known to [make false claims about their tools](https://overlayfalseclaims.com/), their products can actually [make the web worse for people with disabilities](https://www.accessibility.org.au/accessibility-overlays-make-the-web-worse-not-better-new-york-times/), and screen reader users have even created a [Google Chrome extension to block overlays](https://www.accessibyebye.org/). My contributions to WCAG will remain on halt for as long as there is a W3C community group run by overlay vendors.

patrickhlauke:
"allowing"? i'm fairly sure W3C isn't an arbiter here, and a big part of the whole W3C process is pay-to-play when it comes to membership in the first place...

cstrobbe:
@patrickhlauke Recommended reading: [Neutrality isn’t an option when it comes to accessibility](https://karlgroves.com/neutrality-isnt-an-option-when-it-comes-to-accessibility/) by Karl Groves. It's not about the W3C, but I'm sure I don't need to explain how it is relevant.

patrickhlauke:
[...]

PR #1025, opened 2020-01-29
Update reference link in Understanding of SC 2.4.2 (Page Titled)

JAWS-test2:
Fixes https://github.com/w3c/wcag/issues/825


bruce-usab:
Flagged on backlog call 6/6 and left in Drafted.

patrickhlauke:
to clear merge conflict, turned this from "replacing links" to "adding links" (and tweaking one existing link to the latest url it redirects to)

kfranqueiro:
@patrickhlauke Note that the conflict resolution you performed changes the nature of this PR, which is why I was holding off on it until we examined it (as stated in my previous comment). The PR originally intentionally removed the "Writing Better Page Titles" link rather than updating it, as explained in #825. We should critically examine the nature of that new article, because its content may have changed.

While we're at it though, we should probably critically examine each of the other added links as well.

patrickhlauke:
ah, apologies @kfranqueiro ... i misread this to be fixing "broken" links (per the PR title), rather than "still existing, but maybe not that good/changed" ones

understanding/20/page-titled.html
```diff
@@ -122,22 +122,22 @@ <h2>Resources for Page Titled</h2>
...
          <li> 
-            
-            <a href="https://socialpatterns.com/writing-better-titles/">Writing Better Web Page Titles</a>. How to write titles for web pages that will enhance search engine effectiveness.
+            Dey Alexander: <a href="https://4syllables.com.au/articles/accessibility-writers-titles/">Page titles – accessibility for web writers</a>. Write a descriptive title for your web pages and any documents you publish online (2011)
+         </li>
+         <li> 
+            Gosia Mlynarczyk: <a href="https://www.nomensa.com/blog/2013/how-to-write-better-page-titles/">How to write better page titles</a>. A few tips on how to write good and accessible page titles (2013)           							     
+         </li>
+         <li>
+            <a href="https://contentcreation.ai/writing-better-titles/">Writing Better Web Page Titles</a>. How to write titles for web pages that will enhance search engine effectiveness.
```
dbjorge: TF reviewed this today and agreed with updating the links but felt that these new suggestions were either things we weren't comfortable with the longetivity of or had the same issues as the original in being too SEO focused. MDN has some reasonable guidance here that I'd suggest instead:

```suggestion
            <a href="https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/title#accessibility">&lt;title&gt;: The Document Title element: Usage notes (Accessibility)</a>.
```

mbgower:
@JAWS-test2 

Please review the updated PR, which the Task Force supports. We are adding this to a working group review. We can close this one and open a new PR if you are concerned with this change.

Issue #1029, opened 2020-02-01
[Editorial] "Text Alternatives" reference


Malvoz:
In https://www.w3.org/TR/WCAG21/#dfn-text-alternative:

> NOTE
> 
> Refer to [Understanding Text Alternatives](https://www.w3.org/TR/UNDERSTANDING-WCAG20/conformance) for more information.

Would you agree that the "Understanding Text Alternatives" link destination should be changed to https://www.w3.org/TR/UNDERSTANDING-WCAG20/conformance#uc-text-alternatives-head? Since it is the section of interest.

JAWS-test2:
Link to current version would be better: https://www.w3.org/WAI/WCAG21/Understanding/conformance#text-alternatives

kfranqueiro:
Resolved by #1830

PR #1031, opened 2020-02-02
Update link in definition of text alternatives

JAWS-test2:
Fixes https://github.com/w3c/wcag/issues/1029

### DIFF
--- a/guidelines/terms/20/text-alternative.html
+++ b/guidelines/terms/20/text-alternative.html
@@ -10,7 +10,7 @@
       text alternative for the chart indicates that a description follows.
    </p>
    
-   <p class="note">Refer to <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/conformance">Understanding Text Alternatives</a> for more information.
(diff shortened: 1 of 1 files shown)


Malvoz (approved):

patrickhlauke:
@kfranqueiro is there a better, more evergreen way to link from the term to that particular understanding part? I'm always a bit loathe to see fully qualified URLs to a specific version of the spec in these things

kfranqueiro:
This PR was sent over 4 years ago, and since then, [this exact update has already been made](https://github.com/w3c/wcag/commit/009b5257f8#diff-11e2e90427f9e567307f983f38339aa791b908c9d18afeb93fb03e2b789ed49b), so I think this should be closed.

To @patrickhlauke's question, AFAIK this isn't something that is explicitly supported; most importantly, this would also need to work within the guidelines themselves, and I'm not sure whether that's realistic.

One thing that would probably be worth pursuing if at all possible would be rewriting explicit 2.1 links to the appropriate version within the context of the informative documentation.

kfranqueiro:
Already included in #1830

Issue #873, opened 2019-08-26
Explanation of exception in F81 insufficient


JAWS-test:
http://www.w3.org/WAI/WCAG21/Techniques/failures/F81: 

> It would also not fail if the color chosen had sufficient luminosity difference (lightness) from the other text that it would be easily be seen as different if viewed in black and white. In these cases - the information would not be displayed in color (hue) alone but also in "presentation" or "lightness" respectively

What is "sufficient luminosity difference"? 3:1 as in G183? And if: Why do links with 3:1 contrast need further clues to be able to recognize them and required or error fields do not:

> if additional visual confirmation is available when a user points or tabs to the link

alastc:
I think it would be useful to add to the technique:

- A reference to G183.
- A reference to non-text contrast, for things like an extra red border, of change of color for the input border.

JAWS-test:
I'd like that. However, 2 questions remain unanswered:
* Why do links require further features for mouseover and incorrect fields do not?
* Is the reference to 1.4.11 correct if 1.4.11 only considers directly adjacent elements?

alastc:
- You are leaping from one particular technique (G183) to an area that doesn't have a technique (input errors). The short answer is probably that [G183](https://www.w3.org/WAI/WCAG21/Techniques/general/G183) was written pre-touchscreen popularity when mouse & keyboard were the primary inputs, so mousing was included in the procedure. That doesn't mean anything about inputs.
- It is correct in that it would help explain why a change in contrast could be a method. I'm not saying it would be the test for that, but that using a contrast change would be another visual queue compared to color alone.

JAWS-test:
> You are leaping from one particular technique (G183) to an area that doesn't have a technique (input errors). The short answer is probably that G183 was written pre-touchscreen popularity when mouse & keyboard were the primary inputs, so mousing was included in the procedure. That doesn't mean anything about inputs.

The question would be whether the recognizability of links is more important than that of errors. If so, then I think the distinction is good. If no, then the requirement for links and form fields with errors should not be different.

alastc:
It isn't about importance, the context is often quite different. For example, links can be buried in blocks of text, whereas an error message is generally near an input. As the context is generally different, so the requirements can be different.

JAWS-test:
> whereas an error message is generally near an input

But F81 is about the fact that there is no error message. The incorrect fields are only marked with color. And thus just as little to recognize as only colored links in text.

alastc:
[...]

PR #1033, opened 2020-02-04
Update F81

JAWS-test2:
Fixes https://github.com/w3c/wcag/issues/873


Myndex:
# Objection

The line _"A minimum contrast ratio of 3:1 is considered sufficient."_ does not belong in this criterion and creates a conflict with the other SCs that define luminance.

Side note: the term is _luminance_ not _luminosity_.

Also, the word "are" was incorrectly removed from the change in the first paragraph.

JAWS-test:
>The line "A minimum contrast ratio of 3:1 is considered sufficient." does not belong in this criterion and creates a conflict with the other SCs that define luminance.

The exception regarding contrast was previously already in the criterion, without an exact specification of the contrast ratio. The text has been inserted to be compliant with https://www.w3.org/WAI/WCAG21/Techniques/general/G183 and with #873.

techniques/failures/F81.html
```diff
@@ -1,14 +1,17 @@
...
       <p>All technologies</p>
    </section><section id="description"><h2>Description</h2>
-      <p>This objective of this technique is to describe the failure that occurs when a required field or an error field is marked with color differences only, without an alternate way to identify the required field or error field. This can cause problems for people who are blind or colorblind, because they may not be able to perceive the color differences that indicate which field is required or which field is causing an error.</p>
+      <p>This objective of this technique is to describe the failure that occurs when a required field or an error field is marked with color differences only, without an alternate way to identify the required field or error field. This can cause problems for people who colorblind, because they may not be able to perceive the color differences that indicate which field is required or which field is causing an error.</p>
```
yatil, line 4: There's a verb missing @JAWS-test2 

> This can cause problems for people who colorblind

to

> This can cause problems for people who **are** colorblind

techniques/failures/F81.html
```diff
@@ -1,14 +1,17 @@
...
+         <p>In both examples, the color could be used without failure if the text was sufficiently different in visual presentation (e.g. bold or in a different font) that it would be easily differentiated from the surrounding text if the color were removed. It would also not fail if the color chosen had sufficient luminosity difference (lightness) from the other text that it would be easily be seen as different if viewed in black and white. A minimum contrast ratio of 3:1 is considered sufficient. In these cases - the information would not be displayed in color (hue) alone but also in "presentation" or "lightness" respectively. </p>
+      </div>
+      <div class="note">
+         <p>For users of assistive technology, required or error fields must also be perceptible. This is checked with the SCs 1.3.1 and 3.3.1.</p>
```
yatil, line 14: > perceptible

Should that be perceivable (just because that is what WCAG uses elsewhere and it seems to be good to re-use nomenclature :-)

techniques/failures/F81.html
patrickhlauke:
```suggestion
      <p>This objective of this technique is to describe the failure that occurs when a required field or an error field is marked with color differences only, without an alternate way to identify the required field or error field. This can cause problems for people who are colorblind, because they may not be able to perceive the color differences that indicate which field is required or which field is causing an error.</p>
```

techniques/failures/F81.html
patrickhlauke:
```suggestion
         <p>For users of assistive technology, required or error fields must also be perceivable. This is checked with the SCs 1.3.1 and 3.3.1.</p>
```

techniques/failures/F81.html
```diff
@@ -1,14 +1,17 @@
...
+         <p>In both examples, the color could be used without failure if the text was sufficiently different in visual presentation (e.g. bold or in a different font) that it would be easily differentiated from the surrounding text if the color were removed. It would also not fail if the color chosen had sufficient luminosity difference (lightness) from the other text that it would be easily be seen as different if viewed in black and white. A minimum contrast ratio of 3:1 is considered sufficient. In these cases - the information would not be displayed in color (hue) alone but also in "presentation" or "lightness" respectively. </p>
+      </div>
+      <div class="note">
+         <p>For users of assistive technology, required or error fields must also be perceivable. This is checked with the SCs 1.3.1 and 3.3.1.</p>
```
patrickhlauke:
```suggestion
         <p>In both examples, the color could be used without failure if the text was sufficiently different in visual presentation (e.g. bold or in a different font) that it would be easily differentiated from the surrounding text if the color were removed. It would also not fail if the color chosen had sufficient luminosity difference (lightness) from the other text that it would be easily be seen as different if viewed in black and white. A minimum contrast ratio of 3:1 is considered sufficient. In these cases - the information would not be displayed in color (hue) alone but also in "presentation" or "lightness" respectively. However, while this will satisfy the visual requirement for 1.4.1, the example may still fail other requirements, such as 1.3.1, 3.3.1, and 4.1.2.</p>
```

techniques/failures/F81.html
```diff
@@ -1,14 +1,17 @@
...
       <p>All technologies</p>
    </section><section id="description"><h2>Description</h2>
-      <p>This objective of this technique is to describe the failure that occurs when a required field or an error field is marked with color differences only, without an alternate way to identify the required field or error field. This can cause problems for people who are blind or colorblind, because they may not be able to perceive the color differences that indicate which field is required or which field is causing an error.</p>
+      <p>This objective of this technique is to describe the failure that occurs when a required field or an error field is marked with color differences only, without an alternate way to identify the required field or error field. This can cause problems for people who are colorblind, because they may not be able to perceive the color differences that indicate which field is required or which field is causing an error.</p>
```
patrickhlauke:
```suggestion
      <p>This objective of this technique is to describe the failure that occurs when a required field or an error field is marked with color differences only, without an alternate way to identify the required field or error field. This can cause problems for people with color vision deficiency or low vision, because they may not be able to perceive the color differences that indicate which field is required or which field is causing an error.</p>
```

techniques/failures/F81.html
```diff
@@ -1,14 +1,17 @@
...
    </section><section id="examples"><h2>Examples</h2>
       <ul>
-         <li>A user is completing an online form, and the phone number field is required. To indicate that the phone number field is required, the label "Phone Number" is displayed in a color different from the color used for optional fields, without any other indication that "Phone Number" is a required field. A blind or colorblind user may not be able to identify that "Phone Number" is a required field.</li>
+         <li>A user is completing an online form, and the phone number field is required. To indicate that the phone number field is required, the label "Phone Number" is displayed in a color different from the color used for optional fields, without any other indication that "Phone Number" is a required field. A colorblind user may not be able to identify that "Phone Number" is a required field.</li>
```
patrickhlauke:
```suggestion
         <li>A user is completing an online form, and the phone number field is required. To indicate that the phone number field is required, the label "Phone Number" is displayed in a color different from the color used for optional fields, without any other indication that "Phone Number" is a required field. A person with color vision deficiency or low vision may not be able to identify that "Phone Number" is a required field.</li>
```

mbgower:
@yatil do you have time to take a look at this?

patrickhlauke:
@yatil want to give this another look, or you happy with it?

patrickhlauke (approved):
Made a few tweaks for better language/flow, but otherwise think this is good

bruce-usab (approved):

patrickhlauke:
cross-reference to https://github.com/w3c/wcag/pull/4472 where the "colorblind" replacement happens for all other instances

scottaohara (approved):

bruce-usab:
[diff for F81](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FTechniques%2Ffailures%2FF81&doc2=https%3A%2F%2Fdeploy-preview-1033--wcag2.netlify.app%2Ftechniques%2Ffailures%2Ff81): Failure of Success Criterion 1.4.1 due to identifying required or error fields using color differences only

PR #1035, opened 2020-02-04
Update F3

JAWS-test:
The old wording has suggested that background graphics to convey information are fine as long as the information is not important. But even unimportant information must not be displayed with background graphics. Only for decorative content background graphics may be used

### DIFF
--- a/techniques/failures/F3.html
+++ b/techniques/failures/F3.html
@@ -1,6 +1,4 @@
-<!DOCTYPE html><html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml"><head><title>Failure of Success Criterion 1.1.1 due to using CSS to include images that convey
-                    important information</title><link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove"></link></head><body><h1>Failure of Success Criterion 1.1.1 due to using CSS to include images that convey
-                    important information</h1><section class="meta"><p class="id">ID: F3</p><p class="technology">Technology: failures</p><p class="type">Type: Failure</p></section><section id="applicability"><h2>When to Use</h2>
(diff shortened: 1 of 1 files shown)


Myndex:
??? Removing the word important makes this nearly impossible to test. Most images convey information, even if mainly decorative. Using the word important allows a value judgement and a bright line as to where that cut off may be.

mbgower:
**Official Task Force Response**
The numbering of this failure technique implies it was the third failure technique ever generated by the working group. As such, the incorporation of the word "important" in all major parts of the document -- the title, description, and the test procedure -- seems like an intentional cue by the authors of WCAG 2.0 of what would constitute a failure of 1.1.1 in regard to considerations around the "Decoration" situation (Situation F).

Failure techniques generally have a higher bar for the test procedure since any "pass" of the failure technique results in a clear failure of WCAG. This could explain the reason for the more cautious assessment. This failure technique also provides potentially critical information on the nature of content that _could_ be considered decorative. Although it appears to be a more liberal interpretation of 'Decorative' than is contained in the normative wording of the success criterion, the working group believes changes to failure technique wording after 17 years must be very cautiously debated -- and in this case not carried out.

WCAG 3 may wish to carry out a more focused discussion on the types of non-text content that require a text alternative, and whether unimportant non-text content should have a "lower bar" in regard to the both the requirement to provide an alternative _and_ the level of detail such an alternative would need to provide.

mbgower:
Discussed at July 25, 2025 TF meeting. Participants were generally in accord with retaining "important".

It was pointed out that the second sentence of the description is no longer accurate, and that other parts of the description may also need to be updated.
>  The CSS background-image property was designed for decorative purposes and it is not possible to associate text alternatives with images that are included via CSS.

However, that is a distinct challenge from the proposal in this PR. The TF may open a separate issue to deal with this.

bruce-usab:
Demoting to In Progress until alternative PR available.

gundulaniemann:
I agree that no information should be conveyed in background images. It is kind of impossible to judge which kind of information is important.

patrickhlauke:
the line between "information" (be it "important information" or just "information") and "decorative" is fuzzy already as it is.

the current wording seems to have 3 woolily defined buckets: important information, not important/unimportant information, and then decorative content. the cut-off point between these is not defined anywhere and subjective. with this PR, it boils them down to two: information, and decorative content. the cut-off point is still wooly, subjective, undefined.

Whether that is better or not... no idea. Might be good to maybe explicitly say somewhere that this is of course subjective/context-specific

mbgower:
The [draft response](https://github.com/w3c/wcag/pull/1035#issuecomment-3099357479) underwent a working group review, and has now has been marked as the official response. This PR is being closed without implementation.

PR #1037, opened 2020-02-06
WCAG 2.2 SC - Accessible Authentication

alastc:
This is the PR for the new SC, a fairly straightforward addition of:

- [New SC text](https://pr-preview.s3.amazonaws.com/w3c/wcag/pull/1037.html#accessible-authentication), with include on the index page.
- A new term ([cognitive functional tests](https://pr-preview.s3.amazonaws.com/w3c/wcag/pull/1037.html#dfn-cognitive-function-test))
- A new understanding doc ([preview understanding doc](https://raw.githack.com/w3c/wcag/3e05c24b9f1ef2d3538aa3ee8c14f1ca841c916b/understanding/22/accessible-authentication.html)).
- A new technique, [email authentication](https://raw.githack.com/w3c/wcag/wcag22-accessible-auth/techniques/general/email-link-authentication.html).

Transferred from the [draft SC Google doc](https://docs.google.com/document/d/1dk4pYxq9vxj2E1GfeH6RGWvqvtCyGfLGDVFopNf8B_U/edit#) and [technique doc](https://docs.google.com/document/d/14kDhWxqNPMVUj09-TBFrv2oOXFQDCPBv-uigLr1QxGY/edit#).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/pull/1037.html" title="Last updated on Mar 30, 2020, 9:59 PM UTC (829aaba)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/1037/73d0d5b...829aaba.html" title="Last updated on Mar 30, 2020, 9:59 PM UTC (829aaba)">Diff</a>


lauracarlson:
Hi @alastc ,

3 headings in the [Understanding Doc](https://raw.githack.com/w3c/wcag/3e05c24b9f1ef2d3538aa3ee8c14f1ca841c916b/understanding/22/accessible-authentication.html) refer to "Focus Visible" instead of "Accessible Authentication":

1. Benefits of Focus Visible
2. Examples of Focus Visible
3. Techniques for Focus Visible

techniques/general/email-link-authentication.html
```diff
@@ -0,0 +1,64 @@
...
+		<section id="description">
+			<h2>Description</h2>
+			<p>The objective of this technique is to provide an easy way for users to authenticate without needing a password. This technique involves providing an authentication mechanism where the user can enter their email address, and they are sent an email with a link to click. When the user clicks the link in the email, they are directed back to the website and automatically logged in.</p>
+			<p class="note">The security of the email link mechanism is not the focus of this technique, but it generally involves sending a time limited token as part of the email.</p>
+
+		</section>
+		<section id="examples">
+			<h2>Examples</h2>
+			<section class="example">
+				<h3>Author provides a email mechanism to login with a link</h3>
```
patrickhlauke: a email > an email

understanding/22/accessible-authentication.html
```diff
@@ -0,0 +1,120 @@
...
+            
+        <p class="note">Websites can employ username (or email) and password inputs as an authentication method if it meets <a href="#">Success Criterion 1.3.5 Input Purpose</a> and password managers / browser functionality is not blocked. When properly marked up, browser features or password managers can save the user’s information and refill the login.</p>
+
+        <p>If there is more than one step in the authentication process, such as with multi-factor authentication, all steps should comply with this success criterion. There should be a path through authentication that does not rely on cognitive function tests.</p>
+
+
+
+    </section>
+    <section id="benefits">
+      <h2>Benefits of Focus Visible</h2>
```
patrickhlauke: title needs changing

understanding/22/accessible-authentication.html
```diff
@@ -0,0 +1,120 @@
...
+      
+      <p>People with cognitive issues to do with memory, reading (e.g. dyslexia), numbers (e.g. dyscalculia), or perception-processing limitations will be able to authenticate irrespective of the level of their cognitive abilities.</p>
+      
+   </section>
+   
+   <section id="examples">
+      <h2>Examples of Focus Visible</h2>
```
patrickhlauke: title needs changing

understanding/22/accessible-authentication.html
```diff
@@ -0,0 +1,120 @@
...
+         <li>A website uses WebAuthn so the user can authenticate with their device instead of username/password. The user’s device could use any available modality. Common methods on laptops and phones are facial-scan, fingerprint, and pin number. The website is not enforcing any particular use, it is assumed a user will setup a method that suits them.</li>
+         <li>A website offers the ability to login with a 3rd party provider using the OAuth method.</li>
+         <li>A website that requires 2-factor authentication allows for multiple options for the 2nd factor, including a USB-based method where the user simply presses a button to enter a time-based token.</li>
+      </ul>
+      
+   </section>
+
+   
+   <section id="techniques">
+      <h2>Techniques for Focus Visible</h2>
```
patrickhlauke: title needs changing

patrickhlauke:
The understanding document focuses exclusively on login type scenarios

> The purpose of this success criterion is to ensure there is an accessible, easy-to-use, secure method to log in and access content.

what about more general situations where a CAPTCHA is presented? for instance, posting a comment on a blog - and the form may include a CAPTCHA there? while some aspects of this scenario will be already covered by other SCs (e.g. 1.1.1 for those "type in the scrambled text you see here" type scenarios for instance), others won't. and while i agree (and i believe, discussed at length in the past) this type of thing is not "authentication" per se, it would be good to cover this more broadly as a requirement for cases other than logins.

also noting that just the email example as a technique essentially doesn't allow two-factor authentication (as emails are not generally secure enough)

alastc:
Thanks for catching the headings & email typo.

> what about more general situations

I remember the comments, but no one has been able to come up with a useful scope. The key scenario the TF were interested in was logging in, so that was the focus.

Widening it to any process could capture a lot of scenarios, and I don't think anyone has investigated what those would be.

alastc:
> just the email example as a technique essentially doesn't allow two-factor authentication (as emails are not generally secure enough)

Hmm, from a security point of view most systems allow you to change your password via email, so I wouldn't say it is "not secure enough".

That technique could be combined with a 2nd factor, we didn't want complicate it, but you could have a 2nd factor after the email link logs you in as a first factor.

There is another technique (to do) about combining the other techniques as part of 2 factor auth.

patrickhlauke:
the current wording of the normative SC could also apply to scenarios not tied to login ("a task" could be posting a blog comment). basically, that an alternative is available. might need expanding from "authentication" to something more like "authentication and verification".

separately, would be good to have a normative definition of "authentication process" perhaps?

patrickhlauke:
> Hmm, from a security point of view most systems allow you to change your password via email, so I wouldn't say it is "not secure enough" as a general statement.

Sensitive systems (e.g. banking) do generally *require* a second factor. Be it an SMS sent to you with an access code (that you then have to type in/copy ... is that allowed under the wording of this SC?), using a physical card reader where you put in your bank card, or similar.

> That technique could be combined with a 2nd factor, we didn't want complicate it, but you could have a 2nd factor after the email link logs you in as a first factor

The 2nd factor would be at some point before allowing the user to change the password in the first place. The link takes you to the second challenge before letting you go an actually change the password and log in.

I just want to make sure this SC doesn't (inadvertently) try to force less secure solutions

understanding/22/accessible-authentication.html
```diff
@@ -0,0 +1,120 @@
...
+      
+   </section>
+   
+   <section id="examples">
+      <h2>Examples of Accessible Authentication</h2>
+
+      <ul>
+         <li>A website uses a properly marked up email and password as the login authentication.</li>
+         <li>A website uses WebAuthn so the user can authenticate with their device instead of username/password. The user’s device could use any available modality. Common methods on laptops and phones are facial-scan, fingerprint, and pin number. The website is not enforcing any particular use, it is assumed a user will setup a method that suits them.</li>
+         <li>A website offers the ability to login with a 3rd party provider using the OAuth method.</li>
```
cookiecrook: Text should not limit this to OAuth. Prose can be modified to use OAuth as an example, rather than the only acceptable technology.

understanding/22/accessible-authentication.html
```diff
@@ -0,0 +1,120 @@
...
+         <li>A website that requires 2-factor authentication allows for multiple options for the 2nd factor, including a USB-based method where the user simply presses a button to enter a time-based token.</li>
```
cookiecrook: Prose seems to require USB 2FA via pushbutton. Was the intention that this should be an optional example? If so, it should be reworded.

understanding/22/accessible-authentication.html
```diff
@@ -0,0 +1,120 @@
...
+         
+         <ol>
+            <li>
+                <a href="#" class="general">Email link authentication</a>
+            </li>
+            <li>
+                @@ Providing a properly marked up email and password inputs.
+            </li>
+            <li>
+                @@ Providing WebAuth as an alternative to username/password.
```
cookiecrook: 1. typo? "WebAuth"
2. these should be clearly phrased as a list of examples, rather than as currently phrased b/c it appears to be a definitive and finite list.

understanding/22/accessible-authentication.html
```diff
@@ -0,0 +1,120 @@
...
+            </li>
+            <li>
+                @@ Providing a 3rd party login using oAuth.
```
cookiecrook: ditto OAuth commment above

understanding/22/accessible-authentication.html
patrickhlauke: isn't this whole list just a series of examples?

understanding/22/accessible-authentication.html
patrickhlauke: ditto

understanding/22/accessible-authentication.html
patrickhlauke: sufficient techniques (as well as failure techniques) are all just non-exhaustive throughout WCAG understanding docs, i believe...they're just suggestions at that particular point in time, as there's usually many more ways to satisfy the ask of any particular SC

understanding/22/accessible-authentication.html
alastc: These are indeed just examples, it is not normative. What would another 3rd party auth method be?

understanding/22/accessible-authentication.html
alastc: I tried to think of some line of text to put at the top, but the heading starting "Examples" made it redundant.

understanding/22/accessible-authentication.html
alastc: Updated to WebAuthn. Sufficient techniques are always "one possible way" of meeting the guideline.

PR #1052, opened 2020-02-27
Underline links in the head of the spec.

### LINKED ISSUE #562: Link in 2.1 failing F73
The bit near the top: This document is also available in non-normative formats, available from **Alternate Versions of Web Content Accessibility Guidelines 2.1**.

No underline or bold indicating the link. 

Should probably put a bug in for the spec redesign for aspect.

### LINKED ISSUE #1048: Possible 1.4.1 failure in WCAG 2.1 normative document
If you look at the WCAG 2.1 document today, above the abstract, there are a few links that do not have an underline. Some of them are bold, but one is not, making it only distinguishable from the surrounding text by its blue color. This seems to me like it would be a failure of SC 1.4.1 Use of Color. The issue is there on three links:

- "errata" in "Please check the errata for any errors ..."
- "translation" in "See also translations."
- "Alternate Versions of Web Content Accessibility Guidelines 2.1"

Screenshot 2020-02-25 13 04 34

Credit to Eric Velleman at Accessibility Foundation for spotting this.

alastc:
A very constrained bit of CSS to over-ride the base.css

Closes #1048 
Closes #562

### DIFF
--- a/css/additional.css
+++ b/css/additional.css
@@ -162,13 +162,6 @@ div.sc div.note p.prefix {
     margin-bottom: 0;
 }
 
-/* Un-do the removal of underlines from the links in the head area */
-
-.head p:not(.copyright) > a, .head > a:first-child {
-    border-bottom: 1px solid #707070;
-}
-
-
 span.screenreader {position: absolute; left: -10000px}
 
 /* tabbed styles - repeated here to keep cover sheets consistent with /TR/ style */
--- a/css/additional.css
+++ b/css/additional.css
@@ -162,6 +162,12 @@ div.sc div.note p.prefix {
     margin-bottom: 0;
 }
 
+/* Un-do the removal of underlines from the links in the head area */
+
+.head p:not(.copyright) > a, .head > a:first-child {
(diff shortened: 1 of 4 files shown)


michael-n-cooper:
I'm requesting review from PLH as I'm not sure if this violates TR style rules. If it does, we would need to try to get the change made in the central TR styles rather than attempt an override.

michael-n-cooper:
This fix won't work because we can't override central TR styles. I filed https://github.com/w3c/tr-design/pull/187 to address it centrally, though it's likely to be a long time before we see results.

awkawk:
Why can't we make the links bold like the two directly above the one in question? At least that is something...

alastc:
Looking through the recent TR docs, I can't find a single one that has a paragraph between the editors and copyright paragraph, so I'd guess either *that* is a problem or the link underline is an oversight.

Given that the spec-specific styles are after the TR styles, it should technically work.

michael-n-cooper:
Turns out we *can* make links bold, because that's not explicitly turned off in the TR base stylesheet. I've updated the pull request to do that.

PR #1058, opened 2020-02-29
Remove misleading/incorrect statement from 1.4.11 understanding

patrickhlauke:
prompted by https://github.com/w3c/wcag/issues/541#issuecomment-592852826


patrickhlauke:
in the end, i suspect this part will be expanded to point to the new/proposed Focus Visible (Enhanced) as that *will* define requirements for this scenario. But in the meantime, let's remove the misleading statement that was left there as some kind of handwavy backdoor redefinition for what focus visible meant (per the discussion in #541)

bruce-usab:
This is good catch of the technique going beyond what the SC requires.

patrickhlauke:
@alastc not i made a further change here...missed out a "not" in the last sentence by accident.

PR #1065, opened 2020-03-03
Changes to examples in Understanding 1.4.11

### LINKED ISSUE #1014: 1.4.11 Understanding: Additional focus indication of text inputs (default caret visible)
The Understanding doc for 1.4.11 "Non-text contrast" has an example in [section "Active User Interface Component Examples](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast#active-user-interface-component-examples)" > "Text input focus style":  a text field with additional focus indication.

It says:

> "Text input focus style: A focus indicator is required. If the focus indicator is styled by the author, it must meet the 3:1 contrast ratio."

As @patrickhlauke has pointed out in https://github.com/w3c/wcag/issues/680#issuecomment-484517551 

> But the image there shows the text input, and the focus outline on that input yes...but there's a caret also shown. This seems to then suggest that yes, regardless of caret, the outline must have sufficient 3:1 contrast. But if that outline is secondary, and the state of the input (it being focused) is already indicated by the presence of the cursor/caret...then it's irrelevant whether or not that outline is contrasty enough, no?

I believe this issue has not been resolved. The current example would see a site that uses **only** the cursor without **any** additional focus highlight PASS while an implementation with additional highlight (but below 3:1) fails. This seems counter-intuitive to me.

Options:
IMO, either we say "cursor alone is never enough to show focus", or additional focus indications should pass whenever there is already one / any other form of contrasty focus indication (e.g., a blinking caret).

patrickhlauke:
Initial suggestion for changes to Understanding 1.4.11 based on discussion in #1014 

- modify the "Text input focus style" example (including the image) - due to the cursor/caret, this does pass even with a low contrast extra outline
- remove the "Text input using background color focus style" example (and related image), as it's made redundant by the previous example (as it's irrelevant what background color is set, since what matters is the cursor/caret)
- add checkbox example - this better illustrates that the focus indicator needs to be contrasty enough (as no additional indication, like the cursor/caret, are present)

Noting that allowing cursor/caret here as a sufficient focus indication will be ok under 2.4.7 Focus Visible, but will likely NOT be sufficient under proposed WCAG 2.2. 2.4.11 Focus Visible (Enhanced), so that example for focused text input will likely need to be tweaked further with cross-references

Closes #1014


detlevhfischer:
+1

patrickhlauke:
as this came up in a conversation at work...a subtle ping

patrickhlauke:
thank you :)

alastc:
Just workin' on the backlog :-)

PR #1071, opened 2020-03-04
Status message added to 2.4.11

alastc:
As the understanding document for the new WCAG 2.2 SCs are popping up in the regular understanding docs area, we need to highlight the status of those documents.

I've used some CSS from the basic styles available (class=advisement), but not sure where that section would pop out of the build process.

### DIFF
--- a/understanding/22/focus-visible-enhanced.html
+++ b/understanding/22/focus-visible-enhanced.html
@@ -8,6 +8,12 @@
 <body>
    <h1>Understanding Focus Visible (Enhanced)</h1>
    
+    <section id="status" class="advisement">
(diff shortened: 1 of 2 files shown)


michael-n-cooper:
I added support to the XSLT to output the status section.

alastc:
Hi Michael, When it's built, does it look ok?

michael-n-cooper:
Yes, it renders in some kind of orange box. I merged it and it's now 
visible at 
https://w3c.github.io/wcag/understanding/focus-visible-enhanced. Michael

alastc:
Brilliant, thank you.

Issue #1004, opened 2020-01-02
Focus visible for small elements with large click areas


alastc:
An issue @jake-abma raised on email:
<hr>
Please see code below where we use web component structures where the parent / main container, click-area and button are separated and focus indicators are made to fit OR buttons OR placement (like top right corner in dialog OR on touch target area (depending on need of design)

Depending on the 'button' as "user-interface component" is also not fine as we do not use them visually (for visual purposes), can even make them 1px by 1px and so the focus indicator of 2 CSS px would be fine?

![image](https://user-images.githubusercontent.com/216630/71680483-24864880-2d82-11ea-8da7-74a5001bf609.png)

Interaction can be different things:

- for keyboard behaviour the button element is still used
- visual interaction (communication / with colors / icons) can be the <button> element OR the outer element OR click area
- with mouse the click area, is adjustable

alastc:
Hi @jake-abma I put together some more examples for testing & working out how it works:
https://alastairc.uk/tests/wcag22-examples/focus-more-visible-4.html 

I think examples 21 & 22 demontrate what you were aiming to show, I put them right on the edge of the criteria text.

jake-abma:
Thanks @alastc indeed a good one, and this knocks the SC down, as this is weird when the indicator is enough for the button, but with the extra touch target, fails the same button (visually)

jake-abma:
We might see if we can blend the current calculation / SC text with a kind of 'baseline', preventing this discrepancy

alastc:
Rather than a baseline approach, I think it would be better to make an exception where there is a visible boundary smaller than the element.

It is good that the standard way of doing focus is outlines/borders/backgrounds, and generally they would pass in the examples. It is only because the examples are at the minumum it's an issue.

Anyway, for those type of examples, how about adding:

> Note: If the control has a visible boundary smaller than the hit area, the size measure is taken from the visible boundary.

mraccess77:
+1 to Alastair's suggested note to address the issue.

detlevhfischer:
+1 yes, good exception to cover cases of larger click areas

alastc:
Note added to the proposed update ([version 3 in this doc](https://docs.google.com/document/d/1F83m-HXRkXz1QCF6_QNtQGzIULugMiFSLgtqDRgKA-s/edit#heading=h.96ib39rkbmhb)), does that suffice @jake-abma

jake-abma:
yes! thanks you

Issue #1115, opened 2020-04-29
The new text for SC2.4.11: Focus Visible (Enhanced)


guyhickling:
I am very happy that revised texts have now been considered for the Focus Visible (Enhanced) SC, as posted at [https://docs.google.com/document/d/1F83m-HXRkXz1QCF6_QNtQGzIULugMiFSLgtqDRgKA-s/edit#](https://docs.google.com/document/d/1F83m-HXRkXz1QCF6_QNtQGzIULugMiFSLgtqDRgKA-s/edit#). They are a huge improvement. 

Looking at those options, I would advocate a combination of options 1 and 2. I am greatly in favor of the 2px minimum thickness of the line, so would like that from Option 1. For the rest of the SC I like the Option 2 version. It states things clearly, and the 3 to 1 contrast with adjacent colors (which option 1 doesn't have) is definitely necessary.

I strongly favor the 2px line thickness because, for keyboard users navigating using the Tab key, if they have poor vision, it is so difficult to spot where the focus has just just jumped to as they Tab around the page. The more complex and busy the page, the more difficult it becomes. (Even when just testing web pages, with my below average vision, I am often slow to see where it's gone!) Making it 2px in thickness makes it so much easier for these people - and lots of websites from larger organizations and businesses use an outline of 3 or even 4px, so it is a well accepted method. Yes it's prescriptive, but focus outlines are only seen at length by keyboard users - who are the very ones who need a good one. (Other people see them only briefly while a new page loads, and even that can be stopped by a bit of scripting.) 

I hesitate to disturb things any more beyond that in case it leads us too far away from these new options, but regarding the matter of colour-change indicators, where the fore and background colors both change to indicate focus, I think that could still be included with only a few words extra. How about adding a clause to the first bullet point to make it:

> "Is at least the size of a 2 CSS pixel solid border around the control, or is a change of fore and background colors of the component, and"

(An extra three or four words would also be needed in the third bullet point to correspond with that.)

Anyway, congratulations on a hugely improved offering, and many thanks to the Group for listening to all our previous public discussions about it.

alastc:
Thanks Guy, we include that feedback in the upcoming discussion.

alastc:
Hi Guy,

This was discussed at the AG meeting on Tuesday, and the general consensus was to keep some flexibility rather than prescribing a particular focus style/approach. 

[...]

Issue #1075, opened 2020-03-08
2.4.11 Focus Visible (Enhanced) - what counts as "bounding rectangle"/"bounding box"


patrickhlauke:
there is no clear definition of what "bounding rectangle" (and later in the understanding doc, "bounding box") refers to. is it the "clickable"/"active" area of a control? or its visual apparent bounding box?

thinking of situations where the clickable area of a control has been made quite large (perhaps to pass some target size requirement, or just to make it more touch friendly) by using ample padding that goes beyond the "apparent" confines of a control ... is the bounding box the visually apparent box, or the actual clickable/tappable box? if the latter, this would introduce an odd conundrum that two controls that *appear* the same may end up needing differently sized focus indicators, purely because one also made an extra effort to have a more generous invisible "slop" area around it.

detlevhfischer:
Can we point to the latest version of 2.4.11 Focus Visible (Enhanced) from here?

patrickhlauke:
which one, this? https://w3c.github.io/wcag/guidelines/22/#focus-visible-enhanced

patrickhlauke:
(which i see swaps out "bounding box" for "bounding rectangle")

detlevhfischer:
In order to avoid the conundrum you are pointing out I would assume (and have assumed, before you raised the question) that the bounding rectangle would be based on the visible target (as a minimum requirement).

patrickhlauke:
but the problem is...what is the visible target? that's not always clear. what if i have a giant button with no border/background, that only contains a single "GO >" bit of text with a LOT of padding around it, again with no border/background. is the assumed visible target then just the text? (probably yes, but this should be defined/explained)

alastc:
I can see two points here:

1. What should the bounding box be applied to, the visible component or the actual element? (If they differ.)
2. Does the term 'bounding box' need to be defined?

For the first we have #1004 

Does "bounding box" need to be defined? We could do. @awkawk and I discussed that and it seems to be a common term in apps (e.g. design programs), computer science and geometry. The google results for bounding box are reasonable, but if people don't get it I'd propose something like:

"The smallest rectangle that can be drawn around a shape."

patrickhlauke:
[...]

Issue #1067, opened 2020-03-04
2.4.11 Focus Visible (Enhanced) Remove F78 and F55


ShadowBB:
In the understanding document of the working draft:
https://www.w3.org/WAI/WCAG21/Understanding/focus-visible-enhanced.html

F78 and F55 are listed as failure techniques.

However the SC itself is only applicable if(!) there is a visible focus
"**_When_** a User Interface Component displays a visible keyboard focus, all of the following are true: "

If there isn't a visible keyboard focus (which is the case if either F78 or F55 are followed) this SC should be "not applicable", not a "failure".

mraccess77:
I personally feel these techniques are appropriate here.

awkawk:
Interesting observation. The SC language is causing this issue.

If a developer removes the visible focus, do we want 2.4.7 focus visible AND 2.4.11 Focus visible (enhanced) to both fail? Or do we want:

When focus is invisible/gone:
Fail 2.4.7 because the focus isn't visible
Fail 1.4.11 because the focus doesn't meet 3:1
Pass 2.4.11 because there is no visible focus, even though one is needed.

When focus is light (2.5:1):
Pass 2.4.7 because the focus is visible at all
Fail 1.4.11 because the focus doesn't meet 3:1
Fail 2.4.11 because the focus doesn't meet the enhanced criteria.

When focus is dark but doesn't change the color much from non-focus state:
Pass 2.4.7 because the focus is visible at all
Pass 1.4.11 because the focus meets 3:1
Fail 2.4.11 because the focus does not have 3:1 relative to prior appearance.

Because the SC says: "When a User Interface Component displays a visible keyboard focus, all of the following are true:" The situation where there is no focus shown at all would pass this SC. I don't think that is what we want. I think that we want:

2.4.11: Any keyboard operable user interface has a mode of operation where the keyboard focus indicator meets all of the following: (same bullets) 

This would make it crystal clear that failing 2.4.7 means failing 2.4.11.

awkawk:
And then the techniques are unambiguously applicable. We could even get rid of 2.4.7 in WCAG 2.2! (ducks)

alastc:
@awkawk Ha, that's where the language started, I think it was you who updated it to change from the wording of 2.4.7!

Also, I think it best to take 1.4.11 out of the equation, there is another thread/PR on that which removes the focus style examples.

It probably is worth tying into the #1050 discussion when we get to that.

patrickhlauke:
[...]

PR #1121, opened 2020-05-07
WCAG22 - Focus Visible enh updates

alastc:
Previews of:
- [Focus appearance (minimum) understanding](https://raw.githack.com/w3c/wcag/wcag22-focus-visible-enh-updates/understanding/22/focus-appearance-minimum.html)
- [Focus appearance (enhanced) understanding](https://raw.githack.com/w3c/wcag/wcag22-focus-visible-enh-updates/understanding/22/focus-appearance-enhanced.html)

Updates to resolve:
#1004, #1075, #1038, #1053, #1055, #1064, #1067, #1075, #1115, #301

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/pull/1121.html" title="Last updated on Aug 4, 2020, 8:31 PM UTC (6866f0c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/1121/e4dfc4d...6866f0c.html" title="Last updated on Aug 4, 2020, 8:31 PM UTC (6866f0c)">Diff</a>


detlevhfischer:
@alastc The changelog still seems to have a link title "Focus Visible Enhanced" (but with correct fragment identifier)

alastc:
Hi Detlev,

I think we'll leave that there, as it was called that, but then add another entry for the updated name.
However, it did remind me that I should add the (enhanced) version to the 'updates' section at the top, thanks :-)

mraccess77:
I have 2 questions:
* For adjacent contrast - does this require contrast for both sides of the focus rectangle or just one side?   I believe SC 1.4.11 allows for one side.
* Can we overlook intermediate colors like a gray shadow that might be between the focus indicator and the adjacent colors as the eye skips over those (as discussed in SC 1.4.11 understanding document).

alastc:
Hi Jon,

The 3rd bullet is (currently) against "all adjacent colors", so both sides. Remembering the test-cases, if it doesn't contrast in either direction it does need to be thicker. Also, if you move the outline in or out slightly, both sides would be against one colour (barring gradients etc).

We allow for areas of less contrast for the 1st two bullets, so long as they fulfill the overall area & contrast change. 
For the 3rd bullet I'd treat it like testing text-contrast, where you pick the least contrasting bit. 

I'm not sure which bit of 1.4.11 you mean, is it "Gestalt principles such as the "law of continuity" can be used to ignore minor overlaps with other graphical objects or colors." That was in the context of (usually) larger graphical objects where your eye is following a shape. For a focus indicator it is more about tracking that appearance/disappearance, I would be inclinced to keep it strict.

mraccess77:
Thanks @alastc My comment about skipping certain colours comes from this statement in the understanding document of SC 1.4.11
"If components use several colors, any color which does not interfere with identifying the component can be ignored for the purpose of measuring contrast ratio. For example, a 3D drop-shadow on an input, or a dark border line between contrasting backgrounds is considered to be subsumed into the color closest in brightness (perceived luminance)."

ajanec01:
This is the part I was also thinking about. Correct me if I'm wrong, currently, I interpret 1.4.11 and 2.4.11 as if they were complimenting each other because of the idea of the subsumed colours.

 **focus appearance:**
- focus indicator must have the minimum indication area
- the colour changes for the indicator must have sufficient contrast between the focused and unfocused state
-either all adjacent colours for the indicator have sufficient contrast ratio (page background and the UI component) or the indicator is at least 2 css px thick.

The third bullet does not cater for a 2px thick indicator which meets the first two bullet points requirement to have enough contrast ratio against the page background. However, 1.4.11 does. As far as I understand 1.4.11 says that the focus indicator must have enough contrast ratio against the page background but not the UI component itself. This is based on the part of the understanding page which @mraccess77 quoted and the part in the "Relationship with Use of Color and Focus Visible" section which says:

"... the visual focus indicator for a component must have sufficient contrast against the adjacent background when the component is focused,..."

The question is, have I interpreted 1.4.11 correctly. If yes, then they fit nicely with each other. If not, then there's an issue where the bullet point no 3 and 1.4.11 can be mutually exclusive.

patrickhlauke:
GitHub warns me (when going through https://github.com/w3c/wcag/branches/stale) that this PR can't/shouldn't be deleted because another open PR refers to this branch. leaving this intact for now.

PR #1151, opened 2020-06-09
Removed "proposed" from charter

koalie:
... and moved link to GH at the end of the footer.

This is to fix part of [Sam's issue](https://github.com/w3c/charter-drafts/pull/261#issuecomment-640750633) elsewhere.


samuelweiler:
If I understand correctly, the [published charter](https://www.w3.org/2019/12/ag-charter) is a snapshot of this, so that will need to be updated separately.

koalie:
My pull request was assuming that the GH file was what we were serving on the w3c website. Feel free to dispose of the PR as is best. Thanks!

PR #1152, opened 2020-06-09
Updated chairs

koalie:
Per 4 March 2020 announcement: [Chuck Adams and Rachael Montgomery appointed co-Chairs of the Accessibility Guidelines Working Group; Andrew Kirkpatrick steps down](https://lists.w3.org/Archives/Member/chairs/2020JanMar/0052.html).

### DIFF
--- a/charter.html
+++ b/charter.html
@@ -104,7 +104,8 @@
 						<td>
 							<ul>
 								<li>Alastair Campbell (Nomensa)</li>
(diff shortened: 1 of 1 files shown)


samuelweiler:
If I understand correctly, the [published charter ](https://www.w3.org/2019/12/ag-charter) is a static snapshot of this, so it will need to be updated separately.

Issue #869, opened 2019-08-24
1.4.3, 1.4.11 Contrasts of disabled elements


JAWS-test:
I have two questions on this subject:
* If I understood @patrickhlauke correctly (https://github.com/w3c/wcag/issues/857), the WCAG SCs are normative and all other documents (Understanding, Failure etc.) are not normative. This means that the latter can explain the SCs, but cannot extend, restrict or even contradict them. In SC 1.4.11, there is no restriction regarding deactivated elements (as opposed to those in 1.4.3). In the Understanding to 1.4.11 there is the restriction to disabled elements. That would not be allowed, would it? I understand the justification in 1.4.11, but is it permitted at all?
* Why should disabled elements always be excluded from the contrast requirement (for 1.4.3 and 1.4.11)? There are cases where these elements have to transport important information and thus should be recognizable. 
	* Example 1: Completion of the order process - to check the entries, all form fields are displayed disabled. If they are wrong, the form can be processed again with a "Back" function. 
	* Example 2: Application with missing access rights to change certain configurations, but the possibility to read them in disabled form fields.

jake-abma:
First  question:

> In SC 1.4.11, there is no restriction regarding deactivated elements (as opposed to those in 1.4.3). 

This is not correct as 1.4.11 states "except for inactive components", see:

> **User Interface Components**
> Visual information required to identify user interface components and states, **except for inactive components** or where the appearance of the component is determined by the user agent and not modified by the author;

-----------------------------

Second question:

> Why should disabled elements always be excluded from the contrast requirement (for 1.4.3 and 1.4.11)? 

You are right and it is correct that a group of people do want to perceive this information with enough contrast.
On the other site there is also the need from a visual perspective to make button "look" disables in order for the content to be understood in a fast and clear manner.

The result is that none of neither serves both worlds and this has been discussed many, many times in the WG. People in the WG do not disagree with both wishes but can't require one in favour of the other. A way to fix this might be the possibility of adding (disabled) icons, or providing multiple ways of showing disabled elements. In future this might also be done / possible with personalisation, work on this is being performed by the COGA TF.

Also in Silver (WCAG 3.0...) it will be on the agenda to see how we might find wording to solve this question.

JAWS-test:
First question: My fault, sorry!
Second question: That doesn't contradict itself to make something look deactivated and yet to present it with sufficient contrasts.

PR #1154, opened 2020-06-09
Proposed definition for Inactive

awkawk:
Context: https://www.w3.org/2020/06/09-ag-minutes#item07
Survey: https://www.w3.org/2002/09/wbs/35422/2020-05-cwa/results#x302

### DIFF
--- a/guidelines/terms/22/inactive.html
+++ b/guidelines/terms/22/inactive.html
@@ -1,7 +1,7 @@
 <dt><dfn>inactive</dfn> (user interface components)</dt>
 <dd class="new">
 	<p class="change">New</p>	
-   <p>A state of a <a>User Interface Component</a> when it is not available for user interaction, such as a disabled control in HTML. An inactive user interface component is visible but not currently operable using a keyboard, pointer, or assistive technology.
+   <p>A state of a <a>User Interface Component</a> when it is not available for user interaction, such as a disabled control in HTML. An inactive user interface component is visible but not currently operable using standard input mechanisms such as keyboard, pointer, or assistive technology.
     </p>
 <p class="note">An example inactive user interface component is a submit button at the bottom of a form which is visible but cannot be activated until all the required fields in the form are completed.</p>   				
 <p class="note">An inactive user interface component should not be confused with controls which are not currently selected. For example in a tabbed interface, the tabs which are not selected are not regarded as inactive user interface components - they may be interacted with by the user to switch tabs in the tabbed interface, and therefore do not meet the definition of inactive.</p>   				
--- a/guidelines/terms/22/inactive.html
+++ b/guidelines/terms/22/inactive.html
@@ -4,5 +4,6 @@
    <p>A state of a <a>User Interface Component</a> when it is not available for user interaction, such as a disabled control in HTML. An inactive user interface component is visible but not currently operable using standard input mechanisms such as keyboard, pointer, or assistive technology.
     </p>
 <p class="note">An example inactive user interface component is a submit button at the bottom of a form which is visible but cannot be activated until all the required fields in the form are completed.</p>   				
-<p class="note">An inactive user interface component should not be confused with controls which are not currently selected. For example in a tabbed interface, the tabs which are not selected are not regarded as inactive user interface components - they may be interacted with by the user to switch tabs in the tabbed interface, and therefore do not meet the definition of inactive.</p>   				
+<p class="note">An inactive user interface component should not be confused with controls which are not currently selected. For example in a tabbed interface, the tabs which are not selected are not regarded as inactive user interface components - they may be interacted with by the user to switch tabs in the tabbed interface, and therefore do not meet the definition of inactive.</p>
+  <p class="note>A read-only text input field which is focusable using standard input mechanisms is considered to be active.</p>
 </dd>
--- a/guidelines/terms/22/inactive.html
+++ b/guidelines/terms/22/inactive.html
@@ -1,9 +1,8 @@
 <dt><dfn>inactive</dfn> (user interface components)</dt>
 <dd class="new">
 	<p class="change">New</p>	
-   <p>A state of a <a>User Interface Component</a> when it is not available for user interaction, such as a disabled control in HTML. An inactive user interface component is visible but not currently operable using standard input mechanisms such as keyboard, pointer, or assistive technology.
-    </p>
-<p class="note">An example inactive user interface component is a submit button at the bottom of a form which is visible but cannot be activated until all the required fields in the form are completed.</p>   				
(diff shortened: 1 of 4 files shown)


patrickhlauke:
a rewording/restructuring of the proposed PR here? https://github.com/w3c/wcag/pull/638

awkawk:
@patrickhlauke yes, based on the discussion on the call earlier today.

patrickhlauke:
i'll close the other one then...

patrickhlauke:
Just a heads-up ... is this going to be merged?

guidelines/terms/22/inactive.html
```diff
@@ -0,0 +1,8 @@
+<dt><dfn>inactive</dfn> (user interface components)</dt>
+<dd class="new">
+	<p class="change">New</p>	
+   <p>A state of a <a>User Interface Component</a> when it is not available for user interaction, such as a disabled control in HTML. An inactive user interface component is present but not currently operable using standard input mechanisms such as keyboard, pointer, or assistive technology.</p>
+<p class="note">An example inactive user interface component is a submit button at the bottom of a form which is present but cannot be activated until all the required fields in the form are completed.</p>
```
jnurthen, line 5: What about controls that take focus but cannot be activated such as a date in a grid which is not available or  the other control types specified at https://w3c.github.io/aria-practices/#kbd_disabled_controls.

patrickhlauke: i think that's covered by the proposed definition there, since it says nothing about whether or not the inactive component can receive focus or not...just that it's not available for user interaction (unless you mean that even the act of focusing on it is an interaction?) / operable.

jnurthen: yes - I think we could have endless debates on that. Don't we all love endless debates.

guidelines/terms/22/inactive.html
```diff
@@ -0,0 +1,8 @@
+<dt><dfn>inactive</dfn> (user interface components)</dt>
+<dd class="new">
+	<p class="change">New</p>	
+   <p>A state of a <a>User Interface Component</a> when it is not available for user interaction, such as a disabled control in HTML. An inactive user interface component is present but not currently operable using standard input mechanisms such as keyboard, pointer, or assistive technology.</p>
```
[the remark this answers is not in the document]
patrickhlauke: i'd still say that this is unnecessary, but can live with it...

alastc:
It would be useful to resolve whether we think 'inactive' controls can be focusable before taking it back to the group.

Sorry, slipped off my radar as the issue was closed.

patrickhlauke:
personally, still think that the concepts of inactive/disabled/inoperable and focusable are two orthogonal things. thinking for instance about `readonly` inputs that can receive focus by default, but are inoperable in the sense that it's not possible to edit them (unless this now opens up a different can of worms about "is being in a readonly state really 'inactive'?")

patrickhlauke:
so we *may* need some clarification of this aspect too, perhaps?

* inactive means you can't activate it/change the selection (for `<select>` type things)/check it/toggle it
* you *may* be able to focus it
* read-only inputs don't count as inactive because you can still select their content and copy it (and because it would open up too much of a loophole if they were exempted from contrast requirements etc)

jnurthen:
> It would be useful to resolve whether we think 'inactive' controls can be focusable before taking it back to the group.
> 
We go into this a bit in ARIA Authoring Practices https://www.w3.org/TR/wai-aria-practices-1.2/#kbd_disabled_controls

JAWS-test:
The fact that disabled or inactive elements are excluded from the contrast requirements is, in my opinion, a mistake in the WCAG. This is shown by the ARIA example. If the ARIA Guidelines say that sometimes disabled elements should be accessible so that they can be perceived with assistive technology, then these elements should also have sufficient contrast. See https://github.com/w3c/wcag/issues/869

alastc:
> The fact that disabled or inactive elements are excluded from the contrast requirements is, in my opinion, a mistake in the WCAG.

Perhaps, but it is due to there not being a solution to the problem of many people thinking the control is active when it isn't. I've seen that happen myself in usability testing. Without a solution to that problem, it isn't likely to change.

If Patrick's list above is agreeable then perhaps something like (bold for new, at the end):

> A state of a <a>User Interface Component</a> when it is not available for user interaction, such as a disabled control in HTML. An inactive user interface component is present, and in some cases may even be focusable, but is not currently operable using standard input mechanisms such as keyboard, pointer, or assistive technology. **A read-only control is <em>active</em> because the text should be readable and available for copy operations**.

patrickhlauke:
would be good for this to be finalised/clarified (as this topic has come up in discussions again today)

patrickhlauke:
I'd reopen this, as it seemed close to agreed/consensus at the time, and probably still something that would be good having?

patrickhlauke:
@alastc would be good to get this looked at again in the WCAG 2.x calls at some point soon to clear the deck

mbgower:
> I have always understood that readonly controls are NOT inactive. Read-only controls can be operated (you can move the cursor and copy them) so they certainly shouldn't be exempt from any contrast requirements.

I agree that a read-only control is not inactive. Most can take focus, and if a read-only control's content was greyed out, I would consider that a failure of Contrast, since the content is in a state intended for consumption. So I'm inclined to say that something that can take focus is _not_ inactive. That seems to be supported by the final note:
> A read-only text input field which is focusable using standard input mechanisms is considered to be active.

guidelines/terms/22/inactive.html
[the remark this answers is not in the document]
mbgower: Nope, I think this is marring the definition of inactive, which I think is both inoperable AND not navigable. 
```suggestion
   <p>A state of a <a>User Interface Component</a> when it is not available for user interaction, such as a disabled control in HTML. An inactive user interface component is present but not currently operable or navigable using standard input mechanisms such as keyboard, pointer, or assistive technology.</p>
```

guidelines/terms/22/inactive.html
mbgower: That wording from the APG really complicates things. I'm inclined to say that if something is navigable but not operable, it should not be considered disabled (and should not be given a pass on text contrast considerations).  We spent a lot of time at IBM designing read-only versions of components; it was difficult, but it is possible. There should be a way of achieving minimum contrast for unavailable options in a listbox (to use one of the APG examples), if the desire is to make them discoverable by all. Letting them be 'discoverable' programmatically but not visually seems inequitable.

fstrr:
> We go into this a bit in ARIA Authoring Practices https://www.w3.org/TR/wai-aria-practices-1.2/#kbd_disabled_control

If anyone is looking for that content, I believe it's moved to this [Developing A Keyboard Interface](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#focusabilityofdisabledcontrols) page.

Issue #1149, opened 2020-06-08
Link purpose (2.4.4 and 2.4.9) "ambiguous to users in general" is self-defeating


patrickhlauke:
On their face, 2.4.4 Link Purpose (in Context) and 2.4.9 Link Purpose (Link Only) want to ensure that links are sufficiently descriptive of their purpose/the content that they link to (with 2.4.4 allowing for context, and 2.4.9 making the requirement of the link itself in isolation). From the understanding documents, this is about aiding users in being able to decide whether or not they want to follow those links. It mentions users with motion limitations, cognitive disabilities, and visual impairments. All good so far.

However, the "except where the purpose of the link would be ambiguous to users in general" exception for both SCs actually completely guts most of the intent here. if the page itself doesn't provide any more clues/explanations about what a link does, it arguably is "ambiguous to users in general". There's a certain circular logic at work here that if a link's text does not describe its purpose accurately/sufficiently, and there is no other context provided anywhere else on the page, then you can't really say "this link should be more descriptive", because by the very fact that it isn't, it's "ambiguous to users in general".

In practice then, what this SC really boils down to, is: if there's other content in the page (be it other text, or layout that suggest some kind of relationship, or use of color, or anything else - which may also fail other SCs, like 1.3.1 or 1.4.1) that does provide more information about the purpose of a link, it needs to either be associated with / programmatically determinable (such as being part of the same paragraph) for the link (for 2.4.4) or even be included/associated as part of the link itself (for 2.4.9). And for 2.4.9, it's even acceptable to use visually hidden/AT only solutions. So really, the only users who really benefit from these SCs, when there *is* additional information/description in the page, are AT users. The SCs, in these cases, can be satisfied while still not helping users with limited motion or cognitive disabilities at all.

And of course, if there *isn't* anything else in the page, then the link doesn't need to be descriptive (of its purpose, or what it links to) because it's then by definition "ambiguous to users in general". So the whole rationale of the SCs (to make sure it's easy for users to work out what a link does/where it points to) gets defeated outright.

Or am I missing something?

Long story short: these SCs only have any real bite when there is unarguably further context provided in the page already. And they only really help AT users.

patrickhlauke:
extreme example: a page with just a giant "click here" link? passes 2.4.4 and 2.4.9, as it's ambiguous to all users.

[...]

Issue #1097, opened 2020-04-07
2.4.4 and preceding headings as context


patrickhlauke:
I know this is an old itch that keep getting scratched (most recently, mentioned here https://github.com/w3c/wcag/issues/819), but: was the determination that headings do not count as context based on the current state of what AT currently do/support, rather than what is technically possible?

It would be trivial enough to programmatically determine, for a link, what its closest preceding heading would be (under which the link would logically/structurally/hierarchically fall under).

And if the deciding factor was indeed "based on what current AT/UAs do", how would that rationale square up against similar situations where it has been argued that something is "theoretically" possible (e.g. in the case of 2.4.1, where the concerns about accepting ARIA landmarks or headings since they won't help non-AT keyboard users is usually countered with "it would be easy enough to make an extension that keyboard users can install"; or 1.3.5 where a large part of the rationale relies on "once UAs/ATs can do something meaningful with a particular way of identifying the purpose of controls"; and lastly, if memory serves, there are even arguments around 2.4.9 and certain advisory techniques, like using title, which don't have full AT support - where not every AT actually exposes it when getting a list of links, but in that case it was deemed a shortcoming of the ATs because they theoretically could).

mraccess77:
For the record JAWS does support announcing the preceding heading when JAWSKey+T is used to speak the title.  And for 2.4.1 I/ we don't allow sole reliance on headings/landmarks -- skip links need to be present.

JAWS-test:
See also: https://github.com/w3c/wcag/issues/1047

patrickhlauke:
adding here as i forgot about it myself / hadn't clocked it in #1047 - there's still this technique https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/H80 (but that's apparently deemed "advisory" only, though it's difficult to see how/why that'd be the case)

patrickhlauke:
a decision on this would be nice to have. i'd still say that by the very nature of how headings work in HTML (that a heading currently acts as heading for anything that comes after it, until a heading of same or higher hierarchy is encountered), they do provide unambiguous programmatically-defined context (but the fact that not all AT expose this is obviously a problem). and again, squaring this up against other cases where certain approaches have been deemed "ok if the user installs a readily available extension or similar"...

patrickhlauke:
did we ever reach a consensus on this?

detlevhfischer:
[...]

PR #1164, opened 2020-06-17
Update Link Purpose (in Context) understanding benefits to focus on AT

detlevhfischer:
Edited benefits section to focus of users of assistive technologies.

Relates to #1149

[Diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fwcag%2Funderstanding%2Flink-purpose-in-context&doc2=https%3A%2F%2Fdeploy-preview-1164--wcag2.netlify.app%2Funderstanding%2Flink-purpose-in-context)


understanding/20/link-purpose-in-context.html
```diff
@@ -95,21 +95,12 @@ <h2>Benefits of Link Purpose (In Context)</h2>
...
-            navigation to and from content they are not interested in.
-         </li>
-         
+                
          <li>
-            People with visual disabilities will be able to determine the purpose of a link by
-            exploring the link's context. 
-            
-            
+            This Success Criterion helps people people with visual disabilities who use assistive technologies. Using screen reader shortcuts, these users can quickly explore a link's programmatically determinable context.
```
patrickhlauke, line 100: `This Success Criterion helps people people` duplicate word

patrickhlauke:
pro-tip: if you add "Closes #1149" to the PR description, github will auto-close the issue once the PR gets merged. (adding this now myself here, if you don't mind)

patrickhlauke:
i'd say this still has value (admittedly not checked the full context here of whether or not the WG disagreed with the premise)

fstrr:
I'll add it to the board.

bruce-usab (changes requested):
@detlevhfischer , I like your updated phrasing for how meeting this SC this benefits users of screen reading software.

I am not sure about deleting the two bullets mentioning benefits for people with mobility impairments and for coga.

At present, those two bullets are poorly worded, but I would prefer to leave them in, unless you want to take a turn at improving them.  Improving the two bullets could be separate issue/PR if you prefer.

patrickhlauke:
> I am not sure about deleting the two bullets mentioning benefits for people with mobility impairments and for coga.

As discussed in the original issue, I can not see - nor think of - any advantage for these groups coming from satisfying this SC. the "or from the link text together with its [programmatically determined link context](https://www.w3.org/TR/WCAG22/#dfn-programmatically-determined-link-context)," aspect of the SC has absolutely no bearing on these user groups, UNLESS they use assistive technology that does something clever with the programmatically determined relationship. otherwise, it's just behind-the-scenes plumbing that has absolutely no bearing on these users.

The only aspect that would benefit these users is the first part "The [purpose of each link](https://www.w3.org/TR/WCAG22/#dfn-purpose-of-each-link) can be determined from the link text alone", but I'll be honest it's rare that content passes 2.4.4 purely on the merit of that first part - it's the programmatically determined aspect that is relied on in the majority of cases I come across.

However, granted, if those bullet points could be reworded to make the above clear (that it's the "link text alone" aspect that will benefit these two user groups, but *not* the "programmatically determinable"), then fine.

GreggVan:
@patrickhlauke   I  think you are entirely missing the role of AT in ALL disabilities.

There has long been a misinterpretation that all "programmatically determined" provisions are just for people who are blind - so they could use their screen meters. They are not; they are there to enable all sorts of Assistive technologies. In the past, we have not had very many Assistive technologies for groups like cognitive, language, and learning disabilities but that is no longer true. So all of these provisions that talk about Information being programmatically determinable (machine readable) always were for all disabilities -- and that will be increasingly true as we go forward.

patrickhlauke:
>  I think you are entirely missing the role of AT in ALL disabilities.

no, i'm not [and i'm sorry, but that comment was perhaps a shade too patronising perhaps?]. as said [and now clarified further in my comment above to make it even more explicit], the bullets need to make clear that "programmatically determined" will only be exposed to people using AT. whether the people using AT are doing so because they are blind, or have cognitive disabilities, or whatever, doesn't detract from my statement that the bullets need to make clear that the programmatically determined part will only work for people using AT - and not all keyboard users or users with cognitive disabilities use AT, so those bullets are misleading in their current form.

GreggVan:
@patrickhlauke   Sort of.  
it is true today - but we are writing for the future.   One of the things being pushed for is browsers that can provide more AT-like functionality as part of the standard browser features.  (like zoom).    This will be a result of AI enhancements in the browsers.   So that markup can be used by more than AT.  it can be used by any program - any code.     And there are many who need more accessible experiences who do not have and will not be provided with AT.   And those who don't have their own computer and have to use public computers cannot add any AT.     But all the great assistive features mainstream industry is building into their platforms  (os and browsers) will help them and be available to all.    And this can benefit them  and even "mainstream" users who don't know they have a disability but would be helped by features that are built in and may not even look like assistive technologies.   

That is all I  was trying to get at.   That these things are useful to more than we have typically considered in the past.

By the way --   an even more exciting development -- even farther off - will be smart browsers that will take inaccessible content - and make it accessible.     Then browsers that will provide custom interfaces to content tuned to each individual.     Then finally info-bots with Individual User Interface Generators that can take any interface - on or off the web - and do that.   That will open up so many things that are not accessible  to so many new people who we have not been able to reach...      But we need to keep doing what works now til we get there.    Just important to see if we can write things that will gracefully slide into the future -- that is coming.    best

patrickhlauke:
> it is true today - but we are writing for the future

the understanding/explanations for SCs should really focus on the here and now. for somebody who's trying to understand why they should be implementing an SC, or what benefits come from satisfying an SC now, it's a bit pointless to refer to a future hypothetical. unless you explicitly mention "once tools that take advantage of this exist and are widespread". otherwise, it's just misleading the reader.

GreggVan:
I guess we have to agree to disagree.   
1) a major criticism of WCAG is the way we write the provisions.  Why not write them simpler -- when the reason is that they need to stand the test of time.   WCAG 2 will be used for 20 years.    And a bunch of those years are still in front of us.
2) a lot of benefit in the next few years to groups we have wanted to serve better will come from work that was done 15-20 years ago.   We always need to be thinking about the future
3)  we are now working on W3  -- and we need to be thinking about the future all the time. Things are going faster each year and we want (need) W3 to be usable for many years once it is out.  and the Understanding WCAG 2 docs are where all our new people cut their teeth.
4) There is so much misunderstanding about why things are worded as they were. And some provisions are misunderstood as being only about existing problems and solutions looking backward -- when they are in fact the foundation for solutions for many othter groups looking forward

Finally -- the Understanding WCAG 2 docs are one of our best teaching guides to the whole field of accessibility -  for ICT of all types.   We need everyone to be thinking not just about today - but the future.    Today is gone tomorrow  - and time flies by.  People learning today will be the ones doing tomorrow.    They need to understand the world they will be needing to solve problems in --  not just now and the past.       all the best.

patrickhlauke:
understanding and techniques documents can and should be updated as things change - that's been a criticism of WCAG as well, the fact that so much of this non-normative stuff has languished for far too long. no point lying to people now about benefits that are currently only theoretical, because THAT then also leads to embarassing conversations with clients when they question "so here it says this benefits these users. how, exactly?" and you have to admit that well, currently, this actually doesn't.

update the documents once it's actually a reality. OR explicitly say that at time of writing, this is only theoretical.

GreggVan:
Woaw.  Woaw.  Woaw.    Who said anything about lying to people.    Let's stay on topic.   I was talking about the important of not just talking about the benefits today, but also talking about how these things can benefit people in the future - as well.  And how things are structured to support future developments. 

I agree that the informative docs have languished and need to be constantly updated. (easy to say but it is a boatload of work - much more than people realize) 

I strongly disagree that we should not both inform and get input from the field regarding where things are going, and how we are preparing for it,  so we can get input on the priorities of consumers in the field.    Just talking about what is -- and not what is coming or possible - or might be possible does not help them or us figure out what we should be doing in our work -- which is all work for the future.      I am going to step off here.     This is turning into a dialog.  Happy to do this offline if you like     Best

detlevhfischer:
@patrickhlauke more than happy. Delirious!

patrickhlauke:
@detlevhfischer @bruce-usab had a stab at tweaking this PR, to more clearly make the distinction between the two aspects/ways in which this SC can be satisfied - link text alone (which benefits all users), OR link text and its programmatically determined context (which does only benefit users that have an assistive technology that takes advantage of/exposes the link context)

patrickhlauke:
@kfranqueiro not urgent, but: any reason why the netlify deploy preview isn't working on this?

patrickhlauke:
> There's no top-level index page so you need to add /techniques/ or /understanding/ to the URL.

@kfranqueiro ah sorry, that's the bit I was missing. all good

understanding/20/link-purpose-in-context.html
```diff
@@ -1,403 +1,187 @@
...
       <p>In some situations, authors may want to provide part of the description of the link
          in logically related text that provides the context for the link. In this case the
          user should be able to identify the purpose of the link without moving focus from
          the link. In other words, they can arrive on a link and find out more about it without
          losing their place. This can be achieved by putting the description of the link in
-         the same sentence, paragraph, list item, or table cell as the link, or in the table header cell for a link in a data table, because these are directly associated with the link itself. Alternatively, authors may choose to use an ARIA technique to associate additional
-         text on the page with the link.
-      </p>
-      
+         the same sentence, paragraph, list item, or table cell as the link, or in the table header cell for a link in a
```
detlevhfischer, line 55: I know the [H80](https://www.w3.org/WAI/WCAG22/Techniques/html/H2.html) (providing programmatic context via the preceding heading) has been demoted to advisory some years back, and I also know that I am not the only one to disagree with that decision. On the basis that Techniques are all informational content and the practice to provide context via the preceding heading seems both widespread and certainly no less obscure or less understood than things like using the table header to provide context, I wonder if there is a chance to re-promote H80. This would allow us to also mention headings here as a way to provide programmatic context. (I realize this would require a different PR and possibly drag on for quite some time, so I am all for going with these changes before we get there).

understanding/20/link-purpose-in-context.html
```diff
@@ -1,403 +1,187 @@
...
-         and destinations to have different link text.
-      </p>
-      
-     <p>A best practice for links to <a>conforming alternate versions</a> is to ensure that the link text to the conforming alternate version indicates in link text that the page it leads to represents the more accessible version. This information may also be provided in text - the goal is to ensure that the end user knows what the purpose of the link is. </p>
-     
+         (and this is a requirement per
+         <a href="consistent-identification" class="guideline">Success Criterion 3.2.4</a> for pages in a set). It is
+         also a best practice for links with different purposes
+         and destinations to have different link text.</p>
+      <p>A best practice for links to <a>conforming alternate versions</a> is to ensure that the link text to the
```
detlevhfischer, line 68: I know my suggestion would make the Understanding text even longer, but should there be something about the inverse case (links with different text leading to the same target)? I seem to remember that some people have been seen failing content just for that...

understanding/20/link-purpose-in-context.html
```diff
@@ -1,403 +1,187 @@
...
-         <li>
-            People with visual disabilities will be able to determine the purpose of a link by
-            exploring the link's context. 
-            
-            
-         </li>
-         
+         <li>Links whose link text clearly conveys their purpose can help people who navigate with a
+            keyboard or keyboard-like interface, letting them skip links that they are not interested in,
+            avoiding the keystrokes needed to visit the referenced content and then returning to
```
detlevhfischer, line 94: >  avoiding the keystrokes needed to visit the referenced content and then returning to the current content.

Maybe better something like: "...avoiding the effort to visit the content referenced only to find that it is not what they want." The main point being that "keystrokes" is replaced by generic navigation.

understanding/20/link-purpose-in-context.html
```diff
@@ -1,403 +1,187 @@
...
+            the current content.</li>
+         <li>When the link text is clear, people with cognitive limitations will not become disoriented
```
detlevhfischer, line 96: > will not become disoriented by multiple means of navigation

Maybe better: "...are less likely to become disoriented..."

understanding/20/link-purpose-in-context.html
```diff
@@ -1,403 +1,187 @@
...
+            by multiple means of navigation to and from content they are not interested in.</li>
+         <li>Links whose purpose is conveyed through a combination of link text and <a>programmatically determined
+            link context</a> will help people using assistive technologies to decide whether or not to follow a
```
detlevhfischer, line 99: > help people using assistive technologies to decide whether

help people using assistive technologies decide whether...

understanding/20/link-purpose-in-context.html
```diff
@@ -1,403 +1,187 @@
...
+            particular link. For instance, depending on the specific technique an author used to provide the context,
```
detlevhfischer, line 100: > specific technique an author used to provide

specific technique an author has used to provide

detlevhfischer (commented):
See comments inline...

understanding/20/link-purpose-in-context.html
patrickhlauke: Yeah suggest doing this as a separate PR, as this would likely stall this entire effort in endless discussions. (also noting that it's not that using table headers is also obscure, but rather that by default SRs provide a way to - without moving focus - have the relevant table headers that relate to the current cell announced, whereas for "what's the preceding heading" - while eminently doable from a structure point of view - is not an option you currently get, so not "accessibility supported" as such?)

understanding/20/link-purpose-in-context.html
patrickhlauke: This is already mentioned/covered by the "best practice for links with the same destination to have consistent text" ... as that is best practice, it follows (for me at least) that it's NOT a failure if links with different text point to the same destination?

understanding/20/link-purpose-in-context.html
```diff
@@ -1,403 +1,187 @@
...
-         </li>
-         
-         <li>
-            People with visual disabilities will be able to determine the purpose of a link by
-            exploring the link's context. 
-            
-            
-         </li>
-         
+         <li>Links whose link text clearly conveys their purpose can help people who navigate with a
```
scottaohara, line 92:
```suggestion
         <li>Links whose text clearly conveys their purpose can help people who navigate with a
```

understanding/20/link-purpose-in-context.html
```diff
@@ -1,403 +1,187 @@
...
+            keyboard or keyboard-like interface, letting them skip links that they are not interested in,
```
scottaohara, line 93: this seems more like rational that helps everyone rather than just people using keyboard?   As a person who navigates using [ pointer, keyboard, touch, voice dictation, etc. ] if link text is descriptive, I can ignore links that aren't of interest to me.

scottaohara: additionally, this seems to stress on link-text alone being descriptive enough (2.4.9), but being this is link purpose in context, why isn't this at least also making mention of a link and its surrounding text?

understanding/20/link-purpose-in-context.html
```diff
@@ -1,403 +1,187 @@
...
+            avoiding the effort to visit the referenced content only to find it is not what they wanted, and
+            having to then navigate back.</li>
+         <li>When the link text is clear, people with cognitive limitations are less likely to become disoriented
```
scottaohara, line 96: link text, or a link's purpose (again, re: link text _and_ surrounding text are sufficient for this particular SC)

understanding/20/link-purpose-in-context.html
```diff
@@ -1,403 +1,187 @@
...
+            by multiple means of navigation to and from content they are not interested in.</li>
+         <li>Links whose purpose is conveyed through a combination of link text and <a>programmatically determined
```
scottaohara, line 98: per my above comments, the start of this list item is the sort of wording i'd have expected.

understanding/20/link-purpose-in-context.html
```diff
@@ -1,403 +1,187 @@
...
-      
+         link so they can decide whether they want to follow the link.</p>
+      <p>Whenever possible, provide link text that identifies the purpose of the link without
+         needing additional context – this approach satisfies the <q>purpose of each link can be
+         determined from the link text alone</q> aspect of this Criterion, and will benefit
+         all users, whether they use assistive technologies or not. Assistive technologies
+         can provide users with a list of links that are on the Web page. Link text that is as
+         meaningful as possible will aid users who want to choose from this list of links.
+         Meaningful link text also helps those who wish to tab from link to link. Meaningful links
+         help users choose which links to follow without requiring complicated strategies
```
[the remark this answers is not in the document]
patrickhlauke: oh yeah, that's some mysterious complicated original wording. i'll tweak that away...

understanding/20/link-purpose-in-context.html
patrickhlauke: to satisfy the SC you have link text OR link text plus programmatic context (which may not necessarily be surrounding text, but also things like `aria-describedby` pointing somewhere completely differently). so I was trying to make the distinction here for "if you satisfy the SC with the first bit, that'll help all/most people (even without AT), whereas the programmatic bit may only help people WITH AT"

understanding/20/link-purpose-in-context.html
patrickhlauke: surrounding text is not the only programmatically determinable context, which complicates matters. heck, adjacent text (in a separate paragraph, and even when not associated programmatically, can help COGA users who don't rely on AT...that's the complication/weirdness of this SC)

understanding/20/link-purpose-in-context.html
scottaohara: yeah, i get that regarding its link text OR link text + context for the SC.   I think i just globed onto it being perceived more as a 2.4.9 recommendation because I was already questioning rational of the benefit specific to keyboard users that doesn't equally help other people with/without disabilities.  

so never mind then the comment about being more 2.4.9 focused.  But I still question this specific benefit callout in general.

understanding/20/link-purpose-in-context.html
scottaohara: yah again, get that.  i'm fine with resolving this particular comment since i have the other one open still.

understanding/20/link-purpose-in-context.html
detlevhfischer: @patrickhlauke table headers / headings - yes, I see your point.

mraccess77: FWIW - with JAWS - you can press JAWSKey+T to hear the preceding heading (in additional to page title) from where you are on the page.  It also tells you the heading level.

understanding/20/link-purpose-in-context.html
detlevhfischer: I wasn't suggesting this should fail, a note should rather serve to explain that this is not a failure. And if already covered, fine, I didn't spot that.

patrickhlauke: I'd say yes, it's implicit in the fact that this says it's a best practice. i.e. not a requirement, so not doing that won't necessarily fail anything.

understanding/20/link-purpose-in-context.html
patrickhlauke: in which case @mraccess77 we can probably reopen the discussion (separately) about whether or not the technique to use headings as context is advisory or sufficient...

understanding/20/link-purpose-in-context.html
patrickhlauke: @scottaohara following your (very valid, I hasten to add) points, I've had another go at tweaking the benefits. let me know what you think

understanding/20/link-purpose-in-context.html
patrickhlauke: https://github.com/w3c/wcag/issues/1097 is probably the last/latest separate issue about this where discussion can continue

understanding/20/link-purpose-in-context.html
```diff
@@ -84,90 +76,54 @@ <h2>Intent of Link Purpose (In Context)</h2>
...
-         <li>
-            <a href="http://webaim.org/techniques/hypertext/">WebAIM Techniques for Hypertext Links</a>						     
-         </li>
+         <li><a href="https://www.nngroup.com/articles/title-attribute/">Using the Title Attribute to Help Users Predict Where They Are Going</a></li>
+         <li><a href="https://webaim.org/techniques/hypertext/">WebAIM Techniques for Hypertext Links</a></li>
       </ul>
-      
    </section>
-   
+
```
patrickhlauke:
```suggestion
```

understanding/20/link-purpose-in-context.html
```diff
@@ -7,75 +7,67 @@
...
+         needing additional context – this approach satisfies the <q>purpose of each link can be
+         determined from the link text alone</q> aspect of this Criterion, and will benefit
+         all users, whether they use assistive technologies or not. Assistive technologies
+         can provide users with a list of links that are on the web page. Link text that is as
+         meaningful as possible will aid users who want to choose from this list of links.
+         Meaningful link text also helps those who wish to tab from link to link.</p>
+      <p>The second aspect of this Criterion, <q>link text together with its <a>programmatically
+         determined link context</a></q>, primarily benefits assistive technology users.
+         If it's not possible to provide link text that, in isolation, conveys the purpose of the link,
+         authors must make sure that there is at least additional <em>context</em> for the link.
```
scottaohara: word swap to not use 'must' in the understanding doc

```suggestion
         authors will need to make sure that there is at least additional <em>context</em> for the link.
```

patrickhlauke:
Check the overlap/relationship with https://github.com/w3c/wcag/pull/1570

bruce-usab (approved):
[Diff preview](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FUnderstanding%2Flink-purpose-in-context.html&doc2=https%3A%2F%2Fdeploy-preview-1164--wcag2.netlify.app%2Funderstanding%2Flink-purpose-in-context)

PR #1165, opened 2020-06-17
Changed pretty to straight quotes

### LINKED ISSUE #1148: The working example for ARIA9 - Example 3 has problems
https://www.w3.org/WAI/WCAG21/working-examples/aria-labelledby-workshop-booking-timetable/

There are 2 tables in this working example:

- the dinosaur conference table from ARIA9
- a second layout table containing the text field labelled 'Maiden name' (which isn't part of example 3 in ARIA9)

Smart quotes (U+201D) have been used for the role attribute value on the second table instead of ASCII U+0022 double-quote. This triggers the following message in the W3 validator:

> an attribute value must be a literal unless it contains only name characters

This is parsed as `role="”presentation”"` in the DOM (with the smart quotes as part of the role name) so the role is unrecognised and falls back to the native table role in Firefox/Chrome/Edge. In Safari layout table heuristics kick-in which stops it being exposed as a table.

Might be simplest to just remove the second table since it doesn't add anything to the example (and some people may be unhappy with the field label).

detlevhfischer:
Relates to issue [The working example for ARIA9 - Example 3 has problems](https://github.com/w3c/wcag/issues/1148)
I have replaced smart quotes with straight quotes.
Closes #1148

### DIFF
--- a/working-examples/aria-labelledby-workshop-booking-timetable/index.html
+++ b/working-examples/aria-labelledby-workshop-booking-timetable/index.html
@@ -90,12 +90,12 @@ Dinosaur conference workshop booking table
 </tr>
 </table>
 <p>&nbsp;</p>
-<table role=”presentation” cellspacing="0" cellpadding="4" border="0">
+<table role="presentation" cellspacing="0" cellpadding="4" border="0">
 <tr>
 <td><label for="123">Maiden Name:</label></td>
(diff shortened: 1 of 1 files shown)


alastc:
Thanks Detlev, I think that qualifies as editorial so I'll merge that in.

PR #1166, opened 2020-06-17
Qualify Sufficient Techniques

detlevhfischer:
Added a note after h3 Sufficient Techniques to flag the fact that the value of two of Techniques listed is limited to screen reader users: ARIA8 (Using aria-label for link purpose) and C7 (Using CSS to hide a portion of the link text).

### DIFF
--- a/understanding/20/link-purpose-link-only.html
+++ b/understanding/20/link-purpose-link-only.html
@@ -44,16 +44,14 @@
       <p>The Success Criterion includes an exception for links for which the purpose of the
          link cannot be determined from the information on the Web page. In this situation,
          the person with the disability is not at a disadvantage; there is no additional context
-         available to understand the link purpose. However, whatever amount of context is available
-         on the Web page that can be used to interpret the purpose of the link must be made
-         available in the link text to satisfy the Success Criterion.
+         available to understand the link purpose.
       </p>
       
       <p>The word "mechanism" is used to allow authors to either make all links fully understandable
          out of context by default or to provide a way to make them this way. This is done
          because for some pages, making the links all unambiguous by themselves makes the pages
          easier for some users and harder for others. Providing the ability to make the links
-         unambiguous (by them selves) or not provides both users with disabilities with the
+         unambiguous (by themselves) or not provides both users with and without disabilities with the
          ability to use the page in the format that best meets their needs.
       </p>
       
@@ -153,14 +151,7 @@
             								       
             <a href="http://webaim.org/techniques/hypertext/">WebAIM Techniques for Hypertext Links</a>
             							     
-         </li>
-         
-         <li>
-            	
-            <a href="http://www.rnib.org.uk/blogs/expert-series?Name=Hidden%20barriers">Hidden barriers - out of sight</a>
-            							     
-         </li>
-         
+         </li>       
       </ul>
       
    </section>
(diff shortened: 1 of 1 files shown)


detlevhfischer:
In my view, the preferable approach would be to weed out 2.4.9 Techniques that are limited to providing full context to people with specific disabilities (here, screen reader users). Some may be demoted to Sufficient Techniques. This would imply changes in the Techniques' references to SCs - so not something that can be done quickly. That's why this note may serve as a first stop gap.

fstrr:
@detlevhfischer Do you still want to merge this in? If you do, taking out this would be good: ```<div role="heading" class="note-title marker" aria-level="4">Note</div>```

understanding/20/link-purpose-link-only.html
```diff
@@ -44,16 +44,14 @@ <h2>Intent of Link Purpose (Link Only)</h2>
       <p>The Success Criterion includes an exception for links for which the purpose of the
          link cannot be determined from the information on the Web page. In this situation,
          the person with the disability is not at a disadvantage; there is no additional context
-         available to understand the link purpose. However, whatever amount of context is available
-         on the Web page that can be used to interpret the purpose of the link must be made
-         available in the link text to satisfy the Success Criterion.
+         available to understand the link purpose.
```
mbgower: Although I get the concern, I believe the existing wording provides a cue that link context can often be gleaned from other content on the page.
I would prefer to retain the removed sentence.

understanding/20/link-purpose-link-only.html
```diff
@@ -171,16 +162,23 @@ <h2>Techniques for Link Purpose (Link Only)</h2>
       
       <section id="sufficient">
          <h3>Sufficient Techniques for Link Purpose (Link Only)</h3>
+       
+        <div class="note">
+               <div role="heading" class="note-title marker" aria-level="4">Note</div>        
+        <p>Some of the Sufficient Techniques listed below provide self-contained link purpose only to some groups of users. <a class="aria" href="https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA8">ARIA8: Using aria-label for link purpose</a> replaces the visible link text with a more fleshed-out description in the value of the aria-label attribute, providing the purpose of the link to visually impaired users using a screen reader. The Technique, however, does not help users with cognitive disabilities since the visual context does not provide the purpose of the link.</p>
```
mbgower: While I think this may reflect the general state of AT support and features, I don't actually think this is necessarily true. There is no reason that the information provided through aria-label could not be exposed by an AT specifically designed to provide context for users with cognitive disabilities.

understanding/20/link-purpose-link-only.html
```diff
@@ -44,16 +44,14 @@ <h2>Intent of Link Purpose (Link Only)</h2>
...
-         available in the link text to satisfy the Success Criterion.
+         available to understand the link purpose.
       </p>
       
       <p>The word "mechanism" is used to allow authors to either make all links fully understandable
          out of context by default or to provide a way to make them this way. This is done
          because for some pages, making the links all unambiguous by themselves makes the pages
          easier for some users and harder for others. Providing the ability to make the links
-         unambiguous (by them selves) or not provides both users with disabilities with the
+         unambiguous (by themselves) or not provides both users with and without disabilities with the
```
mbgower, line 52: I'm glad you fixed this. The existing wording makes no sense.

understanding/20/link-purpose-link-only.html
patrickhlauke: I actually agree with the removal - if we've established that we're in exception territory here (with the first part of that paragraph, it seems weird to then step back and say any context (even if not sufficient to clarify the purpose of the link, since otherwise we wouldn't be in exception territory) must be added. a link is either exempt or it isn't.

understanding/20/link-purpose-link-only.html
```diff
@@ -13,59 +13,46 @@ <h2>In brief</h2>
...
          Having the link and the title agree, or be very similar, is good practice and provides
-         continuity between the link 'clicked on' and the web page that the user lands on.
-         
-      </p>
-      
+         continuity between the link 'clicked on' and the web page that the user lands on.</p>
       <p>The success criterion includes an exception for links for which the purpose of the
          link cannot be determined from the information on the web page. In this situation,
          the person with the disability is not at a disadvantage; there is no additional context
          available to understand the link purpose. However, whatever amount of context is available
```
patrickhlauke:
```suggestion
         available to understand the link purpose.</p>
```

understanding/20/link-purpose-link-only.html
```diff
@@ -13,59 +13,46 @@ <h2>In brief</h2>
...
          on the web page that can be used to interpret the purpose of the link must be made
```
patrickhlauke:
```suggestion
```

understanding/20/link-purpose-link-only.html
```diff
@@ -13,59 +13,46 @@ <h2>In brief</h2>
...
-         available in the link text to satisfy the success criterion.
-      </p>
-      
+         available in the link text to satisfy the success criterion.</p>
```
patrickhlauke:
```suggestion
```

patrickhlauke:
In resolving the merge conflicts, some aspects were lost.

> Added a note after h3 Sufficient Techniques to flag the fact that the value of two of Techniques listed is limited to screen reader users: ARIA8 (Using aria-label for link purpose) and C7 (Using CSS to hide a portion of the link text).

As techniques are now managed differently (not hardcoded, but generated by 11ty), this aspect needs to be addressed differently /cc @kfranqueiro

kfranqueiro:
I restored the note that was originally added. Note that it now appears at the end of Sufficient Techniques, consistent with how other notes within Sufficient Techniques already appeared.

patrickhlauke:
thanks @kfranqueiro

PR #1179, opened 2020-06-25
WCAG 2.2 Pointer Target Spacing

alastc:
This branch imports the Pointer Target Spacing success criteria from the [google doc](https://docs.google.com/document/d/1sszSUKB8t3VuRzxHtOjLfQZjNYCw-xr_EbuMwW7WiGc/edit#).

You can preview the:
- [Understanding doc (and SC text)](https://raw.githack.com/w3c/wcag/wcag22-pointer-target-spacing/understanding/22/pointer-target-spacing.html)
- [Technique](https://raw.githack.com/w3c/wcag/wcag22-pointer-target-spacing/techniques/css/target-padding.html)

The [last survey](https://www.w3.org/2002/09/wbs/35422/target-spacing/results), and [last minutes](https://www.w3.org/2020/04/14-ag-minutes.html#item05).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/pull/1179.html" title="Last updated on Jul 21, 2020, 11:30 PM UTC (4a7ed23)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/1179/b3b33e0...4a7ed23.html" title="Last updated on Jul 21, 2020, 11:30 PM UTC (4a7ed23)">Diff</a>


JAWS-test:
I have 3 questions, the answers to which, in my opinion, cannot be found in the Understandig at present and which could possibly be supplemented:
* Exception enlarge: Would this also include browser zoom, which according to SC 1.4.4 has to be supported up to 200% anyway (if so, would targets with 22x22 px be sufficient if zoom up to 200% works)?
* Is the SC only intended for mobile devices that are operated with fingers? Or also for desktop devices, where mouse or pen operation may also be difficult for people with tremor if the targets are too small?
* Does this also apply to standard elements that are rendered by the browser? E.g. for `option` elements within a `<select>` dropdown list?

alastc:
Hi @JAWS-test,

I think @detlevhfischer is tackling this, but in short:

- The enlargement is to make it 44 CSS px, so zoom does not count towards that.
- It is generally aimed at touch, although mouse users may benefit as well, if you click in the space between targets it doesn't automatically apply a heuristic to work out which link you meant to hit.
- I assume it would apply to browser-default components, but it's an interesting point.

patrickhlauke:
noting that from an actual practical/user point of view - assuming zoom is not suppressed - using browser zoom will make things physically larger for users making targets bigger - but that it's really more a last resort for users having to enlarge all content just to be able to comfortably operate a control (and otherwise, if browser zoom was an acceptable solution, it would essentially make target size and target spacing SCs irrelevant/always passable unless zoom was prevented or capped)

JAWS-test:
What happens if elements overlap each other, e.g. a pop-up, a submenu, a context menu, autocomplete list entries or an open selection list is located before links in the page. Do the pop-up elements, menu or list items then also need to have sufficient distance from the links? It should be noted that in many cases the links can also be activated if they are partially hidden by the above mentioned elements. Only rarely is it necessary to close the hiding element first in order to activate the links below it

> The requirement holds for all interactive elements that do not fall under the inline exception (including native elements like selects, buttons or inputs of type radio or checkbox).

I didn't mean the distance between select elements, but the distance between option elements within select. In Chrome these are 19 px high and have no spacing. So they would violate this. The same applies to all other browsers. Would the new SC really invalidate all select lists?

> The requirement holds independently of the zoom factor of the page. This means that authors cannot meet it by claiming that the target will have enough spacing or sufficient size if the user zooms into the page

But what if the web developer does not rely on the browser zoom, but implements his own zoom function to meet the exception "Enlarge: A mechanism is available to change the CSS pixel size of all targets so that the width and height are each at least 44 CSS pixels"? The effect would be the same as with the browser zoom, but it would satisfy SC, right?

> While the Success Criterion primarily helps touch users by ensuring that activation with the 'fat finger' will not accidentally trigger adjacent targets, it is also useful for mouse or pen users

That something is useful does not answer the question whether the new SC can only be violated on mobile devices, or even if I have a WCAG test where mobile devices are explicitly excluded from the test

understanding/22/pointer-target-spacing.html
```diff
@@ -0,0 +1,151 @@
...
+            <p class="note">This criteria has been formulated to increase the hit-area of small targets, but the group would like feedback from providers of touch-screen devices if there is another way of forming the criteria to better complement the tap-heuristics used.</p>
+            <p class="note">Are there issues with internationalization when describing inline links?</p>
+            <p class="note">Are there issues with pop-over content overlapping targets triggering failures?</p>
+        </blockquote>
+    </section>
+   
+    <section id="intent">
+        <h2>Intent of Pointer Target Spacing</h2>
+
+        <p>The intent of this success criteria is to ensure targets can easily be activated without accidentally activating an adjacent target. When touch targets are small, it is difficult for users with hand tremors and those who have difficulty with fine motor movement to activate them accurately. Providing spacing between touch targets will reduce the likelihood of accidentally activating the wrong control.</p>
```
patrickhlauke, line 40: "this success criteria" > "this success criterion" (singular)

change mention of "touch targets". this is about targets in general (touch, mouse, stylus, etc).

understanding/22/pointer-target-spacing.html
```diff
@@ -0,0 +1,151 @@
...
+       <li>Users with large fingers, or who are operating the device with only a part of their finger or knuckle.</li>
+     </ul>
+      
+   </section>
+   
+   <section id="examples">
+        <h2>Examples of Pointer Target Spacing</h2>
+
+        <ul>
+            <li>Three buttons are on-screen and the touch target area of each button is 44 by 44 CSS pixels.</li>
```
patrickhlauke, line 111: remove "touch"

understanding/22/pointer-target-spacing.html
```diff
@@ -0,0 +1,151 @@
...
+            <li>Links within a paragraph of text have varying touch target dimensions. Links within paragraphs of text do not need to meet the 44 by 44 CSS pixels requirements.</li>
```
patrickhlauke, line 112: remove "touch"

patrickhlauke:
the SC isn't meant to be explicitly about touch, but the current wording is still riddled with talk of touch. it applies, just like target size SC, to all possible pointer interactions.

understanding/22/pointer-target-spacing.html
```diff
@@ -0,0 +1,152 @@
...
+            <p class="conformance-level">AA</p>
+            <p class="change">New</p>
+            
+            <p>For each target, there is an area with a width and height of at least 44 CSS pixels that includes it, and no other targets, except when:</p>
+            <ul>
+                <li><strong>Enlarge:</strong> A mechanism is available to change the CSS pixel size of all targets so that the width and height are each at least 44 CSS pixels;</li>
+                <li><strong>Inline:</strong> The target is in a sentence or block of text;</li>
+                <li><strong>Essential:</strong> A particular presentation of the target is essential to the information being conveyed.</li>
+            </ul>
+            <p class="note">This criteria has been formulated to increase the hit-area of small targets, but the group would like feedback from providers of touch-screen devices if there is another way of forming the criteria to better complement the tap-heuristics used.</p>
```
patrickhlauke, line 31: criteria > criterion

understanding/22/pointer-target-spacing.html
```diff
@@ -0,0 +1,152 @@
...
+            <p class="note">Are there issues with internationalization when describing inline links?</p>
+            <p class="note">Are there issues with pop-over content overlapping targets triggering failures?</p>
+        </blockquote>
+    </section>
+   
+    <section id="intent">
+        <h2>Intent of Pointer Target Spacing</h2>
+
+        <p>The intent of this success criteria is to ensure targets can easily be activated without accidentally activating an adjacent target. When  targets are small, it is difficult for users with hand tremors and those who have difficulty with fine motor movement to activate them accurately. Providing spacing between targets will reduce the likelihood of accidentally activating the wrong control.</p>
```
patrickhlauke, line 40: criteria > criterion

patrickhlauke:
as pedantic as it is, there are still many instances of "criteria" (plural) where it should be "criterion" (singular). started commenting on them individually, but for brevity, just worth doing a quick once-over.

detlevhfischer:
@JAWS-test 

> What happens if elements overlap each other, e.g. a pop-up, a submenu, a context menu, autocomplete list entries or an open selection list is located before links in the page. Do the pop-up elements, menu or list items then also need to have sufficient distance from the links? 

I think the SC should certainly apply to targets in additional layers, including targets below that are possibly overlapped, but that that target spacing of elements on the page below should be assessed in the "default state", i.e., without any dynamic content opened on top of that default state. Not quite sure how to make this clear, and whether it is enough to add this to the understanding text and leave the SC text unchanged.

> I didn't mean the distance between select elements, but the distance between option elements within select. In Chrome these are 19 px high and have no spacing. So they would violate this. The same applies to all other browsers. Would the new SC really invalidate all select lists?

Good point. I think it makes sense to add an exception for targets **inside** unstyled native elements (e.g., lists of select options). I would assume once the author changes these, the target spacing requirement would apply. I have changed the understanding text to include that exception.

> But what if the web developer does not rely on the browser zoom, but implements his own zoom function to meet the exception "Enlarge: A mechanism is available to change the CSS pixel size of all targets so that the width and height are each at least 44 CSS pixels"? The effect would be the same as with the browser zoom, but it would satisfy SC, right?

Yes, IMO.

detlevhfischer:
I note that the exception 
> Enlarge: A mechanism is available to change the CSS pixel size of all targets so that the width and height are each at least 44 CSS pixels;

is really too specific: it could also be enough to change the spacing without increasing the target size, or change both, so we need something like 

**Enlarge: A mechanism is available to change the CSS pixel size of targets and/or spacing between targets so that the width and height of both target dimension and spacing combined are each at least 44 CSS pixels**;

guidelines/sc/22/pointer-target-spacing.html
```diff
@@ -0,0 +1,18 @@
...
+    <p class="conformance-level">AA</p>
+	<p class="change">New</p>
+    
+    <p>For each target, there is an area with a width and height of at least 44 CSS pixels that includes it, and no other targets, except when:</p>
+    <ul>
+        <li><strong>Enlarge:</strong> A mechanism is available to change the CSS pixel size of all targets so that the width and height are each at least 44 CSS pixels;</li>
+        <li><strong>Inline:</strong> The target is in a sentence or block of text;</li>
+        <li><strong>Essential:</strong> A particular presentation of the target is essential to the information being conveyed.</li>
+    </ul>
+    <p class="note">This criteria has been formulated to increase the hit-area of small targets, but the group would like feedback from providers of touch-screen devices if there is another way of forming the criteria to better complement the tap-heuristics used.</p>
```
patrickhlauke, line 14: remaining two instances of "criteria" > "criterion" ... sorry

detlevhfischer: @patrickhlauke only found one? In "Note: This criteria..."

Hope they are all weeded out now - cannot find another instance of 'criteria' when searching https://raw.githack.com/w3c/wcag/wcag22-pointer-target-spacing/understanding/22/pointer-target-spacing.html

alastc:
Thanks for all the updates Detlev.

I'm a little concerned by adding an exception to the understanding document, I think we need to make that explicit in the SC text if we do, e.g:

> User agent:The size of the element is controlled by the user agent and is not modified by the author.

Does that do the job? If so, we'll put back to the group.

detlevhfischer:
Hi Alastair,
Yes, i think that would do.
Best, Detlev

Sent from phone

alastc:
Just noting that this PR was [discussed today](https://www.w3.org/2020/07/14-ag-minutes.html#item02) and no issues were raised with the changes. This will go out for CFC shortly.

understanding/22/pointer-target-spacing.html
```diff
@@ -0,0 +1,153 @@
...
+        <p>As mobile devices generally have internal heuristics to identify which link is closest, spacing can work as effectively as a larger target size.</p>
+
+        <figure id="pointer-target-44x44">
+            <img src="img/pointer-target-example1.png" alt="A series of icons with measures showing they are each 44 by 44 pixels wide and high." />
+            <figcaption>The dimension of each icon target in a row is 44 by 44 CSS pixels, so the targets pass this Success Criterion without additional spacing.</figcaption>
+        </figure>
+
+        <figure id="pointer-target-38x38-spacing">
+            <img src="img/pointer-target-example2.png" alt="A row of icons with measures showing they are each 38 by 38 pixels wide and high, with 6 pixel spacing between them." />
+            <figcaption>The dimension of each icon target in a row is 38 by 38 CSS pixels, so a spacing of 6 CSS pixels is required to arrive at 44 CSS pixels for the combinded width and spacing of the target. No adjacent targets are present above and below, i.e., there is sufficient vertical spacing for the height of 38 CSS pixels to be sufficient.</figcaption>
```
patrickhlauke, line 68: typo: "combinded" > "combined"

understanding/22/pointer-target-spacing.html
```diff
@@ -0,0 +1,153 @@
...
+        </figure>
+      
+        <figure id="pointer-target-44-20-nospace">
+            <img src="img/pointer-target-example4.png" alt="A row of icons which are 44px wide and 20px high, with other targets above and below with only 2px spacing." />   
+            <figcaption><strong>Fail:</strong> Two rows of buttons, all with a height of only 20 CSS pixels. In the top row, buttons all have a width of more than 44 CSS pixels. For the top row, targets above and below are in close proximity. The top spacing of the first row is 4 px while the bottom spacing is 2 px, which adds up to 26 px for target height and vertical spacing combined. Therefore, the buttons in the top row fail the Success Criterion. The button "Pop" in the second row is only 32px wide with no horizontal spacing, so it also fails the Success Criterion even when there is sufficient space below to meet the vertical spacing requirement.</figcaption>
+        </figure>
+
+        <p>Where targets are embedded inside other targets, the inner target either needs a size of at least 44 x 44 CSS pixels or must placed at the edge or corner in a way that sufficient adjacent spacing is available in each dimension.</p>
```
patrickhlauke, line 81: "or must placed at the edge" > "or must be placed at the edge"

patrickhlauke: perhaps slightly awkward wording "in a way that sufficient adjacent spacing is available in each dimension".

maybe "to make sure that there is sufficient adjacent spacing in each dimension"

understanding/22/pointer-target-spacing.html
```diff
@@ -0,0 +1,153 @@
...
+
+        <figure id="pointer=target-embedded">
+            <img src="img/pointer-target-example5.png" alt="Three examples of a target adjacent to another target (see legend below)" />
+            <figcaption>Three examples of a person's profile image, which is supposed to be a target leading to another page, and an magnifying glass icon, which is supposed to open a lightbox view of the profile image. In the first example, the magnifying icon is outside the picture and passes. In the second example, the icon overlaps with the top-left corner of the image, but passes due to sufficient spacing top and left. In the third example, the icon sits inside the linked image. Since it is surrounded by an active click area, the icon target is not large enough to pass the success criterion.</figcaption>
```
patrickhlauke, line 85: "and an magnifying glass" > "and a magnifying glass"

understanding/22/pointer-target-spacing.html
```diff
@@ -0,0 +1,153 @@
...
+        </figure>
+      
+        <p>Users with different disabilities have different needs for control sizes. It can be beneficial to provide an option to increase the active target area without increasing the visible target size. Another option is to provide a mechanism to control the density of layout and thereby change target size or spacing, or both. This can be beneficial for a wide range of users. For example, users with visual field loss may prefer a more condensed layout with smaller sized controls while users with other forms of low vision may prefer large controls.</p>
+      
+      <p>Where user-agents control the size of a target, such as the options in a <code>select</code> drop-down, and the author has not tried to change the size of target with size or padding adjustments, the target is excempted from this criterion. It would be beneficial for users if the minumum size &amp; spacing requirement is meet, but where the user-agent has defined the target it is not the authors responsibility.</p>
```
patrickhlauke, line 90: typo: "excempted" > "exempted".

"not the authors responsibility" > "not the author's responsibility"

understanding/22/pointer-target-spacing.html
```diff
@@ -0,0 +1,153 @@
...
+        
+        
+    </section>
+    <section id="benefits">
+      <h2>Benefits of Pointer Target Spacing</h2>
+      
+      <ul>
```
patrickhlauke, line 97: could do with a single line of text before the list...as this currently doesn't list "Benefits", but "Who benefits".

something like "Having targets with sufficient target spacing can help all users who may have difficulty in confidently targeting/operating small controls. These include, but aren't limited to:"

techniques/css/target-padding.html
```diff
@@ -0,0 +1,113 @@
...
+		&lt;li&gt;&lt;span&gt;previous&lt;/span&gt;&lt;/li&gt;
+		&lt;li&gt;&lt;strong&gt;1&lt;/strong&gt;&lt;/li&gt;
+		&lt;li&gt;&lt;a href=&quot;#&quot;&gt;2&lt;/a&gt;&lt;/li&gt;
+		&lt;li&gt;&lt;a href=&quot;#&quot;&gt;3&lt;/a&gt;&lt;/li&gt;
+		&lt;li&gt;&lt;a href=&quot;#&quot;&gt;4&lt;/a&gt;&lt;/li&gt;
+		&lt;li&gt;&lt;a href=&quot;#&quot;&gt;5&lt;/a&gt;&lt;/li&gt;
+		&lt;li&gt;&lt;a href=&quot;#&quot;&gt;next&lt;/a&gt;&lt;/li&gt;
+	&lt;/ol&gt;
+&lt;/nav&gt;</code></pre>
+				<p>The following CSS would ensure the minumum spacing is applied:</p>
```
patrickhlauke, line 47: typo: "minumum" > "minimum"

understanding/22/pointer-target-spacing.html
patrickhlauke: typo: "minumum" > "minimum"

understanding/22/pointer-target-spacing.html
```diff
@@ -0,0 +1,153 @@
...
+        </figure>
+      
+        <p>Users with different disabilities have different needs for control sizes. It can be beneficial to provide an option to increase the active target area without increasing the visible target size. Another option is to provide a mechanism to control the density of layout and thereby change target size or spacing, or both. This can be beneficial for a wide range of users. For example, users with visual field loss may prefer a more condensed layout with smaller sized controls while users with other forms of low vision may prefer large controls.</p>
+      
+      <p>Where user-agents control the size of a target, such as the options in a <code>select</code> drop-down, and the author has not tried to change the size of target with size or padding adjustments, the target is exempted from this criterion. It would be beneficial for users if the minumum size &amp; spacing requirement is meet, but where the user-agent has defined the target it is not the authors responsibility.</p>
```
patrickhlauke, line 90: sorry, missed one instance of "minumum" > "minimum" here

PR #1204, opened 2020-07-09
Adding fstorr's updates - border note in focus-vis

alastc:
I needed to set this up to merge with the feature branch (rather than master) and my git-fu was not enough, so re-creating.


alastc:
Approved in the meeting: https://www.w3.org/2020/07/21-ag-minutes.html#item04

PR #1211, opened 2020-07-14
Clarify title of Failure F108

detlevhfischer:
The title "Failure of Success Criterion 2.5.X Dragging due to not providing a single pointer method for the user to operate a function that does not require a dragging movement" is ambiguous since the last part, "that does not require a dragging movement" invites the interpretation that the single pointer operation alternative has to be provided for a function that does not use dragging - which is clearly nonsense.

Simplest solution: Delete "that does not require a dragging movement".


gundulaniemann:
What about "Failure of Success Criterion 2.5.X Dragging due to not providing an alternative  single pointer method for the user to operate a function" ?

kfranqueiro:
We should also update the same wording in the `h1` element in the same file

bruce-usab:
Discussed on call `11/25, worth incorporating.  Question about Title versus H1 in build process.

kfranqueiro:
> Discussed on call `11/25, worth incorporating. Question about Title versus H1 in build process.

Confirmed that the build process does synchronize the h1 with the title, so updating both consistently is mostly to avoid confusion and not critical to the build output.

kfranqueiro:
I have rebased the branch to synchronize it with `main`, and have also changed the `h1`. Note that subsequent to this PR, the title of this technique had been updated from "2.5.X Dragging" to "2.5.7 Dragging Movements"; that change is retained.

Gundula's suggestion of possibly adding "alternative" before "single pointer method" is still open.

patrickhlauke:
not sure if "alternative" is necessary - if you're not providing a single pointer method, you're not providing it full stop (not as a primary, not as an alternative)

techniques/failures/F108.html
```diff
@@ -1,15 +1,15 @@
 <!DOCTYPE html>
 <html xmlns="http://www.w3.org/1999/xhtml" lang="en">
 <head>
-      <title>Failure of Success Criterion 2.5.7 due to not providing a single pointer method for the user to operate a function that does not require a dragging movement</title>
+      <title>Failure of Success Criterion 2.5.7 Dragging Movements due to not providing a single pointer method for the user to operate a function</title>
```
patrickhlauke, line 4: I do think the "does not require dragging" is important ... as otherwise this loses its meaning - if the single pointer method currently there DOES rely on dragging, then that's the failure. it's not whether or not it's single pointer that's the issue.

maybe better ...
```suggestion
      <title>Failure of Success Criterion 2.5.7 Dragging Movements due to not providing a single pointer method that does not require a dragging movement</title>
```

techniques/failures/F108.html
```diff
@@ -1,15 +1,15 @@
...
 
       <link rel="stylesheet" type="text/css" href="../../css/editors.css"/>
 
 </head>
 <body>
 
 
-    <h1>Failure of Success Criterion 2.5.7 Dragging Movements due to not providing a single pointer method for the user to operate a function that does not require a dragging movement</h1>
+    <h1>Failure of Success Criterion 2.5.7 Dragging Movements due to not providing a single pointer method for the user to operate a function</h1>
```
mbgower, line 12:
```suggestion
    <h1>Failure of Success Criterion 2.5.7 Dragging Movements due to not providing a single pointer method that does not require a dragging movement</h1>
```

patrickhlauke (approved):

patrickhlauke:
Discussed my counter-proposed rewording in backlog call 21/02 ... group agreed that instead of dropping the "does not require dragging" (which then loses the meaning), it's better to drop the "user to operate a function" part...

Issue #1195, opened 2020-07-07
Intermediate points in definition of dragging (for SC2.5.7)


guyhickling:
In the definition of "dragging movement" in the WCAG 2.2 Glossary what does "without requiring movement through an intermediate point" mean? It is not explained anywhere, and does not appear to have any relevance to the matter of dragging (that I can see). Is this a carry-over from some past discussion or early thoughts in creating this SC?

Also, why should drags through intermediate points be excluded from this SC? I can think of at least one scenario where going through an intermediate point would be a perfectly valid thing to do that should not be excluded. Take a children's game, where they have a row of objects along the top of the page, and they have to drag down through the correct one of a choice of gates along the middle of the page, and leave them in piles at the foot of the page. The gates are  intermediate points, but this drag operation should still be possible using single pointers (e.g. providing up/down/left/right buttons to click or tap on) same as any other drag operation.

So I suggest deleting the phrase "without requiring movement through an intermediate point" from the definition of dragging, or else explain what it means.

detlevhfischer:
@guyhickling You are right that the phrase "without requiring movement through an intermediate point" may be hard to understand without reference to 2.5.1 Pointer Gestures, where the requirement of the movement to go through an intermediate point was a technical expression of the directionality of the gesture - see [Understanding Pointer Gestures](https://w3c.github.io/wcag/understanding/pointer-gestures.html). The phrase was chosen to separate swiping gestures from dragging, which could go in any direction (the element under the pointer just 'latching on' and following the pointer movement). 

> Also, why should drags through intermediate points be excluded from this SC? 

They are not excluded - the phrase just says that in contrast to single pointer gestures, dragging is a movement that does not require (but also, does not exclude) going through an intermediate point.  Maybe this should be made clearer in the definition,

> So I suggest deleting the phrase "without requiring movement through an intermediate point" from the definition of dragging, or else explain what it means.

I think the phrase is useful to separate Dragging from Pointer Gestures (as understood in 2.5.1). As to providing an explanation, the Understanding texts are typically the place where such explanations should be given. I'm not sure that a longish explanation in the glossary would be appropriate. Maybe a cross reference to "single pointer "would help.

guyhickling:
Thank you. If I am understanding you correctly, you are saying that the words "without requiring movement through an intermediate point", in that "dragging movement" glossary definition, are referring to path-based gestures on touch screens - is that correct?

[...]

PR #1214, opened 2020-07-15
Dragging movement glossary definition: Deleting reference to intermediate point

detlevhfischer:
Deleted the qualification "without requiring movement through an intermediate point". This was initially included in the definition to separate 'dragging movement' more clearly from single pointer gestures. Since the 'single pointer' definition has no reference to the intermediate point requirement (see Understanding text for 2.5.1) , including it in the definition of dragging movement is likely to cause confusion.

Closes issue #1195

### DIFF
--- a/guidelines/terms/22/dragging-movement.html
+++ b/guidelines/terms/22/dragging-movement.html
@@ -2,6 +2,6 @@
 <dd class="new">
 	<p class="change">New</p>
    					
-   <p>an operation where the pointer engages with an element on the down event and the element (or a representation of its position) follows the pointer until the up event</p>
+   <p>an operation where the pointer engages with an element on the down event and the element (or a representation of its position) follows the pointer</p>
    <p class="note">The element could be, for example, a list item, a text element, or an image.</p>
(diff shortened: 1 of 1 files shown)


guyhickling:
The phrase "without requiring movement through an intermediate point" from the dragging definition in the Understanding doc for 2.5.7 (in the 28 July draft). However it now differs from the definition of dragging movement in the glossary, and the SC itself references the glossary definition by linking to it from the words "dragging movement", which leaves the SC still confusing. 

So is that glossary definition going to be changed as well, or is the link going to be removed from the SC, or some other solution (such as adding "without a path-based gesture" to the SC like 2.5.1?)

alastc:
Hi Guy,

It was updated in the correct place, so the to-be-published draft is correct:
https://w3c.github.io/wcag/guidelines/22/#dfn-dragging-movements
https://w3c.github.io/wcag/understanding/dragging.html#key-terms

guyhickling:
That's ok then, I just wanted to be sure it hadn't been overlooked!

PR #1215, opened 2020-07-15
Reflow for table cells - Addressing 1129

### LINKED ISSUE #1129: 1.4.10 Reflow and having more nuance for tables
Prompted by @WayneEDick's comment on the wai-ig mailing list https://lists.w3.org/Archives/Public/w3c-wai-ig/2020AprJun/0112.html in a discussion about reflow and tables

> The big problem with scroll bars is they are not useful for large print. If
the data in the cell is text the scroll bar forces horizontal scrolling.
Evan worse scroll bars often blow off the screen in large print (or low
resolution). Scrolling to get from column to column is ok in tables, but
scrolling within columns is not.

and my comment on that https://lists.w3.org/Archives/Public/w3c-wai-ig/2020AprJun/0113.html

> noting that the normative wording of the SC (and even the 
informative understanding document) doesn't make that particular 
distinction in its exception. It exempts tables pretty much wholesale - 
when it should probably have been a lot more nuanced.

I'm wondering if it may be possible to patch/address this in a bit more detail at least in the understanding document - clarifying that the exemption for tables is for keeping/allowing the two-dimensional relationship to remain intact, but that individual cells should not, by virtue of their content itself, require two-dimensional scrolling...i.e. you should be able to read/consume the content of an individual cell without needing to scroll in two dimensions.

related, this older issue about carousels (and similar widgets) https://github.com/w3c/wcag/issues/668 could possibly also be clarified along similar lines?

alastc:
Removing data cells from the the reflow exclusion.

Closes #1129


alastc:
Thanks, added.

Issue #504, opened 2018-10-06
SC 1.3.5 Technique "Use type attribute on fields in simple login forms"?


detlevhfischer:
In rating some forms against 1.3.5 Identify Input Purpose, I wondered whether the frequent two-field login forms that ask for just email and password would pass this SC already by specifying the type of the field (type=email, type=password). I reckon assistive technology presenting visual cues would not miss the opportunity to display the appropriate  image (not sure what these would be, though) to identify the purpose not just on forms that use the autocomplete attribute but also for (those parts of) forms that betray purpose in a programmatically determinable way via the type attribute.
So is there a case for a sufficient technique  "Use type attribute on fields in simple login forms", making it clear that it is scoped to just forms which contain no additional fields without unambiguous type?

mraccess77:
We had this discussion on WebAIM recently.  The consensus from some seems to be that type can't be used.

https://webaim.org/discussion/mail_thread?thread=8936

detlevhfischer:
Yes, I saw that discussion (and the epic exchange between JF and Jared) but I still feel that in a narrowly constrained context such as login, where it is clear that just you email and your password are needed, type should suffice to meet the SC. Why should cogs AT not be smart enough to look out for type as well, in the absence of autocomplete and supply fitting icons when there is an unambiguous match? Once you also have other fields like name that would be of type text, this obviously stops working. But I am talking here about a very common, restricted and conventional scenario.

patrickhlauke:
i would say `type="email"` and `type="password"`, in isolation and without any other fields of the same type, should likely be treated as acceptable for a login (compared to, say, two fields just with `type="text"`. i'd say it's pretty unambiguous what purpose those two inputs would have.

though i could see a more pedantic reading that would say the only acceptable way here would be:
```
<input type="email" ...>
<input type="password" autocomplete="current-password">
```

detlevhfischer:
@patrickhlauke ....which raises the interesting question what you’d have to do in the frequent case that an input field (just like the one I just used to login to Github) accepts BOTH user name AND email. Is it valid for autocomplete to have two values? I guess not (but too lazy to check). So you would, when using autocomplete, misrepresent the purpose either way, simply by omission. What would be the right thing to do on those form fields, and in turn, the right approach in conformance testing?

patrickhlauke:
[...]

PR #1217, opened 2020-07-17
1.3.4 - added exception for two-value fields

detlevhfischer:
This addition to the 1.3.4 Understanding text addresses issue #504 on how to assess input fields that allow two different types of input purpose, like combined user name/e-mail fields.
Closes https://github.com/w3c/wcag/issues/504


patrickhlauke:
wondering now, however...the limitation of only being able to give one specific input purpose value came up for `autocomplete` attribute in HTML. as supposedly this SC can also be satisfied in "other" ways (RDFa etc), is the same limitation present in those technologies too? or do they allow for greater flexibility (e.g. providing a comma or space separated list of potential values, or even a whole logical "this OR the other" syntax)?

Maybe this needs some more qualification, a la "if the technology used does not allow multiple purpose values to be defined (as is the case, for instance, with the `autocomplete` attribute in HTML, which only allows a single purpose to be defined)"

PR #1223, opened 2020-07-21
Duplicate of G18.html using correct maths

### LINKED ISSUE #1213: Create new technique for new contrast formula (threshold value)
One approach to dealing with the updated colour contrast value is to add a new technique using the new value, making it a sufficient one for the contrast criteria.
https://www.w3.org/2020/07/14-ag-minutes.html#item07

bruce-usab:
Closes #1213


bruce-usab:
This is a very light copy/change of [G18](https://www.w3.org/WAI/WCAG21/Techniques/general/G18).
- corrected number in formula at bottom of page (step 1 of procedure)
- deleted resources (since those all use deprecated math)

alastc:
I think this would be a valid technique for 2.1 as well, will propose that as part of the survey.

PR #1252, opened 2020-08-04
Issue 1187

### LINKED ISSUE #1187: Understanding 1.4.11 - clarify what does _not_ need such contrast
> I also see lots of false positives from testers for Non-text Contrast.
...
> I think the understanding document needs to define more of what does Not have contrast requirements, I see testers going over board on non-text contrast.

from https://twitter.com/pauljadam/status/1277979852205785088

Examples I've been asked about:

image

image

fstrr:
Closes #1187


fstrr:
@mbgower @gundulaniemann 

Here's the non-text-contrast edit we agreed in the call today. Note: there are numerous other changes in this PR.

understanding/21/non-text-contrast.html
```diff
@@ -82,7 +82,7 @@ <h4 id="related-scs">Relationship with Use of Color and Focus Visible</h4>
 				<figure id="figure-two-star-ratings-failing">
 					<img src="img/star-examples-fail.png" alt="Two star ratings, the first uses 5 stars with a black outline and a yellow or white fill, where yellow indicates checked. The second uses only pale yellow stars on white." width="300" />
 					<figcaption>
-						Two examples which fail a success criteria, the first fails the Use of color criteria due to relying on yellow and white hues. The second example fails the Non-text contrast criteria due to the yellow (#fff000) to white contrast ratio of 1.2:1.
+						Two examples which fail a success criteria, the first fails the Use of color criteria due to relying on yellow and white hues. The second example fails the Non-text contrast criteria due to the yellow (#fff000) to white background of 1.2:1.
```
patrickhlauke, line 85: the sentence now reads very awkwardly without the "contrast ratio" to me. should it perhaps be "the yellow (#fff000) to white background contrast ratio of 1.2:1"?

or turn it around and expand it as "fails the Non-text contrast criterion due to the low contrast ratio of 1.2:1 of the yellow (#fff000) against the white background."

patrickhlauke: additionally, while we're making edits...there seem to be - once again - instances in this document that use "success criteria" (plural) when they mean "success criterion" (singular).

taking the above second suggestion, and fixing up the start of the sentence as well, it should probably be more:

"Two examples which fail two different success criteria: the first fails the Use of color criterion due to relying on yellow and white hues; the second example fails the Non-text contrast criterion due to the low contrast ratio of 1.2:1 of the yellow (#fff000) against the white background."

fstrr: Hey

We agreed in today's AGWG call that there's a little more editing to be done on that "two examples" update, but I was asked to push the wording as is and @mbgower and @gundulaniemann were going to work on improving in.

Thanks for the feedback.

gundulaniemann: Hello, 
I understand one point of irritation was naming the hues without pointing out that the contrast does not contribute to the distinction.
As for my impression the most recent suggested version reads very fluently and clearly, I picked it up to suggest the following.

"Two examples which fail two different success criteria: the first fails the Use of color criterion due to relying on variation in color alone (yellow and white hues, without adequate contrast); the second example fails the Non-text contrast criterion due to the low contrast ratio of 1.2:1 of the yellow (#fff000) against the white background."

patrickhlauke:
@fstrr this is now 6 years old, and heavily conflicted. could I suggest, if this is still something we want to pursue, closing this and doing a fresh PR?

gundulaniemann:
> @fstrr this is now 6 years old, and heavily conflicted. could I suggest, if this is still something we want to pursue, closing this and doing a fresh PR?

I understand the paragraph on the two examples involving yellow is the only one discussed. the other changes seem to have agreement.

What is the benefit of creating a new PR?

patrickhlauke:
well either creating a new PR, or reworking this PR to resolve the merge conflicts that make this currently unmergeable

PR #1322, opened 2020-08-19
Not fully obscured

### LINKED ISSUE #1291: SC 2.4.11/2.4.12 Clarify third bullet point
> Success Criterion 2.4.12 Focus Appearance (Enhanced)
> …
> * **Unobscured:** No part of the focus indicator is hidden by author-created content.

I think this point in the SC is already addressed by 2.4.11 Focus Appearance (Minimum). I don't think it is possible to say that a user interface component meets a AAA criterium and therefor does not need to meet the AA criterium that also applies to it. AAA criteria are additive.

For example, I would assume that the bullet point **Adjacent Contrast** from 2.4.11 needs to be met in order to conform to 2.4.12. Or, putting it the other way around, if you failed a AA criterion, you cannot get AAA conformance, so claiming meeting the AAA criterion is not useful on its own.

alastc:
Closes #1291


alastc:
Agreed 25/08/2020

PR #1325, opened 2020-08-20
Updating the understanding - redundant entry

### LINKED ISSUE #1239: SC3.3.8: Redundant entry - When information previously entered has been saved
This is to recommend adding another use case to the scope of this SC. The Understanding limits coverage of this SC to a single session, so if the user aborts the process then returns later, they have to reenter all the data. Which is fine for that situation.

However there is another situation where, in a long and complicated data entry process, the website deliberately allows you to save your data part way, and resume entry at a later date. Banks often do this for more complicated applications. The income tax return in the UK also allows this.

In these cases the organisation has saved your data in their database, and will almost certainly have reloaded some or all of it when you return to finish the job, in order both to display parts of it (for instance user name and so on), and to find out where to resume input. So it would be quite reasonable to extend the coverage of this SC to such cases since the website already has that data available. 

This could be done very simply by adding a few words to cover this case in the relevant places in the Understanding. There are two places. The first is where it says:
> This Success Criterion pertains to the current user session, and is not applicable a user returns after closing a session or navigating away

and lower down it says:
> This Success Criterion does not add a requirement to remember information between sessions, a process is defined on the basis of an activity and does not continue across sessions

In both cases we could add the following words or similar:
> ....except where the user has been allowed to save their entered data part way through the process.

alastc:
Closes #1239

### DIFF
--- a/understanding/22/redundant-entry.html
+++ b/understanding/22/redundant-entry.html
@@ -25,16 +25,15 @@
     <section id="intent">
         <h2>Intent of Redundant Entry</h2>
 
-
         <p>The intent of this Success Criterion is to ensure that users can successfully navigate multi-step processes. It reduces cognitive effort where  information is asked for more than once during steps in a process. It also reduces the need to recall information provided in a previous  step.</p>
 
-        <p>This Success Criterion pertains to the current user session, and is not applicable  a user returns after closing a session or navigating away.</p>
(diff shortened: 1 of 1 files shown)


alastc:
Did not agree to this, 25/08/2020

alastc:
The group did not agree to add the requirements/changes, so closing.

PR #1330, opened 2020-08-20
Adding an 8px thick indicator example

### LINKED ISSUE #1279: Add example of 8px side border
From Nicaise:
> Regarding the “minimum area” requirement, One feedback I got is that folks wish to see in the understanding document, an example of how one meets this requirement with “a thickness of at least 8 CSS pixels along the shortest side of the element”. Is it possible to add an example specifically illustrating that in the understanding doc?
I checked the understanding doc for this criterion and there does not appear to be one. If there is, pls forgive me and point it out to me so I can pass on.

Good point, it was a late addition, I haven't had a chance to update the understanding doc so adding 'needs assigning'.

alastc:
Closes #1279


alastc:
Agreed 25/08/2020

PR #1331, opened 2020-08-20
Updating user-interface-component def

### LINKED ISSUE #1296: SC 2.4.11: Focus Appearance (Minimum) -  Possible grammatical error on Understanding page
On the Understanding page, in the notes under the definition of "[user interface component](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance-minimum.html#dfn-user-interface-component)", the start of the second sentence in the following note:

>  NOTE
Multiple user interface components may be implemented as a single programmatic element. Components here is not tied to programming techniques, but rather to what the user perceives as separate controls.

seems to be grammatically incorrect, as the word "Components" is plural, but the use of "is" implies that it is singular.

The note should probably read as either:

>  NOTE
Multiple user interface components may be implemented as a single programmatic element. Components here are not tied to programming techniques, but rather to what the user perceives as separate controls.

or

>  NOTE
Multiple user interface components may be implemented as a single programmatic element. The term "Components" here is not tied to programming techniques, but rather to what the user perceives as separate controls.

alastc:
Closes #1296.

### DIFF
--- a/guidelines/terms/20/user-interface-component.html
+++ b/guidelines/terms/20/user-interface-component.html
@@ -6,7 +6,7 @@
    </p>
    
    <p class="note">Multiple user interface components may be implemented as a single programmatic element.
-      Components here is not tied to programming techniques, but rather to what the user
+      The term "components" here is not tied to programming techniques, but rather to what the user
       perceives as separate controls.
    </p>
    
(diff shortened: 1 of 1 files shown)


alastc:
Agreed 25/08/2020. Slightly adjusted to match other uses of "compenent".

PR #1332, opened 2020-08-20
Adding an example of wrapping link

### LINKED ISSUE #1328: How should 2.4.11 / 2.4.12 be applied to inline elements that wrap over multiple lines?
Given an inline elements that wraps across multiple lines, such as a link, how should the minimum focus indication area be calculated?

Should each part of the control be treated separately, or should the line break be ignored – effectively 'glueing' the parts of the control together – and then calculated as a whole?

I believe a 1px border on a link that wraps across multiple lines could pass or fail depending on which interpretation you use:

Link with a red border that wraps over two lines. The borders on the right edge of the first part of the link and the left edge of the second part of the link are missing, because of the way the element is wrapped.

If each part of the control should be considered separately, do all parts of the control then need to meet the criteria, or at least one?

If nothing else, it would be useful to have an example of how to do the calculations in this situation.

alastc:
Closes #1328.

[Preview the additional example](https://raw.githack.com/w3c/wcag/Issue1328-focus-appearance/understanding/22/focus-appearance-minimum.html#focus-indicator-box-shadow-only).


understanding/22/focus-appearance-minimum.html
```diff
@@ -133,6 +133,11 @@ <h4>Unusual shapes and gradients</h4>
                     <figcaption>The focused button with the non-contrasting areas removed, showing that it is a thick indicator that meets the requirements.</figcaption>
                 </figure>
 
+                <p>If an inline-link is broken over two lines the indicator can be split between them. In this case the link element is still treated as one instance.</p>
```
36degrees, line 136: A couple of minor suggestions, if that's OK:

- inline link doesn't appear to be hyphenated anywhere else in this repository, so remove hyphenation
- explicitly allow for links that span more than two lines
- use 'control' rather than 'instance' as 'instance' is not used anywhere else, and 'control' is defined.

Are there any other inline elements that might take keyboard focus? Should 'inline link' be 'inline element'

```suggestion
                <p>If an inline link is broken over two or more lines the indicator can be split between them. In this case the link element is still treated as a single control.</p>
```

36degrees, line 136: A couple of minor suggestions, if that's OK:

- inline link doesn't appear to be hyphenated anywhere else in this repository, so remove hyphenation
- explicitly allow for links that span more than two lines
- use 'control' rather than 'instance' as 'instance' is not used anywhere else, and 'control' is defined.

Are there any other inline elements that might take keyboard focus? Should 'inline link' be 'inline element'?

```suggestion
                <p>If an inline link is broken over two or more lines the indicator can be split between them. In this case the link element is still treated as a single control.</p>
```

understanding/22/focus-appearance-minimum.html
```diff
@@ -133,6 +133,11 @@ <h4>Unusual shapes and gradients</h4>
...
+                <figure id="focus-indicator-inline-link">
+                    <img src="img/focus-inline-link.png" alt="A paragraph of text with a text link that wraps around. The red focus outline has a gap on the right of one part of the link, and on the left of the other part of the link." />
```
36degrees, line 138: For consistency with the preceding paragraph:

```suggestion
                    <img src="img/focus-inline-link.png" alt="A paragraph of text containing a link broken over two lines. The red focus outline has a gap on the right of one part of the link, and on the left of the other part of the link." />
```

alastc:
Agreed 25/08/2020

PR #1334, opened 2020-08-21
2.5.7 Dragging: adding reference to 2.5.1 Pointer Gestures and v.v.

### LINKED ISSUE #1316: Suggestion - the understanding for 2.5.7 Dragging and 2.5.1: Pointer Gestures should cross reference
When reading 2.5.7: Dragging for the first time, my first thought was that it should be clearly differentiated from 2.5.1: Pointer Gestures. In turn, in 2.5.1: Pointer Gestures it says...

> Gestures that involve dragging in any direction are not in scope for this SC because only the start and end points matter in a dragging operation. 

I think that this should probably signpost 2.5.7 Dragging. In turn, that SC should probably signpost 2.5.1: Pointer Gestures and explaining that it is not concerned with multi-point or path gestures.

Potentially something along the lines of Non Text Content has with "Relationship with Use of Color and Focus Visible".

detlevhfischer:
Added a paragraph to Understanding of 2.5.7 Dragging to differentiate the new SC from 2.5.1 Pointer Gestures.
Added a paragraph to Understanding of 2.5.1 Pointer Gestures to differentiate the SC from the new SC 2.5.7 Dragging.

Closes #1316


michael-n-cooper:
Yes, we can do an update that covers both a 2.1 and 2.2 version, or just a 2.2 version. I would question doing just a 2.1 version without 2.2 though.

As far as how to manage these versions, I haven't gotten that fully figured out yet. But for now, just doing them in the 2.2 (master) branch is good.

PR #1371, opened 2020-08-30
Change "criteria" to "criterion"

patrickhlauke:
Amends most (all?) instances where "criteria" (plural) is incorrectly used to mean "criterion" (singular)


alastc:
@michael-n-cooper this is probably best to merge sooner than later then, so there are minimal clashes?

michael-n-cooper:
Yup. Since it's editorial it should be ok. Michael

alastc:
Merged, this will be one to pull into WCAG 2.1 as well though.

PR #1379, opened 2020-09-02
Update pointer-target-spacing.html

### LINKED ISSUE #1362: Pointer Target Spacing (2.5.8) understanding note gives the wrong rationale?
The first note in the understanding document https://www.w3.org/WAI/WCAG22/Understanding/pointer-target-spacing.html states

> This criterion has been formulated to increase the hit-area of small targets,...

but this is not strictly true. The whole rationale for the SC is not to "increase the hit-area" (and indeed, see https://github.com/w3c/wcag/issues/1361 for an example where to pass 2.5.8 authors may actually *decrease* the hit area), but rather

> This criterion has been formulated to decrease the likelihood of users accidentally activating the wrong target...

or similar (as is indeed stated in the actual intent later on)

> The intent of this Success Criterion is to ensure targets can easily be activated without accidentally activating an adjacent target.

detlevhfischer:
Replaced the misleading bit "increase the hit-area of small targets" in the first note to emphasise separation of targets to minimise accidental activation.
Closes #1362

### DIFF
--- a/understanding/22/pointer-target-spacing.html
+++ b/understanding/22/pointer-target-spacing.html
@@ -29,7 +29,7 @@
                 <li><strong>User agent:</strong> The size of the target is controlled by the user agent and is not modified by the author;</li>
                 <li><strong>Essential:</strong> A particular presentation of the target is essential to the information being conveyed.</li>
             </ul>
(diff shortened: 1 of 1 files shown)


alastc:
Accepted 8th Aug 2020

Issue #1156, opened 2020-06-09
Loophole in SC2.4.11 allows almost invisible focus indicators


guyhickling:
The new SC2.4.11 Focus Visible (Enhanced) has a major loophole in the wording. The definition of Adjacent Contrast uses "or" between the two requirements for 3 to 1 contrast and 2px thickness. This allows a 2px outline but without any requirement at all for colour contrast. So a developer can add an outline that is much less than 3 to 1 contrast against the adjacent background. In fact it need be only 0.1 to 1 contrast, which would be completely invisible to every one. 

One way this will work in actual practice, is that developers would not have to bother changing the focus indicator's colour as it moves around the page from one background to another. You can see an example of how this works by just looking at the default focus outline in the Google Chrome browser. It is blue, and it remains blue even when the item in focus is on any close shade of blue background. So invisible to all ordinary people! 

(In fact the 2px thickness is already required by the Minimum Area clause, so I don't think it needs to be repeated here, does it, unless I'm misunderstanding the meanings of these different clauses?) 

Change the "or" to "and" in the body of the clause to prevent this loophole, and change the name of the clause to "Contrast and thickness" to match.

patrickhlauke:
i thought the idea was that the contrast requirement would come nonetheless from 1.4.11 and the idea was to avoid overlap in this part?

mraccess77:
I believe a contrast ratio is excluded from that bullet on purpose because it's 2px so it's wide enough that even with less contrast it will be noticeable by size - at least that is how I understood it from Alastair.

alastc:
Hi @guyhickling,

I think the foundation of this question is not seeing the difference between a 'change' of contrast and 'adjacent' contrast. Without that confusion I can't see a loophole.

The fact you raised this indicates there might be something to add to the Understanding doc about that difference, but could you confirm that was the crux of the issue first?

Cheers,

-Alatair

guyhickling:
Alastair, No, the foundation and reason for raising this issue is that SC2.4.11, as currently worded, does not say what you intend it to mean. And when the WCAG doesn't say what it really means (of which there are already too many examples as it is!), that's when developers push back against audits.  They refuse to fix things the tester has found, and claim that false issues are being raised,  because they can use the wording of a Success Criterion to claim what they have done passes. 

[...]

PR #1388, opened 2020-09-08

mbgower:
This is a proposed wording change to address the outstanding matter of a 1.4.11 referral in https://github.com/w3c/wcag/issues/1156#issuecomment-669309673

### DIFF
--- a/understanding/22/focus-appearance-minimum.html
+++ b/understanding/22/focus-appearance-minimum.html
@@ -175,14 +175,14 @@
         <section id="Contrast-thickness">
             <h3>Adjacent contrast</h3>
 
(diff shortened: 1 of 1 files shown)


alastc:
> This requirement complements 1.4.11 Non-text Contrast, where a component in a focused state must achieve a 3:1 contrast against the adjacent background.

This isn't always true though, links & buttons are not required to have borders/backgrounds that contrast with the adjacent colours, and it's focus state might contrast with *either* the background or the element (e.g. with an internal focus indicator).

mbgower:
Does this address your concerns?
> This requirement complements 1.4.11 Non-text Contrast, where a component with an author-indicated focus state must achieve a 3:1 contrast against an adjacent color.

mbgower:
This has been addressed in subsequent changes to the normative text.

Issue #1372, opened 2020-08-31
Comment from IBM: Reword Focus Appearance (Minimum) to provide clarity


mbgower:
The current preamble of [2.4.11](https://www.w3.org/TR/WCAG22/#focus-appearance-minimum) reads as follows:
> For the keyboard focus indicator of each User Interface Component, all of the following are true:

This wording does not align with the 'consequence' of taking focus, which is particularly inherent in the fourth bullet, which reads:
> Unobscured: The item with focus is not entirely hidden by author-created content.

It is possible to misinterpret this bullet that the item with focus must _always_ be visible, even if the user, for instance, scrolls down a page after bringing the component into focus (thus preventing the user from scrolling).

All four bullets are considerations for when an item receives keyboard focus, so we believe the following provides clarity, and makes the fourth bullet more easily understood in conjunction with the other bullets:

>**When user interface components receive keyboard focus, all of the following are true:**

Note that we have also put "user interface components" in lower case, to match both the definition and other occurrences in WCAG 2.1. Potentially this could also be a link to the definition.

awkawk:
@mbgower Have you seen my suggestions in https://github.com/w3c/wcag/issues/1341? Seems to be some overlap

mbgower:
There is definitely some overlap, and some other things you tackle that we have not seem reasonable. However, I also agree with Alastair's comment that changing the 4th bullet to being about the indicator instead of the item with focus is problematic.

guyhickling:
IBM's and @mbgower's wording is much clearer and much easier for people to read and understand. Always go for clarity, assuming the meaning is unchanged! Even if it takes a few words more. The most usual excuse people offer for making the WCAG so abstruse is to have fewer words. But here the clearer version is also 4 words less........!

RavenAlly:
Could the "receive keyboard focus" wording be a bit too absolute in terms of timing?

What if my focus style starts out conformant "on receiving focus" (pass!) but transitions to a non-conformant colour / size? Still got focus but no longer meets the spec.

guyhickling:
@RavenAlly Good point. Perhaps "have keyboard focus" would be better than "receive...."

mbgower:
>Could the "receive keyboard focus" wording be a bit too absolute in terms of timing?
What if my focus style starts out conformant "on receiving focus" (pass!) but transitions to a non-conformant colour / size? Still got focus but no longer meets the spec.

[...]

Issue #1324, opened 2020-08-20
Minimum focus indicator definitions for 2.4.11/2.4.12 are ambiguous; example calculations do not account for corners


36degrees:
Success Criterion 2.4.11 says:

> The focus indication area is greater than or equal to a 1 CSS pixel border of the focused control

It's unclear whether this refers to:

- a border **along the existing edge of** the control, ‘overlapping’ the outermost pixels of the control or;
- a border **around the outside of** the control.

(Effectively, it depends whether you’re thinking about borders using the ‘border-box’ or ‘content-box’ box-sizing model)

In Understanding Success Criterion 2.4.11: Focus Appearance (Minimum), the guidance then says:

> The minimum size of the focus indicator must be at least the area taken by a 1 CSS pixel border around the element.

This seems to suggest that the intent is for the minimum focus indication area to be equivalent to a border around the outside of the control, but it's still not 100% clear.

The example calculations do not help either, because they fail to take corners into account and would be incorrect for either case.

> For example, if a control is a rectangle of 90px wide and 30px tall, the size of the outer border is 90 + 90 + 30 + 30 = 240 CSS pixels.

Every ‘corner pixel’ is being counted twice.

The area of a 1px border would in fact be either +/- 4px depending on whether the requirement is for a border along the existing edge of, or around the outside of the control.

For example, applying the same calculation for a 10px square you’d end up with 10 + 10 + 10 + 10 = 40 CSS pixels, but it’d actually be either 36 or 44px:

![A 1px border along the existing edge of a 10px square has an area of 36px (4 × 8px edges + 4 × 1px corners). A 1px border around the outside of a 10px square has an area of 44px (4 × 10px edges and 4 × 1px corners)](https://user-images.githubusercontent.com/121939/90748324-f1210b80-e2c9-11ea-80d3-00acef129bd8.png)

Similarly, to meet the 2px requirement for 2.4.12, a border around a 10px square would not be 20 + 20 + 20 + 20 = 80 but instead 80px +/- 16px:

![A 2px border along the existing edge of a 10px square has an area of 64px (4 × 12px edges + 4 × 4px corners). A 2px border border around the outside of a 10px square has an area of 96px (4 × 20px edges and 4 × 4px corners)](https://user-images.githubusercontent.com/121939/90748374-05650880-e2ca-11ea-9653-815f213bdc39.png)

[...]

PR #1403, opened 2020-09-11

### LINKED ISSUE #1341: 2.4.11 Focus Appearance (Minimum) (AA) - clarification needed
Regarding the new sc 2.4.11 Focus Appearance (Minimum) (AA), I'd like clarification around the change of contrast requirement: "Change of contrast: The color change for the focus indication area has a contrast ratio of at least 3:1 with the colors of the unfocused state." 

If a focus indicator is added that 1.) does not rely on color and 2.) meets minimum adjacent contrast requirements, is it really necessary for the indicator to also have a 3:1 contrast with the unfocused state? 

If so, this contradicts past precedent of allowing for indicators besides color. It will also severely restrict the colors designers are able to use for default and focus states.

If not, could some clarifying language be added to this sc? 

Also, browser defaults have historically be treated as the exception, as they are not dictated by content creators. Will that continue to be the case regarding focus indicators? If so, language around that exception should be added.

mbgower:
Rewording of intro and all bullets except the last one, in an attempt to address https://github.com/w3c/wcag/issues/1341

Closes #1341, #1372

### DIFF
--- a/guidelines/sc/22/focus-appearance-minimum.html
+++ b/guidelines/sc/22/focus-appearance-minimum.html
@@ -8,7 +8,7 @@
     <p>When user interface components receive keyboard focus, all of the following are true:</p>
     <ul>
         <li><strong>Minimum area:</strong> The area of the focus indicator is at least as large as as the area of a 1 <a>CSS pixel</a> thick perimeter of the focused control, or the focus indicator has a thickness of at least 8 CSS pixels along the shortest side of the user interface component.</li>
-        <li><strong>Change of contrast:</strong> The pixels making up the focus indicator achieve at least a 3:1 contrast between their colors in the focused and unfocused states. </li>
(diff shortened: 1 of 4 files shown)


awkawk:
I disagree that we should be including the formula for perimeter. This is well-understood outside of WCAG.

guidelines/sc/22/focus-appearance-minimum.html
```diff
@@ -5,11 +5,11 @@ <h4>Focus Appearance (Minimum)</h4>
     <p class="conformance-level">AA</p>
 	<p class="change">New</p>
     
-    <p>For the keyboard focus indicator of each User Interface Component, all of the following are true:</p>
+    <p>When user interface components receive keyboard focus, all of the following are true:</p>
```
awkawk, line 8: Will need to be User Interface Component

alastc: Someone tackled this in #1469, most are lower case so that PR was to make all lowercase.

alastc:
Noting for the record that we discussed the perimeter aspect during the call, and there is an advantage to defining it normative, see #1324

Issue #1344, opened 2020-08-25
Cognitive function test exception is arbitrary


WilcoFiers:
The definition of Cognative function test lists remembering name, e-mail and phone number as exceptions. This seems arbitrary, based on current (arguably even outdated) methods of communication. A lot of people use their social media accounts much more than they do e-mail. Why would remembering your Twitter handle be any more difficult then remembering your e-mail address?

My biggest concern with it isn't the arbitrary nature of it, but the problem that phone numbers and e-mail addresses aren't universal. Not everyone has an e-mail address, and a lot of people do not know their own phone number.

I think a better solution for this would be to make this abstract, something like "Except for common personal identifiers such as name, e-mail, phone number or handle on social media." That gets to the issue that someone might actually need their handle to sign in, but that shouldn't be hard to find a solution to. The platform needs to support two such identifiers, (i.e. handle and e-mail, or handle and phone number).

alastc:
As noted in a previous issue, I think the logical line we are trying to draw is between:

- Things which are about you, and do not vary by site. E.g. Email address, Name, postal address.
- Things which vary by site (or at least should). E.g. username, password, ID for that organisation.

I think that can be clarified and updated in the definition, your suggestion works for me.

alastc:
I've updated [#1411](https://github.com/w3c/wcag/pull/1411/files) with this wording.

alastc:
Closing based on the changes here:
https://github.com/w3c/wcag/pull/1411/files

Minutes: https://www.w3.org/2020/09/22-ag-minutes.html#item03

Issue #1320, opened 2020-08-19
3.3.7 Accessible Authentication queries on personal info


alastc:
From Aaron Armstrong:
<hr>
This feedback is re: 3.3.7 Accessible Authentication, which states "If an authentication process relies on a cognitive function test, at least one other method must also be available that does not rely on a cognitive function test." Cognitive function tests include "memorization, such as remembering a username, password..."

https://www.w3.org/TR/2020/WD-WCAG22-20200811/#dfn-cognitive-function-test

However, the Sufficient Techniques [permit](https://www.w3.org/WAI/WCAG21/Understanding/accessible-authentication) "providing a properly marked up email and password inputs."

A couple issues I perceived upon first reading:

1.	One criteria mentions "username" and the other mentions "email." Unless the intent is to specifically recommend the use of email addresses and not unique usernames for login purposes, this wording should perhaps be standardized.
2.	More importantly, the criteria seem to list user/pass fields as both sufficient login technique and also a cognitive function test. I believe the intent here is to emphasize "properly marked up," which would allow the use of tools like saving login credentials in the browser or using external password managers and bypass the user being required to remember them. If this is the case, I would suggest expanding upon this in the Sufficient Techniques section and making these expectations clear. 

I hope this is helpful. Thank you for the work you're doing!

Aaron

alastc:
There is a pull request to update the definition to make the difference clearer between personal info and site specific info in [#1412](https://github.com/w3c/wcag/pull/1412/files).

I'll try and put something together for the email/username differences.

alastc:
For the email/username aspect, we aren't trying to say anything specific about those, and the highlighted note in the understanding starts with:
> Web sites can employ username (or email) and password inputs as an authentication method...

The only difference between a username and email address (used as username) is that remembering your own email is not considered a cognitive function test. However, there is inevitably a password or other means of authenticating, so it doesn't make much difference in general.

I'm not seeing anything that needs updating, unless you can expand on what the misunderstanding is?

wardav:
To you think remembering and inputting things such as:

> Things which are about you, and do not vary by site. E.g. Email address, Name, postal address.

Should be in the normative scope. Just looking at that tweak #1412, I don't know that there is some clear language explaining that they are in scope / not in scope. 

For example

[...]

PR #1411, opened 2020-09-15
Moving note into the definition

### LINKED ISSUE #1310: 3.3.7: Clarify exceptions in definition
The definition of Cognitive function test is:
> task that requires the user to remember, manipulate, or transcribe information.
> Examples include, but are not limited to:
> - memorization, such as remembering a username, password, set of characters, images, or patterns;
> - transcription, such as typing in characters;
> - use of correct spelling;
> - performance of calculations;
> - solving of puzzles.
> Note: Remembering your own name, email address, or phone number is not considered a cognitive function test.

Comment from Gregg Vanderheiden about the name & email address aspect:
> If those are exceptions - they need to be added to the definition proper of cognitive test.

The rational was that for _your_ information, you only have to remember one name/email/phone number. Things like usernames and passwords are (or at least should be!) per-site, so you would have to remember many such things.

I can see two solutions to this, either make the 1st part more specific, e.g:
> task that requires the user to remember, manipulate, or transcribe information for the website.

Or add it to the 1st bullet, e.g:
> > - memorization, such as remembering a username, password, set of characters, images, or patterns. Information not specific to the website (such as your name, email and phone number) are not considered memorization.

alastc:
Closes #1310, #1320, #1344

### DIFF
--- a/guidelines/terms/22/cognitive-function-test.html
+++ b/guidelines/terms/22/cognitive-function-test.html
@@ -5,7 +5,7 @@
    					
    <p>A task that requires the user to remember, manipulate, or transcribe information. Examples include, but are not limited to:</p>
    <ul>
-       <li>memorization, such as remembering a username, password, set of characters, images, or patterns. Common personal identifiers such as name, e-mail, phone number or handle on social media are not considered memorization as they are consistent across websites;</li>
+       <li>memorization, such as remembering a username, password, set of characters, images, or patterns. Common personal identifiers such as name, e-mail, phone number, and handle on social media are not considered memorization as they are consistent across websites;</li>
        <li>transcription, such as typing in characters;</li>
        <li>use of correct spelling;</li>
        <li>performance of calculations;</li>
--- a/guidelines/terms/22/cognitive-function-test.html
+++ b/guidelines/terms/22/cognitive-function-test.html
@@ -5,12 +5,11 @@
    					
    <p>A task that requires the user to remember, manipulate, or transcribe information. Examples include, but are not limited to:</p>
    <ul>
-       <li>memorization, such as remembering a username, password, set of characters, images, or patterns;</li>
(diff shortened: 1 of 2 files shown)


guidelines/terms/22/cognitive-function-test.html
```diff
@@ -5,12 +5,11 @@
    					
    <p>A task that requires the user to remember, manipulate, or transcribe information. Examples include, but are not limited to:</p>
    <ul>
-       <li>memorization, such as remembering a username, password, set of characters, images, or patterns;</li>
+       <li>memorization, such as remembering a username, password, set of characters, images, or patterns. Common personal identifiers such as name, e-mail, phone number or handle on social media are not considered memorization as they are consistent across websites;</li>
```
awkawk, line 8: Should be "phone number, and handle on social media", making this an "and" list since all could apply and also adding the Oxford comma since we do everywhere else for clarity.

understanding/22/accessible-authentication.html
```diff
@@ -25,13 +25,12 @@ <h2>Cognitive function test definition</h2>
         <blockquote>
             <p>A task that requires the user to remember, manipulate, or transcribe information. Examples include, but are not limited to:</p>
             <ul>
-                <li>memorization, such as remembering a username, password, set of characters, images or patterns;</li>
+                <li>memorization, such as remembering a username, password, set of characters, images, or patterns. Information not specific to the website (such as your name, email and phone number) are not considered memorization;</li>
```
awkawk, line 28: email, and phone number

guidelines/terms/22/cognitive-function-test.html
alastc: Thanks, also not sure why the understanding doc didn't match, I'm sure I did both at the same time.

Issue #1292, opened 2020-08-12
Accessible Authentication: Does copy-paste or autofills pass?


wardav:
Reviewing 3.3.7: Accessible Authentication, TFA appears to be in scope. For example, "A web site that requires 2-factor authentication allows for multiple options for the 2nd factor, including a USB-based method where the user simply presses a button to enter a time-based token.".

However, with mobile websites, you can autofill OTPs that are sent via SMS. So there is no cognitive requirement. But this is specific to someone accessing the website on mobile. 

![example of autofilled OTP](https://user-images.githubusercontent.com/38493807/90010938-a1728c80-dc98-11ea-9bfa-e66d0af733e5.png)

That said, more broadly, on mobile the user may have an authentication app. In this case, they just need to go to the app, copy the code, and paste it into the website. 

On desktop, the user may just have to copy and paste code from an email. I don't know if this would be in scope. I would guess if the email is just a link ("click this link to activate your account"), it would not be in scope. So would it be in scope if they had to copy a code from the email and paste it into a field, to perform the same task.

alastc:
I think the crux here is whether copy-paste is a recognised method of bypassing the need to transcribe a set of numbers/characters.

In the example above I've not seen that short-cut myself (where the phone grabs the numbers out of a text message), I wonder how wide-spread & reliable that is?

mraccess77:
iOS provides me the option to paste codes from text messages on my iPhone.  It appears as an option above the keyboard on screen.

patrickhlauke:
this would be iOS specific, so not something that can be generally relied on. it's essentially an OS addition to the more general concept of copy/paste operations (iOS "preloads" your clipboard with / automatically copies what it guesses is a confirmation code in a text message, and lets you paste it easily, i a single operation). if the concept of copy/paste is disallowed (which I'd say would be far more restrictive), then this niceness in iOS makes no difference overall (as you couldn't say "offer these various authentication alternatives...oh except on iOS)

wardav:
It is supposedly supported on Android 9 as well, 

* [Google rolling out SMS verification code autofill to Android](https://www.androidpolice.com/2019/09/17/android-autofill-sms-verification-2fa/)
* [How to enable SMS verification code autofill in Android[(https://www.techrepublic.com/article/how-to-enable-sms-verification-code-autofill-in-android/).

I think that @alastc is correct, though. 

> I think the crux here is whether copy-paste is a recognised method of bypassing the need to transcribe a set of numbers/characters.

[...]

PR #1419, opened 2020-09-16
Adding copy-paste references

### LINKED ISSUE #1359: Accessible Authentication (3.3.7): clarify whether copy/paste is a cognitive function test
There are some platforms and environments where autofill is not possible, such as a web-based command line interface where authentication is text-only, with no GUI controls or other options.

The definition of a cognitive function test includes transcription, but it isn't entirely clear if the ability to copy and paste a string is a passing method. Allowing copy/paste would completely solve the command line problem, and other non-standard UI examples we've been able to think of (including desktop and mobile apps). It would also still prohibit the following less-than-ideal examples:
- A required text message authentication step that forces a user to transcribe numbers from their phone onto a computer
- A username/password form that disables pasting functionality
- Captchas
- Images of pin numbers or other 2FA strings

The current wording does not seem to explicitly prohibit copy/paste authentication flows, but it would be helpful to note that a single copy/paste step does not count as transcription in the definition of cognitive function test.

alastc:
Closes #1359, #1292.

### DIFF
--- a/guidelines/terms/22/cognitive-function-test.html
+++ b/guidelines/terms/22/cognitive-function-test.html
@@ -6,7 +6,7 @@
    <p>A task that requires the user to remember, manipulate, or transcribe information. Examples include, but are not limited to:</p>
    <ul>
        <li>memorization, such as remembering a username, password, set of characters, images, or patterns;</li>
-       <li>transcription, such as typing in characters. Copy and paste, such as that automated by a password manager, is not considered transcription but requiring multiple steps or applications to complete a paste is a cognitive functional test;</li>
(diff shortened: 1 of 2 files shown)


guidelines/terms/22/cognitive-function-test.html
```diff
@@ -6,11 +6,11 @@
    <p>A task that requires the user to remember, manipulate, or transcribe information. Examples include, but are not limited to:</p>
```
anubes19, line 6: ](https://github.com/mozilla-mobile/firefox-ios/issues/7330)

PR #1457, opened 2020-09-19
Text spacing understanding reword

mbgower:
I was working on something else, merely referencing this, but got pulled up short by the second paragraph of the Intent -- then got drawn into some other minor fixes. I don't believe anything I'm proposing in the first commit is adventurous.
The second commit, which strips out the second paragraph is more so, but I think is supportable.

### DIFF
--- a/understanding/21/text-spacing.html
+++ b/understanding/21/text-spacing.html
@@ -9,15 +9,20 @@
       <h1>Understanding Text Spacing</h1>
       <section id="intent">
          <h2>Intent</h2>
-         <p>The intent of this Success Criterion (SC) is to ensure that people can override author specified text spacing to improve their reading experience. Each of the requirements stipulated in the SC's four bullets helps ensure text styling can be adapted by the user to suit their needs.</p>
+         <p>The intent of this Success Criterion (SC) is to ensure that people can override author-specified text spacing to improve their reading experience. Each of the requirements stipulated in the SC's four bullets helps ensure text styling can be adapted by the user to suit their needs.</p>
          <p>The metrics set a target for a minimum increase in text spacing that must be met. Starting from the author's baseline implementation, any increase in spacing for any of these four style properties up to and including these targets should not result in the loss of content or functionality.</p>
 
          <p>This SC focuses on the ability to increase spacing between lines, words, letters, and paragraphs. Any combination of these may assist a user with effectively reading text. As well, ensuring users can override author settings for spacing also significantly increases the likelihood other style preferences can be set by the user. For example, a user may need to change to a wider font family than the author has set in order to effectively read text. </p>
       	<section>
          <h3 id="authorresp">Author Responsibility </h3>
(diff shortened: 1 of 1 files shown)


patrickhlauke:
how is the removal of that paragraph justifiable, since in my mind that paragraph addressed the "to at least" part of all the metrics in the bullet points?

mbgower:
Created a separate issue for what is some kind of generator glitch on the figures https://github.com/w3c/wcag/issues/1458

mbgower:
@patrickhlauke, please see my individual comments on commits. I put the removal in a separate commit. I found this paragraph nonsensical when I read it, and tried it just do a minor change, which I _think_ makes it more readable, but it remains clear to me exactly what it is saying (if it is more than what is talked about in the very next paragraph, which I also addressed in my commit and PR comments).

mbgower:
@patrickhlauke Is this what it's meant to say?
> The metrics set a target for a minimum increase in text spacing that must be met in. Any increase in spacing for any of these four style properties up to and including these targets should not result in the loss of content or functionality.

That seems to me to be self-evident from the wording in the SC, but this at least makes sense to me, which the existing wording on its own did not.

I suggest this would fit better as the third paragraph of the Intent, rather than the second.

Let me know, and I'll push it as another commit.

patrickhlauke:
> Is this what it's meant to say?

I believe so, yes (if i'm reading/understanding the lengthy to-ing and fro-ing from #635 and the resultant #860 right)

mbgower:
On review, shifted the user responsibility section to immediately follow the author responsibility section, and added in clarification that the author is not responsible for providing a mechanism to alter text spacing. I think this is good to go now.

fstrr:
@mbgower You still want this PR?

patrickhlauke:
This is so conflicted at this point, that I'd recommend redoing this as a fresh PR

understanding/21/text-spacing.html
```diff
@@ -9,47 +9,50 @@
       <h1>Understanding Text Spacing</h1>
       <section id="intent">
          <h2>Intent</h2>
-         <p>The intent of this Success Criterion (SC) is to ensure that people can override author specified text spacing to improve their reading experience. Each of the requirements stipulated in the SC's four bullets helps ensure text styling can be adapted by the user to suit their needs.</p>
-         <p>The specified metrics set a minimum baseline. The values in between the author's metrics and the metrics specified in this SC should not have loss of content or functionality.</p>
+         <p>The intent of this Success Criterion (SC) is to ensure that people can override author-specified text spacing to improve their reading experience. Each of the requirements stipulated in the SC's four bullets helps ensure text styling can be adapted by the user to suit their needs.</p>
+         <p>The metrics set a target for a minimum increase in text spacing that must be met. Starting from the author's baseline implementation, any increase in spacing for any of these four style properties up to and including these targets should not result in the loss of content or functionality.</p>
+
```
mbgower, line 14:
```suggestion
```

mbgower:
Redid this cleanly against main as #4125

Issue #1478, opened 2020-10-14
I think there's a typo on w3.org


Sylvie123:
The typo is on this page: https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA11.html
- typo is “navigation” in: role="navigaton"
 it is missing the “i”. It is listed more than once within the page.
Thank you
Sylvie Faulkner

alastc:
Thanks for highlighting this, the update has been merged with master and should appear soon.

Sylvie123:
Hello, will it take a while before the fix shows up on the website? Thank you
<div id="leftnav" role="navigaton" aria-labelledby="leftnavheading">
<div id="leftnav" role="navigaton" aria-label="Primary">

PR #1479, opened 2020-10-14
Fixes for #1478

awkawk:


alastc:
@michael-n-cooper This is for 2.2 & 2.1, I've labelled it as such, is there anything else I should do to highlight it is for 2.1 as well?

PR #1494, opened 2020-10-29
Link page about avoiding new window links to new window link warning

lkmorlan:
Link G200 to G201.

These are related and should be linked. The link in the other direction is already present.


mbgower:
This appears to be a duplicate (or at least a part) of #314, so closing (especially since I have a draft response in place).
If that is incorrect, feel free to reopen.

PR #1496, opened 2020-10-31
Adding the hex colour value to an example

### LINKED ISSUE #1305: Example 2 for G195 is too inaccurate
[G195](https://w3c.github.io/wcag/techniques/general/G195):

> A Web page includes a form inside a table. The borders of both the table and the form elements are thin, black lines. When focus lands on a form element, the element is outlined with a 5 pixel red line that is partially transparent. 

As long as "red" is not defined more precisely, it is unclear whether this color meets the contrast requirements, especially since part of the red is transparent (how large is this part?). The example should be removed or clarified to comply with SC 2.4.11 and 2.4.12

alastc:
Closes #1305

### DIFF
--- a/techniques/general/G195.html
+++ b/techniques/general/G195.html
@@ -13,7 +13,7 @@
       <section class="example">
          <h3>Form Elements</h3>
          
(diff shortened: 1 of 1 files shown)


alastc:
Noting that we'd like to over-haul this to include visual examples.

PR #1498, opened 2020-10-31
Adding -4 to the measure

### LINKED ISSUE #1324: Minimum focus indicator definitions for 2.4.11/2.4.12 are ambiguous; example calculations do not account for corners

alastc:
Closes #1324


alastc:
Agreed https://www.w3.org/2020/11/10-ag-minutes.html#item17

Issue #1076, opened 2020-03-09
1.4.1: Use of Color - understanding document needs to be clear if it's about "visual" or more


patrickhlauke:
https://www.w3.org/WAI/WCAG21/Understanding/use-of-color.html currently seems to mix and match different concepts. while the SC normatively talks about "visual means" and notes specifically that it's about *color perception*, the understanding document starts talking about "text only displays" (which provide NO other cues other than pure text) and Braille displays.

To me, this makes the SC veer far more into "a text-based method of conveying information/distinguishing content must be provided", which I don't believe is the intention of the SC which was dealing purely with color perception.

In that light, this also seems to bump against that controversial escape clause in F73 https://www.w3.org/WAI/WCAG21/Techniques/failures/F73 where it talks about red/pink not just being a difference in color, but lightness ... since in the context of text-only or braille browsers, that excuse ("it's not just a change of color, it's also a change in lightness") wouldn't hold up.

does the understanding document need to be tightened to remove what I would suggest are actually misleading examples/benefits that take this into 1.1.1, 1.3.1 and 4.1.2 territory?

patrickhlauke:
cross-reference this discussion on WebAIM https://webaim.org/discussion/mail_message?id=41883

patrickhlauke:
and from that, as I note, if this DOES apply to text-only displays as well, then "even an additional icon or border or different text visual treatment or similar would not be sufficient to pass", arguably, as there's no guarantee a text-only display - assuming this means a terminal or similar - can display those.

patrickhlauke:
lastly, which is the question from that discussion: that whole aspect of "color is hue" ... if an indication is given by completely inverting foreground/background, is that a valid way to pass 1.4.1?

JAWS-test:
SC 1.4.1 is only about visual in standard view (e.g. not in HCM, with custom style sheet or anything else). 

See also https://github.com/w3c/wcag/pull/1020

patrickhlauke:
thank you @JAWS-test but that was not the question

mraccess77:
Use of > 3:1 contrast can be a way to pass SC 1.4.1 when the difference in luminance is used to communicate meaning visually.    The same thing goes for anything like a black caret or a black checkmark, blue background selection of a list item, or a black border  -- they are all differences in luminance and not color that is communicating something -- a checked state, border, caret, selection.       

[...]

Issue #1032, opened 2020-02-03
Note and benefits contradict each other in 1.4.1 (Use of color)


JAWS-test:
Note in [SC 1.4.1](https://w3c.github.io/wcag/understanding/use-of-color.html):

> This success criterion addresses color perception specifically. Other forms of perception are covered in Guideline 1.3 including programmatic access to color and other visual presentation coding. 

Benefits

> People using Braille displays or other tactile interfaces can detect text cues by touch
> People using text-only, limited color, or monochrome displays may be unable to access color-dependent information. 

I understand the note to mean that SC 1.4.1 checks whether colour-coded content is distinguishable without assistive technology. That is why there is also a requirement that the contrast ratio must be at least 3:1 so that sighted users can distinguish different colours (F81, G183). On the other hand, SC 1.3.1 checks whether colour-coded content is transmitted programmatically.

The chapter "Benefits" contradicts this because it suddenly mentions users of assistive technology who cannot see or who do not use the standard view. For these users, however, even a minimum contrast of color-coded content would not be useful. I would therefore like to delete the corresponding users. Does anyone have any objections? 

See my related PRs https://github.com/w3c/wcag/pull/1020 and https://github.com/w3c/wcag/pull/1033

patrickhlauke:
Agree, 1.4.1 has nothing to do with AT users.

alastc:
With #1500 merged, closing this issue.

PR #1500, opened 2020-10-31
Rework understanding for 1.4.1 to be firmly only about visual aspect

patrickhlauke:
Closes https://github.com/w3c/wcag/issues/1076, https://github.com/w3c/wcag/issues/1032

* removes mentions of braille displays and assistive technologies - this is not what SC 1.4.1 is about. while related, these are covered in separate SCs ... 1.4.1 is purely about visual aspects for users who cannot perceive color (but are still sighted)
* adds the "oh no, this is not just a difference of color - they have different lightness/luminosity, so as long as contrast is 3:1 or higher that counts as another visual cue" redefinition of what "color" means from being a side note in a technique ([F73: Failure of Success Criterion 1.4.1 due to creating links that are not visually evident without color vision](https://www.w3.org/WAI/WCAG21/Techniques/failures/F73)) to actually being part of the understanding document itself

### DIFF
--- a/understanding/20/use-of-color.html
+++ b/understanding/20/use-of-color.html
@@ -266,7 +266,7 @@
             
             <li>
                									         
-               <a href="https://www.w3.org/WAI/WCAG21/Techniques/css/C15" class="css"></a>
+               <a href="https://www.w3.org/WAI/WCAG21/Techniques/css/C15" class="css">Using CSS to change the presentation of a user interface component when it receives focus</a>
                								       
             </li>
             
--- a/understanding/20/use-of-color.html
+++ b/understanding/20/use-of-color.html
@@ -13,7 +13,7 @@
       <h2>Intent of Use of Color</h2>
       
       
-      <p>The intent of this Success Criterion is to ensure that all users can access information
+      <p>The intent of this Success Criterion is to ensure that all sighted users can access information
          that is conveyed by color differences, that is, by the use of color where each color
          has a meaning assigned to it. If the information is conveyed through color differences
          in an image (or other non-text format), the color may not be seen by users with color
@@ -24,7 +24,7 @@
       <p>Color is an important asset in design of Web content, enhancing its aesthetic appeal,
          its usability, and its accessibility. However, some users have difficulty perceiving
          color. People with partial sight often experience limited color vision, and many older
-         users do not see color well. In addition, people using text-only, limited-color or
+         users do not see color well. In addition, people using limited-color or
          monochrome displays and browsers will be unable to access information that is presented
          only in color.
       </p>
@@ -44,6 +44,24 @@
          </p>
          
       </div>
+
+      <div class="note">
+
+         <p>This criterion does not directly address the needs of users with assistive technologies.
+            It aims to ensure that sighted users who cannot perceive color can still understand content.
+            How information is conveyed to assistive technology users is covered separately in other
+            criteria, including (but not limited to)
+            <a href="non-text-content">1.1.1 Non-text Content</a>,
+            <a href="info-and-relationships">1.3.1 Info and Relationships</a>, and
+            <a href="name-role-value">4.1.2 Name, Role, Value</a>.
+         </p>
+         <p>Conversely, even if information that is conveyed by color differences is appropriately conveyed
+            to assistive technologies, it does not necessarily pass this criterion, as sighted users who cannot
+            perceive color may not necessarily be using any assistive technologies. This criterion requires
+            a visible alternative to color.
+         </p>
+
+      </div>
       
       
    </section>
@@ -61,16 +79,12 @@
             in other visual ways.
          </li>
          
-         <li>People using text-only, limited color, or monochrome displays may be unable to access
+         <li>People using limited color monochrome displays may be unable to access
             color-dependent information.
          </li>
          
          <li>Users who have problems distinguishing between colors can look or listen for text
             cues.
-         </li>
-         
-         <li>People using Braille displays or other tactile interfaces can detect text cues by
-            touch.
          </li>
          
       </ul>
@@ -92,11 +106,10 @@
             </p>
             
             <p>A form contains both required and optional fields. Instructions at the top of the
-               form explain that required fields are labeled with red text and also with an icon
-               whose text alternative says, "Required." Both the red text and the icon are programmatically
-               associated with the appropriate form fields so that assistive technology users can
-               determine the required fields. 
-               
+               form explain that required fields are labeled with red text and also with an icon.
+               Users who cannot perceive the difference between the optional field labels and the
+               red labels for the required fields will still be able to see the icon next to the
+               red labels.               
             </p>
             
          </li>
@@ -110,32 +123,9 @@
             </p>
             
             <p>Students view an SVG image of a chemical compound and identify the chemical elements
-               present based on the colors and numbers used in the diagram. The text alternatives
-               associated with each element name the color of the element and indicate the element's
-               position in the diagram. Students who cannot perceive color have the same information
-               about the compound as their classmates. (This technique also satisfies 
-               <a href="non-text-content" class="guideline">Guideline 1.1</a> Level A.)
-            </p>
-            
-         </li>
-         
-         <li>
-            
-            <p>
-               								       
-               <strong> Disabled Form elements.</strong>
-               							     
-            </p>
-            
-            <p>
-               Form elements which are disabled via markup or scripting, are greyed out and made
-               inactive by the user agent. When in the disabled state these elements do not receive
-               focus. Assistive technologies can programmatically determine the state of disabled
-               elements and will provide this information to the user as the elements are encountered
-               on the page. The change in color and loss of focus provides redundant, visual information
-               about the state of the control.
-               
-               
+               present based <strong>both</strong> on the colors used, as well as numbers next to each
+               element. A legend shows the color and number for each type of element. Sighted users who
+               cannot perceive color can still understand the image by relying on the numbers.
             </p>
             
          </li>
--- a/understanding/20/use-of-color.html
+++ b/understanding/20/use-of-color.html
@@ -58,9 +58,8 @@
             this would pass (again, assuming that the foreground and background colors have a sufficient contrast).
          </p>
          <p>However, if content relies on the user's ability to accurately perceive or differentiate a particular color
-            (either implicitly - e.g. green for success, red for failure - or explicitly - e.g. "the blue items represent
-            the most recently updated entries"), an additional visual indicator will be required, regardless of
-            the contrast ratio between those colors.
+            an additional visual indicator will be required regardless of the contrast ratio between those colors. For example, 
+           knowing whether an outline is green for valid or red for invalid.
          </p>
 
       </div>
--- a/understanding/20/use-of-color.html
+++ b/understanding/20/use-of-color.html
@@ -45,6 +45,20 @@
          
       </div>
 
+
+      <div class="note">
+
+         <p>If content is conveyed through the use of colors that differ not only in their hue,
+            but that also have a significant difference in lightness, then this counts as an additional
+            visual distinction, as long as the difference in contrast between the colors is 3:1 or greater.
+            For example, a light green and a dark red differ <strong>both</strong> by color (hue)
+            <strong>and</strong> by lightness, so they would pass if the contrast difference is at least 3:1.
+            Similarly, if content is distinguished by inverting an element's foreground and background colors,
+            this would pass (again, assuming that the foreground and background colors have a sufficient contrast).
+         </p>
+
+      </div>
+
       <div class="note">
 
          <p>This criterion does not directly address the needs of users with assistive technologies.
--- a/understanding/20/use-of-color.html
+++ b/understanding/20/use-of-color.html
@@ -56,6 +56,11 @@
             Similarly, if content is distinguished by inverting an element's foreground and background colors,
             this would pass (again, assuming that the foreground and background colors have a sufficient contrast).
          </p>
+         <p>However, if content relies on the user's ability to accurately perceive a particular color (either
+            implicitly - e.g. green for success, red for failure - or explicitly - e.g. "the blue items represent
+            the most recently updated entries"), an additional visual indicator will be required, regardless of
+            the relative contrast between those colors.
+         </p>
 
       </div>
 
--- a/understanding/20/use-of-color.html
+++ b/understanding/20/use-of-color.html
@@ -56,8 +56,8 @@
             Similarly, if content is distinguished by inverting an element's foreground and background colors,
             this would pass (again, assuming that the foreground and background colors have a sufficient contrast).
          </p>
-         <p>However, if content relies on the user's ability to accurately perceive a particular color (either
-            implicitly - e.g. green for success, red for failure - or explicitly - e.g. "the blue items represent
+         <p>However, if content relies on the user's ability to accurately perceive or differentiate a particular color
+            (either implicitly - e.g. green for success, red for failure - or explicitly - e.g. "the blue items represent
             the most recently updated entries"), an additional visual indicator will be required, regardless of
             the relative contrast between those colors.
          </p>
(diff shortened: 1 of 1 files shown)


patrickhlauke:
In addition, it would be nice to pull in the key term/definition for "contrast ratio" into the understanding document (and ideally link it from the new note)...but I'm not quite sure how that can be done easily - seems there's some automagic behavior that pulls in key terms somehow that currently escapes me.

JAWS-test:
I am unsure whether in addition to the 3:1 exception it should  be mentioned that color coding can also be an error if the contrasts of the used colors are higher than 3:1. This is always the case when the color itself conveys information and not only the color difference:
* Example for color coding: red for error, green for success - the color itself is important
* Example for color difference: Blue for active menu item, white for inactive menu item - the color is not important, only the contrast

patrickhlauke:
> Example for color coding: red for error, green for success - the color itself is important

I believe in these cases, relying on the user's ability to *recognise* the intrinsic meaning of a particular color (rather than the more narrow case of being able to simply *distinguish* between colors) falls more under 1.1.1 and 1.3.1.

mraccess77:
In my opinion not being able to distinguish a certain color such as activate the green button - would fall under SC 1.4.1.  The understanding doc for 1.3.3 for sensory characteristics points users to guideline 1.4 which seems appropriate to me.

patrickhlauke:
ah, on reflection, yes that seems appropriate as this is still about providing a *visual* thing (where 1.1.1 and 1.3.1 are mainly AT-based). ok, wondering now if that should be another note, or at which point it's not worth making notes and instead just plain core understanding text itself.

patrickhlauke:
added an extra clarification to the note about situations where sufficient contrast still doesn't absolve the need for something else. maybe a bit verbose, but hopefully gets the point across.

JAWS-test:
@patrickhlauke Thanks a lot! I think the addition is very good and not at all to verbose

yatil:
I’m not comfortable with a change in that direction. While it aligns with the G183 loophole, color as hue-only is not defined in WCAG proper. That is emphasized that the definition of relative luminance uses color for all aspects of a color not only for hue. I don’t think color should mean something different in different places in the same document.

I think a better, easier to teach and implement version is to interpret color in its commonly used meaning as the combination of all properties of a color, thus rendering G183 a failure, which it IMHO always should have been.

JAWS-test:
> I think a better, easier to teach and implement version is to interpret color in its commonly used meaning as the combination of all properties of a color, thus rendering G183 a failure, which it IMHO always should have been.

I think that would not be correct. Every visual representation on the web works with color differences, e.g. black text on a white background is also just a color difference with contrast 21:1, i.e. a minimum contrast must be defined, otherwise each content would violate 1.4.1

yatil:
The minimum contrast for text is already defined elsewhere. Also the form of the letters conveys the information of the text, not the color alone. This issue is about 1.4.1.

JAWS-test:
@yatil 

> The minimum contrast for text is already defined elsewhere. 

That's right. Maybe the example was not so good. But if I do not define a minimum contrast in 1.4.1, I would not allow e.g. the active menu item to be white on black and the inactive menu items to be black on white, although the contrast difference between active and inactive is 21:1. Is this intended? If so, probably many sites would violate 1.4.1 that have always claimed the 3:1 exception so far

yatil:
> But if I do not define a minimum contrast in 1.4.1, I would not allow e.g. the active menu item to be white on black and the inactive menu items to be black on white, although the contrast difference between active and inactive is 21:1. Is this intended?

From 1.4.1:

> Color is not used as the only visual means of conveying information, indicating an action, prompting a response, or distinguishing a visual element.

So, yes. This is what the SC says. Of course that somehow would make 1.4.11 redundant, but it just shows how generalized the SC 1.4.1 is. If we want to change it to

> Hue is not used as the only visual means of conveying information, indicating an action, prompting a response, or distinguishing a visual element.

then go for it. I stand by my statement that having different combinations of color attributes labeled with the word ”color” in the same spec is bad for understanding.

bruce-usab:
@yatil wrote:
> I think a better, easier to teach and implement version is to interpret color in its commonly used meaning as the combination of all properties of a color

I was struck by how remarkably sensible this is!  It needs a little tweaking, as we would not want to have a formal definition of color that uses the word color as part of the definition.  But the WCAG 2.x definition of color could mention hue, brightness, and both <q>commonly used meaning</q> and <q>all properties</q>.

> thus rendering G183 a failure, which it IMHO always should have been.

Nuts, you lost me right there!  Please explain, as it very much seems to me that G183 relies on the commonly used meaning of color.

> I stand by my statement that having different combinations of color attributes labeled with the word ”color” in the same spec is bad for understanding.

+1

patrickhlauke:
> it very much seems to me that G183 relies on the commonly used meaning of color

do mere mortals, when asked, say "red and pink are the same *color* ... they just have different *lightness*" ?

i personally don't think so...

and to be clear, I've not been a total fan of this reinterpretation of color in F73 ... but if this is indeed the party line, then this PR aims to bubble up this unorthodox interpretation at least one level up to the understanding, not buried in a technique

Myndex:
### I disagree with these changes IN PART.

While I certainly agree with the spirit of some of the changes, there is need of further revision IMO.

On commit **2622d7f**:

I agree with the added word sighted: _...ensure that all **sighted** users..._

But I disagree with the removal of "text only displays". These types of displays certainly affect sighted users. The world is a big place, while here in the US we may rarely see text only and limited color displays in direct use, we can't say that is true for the rest of the world. 

And we do have a "text-only" here when we enter reader mode that ignores much of the CSS styling including color as in hue.

And if we are going to open this up for a change, we can clarify and modernize further. Here is a suggested alternate paragraph:

      <p>Color is an important asset in the design of Web content, enhancing its aesthetic appeal,
         its usability, and its accessibility. However, some users have difficulty perceiving color. 
         People with Color Vision Deficiency may be unable to recognize different hues, such as an
         inability to tell the difference between red and green. People with partial sight, cognitive and neurological
         impairments, as well as many older users, can have functionally limited color vision. People with these and
         other visual needs may use their own CSS styles that overwrite the author's original colors. Further, people
         using text-only, limited-color, or monochrome displays and browsers will be unable to access
         or understand information that is presented only as a difference in color.
      </p>

Later in this commit is this note, quoted only in part:

        <p>This criterion does not directly address the needs of users with assistive technologies.
            It aims to ensure that sighted users who cannot perceive color can still understand content.
            How information is conveyed to assistive technology users is covered separately ......

This may be true if assistive technologies only means screen-readers. However I consider extensions like "DarkMode" and other user adjustments to be assistive technology as well, and if we consider those then this SC does still apply, and I suggest it does apply **_to all sighted users_**. WHich brings us to this sentence:

         ...ensure that sighted users **who cannot perceive color** can still understand...

All sighted users can perceive color with the very rare exception of "Blue Cone Monochromats" who have no red or green cones and are the only ones that are truly "color blind". But they are not only incredibly rare (one in a million) they also suffer a profound impairment as blue cones are very poor in resolution among other things. They need assistive technology regardless.

The remaining individuals with Color Vision Deficiency still perceive color. They have a more limited gamut of color, and difficulty in differentiating between hues. [I have a CVD simulator that demonstrates this clearly.](https://www.myndex.com/CVD/sRGBCVD) But again, this SC does not only apply to CVD, it applies to all sighted users, and is primarily a restriction from coding information strictly with hue, which is bad design practice and I'll get to that in a moment. The statement "...who cannot perceive color" is repeated a couple times. All need to be revised to something like _"those who cannot differentiate between some color hues."_

### Objection RE: Note on Coding Information

**I most strongly object** to the note of commit 9b74ac0, which implies that luminance contrast is useful for coding. **IT IS NOT.** Multiple international standards including the FAA/NASA have restrictions on coding using luminance contrast, for good reason.

### Let's discuss.

Luminance contrast (light/dark, disregarding hue and saturation) is critical for things like letters, fonts, objects, so that the visual stimuli gets through the thalamus, and in the visual cortex is filtered by V1, sent to  V4, and then to the VWFA (Visual Word Form Area) for decoding as whole words and letter pairs. For this to happen efficiently, we need ample luminance contrast, and hue/chroma contrast is not much a part of this.

But that does not mean that luminance contrast is useful for information coding based on contrast alone, and if in combination with hue only to support the hue coding, and the 3:1 reference is unsupported here. And luminance contrast should not be a part of this SC at all, it is not relevant (other than luminance contrast shall not be used for coding information). Luminance contrast can be useful for things like "focus" but not so much for things like "good/bad".

Something this SC does not appear to touch on (perhaps I missed it) is that color hues are also NOT absolute in meaning and are interpreted differently in different cultures. **Words and symbols convey meaning.** Color hues do not reliably convey meaning, and luminance contrast does not either (with the possible exception of "on/off").

### Summary:
- Luminance contrast is important for details of content such as fonts, words, drawings of objects... But luminance contrast itself does not convey the meaning, the meaning is conveyed by the details in the stimuli. 

- Color as in hue/chroma/saturation is less important for "details" and lexical processing. Hue/chroma is a third the strength of luminance in terms of contrast and resolution (this is how image data compression like jpeg gets away with throwing out most of the color information in an image).
    - Color hue & sat can be useful for coding information, but a number of limitations exist, some of which affect all users regardless of impairment. 
    - Thus color hue can not be the only informational coding.
    - A STOP sign is not just a big red octagon. It is emblazoned with the word STOP. And here again, we have multiple information codings. The sign is reflective red, but it is also an octagon, and it also has a lexical, literal statement of meaning. 

- Meaning is found in words and symbols.
    - Luminance contrast serves to makes _THOSE_ words and symbols readable, but does not add meaning. However, consistent contrast of specific items (such as roads on a map) can help make that information more clear and organized.
    - Color hue/chroma can add meaning in some cases, but it can also detract and interfere with understanding if used improperly. For instance, what does Cerulean Blue mean? The answer is unknown without a legend. Color does not have universal meaning on its own.

- There are other kinds of contrast: size, shape, motion — but contrasts are not really relevant for this SC, which is about information coding. Contrast is better covered elsewhere.

### SIde Bar on CVD 

Keep in mind that color as in hue is already full of luminance contrast. Full green #0F0 is more than three times the luminance of full red #F00. But someone with protanopia won't even see red on one of the new UHD monitors. in eRGB, they see red a bit darker than we do, but most important can not distinguish between red and green. Protan and Deutan are the most common CVD types, and thus colors that vary only in their red to green ratios present a significant problem for them.

Again, luminance contrast does not convey information, but makes lexical information more clear. Examples:

Here is red and green, set to the same luminance (no luminance contrast). It does not matter if you have CVD or not, it is plainly hard to read:

Screen Shot 2020-11-09 at 8 21 41 PM

But how do people with CVD see it?

Screen Shot 2020-11-09 at 8 19 55 PM

There's an odd fact: the protanopia sees it BETTER because they see red darker, it actually improves their contrast. But it is literally unreadable to the deuteranopia and might as well be invisible.

Now let's boost the luminance contrast to a reasonable 80 (APCA):

Screen Shot 2020-11-09 at 8 21 04 PM

This contrast boost did not add any real "information" it just made the words easier to read.

Screen Shot 2020-11-09 at 8 20 49 PM

And both protan and deutan see it better as well. Now let's flip it around:

Screen Shot 2020-11-09 at 8 55 56 PM

About the same APCA 80 contrast here...

Screen Shot 2020-11-09 at 8 56 02 PM

But again the deutan and protan can really glean no "information" because they cannot distinguish red from green. The luminance contrast does not convey information, it just make information clearer. (Sounds like a BASF commercial, LOL).

Andy

understanding/20/use-of-color.html
```diff
@@ -44,6 +44,43 @@ <h2>Intent of Use of Color</h2>
          </p>
          
       </div>
+
+
+      <div class="note">
+
+         <p>If content is conveyed through the use of colors that differ not only in their hue,
+            but that also have a significant difference in lightness, then this counts as an additional
+            visual distinction, as long as the difference in contrast between the colors is 3:1 or greater.
```
alastc, line 53: The explanation uses hue, lightness and contrast without explaining that (we define) contrast as a difference in lightness. Suggest:  "as long as the difference in lightness (contrast) between the colors is 3:1 or greater."

understanding/20/use-of-color.html
```diff
@@ -44,6 +44,43 @@ <h2>Intent of Use of Color</h2>
...
+            the most recently updated entries"), an additional visual indicator will be required, regardless of
+            the relative contrast between those colors.
+         </p>
+
+      </div>
+
+      <div class="note">
+
+         <p>This criterion does not directly address the needs of users with assistive technologies.
+            It aims to ensure that sighted users who cannot perceive color can still understand content.
```
alastc, line 70: Somewhat agree with Andy's comments, "cannot perceive color" is very binary, suggest "users who cannot distinguish between colors".

understanding/20/use-of-color.html
patrickhlauke: to be pedantic, should it perhaps more accurately be "the difference in relative luminance between the colors leads to a contrast ratio of 3:1 or greater" (thinking of the actual glossary definition here https://www.w3.org/TR/WCAG/#dfn-contrast-ratio)

understanding/20/use-of-color.html
Myndex: Yes, though even then it's "users who cannot distinguish between **some** colors".   

While normal vision can see millions of colors, most CVD types can still see hundreds of thousands of colors or more.

patrickhlauke: done (both here and in the other places where the understanding document talked about "perceive color")

understanding/20/use-of-color.html
patrickhlauke: made changes here to use the more accurate glossary definition. would be nice to link to/include the glossary definition of "contrast ratio", but beyond adding a hard link (which would then potentially link to wcag 2.0 or 2.1 or whatever, making it immediately broken when this SC is reused in other versions) i don't think there's some magic incantation to add in the definition directly/automatically in the understanding itself

carlosapaduarte:
The ACT CG is working on a rule to partially check SC 1.4.1 (https://github.com/act-rules/act-rules.github.io/pull/1010)

Currently we have a blocking issue that is understanding what is meant by color in SC 1.4.1 (hue only?)
This PR would make it cleat what the SC applies to. Is there any way that we can help this move forward, or help in reaching out an understanding (even it is the opposite of what the PR currently proposes)?

StommePoes:
> All sighted users can perceive color

This assumes we're only talking about what a human's eye can do, and that everyone who sees is seeing through a human eye. Other SCs consider the limits of AT (ie what can a screen reader do, what can a Braille display do, what can speech recognition do... and what kinds of users most commonly use them).

Adjusting things to help me see them may reduce or eliminate colours, just as when I magnify I'm changing the size (oh and we have 2 SCs about that, and both mention how technology may work when users do this). 

I feel this point cannot be forgotten.

understanding/20/use-of-color.html
```diff
@@ -276,7 +288,7 @@ <h3>Additional Techniques (Advisory) for Use of Color</h3>
             
             <li>
                									         
-               <a href="https://www.w3.org/WAI/WCAG21/Techniques/css/C15" class="css"></a>
+               <a href="https://www.w3.org/WAI/WCAG21/Techniques/css/C15" class="css">Using CSS to change the presentation of a user interface component when it receives focus</a>
```
awkawk, line 291: @patrickhlauke the generation process should take care of this - we often leave the title out so if it changes it isn't incorrect (but the generator should replace the anchor content anyway).

patrickhlauke: ah, good to know. would you prefer me to remove the text, for consistency/style's sake?

patrickhlauke: noting that all other links to techniques in this understanding doc have explicit link text, so perhaps for consistency within this document itself probably worth keeping (or changing the other ones to be empty)

alastc:
Noting this PR was accepted, with an update (to be done) to replace the 'blue' example with a filled-stars one. I can tackle that later tonight.

alastc:
Actually the example wouldn't make sense as a filled star, so I adjusted it to remove the 'instructions' aspect, so it doesn't obviously fail 1.3.3.

patrickhlauke:
> Actually the example wouldn't make sense as a filled star, so I adjusted it to remove the 'instructions' aspect, so it doesn't obviously fail 1.3.3.

yup, looking good to me

PR #1501, opened 2020-11-02
Focus-appearance 2.4.11 SC & Understanding rewording suggestions

mbgower:
Still in progress. I'll add Alastair as a reviewer when I'm done.


mbgower:
@alastc I worked through the document. Stopped at Benefits.
I feel like it will need some more work, but hopefully you view this as a solid improvement.

guidelines/sc/22/focus-appearance-minimum.html
```diff
@@ -5,11 +5,11 @@ <h4>Focus Appearance (Minimum)</h4>
...
-        <li><strong>Change of contrast:</strong> The color change for the focus indication area has a contrast ratio of at least 3:1 with the colors of the unfocused state.</li>
-        <li><strong>Adjacent contrast:</strong> The focus indication area has a contrast ratio of at least 3:1 against all adjacent colors for the minimum area or greater, or has a thickness of at least 2 CSS pixels.</li>
+        <li><strong>Minimum area:</strong> The area of the focus indicator is at least as large as the area of a 1 <a>CSS pixel</a> thick perimeter of the focused control, or the focus indicator has a thickness of at least 8 CSS pixels along the shortest side of the user interface component.</li>
+        <li><strong>Change of contrast:</strong> The pixels making up the minimum focus indicator area achieve at least a 3:1 contrast between their colors in the focused and unfocused states. </li>
+        <li><strong>Adjacent contrast:</strong> The focus indicator contrasts at least 3:1 against all adjacent colors, or has a thickness of at least 2 CSS pixels.</li>
         <li><strong>Not fully obscured:</strong> The item with focus is not entirely hidden by author-created content.</li>
```
alastc, line 13: I think most/all of these changes have already gone in, we'll just have to check for merge issues.

understanding/22/focus-appearance-minimum.html
```diff
@@ -197,15 +198,15 @@ <h3>Adjacent contrast</h3>
...
 
         </section>
         <section id="relation-non-text-color-contrast">
-            <h3>Relation to non-text color contrast</h3>
-            <p><a href="non-text-contrasast" class="sc">Success Criterion 1.4.11 Non-text contrast</a> requires that a UI component maintains contrast across it’s states, including when focused. For example, a checkbox must contrast with its adjacent colors, and the check within the checkbox must contrast with its adjacent colors. </p>
+            <h3>Relationship to Non-text Contrast</h3>
+            <p><a href="non-text-contrast" class="sc">Success Criterion 1.4.11 Non-text Contrast</a> requires that a UI component maintains contrast against the adjacent <em>background</em> for each of its states. When it has does not have focus, it must be visible against the background. When it receives focus, it still must be visible. However, Non-text Contrast does require contrast <em>between</em> the UI component's focused and non-focused states. Merely that in each state it contrasts against the adjacent background. </p>
```
alastc, line 207: Missing a "not" from "However, Non-text Contrast does NOT require"

understanding/22/focus-appearance-minimum.html
```diff
@@ -214,17 +215,24 @@ <h3>Relation to non-text color contrast</h3>
...
-                <li>The check (tick) within the checkbox contrasts with its adjacent color; </li>
-                <li>The circular outline around the checkbox that indicates focus has a sufficient <em>change</em> of contrast.</li>
+                <li>The dark square outline of the checkbox has sufficient contrast with its adjacent light background in the default state (meeting 1.4.11 Non-text Contrast).</li>
+                <li>When the checkbox receives focus in the second image, a grey ring appears around it. The ring contrasts against the light page background sufficiently (still meeting 1.4.11). The size of the ring's dark outline is larger enough to meet the <strong>Minimum area</strong> bullet of 2.4.11 Focus Appearance. The color of the pixels making up the ring's outline versus the same pixels in the unfocused state meets the 2.4.11 <strong>Change of contrast</strong> bullet. The ring's outline also contrasts against the inside shading (another more subtle indicator of focus), so both the inner and outer edges of the ring's outline meet 2.4.11's  <strong>Adjacent contrast</strong> bullet.</li>
```
alastc, line 219: "largeR enough"
Also, with the change to the 3rd bullet now, I don't think the outer ring needs to contrast against "all" adjacent colours. Because it is separated from the component, it doesn't have to meet any adjacent colour.

alastc (changes requested):
I'm going to try to resolve the merge conflicts now, it was for the SC text specifically. 
There are a couple of other little things I left comments on though.

alastc:
Also note the &Pi; entity causes a build error in the Java thing-a-majig, so I put in the Unicode character instead.

Otherwise, looks good, I'll add it to the survey.

understanding/22/focus-appearance-minimum.html
mbgower: added in a new commit.

understanding/22/focus-appearance-minimum.html
mbgower: I made it "large".
 Has the 3rd bullet change been approved? I'll need to look at the final wording, but my sense is maybe there still needs to be some comment here, as they have added in that shading as part of the focus indicator. Or are we now saying that it wouldn't matter what the degree of shading here, it would still meet? (i.e., no shading, meets; shading matches the ring color -- becoming a massive ring -- still meets) It seems to me we still care about a contrast between the light shading and the edge of the control itself. So maybe this needs to be rewritten.

alastc: The focus indicator (that meets the 1st two bullets) is sufficient, and as the indicator isn't adjacent to the component, it missed the 3rd bullet (now). We did resolve to use that text yesterday, but as it's normative the SC (with all it's changes) will go through CFC.

guidelines/terms/22/perimeter.html
```diff
@@ -1,7 +1,6 @@
 <dt class="new"><dfn>Perimeter</dfn></dt>
 <dd class="new">
 	<p class="change">New</p>	
-   <p>the continuous line forming the boundary of a shape. The perimeter calculation for a rectangle is 2<em>h</em>+2<em>w</em>, where <em>h</em> is the height and <em>w</em> is the width. The perimeter of a circle is 2𝜋<em>r</em>.
-    </p>
+   <p>the continuous line forming the boundary of a shape. The perimeter calculation for a rectangle is 2<em>h</em>+2<em>w</em>, where <em>h</em> is the height and <em>w</em> is the width. The perimeter of a circle is 2𝜋<em>r</em>.</p>
```
awkawk, line 4: to be consistent with other definitions, suggest:
<p>continuous line forming the boundary of a shape</p>
<p>The perimeter calculation for a rectangle is 2<em>h</em>+2<em>w</em>, where <em>h</em> is the height and <em>w</em> is the width. The perimeter of a circle is 2𝜋<em>r</em>.</p>

understanding/22/focus-appearance-minimum.html
```diff
@@ -22,16 +22,17 @@ <h2>Focus Appearance (Minimum) Success Criteria text</h2>
...
             </ul>
             <p class="note">A keyboard focus indicator which has a pattern or gradient may have parts that do not meet the 3:1 contrast ratio for the change of contrast, as long as an area equal to the minimum does meet the contrast ratio.</p>
             <p class="note">If the control has a visible boundary smaller than the hit area, the size measure is taken from the visible boundary.</p>
-            <dt class="new"><dfn>Focus indication area</dfn></dt>
+            <p class="note">The default user agent focus indicator is not excepted from the requirements of Focus Appearance (Minimum). If a user agent implementation fails to meet these requirements, any page relying on that user agent would fail.</p>
+            <dt class="new"><dfn>Focus indicator</dfn></dt>
             <dd class="new">
-	            <p>the pixels that change between the focused and unfocused states of a <a>User Interface Component</a></p>   				
+	            <p>The pixels that are changed to visually indicate when a user interface component is in a focused state.</p>
```
awkawk, line 35: Again, for consistency:

 <p>pixels that are changed to visually indicate when a user interface component is in a focused state</p>

understanding/22/focus-appearance-minimum.html
```diff
@@ -116,7 +117,7 @@ <h4>Unusual shapes and gradients</h4>
                     <figcaption>A circular button in the default state, and then with a focus indicator.</figcaption>
                 </figure>
 
-                <p>If a focus indicator is an irregular shape, such as a drop-shadow under a star icon, it helps if the contrasting area is obviously quite large.</p>
+                <p>If a focus indicator is an irregular shape, such as a drop-shadow under a star icon, it obviously helps if the contrasting area is quite large.</p>
```
awkawk, line 120: suggest removing "obviously"

mbgower:
I added some minor wording updates from comments in https://www.w3.org/2020/11/24-ag-minutes.html#item02
I do not believe these need to be reviewed by group, as they are discussed within confines of the topic, and seem to fit within the resolution.
@alastc I recommend this be implemented, as per the resolution.

understanding/22/focus-appearance-minimum.html
mbgower: adopted.

alastc:
Finally got around to resolving the conflict and merging.

PR #1505, opened 2020-11-03
Updates from 1418

### LINKED ISSUE #1352: Non-modal dialogues and Unobscured 2.4.11: Focus Appearance
non-modal dialogues may be movable by users (dragged to a unknown place on screen) and thus not clear for an author where they are on the page.

The "Unobscured" component (non-modal) should not be applicable in this case as it is not possible for authors (or very hard...) to prevent focussed items to appear under such non-modal dialogs.

### LINKED ISSUE #1418: 3.3.7 Accessible Authentication - Add mechanism text
I've reviewed Accessible Authentication with Adobe's security team and the team that manages our centralized login experience.

They shared concerned with this SC in that while WebAuthn is great and passwordless is the direction that we intend to be moving over the coming years, they did a review in FY19-Q4 and found that the browser support is not quite where it needs to be to depend on it yet.

They also indicated that even where passwordless authentication is in place, there is still a fallback to traditional username/password types of authentication, so we need to make sure that it is very clear that username / password is ok.

We would like to make sure that the information about passwords being ok as long as the password managers can support the end user. My suggestion is to modify the SC from:

> If an authentication process relies on a cognitive function test, at least one other method must also be available that does not rely on a cognitive function test.

to

If an authentication process relies on a cognitive function test, at least one other @@authentication@@ method must also be available that does not rely on a cognitive function test or a mechanism must be available to assist the user in completing the cognitive function test.

Note: Support for password entry by password managers and copy and paste are examples of mechanisms which can be used to help address memorization and transcription cognitive function tests.

alastc:
Closes #1418

### DIFF
--- a/guidelines/sc/22/accessible-authentication.html
+++ b/guidelines/sc/22/accessible-authentication.html
@@ -6,6 +6,6 @@
 	<p class="change">New</p>
     
     <p>For each step in an authentication process that relies on a <a>cognitive function test</a>, at least one other authentication method is available that does not rely on a cognitive function test, or a mechanism is available to assist the user in completing the cognitive function test.</p>
-    <p class="note">Examples of mechanisms include support for password entry by password managers to address the memorization cognitive function test and copy-paste to help address transcription cognitive function test.</p>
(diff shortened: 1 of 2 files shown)


alastc:
Approved in the meeting today.

Issue #1506, opened 2020-11-04
I think the forward slash is missing


Sylvie123:
https://www.w3.org/WAI/WCAG21/Techniques/html/H65.html
I'm just learning and was wondering if the space at the end of each input needs to be removed or add a forward slash.
<input id="areaCode" name="areaCode" title="Area Code" type="text" size="3" value="" >
<input id="exchange" name="exchange" title="First three digits of phone number" type="text" size="3" value="" >
<input id="lastDigits" name="lastDigits" title="Last four digits of phone number" type="text" size="4" value="" >

patrickhlauke:
it doesn't do any harm having that space there...just looks a bit messy. and no, slash (to make it self-closing) is not needed unless you're being strict about xhtml (which really isn't a thing anymore)

PR #1507, opened 2020-11-04
Trivial: remove unnecessary space before tag closing bracket

patrickhlauke:
Code cleanup, nothing normative. Not high priority, of course, but would be good to have clean markup

Closes https://github.com/w3c/wcag/issues/1506

### DIFF
--- a/wcag20/sources/quickref.xslt
+++ b/wcag20/sources/quickref.xslt
@@ -192,7 +192,7 @@ script type="text/javascript" src="annotate.js" xmlns="http://www.w3.org/1999/xh
       <xsl:if test="$show.techniques != '0'">
         <!-- Sufficient Techniques -->
                <xsl:processing-instruction name="php"><![CDATA[ if($bSufficient) { ]]></xsl:processing-instruction>   
-        <div class="sufficient" xmlns="http://www.w3.org/1999/xhtml" >
+        <div class="sufficient" xmlns="http://www.w3.org/1999/xhtml">
         <xsl:apply-templates select="$guide-src//spec/body/div1/div2[@id=$anchor2]/div3[@role='techniques']/div4[@role='sufficient']"/>
         </div>
         <xsl:processing-instruction name="php"><![CDATA[ } ]]></xsl:processing-instruction> 
@@ -652,7 +652,7 @@ script type="text/javascript" src="annotate.js" xmlns="http://www.w3.org/1999/xh
     </xsl:for-each>
   </xsl:template>
   <xsl:template match="p[@role='i'] | p[@role='v']">
-    <p  >
+    <p>
       <xsl:if test="@id">
         <xsl:attribute name="id"><xsl:value-of select="@id"/></xsl:attribute>
       </xsl:if>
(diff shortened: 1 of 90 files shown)


wcag20/sources/quickref.xslt
```diff
@@ -652,7 +652,7 @@ script type="text/javascript" src="annotate.js" xmlns="http://www.w3.org/1999/xh
     </xsl:for-each>
   </xsl:template>
   <xsl:template match="p[@role='i'] | p[@role='v']">
-    <p  >
+    <p>
       <xsl:if test="@id">
         <xsl:attribute name="id"><xsl:value-of select="@id"/></xsl:attribute>
```
patrickhlauke, line 657: i'm assuming/hoping that in this case the extra space wasn't actually needed to avoid the xsl operation just cramming in that attribute without space, i.e. `<pname="...">`

alastc:
I'm fairly sure that anything in the .xml files is not used anymore, there is no intent to update the WCAG 2.0 versions now. On the other hand, seeing as we won't be using them I don't think it is worth spending time undoing that in the PR.

Anything in the XSL is potentially dangerous, purely because XSL is weird and I don't like touching it!

Michael, if that XSL example if fine, I'm happy to merge and close.

patrickhlauke:
@michael-n-cooper ping

michael-n-cooper (commented):
It's hard for me to approve this for a few reasons:

* There are 90 files changed in one commit and it's impractical to review them all;
* On a spot-check, many of the changes remove spaces at the end of self-closing elements, which are a common convention in XML-compatible HTML, which this is, and would be likely to crop up again.
* Other changes remove spaces at end of opening tags of elements. While those spaces probably cropped up as a result of a global find-replace long ago, they don't cause harm.

Given that I don't see what real problems this solves, I can see the likelihood of the same thing reappearing, and I can't review it comprehensively, I can't submit an approving review at this time.

patrickhlauke:
> On a spot-check, many of the changes remove spaces at the end of self-closing elements, which are a common convention in XML-compatible HTML, which this is, and would be likely to crop up again.

just on that point, nowhere does this PR remove the space before an actual self-closing slash. this only removes the space from ` >` (space and closing angle bracket), which i don't believe is a convention (at least not one I've heard of).

beyond that sure, these extra spaces don't have a material effect. it's more about code quality/consistency, hence why this is a fairly trivial PR. but yeah, take it or leave it, won't make any difference.

patrickhlauke:
on the "reviewing 90 file changes" aspect, github's files view should be able to help quite easily here to show the diff, no? https://github.com/w3c/wcag/pull/1507/files

alastc:
@michael-n-cooper - I've reviewed the XML & HTML files, that's fine. 

Could you just check the 4 XSL files? 
- [XSL 1-3](https://github.com/w3c/wcag/pull/1507/files#diff-7930e655bf73adc22274d7bbb6d796935ca7846b09e129743adabc20ecf9b7ae)
- [XSL 4](https://github.com/w3c/wcag/pull/1507/files#diff-039bcf31eb181f50697dd5d8696b9b8e2382af9efff3d37339e269c2a852a93f)

It looks fine to me. As there's no merge conflict I'm inclined to hit the 'merge' button.

patrickhlauke:
Abandoning this, as it's primarily about the old/outdated xml files

PR #1526, opened 2020-11-13
Replace "ACT Rules" with "Test Rules"

nitedog:
Replace "ACT Rules" with "Test Rules" for easier flow and to match the description in the Techniques that has been agreed on.


maryjom (approved):
These changes look good to me.

Issue #1548, opened 2020-11-30
SC 1.4.8 Visual Presentation - Examples Seem to be Missing Images


Grunet:
The understanding page for SC 1.4.8 Visual Presentation (https://w3c.github.io/wcag/understanding/visual-presentation.html) has an ["Examples" section](https://w3c.github.io/wcag/understanding/visual-presentation.html#examples) that includes the words "The following images", but there aren't any images that come afterwards (seems like they should probably be ones of text spacing in paragraphs?).

Here's a screenshot of that section
![image](https://user-images.githubusercontent.com/29322882/100566197-00313580-328b-11eb-8454-3ed0890ef20a.png)

And here's the html of that section
```html
<section id="examples">
   <h2>Examples</h2>
   <p>The following images show examples of single-spacing, space-and-a-half and double-spaced text in a paragraph.</p>
   <p>Examples of glyphs include  "A", "→" (an arrow symbol), and "さ" (a Japanese character).</p>
</section>
```

JAWS-test:
In the [WCAG 2.0](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-visual-presentation.html#visual-audio-contrast-visual-presentation-examples-head) there is still the image, but not in [2.1](https://www.w3.org/WAI/WCAG21/Understanding/visual-presentation.html#examples)

PR #1549, opened 2020-11-30
Fix understanding images

patrickhlauke:
- add images that went missing for 1.4.8 visual presentation (see https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-visual-presentation.html#visual-audio-contrast-visual-presentation-examples-head)
- add images for 1.4.12 text spacing directly into the repo (rather than pointing the image `src` to the wiki)

Closes https://github.com/w3c/wcag/issues/1548


alastc:
Thanks for that, must have been missed during a template update.

@michael-n-cooper could you just check the figure aspect? I assumed the images wouldn't come through without a figure, and that having 3 in 1 is ok?

patrickhlauke:
@michael-n-cooper ping

patrickhlauke:
@michael-n-cooper ?

michael-n-cooper (approved):

PR #1553, opened 2020-12-03
G183: Removing hover/focus steps from procedure

### LINKED ISSUE #201: Technique G183 (and note 1 in F73) provide loophole resulting in inaccessible content (was: Technique G183 not applicable to touch/inputs that lack hover/focus)
Currently technique G183 https://www.w3.org/TR/WCAG20-TECHS/G183.html can be used as a "loophole" to stick with lower color contrast requirement for things like links in regular text, under the assumption that a user who has trouble discerning the 3:1 contrast can always set focus (with the keyboard) or hover (with the mouse) over the dubious item and get visual confirmation that it is indeed a link.

I would say the technique requires an additional note/warning that the technique itself cannot be relied upon for situations where a user may be using a touchscreen (or similar inputs that lack the concept of hover/focus).

Generally, touchscreens don't have a concept of "hover" (though there are a few experimental touchscreens that do recognise a non-touching/hovering finger, these are certainly not the norm - this also applies to stylus/pen type interfaces, which don't always sense a hovering stylus/pen). A user is either touching or not touching the screen. Additionally, while most browsers will react to a touch by then setting the focus (firing the `focus` event in JS, applying the `:focus` styles), this is done as part of the actual activation - i.e. the focus is applied, but the browser is already busy following the link (so either the user DID work out, despite the 3:1 ratio, that this was in fact a link, OR the user accidentally tapped it, NOT knowing it was a link, by which point it's already too later anyway as the link is being followed).

### LINKED ISSUE #1272: Discrepancy between F73 and G183
G183 says links in text pass if they have a contrast of 3:1 or more to the surrounding text, and a distinct hover and focus style.

F73 says links in text fail if they have a contrast of less than 3:1 with it surrounding text.

Notice that F73 says nothing about whether or not a hover / focus style is required. This creates a gap. On the one hand you could argue that just because F73 doesn't mention it, doesn't mean it passes. Alternatively you could argue that because G183 is a sufficient technique that just because you didn't do everything it says, doesn't mean you've failed the SC.

It isn't clear which way to go, because even suggesting that 1.4.1 can be passed with color contrast (with or without hover/focus) is not based on any normative language (as [Patrick pointed out here](https://github.com/w3c/wcag/issues/1118#issuecomment-623693442)). Guy's suggestion in #1118 to delete G183 would address this issue, but if we're not deleting it, at least we should align these two SCs.

alastc:
Closes #1272
Related to #201


yatil:
So this PR means that it is now OK to have a hardly visible link that even does not change on hover/focus… I guess that is only consequential from PR #1500 and the discussion around using this, but it is still disappointing to me.

alastc:
There are two necessary reasons for this:
1. Having an SC for non-text contrast does not work unless you are differentiated colours by contrast. I.e. find an icon that is not using colour do differentiate the icon from the background.
2. There would be too many false positives, i.e. examples where people (including with low vision) can differentiate links by 'colour' (in the general sense). 

So yes, it is possible to create very difficult examples, but the opposite is also true so the metric is not enough to make it binding.

understanding/20/use-of-color.html
```diff
@@ -36,11 +36,12 @@ <h2>Intent of Use of Color</h2>
          of prompting a response would be: using highlighting on form fields to indicate that
          a required field had been left blank.
       </p>
+
+      <p>Contrast between colors can be considered an alterantive method of conveying information.  Color (generally defined as combination of hue, saturation and luminance) can be differentiated if there is a sufficient difference in contrast. If the contrast ratio (<a>relative luminance</a>) between the colors differ by at least 3:1, or if there is another method of conveying that information it would pass this criterion. For example, a text link that only differs from adjacent text using color and there is no other visual indication that the text is linked (e.g. the link underline is removed) needs to ensure that the link color meets the 3:1 luminosity contrast ratio relative to the non-linked text color in order to meet this Success Criteria.</p>>
```
patrickhlauke, line 40: "alterantive" > "alternative"

i'd add commas and a "where"

"only differs from adjacent text using color and there is no other visual indication that the text is linked (e.g. the link underline is removed) needs to ensure" > "only differs from adjacent text using color*,* and *where* there is no other visual indication that the text is linked (e.g. the link underline is removed)*,* needs to ensure"

understanding/20/use-of-color.html
```diff
@@ -36,11 +36,12 @@ <h2>Intent of Use of Color</h2>
...
       
       <div class="note">
          
-         <p>This should not in any way discourage the use of color on a page, or even color coding
-            if it is redundant with other visual indication.
+         <p>This should not in any way discourage the use of color on a page, or even color coding if it is redundant with other visual indication.
```
patrickhlauke, line 44: i'd change "redundant with" to "complemented by"

patrickhlauke (approved):

patrickhlauke:
Suggest tweaking the title of this issue to also indicate it touches on the understanding doc itself

x-ref also #1500

techniques/general/G183.html
```diff
@@ -1,11 +1,11 @@
...
    </section><section id="description"><h2>Description</h2>
-      <p>The intent of this technique is to provide a redundant visual cue for users who may not be able to discern a difference in text color. Color is commonly used to indicate words that are links within a paragraph or other block of text. For example, scattered words in text may be hypertext links that are identified only by a difference in color with surrounding text. This technique describes a way to provide additional cues on hover and focus so that users who may have difficulty perceiving color differences or have low vision can identify them.</p>
+      <p>The intent of this technique is to provide a redundant visual cue for users who may not be able to discern a difference in text color. Color (generally defined as combination of hue, saturation and luminance) is commonly used to indicate words that are links within a paragraph or other block of text. For example, scattered words in text may be hypertext links that are identified only by a difference in color with surrounding text. This technique describes a way to provide a difference in contrast rather than relying on hue.</p>
```
patrickhlauke, line 4: "defined as combination of hue" > "defined as *a* combination of hue"

understanding/20/use-of-color.html
alastc: As I went through the other issues, I think the explanation is covered in the changes from #1500, so I'll remove that.

patrickhlauke (approved):

patrickhlauke:
What was the state of this? Will it be merged, or does it need further discussion?

patrickhlauke:
@alastc so now that you've added the step back in ... are we saying that the 3:1 ratio is not sufficient?

"For this technique, other visual highlights are also required"

so does it pass or not without the extra highlighting? this is now contradictory

techniques/general/G183.html
```diff
@@ -28,11 +32,12 @@ <h3>Colors that would provide 3:1 contrast with black words and 4.5:1 contrast w
          <ol>
             <li>Locate all instances where color alone is used to convey information about text.</li>
             <li>Check that the <a>relative luminance</a> of the color of the text differs from the relative luminance of the surrounding text by a contrast ratio of at least 3:1.</li>
+            <li>Check that pointing (mouseover) to the link causes a visual enhancement (such as an underline, font change, etc.)</li>
```
patrickhlauke, line 35: "pointing (mouseover) to the link" ... why not hovering? don't think anybody does anything explicitly on `mouseover` from a JS perspective, it's usually `:hover` from CSS

patrickhlauke:
if it's more a case of providing an additional cue as a technique that goes beyond the normative requirement, then i'd suggest not using the word "required" in the phrase there. instead, making it clear that although the 3:1 ratio is sufficient to pass the minimum requirement of use of color, this technique suggests adding an additional highlight/cue that goes beyond the normative requirement, for increased usability

patrickhlauke:
👍

yatil:
Super concerned about adding back `hover`/`focus` into what is supposed to be a sufficient technique. `hover`/`focus` are not needed for meeting 1.4.1 (as we have now – sadly – determined), so sufficient techniques should not include further advisory advice.

Consider removing `hover`/`focus` from this technique and instead creating an advisory technique that advices on using `hover`/`focus` in addition to the sufficient technique.

patrickhlauke:
agree with @yatil this should be an advisory technique. the sufficient part is the 3:1 contrast, the rest is window dressing.

may also be good to add a note explicitly clarifying that while this will be of benefit to mouse users, it won't help on touchscreen or stylus devices where there is no concept of hover (well, the vast majority of stylus devices ... some fancy ones do have hover stylus support, like Samsung Note ones)

techniques/general/G183.html
```diff
@@ -29,16 +31,13 @@ <h3>Colors that would provide 3:1 contrast with black words and 4.5:1 contrast w
       <section class="procedure"><h3>Procedure</h3>
          <ol>
             <li>Locate all instances where color alone is used to convey information about text.</li>
```
awkawk, line 33: This should probably be in a paragraph before the list: 
<p>For each instance where color is used to convey information about text:</p>

This will also affect the "checks #2 and #3 are true"

techniques/general/G183.html
```diff
@@ -1,11 +1,13 @@
-<!DOCTYPE html><html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml"><head><title>Using a contrast ratio of 3:1 with surrounding text and providing additional visual cues on focus for links or controls where color alone is used to identify them</title><link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove"></link></head><body><h1>Using a contrast ratio of 3:1 with surrounding text and providing additional visual cues on focus for links or controls where color alone is used to identify them</h1><section class="meta"><p class="id">ID: G183</p><p class="technology">Technology: general</p><p class="type">Type: Technique</p></section><section id="applicability"><h2>When to Use</h2>
+<!DOCTYPE html><html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml"><head>
+   <title>Using a contrast ratio of 3:1 with surrounding text and providing additional visual cues on hover for links or controls where color alone is used to identify them</title>
```
awkawk, line 2: This would require a change in the title also

techniques/general/G183.html
```diff
@@ -1,11 +1,13 @@
...
+   <link rel="stylesheet" type="text/css" href="../../css/editors.css" class="remove"></link>
+</head>
+<body><h1>Using a contrast ratio of 3:1 with surrounding text and providing additional visual cues on hover for links or controls where color alone is used to identify them</h1><section class="meta"><p class="id">ID: G183</p><p class="technology">Technology: general</p><p class="type">Type: Technique</p></section><section id="applicability"><h2>When to Use</h2>
```
awkawk, line 5: And a change in the H1

techniques/general/G183.html
```diff
@@ -1,11 +1,13 @@
...
+      <p>The intent of this technique is to provide a redundant visual cue for users who may not be able to discern a difference in text color. Color (generally defined as a combination of hue, saturation and luminance) is commonly used to indicate words that are links within a paragraph or other block of text. For example, scattered words in text may be hypertext links that are identified only by a difference in color with surrounding text. This technique describes a way to provide a difference in contrast rather than relying on hue.</p>
+      <p>To meet success criterion 1.4.1: Use of Color a <a>relative luminance</a> (lightness) difference of 3:1 or greater with the text around can be used. This technique goes beyond the success criterion and asks for visual highlights when the user hovers over each link such as an underline, a change in font style such as bold or italics, or an increase in font size.</p>
```
awkawk, line 9: This technique no longer goes beyond the SC, so the sentence needs to be changed.

mbgower:
@alastc I've added in a modification which I think addresses my concern. I didn't put in a link to 2.4.7, as I wasn't sure on the syntax.

alastc:
@yatil and @patrickhlauke, we discussed this in the meeting, and there's a key point that sufficient techniques can go beyond the minimum requirements. 

The description now includes:
> To meet success criterion 1.4.1: Use of Color a relative luminance (lightness) difference of 3:1 or greater with the text around can be used. This technique goes beyond the success criterion and...

We also discussed removing the technique entirely, but decided not to as it helps to have sufficient techniques, and this one in particular helped the SC overcome objections originally.

patrickhlauke:
ah, shamefully i have to admit i was unaware of the exact definition of advisory techniques (that, in essence, they may not even address the actual ask of an SC). somehow got it in my head that advisory was "sufficient to pass the SC, plus something extra on top of that". so i can see how this is a sufficient technique, but not strictly the "minimum sufficient technique" as it incorporates extra stuff.

long way of saying yup, agree (now that i delved into the actual definitions of advisory vs sufficient)

yatil:
Sometimes I just wished we would use words in a way that they are used commonly outside of our accessibility bubble where sufficient does not mean going beyond. This conflating of meaning is a constant struggle, especially for people learning to differentiate between best practices and absolute requirements. Having Pass/Fail requirements but muddying the water in examples does not really help learning.

PR #1555, opened 2020-12-07
Update for #1535, motion animation

### LINKED ISSUE #1535: "Changes of color" applied to motion definition
Since the new focus indicator SCs discussions the last couple of years, every addition / change in pixels is seen as "A change of color"

@alastc introduced this view if I remember well (for the 'addition of focus indicators on the outside of a component as an example), and after a couple of seconds thinking of this it seemed logical and clear.

Introducing this insight to others in training, EVERYONE is a bit surprised at first with a shock-and-awe effect but also get it pretty fast.

NOW...

In https://www.w3.org/WAI/WCAG22/Understanding/animation-from-interactions.html / 2.3.3: Animation from Interactions there is a definition for:

> motion animation
> addition of steps between conditions to create the illusion of movement or to give a sense of a smooth transition
> 
> For example, an element which moves into place or changes size while appearing is considered to be animated. An element which appears instantly without transitioning is not using animation. Motion animation does not include changes of color, blurring or opacity.

Which clearly states **"Motion animation does not include changes of color"**

It also states in the NOTE:

> For example, an element which moves into place or changes size while appearing is considered to be animated.

Fundamentally THIS ALSO IS A CHANGE OF COLOR as they are change of pixels.

To confirm this we have the following in the Understanding of 2.4.11: Focus Appearance (Minimum):
https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance-minimum#change-of-color

Where it states:

> When a component changes to include a focus indicator, that change can always be measured as a change of color contrast. For example, if a yellow outline is added to a button on a blue background, the change of color is from blue to yellow.

Now this creates a contradiction in the approach, although fully understandable it's not consistent.

Changes in pixels ARE or ARE NOT a change in color and we need to set a clear line here in when we mean what.

alastc:
Closes #1535

### DIFF
--- a/guidelines/terms/21/motion-animation.html
+++ b/guidelines/terms/21/motion-animation.html
@@ -2,6 +2,6 @@
 <dd>
    					
    <p>addition of steps between conditions to create the illusion of movement or to give a sense of a smooth transition</p>
-	<p class="example">For example, an element which moves into place or changes size while appearing is considered to be animated. An element which appears instantly without transitioning is not using animation. Motion animation does not include changes in an element's color, blurring or opacity, unless used to change the perceived size, shape, or position of the element.</p>
+	<p class="example">For example, an element which moves into place or changes size while appearing is considered to be animated. An element which appears instantly without transitioning is not using animation. Motion animation does not include changes of color, blurring, or opacity which do not change the perceived size, shape, or position of the element.</p>
    				
 </dd>
--- a/guidelines/terms/21/motion-animation.html
+++ b/guidelines/terms/21/motion-animation.html
@@ -2,6 +2,6 @@
 <dd>
    					
    <p>addition of steps between conditions to create the illusion of movement or to give a sense of a smooth transition</p>
-	<p class="example">For example, an element which moves into place or changes size while appearing is considered to be animated. An element which appears instantly without transitioning is not using animation. Motion animation does not include changes of color, blurring or opacity which do not change the perceived size, shape, or position of the element.</p>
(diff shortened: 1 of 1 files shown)


guidelines/terms/21/motion-animation.html
```diff
@@ -2,6 +2,6 @@
 <dd>
    					
    <p>addition of steps between conditions to create the illusion of movement or to give a sense of a smooth transition</p>
-	<p class="example">For example, an element which moves into place or changes size while appearing is considered to be animated. An element which appears instantly without transitioning is not using animation. Motion animation does not include changes of color, blurring or opacity.</p>
+	<p class="example">For example, an element which moves into place or changes size while appearing is considered to be animated. An element which appears instantly without transitioning is not using animation. Motion animation does not include changes of color, blurring or opacity which do not change the perceived size, shape, or position of the element.</p>
```
patrickhlauke, line 5: per my additional concern in https://github.com/w3c/wcag/issues/1535#issuecomment-737906877 would it be worth also changing/expanding

> moves into place or changes size while ...

to

> moves into place or changes size or shape while ...

(as "size" may remain the same - in terms of area - but proportions/shape may morph)

alastc: The first part is 'for example' so can include more. Also, the context is 'whilst appearing', so changing shape whilst appearing seems less common than changing size (e.g. expanding from an icon to full size).

patrickhlauke: would just be nice to have it at least mentioned in passing (for things that do subtle stretchy animation things that do maintain size, but futz around with aspect ratio for instance), but yeh...not too fussed

alastc:
Slightly updated to remove the double-negative, and approved: https://www.w3.org/2021/06/29-ag-minutes.html#item07

Issue #985, opened 2019-12-05
Can text buttons fail 1.1.1 and 3.3.2?


WilcoFiers:
This came up during the the ACT TF call this week. We were looking at a rule that checks if buttons have an accessible name. It was clear buttons are tested under 4.1.2, but the group seemed split about whether or not 1.1.1 and/or 3.3.2 can fail for something like this:

```
<button type="button" />
```

## 1.1.1 Non-text content

For 1.1.1 the argument was that buttons may be considered non-text content. They may contain text, but aren't themselves text. They can also be considered "Controls, Input". Technique H44 suggests that things such as select, textarea and text input are non-text content. On the other hand, the understanding document mentions this:

> *For non-text content that is a control or accepts user input*, such as images used as submit buttons, image maps or complex animations, a name is provided to describe the purpose of the non-text content so that the person at least knows what the non-text content is and why it is there.

While it doesn't outright say buttons without images don't apply, that it mentions images / animation in all three example suggests that might be the intention. The ACT Task Force would like to know what the position of the AGWG. Are buttons without images still considered non-text content? And perhaps in extension to this, should form fields be considered non-text content?

## 3.3.2 Labels or Instructions

For the same button, would the lack of a visible label be considered a failure of 3.3.2? Since there is no definition for "requires user input". Clicking a button is (usually) required for user input to be submitted, however (usually) buttons themselves aren't used as a mechanism for the user to provide information.

Either way it would be helpful to include information about how 3.3.2 relates to buttons in the understanding document.

mraccess77:
I will say that a label can be an icon and 3.3.2 applies to visual labels (icons, text, or even another control that acts as a label).  So the situation you are talking about is a button with no visual label text or icon or other associated control - this seems like an issue for all users.

In regards to whether 1.1.1 applies to the non-text portion of things like buttons and input fields -- I asked a similar question in #929 https://github.com/w3c/wcag/issues/929

awkawk:
My interpretation of 1.1.1 is that for controls 4.1.2 handles the requirement for a name. We should look at H44 to make that more clear. 4.1.2 handles the "what kind of thing is this and what is it called?" questions and I don't think that 1.1.1 has anything to do. I wish 1.1.1 said "If non-text content is a control or accepts user input, refer to Success Criterion 4.1.2 for additional requirements for controls and content that accepts user input." as this would be more clear.

[...]

PR #1568, opened 2020-12-26
Expand 3.3.2 Labels or Instructions understanding

patrickhlauke:
Based on discussion in https://github.com/w3c/wcag/issues/985#issuecomment-562504332 this expands the intent given for 3.3.2 to cover buttons involved in form submission/proceeding to the next step. It also covers radio buttons, checkboxes, selects etc which previously weren't clearly covered (though I believe in practice people considered those as part of 3.3.2).
Includes an explanation that links and "immediate action" type buttons like disclosure widget buttons don't necessarily represent "user input" (though this seems to be a controversial point?)

### DIFF
--- a/understanding/20/labels-or-instructions.html
+++ b/understanding/20/labels-or-instructions.html
@@ -45,8 +45,8 @@
          and therefore pass Success Criterion 4.1.2, but to still fail this Success Criterion (if the labels or instructions
          aren't presented to all users, not just those using assistive technologies).
       </p>
-      <p>Lastly, this Success Criterion does not apply to links or other controls (such as controls for disclosure widgets
-         or similar interactive components) that are not associated with data entry.
+      <p>This Success Criterion does not apply to links or other controls (such as an expand/collapse widget, or similar
+         interactive components) that are not associated with data entry.
       </p>
       <p>While this Success Criterion requires that controls and inputs for data entry and submission have labels, whether or not
          these labels are sufficiently clear or descriptive is covered separately by
--- a/understanding/20/labels-or-instructions.html
+++ b/understanding/20/labels-or-instructions.html
@@ -16,8 +16,7 @@
          or labels that identify the controls in a form so that users know what input data
          is expected. In the case of radio buttons, checkboxes, comboboxes, or similar controls
          that provide users with options, each option must have an appropriate label so that
-         users know what they are actually selecting. Lastly, controls that submit the user input,
-         or proceed to the next step in a multi-step process, must also provide a label.
+         users know what they are actually selecting. 
          Instructions or labels may also specify data formats for data entry fields, especially
          if they are out of the customary formats or if there are specific rules for correct
          input. Content authors may also choose to make such instructions available to users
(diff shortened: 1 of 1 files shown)


detlevhfischer:
It is good to see buttons (or links to next step) more assertively included in the scope of 3.2.2. I am just not entirely sure that the chosen wording does not get a bit too specific:
> each option must have an appropriate label so that
         users know what they are actually selecting. Lastly, controls that submit the user input,
         or proceed to the next step in a multi-step process, must also provide a label.

"appropriate" seems a bit vague. Also, with radio and checkbox inputs it is not rare to have both a legend (or a group label) and the label of the control together describe input options. Even an additional description may suffice from a normative point of view.
Lastly, regarding controls that submit user input (or proceed to next step): do they not always provide a label of some kind to be detectable / usable at all? Do we need to require them to be 'descriptive'? Would a right-facing triangle in a button with aria-label="next" cut it?
I should have a suggestion for rewording, but I can't think of one right now... I can certainly live with these changes and think they close a gap, I just feel the language can still be honed a bit.

patrickhlauke:
for sure, this was a first stab. happy to tweak further. i was just following up on various issues that i earmarked as interesting gaps that should be addressed at some point, and it's easier to progress them when there's at least one proof of concept PR to kick things off 👍

patrickhlauke:
any more thoughts on this?

bruce-usab:
Just picayune word smithing:

> ... or other controls (such as controls 

How about:

This Success Criterion does not apply to links or other controls (such as an expand/collapse widget, or similar interactive components) that are not associated with data entry.

patrickhlauke:
good call. also, "picayune" tickled me there :+1:

patrickhlauke:
I note from yesterday's AGWG meeting that there were concerns/questions about why links, disclosure widget triggers, etc were explicitly noted as not falling under the SC, and a request to remove that particular part.

I don't agree, as that part comes directly from how the normative SC is scoped: "Labels or instructions are provided *when content requires user input*." - links don't require user input. disclosure widget triggers don't require user input. Aspects such as "every control needs an accessible name" etc are already covered by other SCs like 4.1.2.

alastc:
This was approved today, after amending to remove "Lastly, controls that submit the user input, or proceed to the next step in a multi-step process, must also provide a label."

Wilco's point was that, whilst true, it comes under a different SC so doesn't need to say that here.

mraccess77:
I misunderstood the resolution on today's call.  I didn't know it included removing that bit - that was not clear.  I voted on the PR as was written.  SC 2.4.6 could be read the headings and labels only need to be descriptive - not that you have to provide them.

patrickhlauke:
> SC 2.4.6 could be read the headings and labels only need to be descriptive - not that you have to provide them.

indeed, the intent for 2.4.6 actually states https://www.w3.org/WAI/WCAG22/Understanding/headings-and-labels.html

> This Success Criterion does not require headings or labels. This Success Criterion requires that if headings or labels are provided, they be descriptive.

Issue #747, opened 2019-05-20
Benefits 2.5.1: split benefit and remove redundant one (for other SC)


jake-abma:
https://www.w3.org/WAI/WCAG21/Understanding/pointer-gestures.html#benefits

> Benefits
> - Users who cannot (accurately) perform complex pointer gestures will have alternative means for operating the content.
> - The availability of user interface elements provided as a fallback for complex gestures helps users who may often be unaware of the support for complex pointer gestures. This can be beneficial especially for users with cognitive or learning disabilities.

Benefit 1: Fine, covers the SC text but would expect "multipoint" and "path-based" to be separated into their own benefit for clarity

Benefit 2: this one may be a side effect but is not what the intent is about, keeping focus on the SC itself and not all side effects I prefer to see this one at another SC, not here. It may be confusing and have people give wrong comprehension of the SC. This benefit would be fine for the 2.2 "Indication of Gestures and Movement". The intent text is also not mentioning "user interface elements (in text)" as the mandatory alternative and the SC is not focussed at: "beneficial especially for users with cognitive or learning disabilities".

patrickhlauke:
Agree about Benefit 2. It could still be kept around, but prefixed clearly as being a side effect, and not the goal of the SC

patrickhlauke:
Made a slightly wider PR that covers this, as well as a few other niggles. See https://github.com/w3c/wcag/pull/1569

jake-abma:
eager to see this on the agenda for discussion...

PR #1569, opened 2020-12-27
Changes to 2.5.1 Pointer Gestures understanding

patrickhlauke:
* Add extra example that does not involve adding an actual button/control, but provides a single-pointer alternative to a multi-pointer gesture (basically what Google Maps does)
* Expand/clarify the slider/carousel example to make clear it's a flick/swipe, not a drag/swipe (which *is* permitted)
* Split the benefit to better illustrate how addressing path-based gestures, *and* multi-pointer gestures, can benefit users
* Remove the indirect benefit from adding extra buttons/controls, as the ask of the SC is not to always add extra buttons

The last two points close https://github.com/w3c/wcag/issues/747


detlevhfischer:
Looks good to me!

mbgower:
@alastc and @patrickhlauke 
Added in some slight changes in a [new commit](https://github.com/w3c/wcag/pull/1569/commits/8059364fc66340ea622861d158b909774d163071) based on today's survey and discussion, as directed.

patrickhlauke:
@mbgower works for me

alastc:
Approved: https://www.w3.org/2021/06/01-ag-minutes.html#t07

PR #1570, opened 2020-12-27
Add clarification in 2.4.4 and 2.4.9 understanding that link text is/includes accessible name

### LINKED ISSUE #1202: Links ambiguous to AT users only can pass 2.4.4 and 2.4.9
> Success Criterion 2.4.4 Link Purpose (In Context) (Level A): The purpose of each link can be determined from the link text alone or from the link text together with its programmatically determined link context, except where the purpose of the link would be ambiguous to users in general.
https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html

> Success Criterion 2.4.9 Link Purpose (Link Only) (Level AAA): A mechanism is available to allow the purpose of each link to be identified from link text alone, except where the purpose of the link would be ambiguous to users in general.
https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-link-only

Since `aria-label` and `aria-labelledby` completely replace the visible link text you can get this situation:

```
<a href="http://facebook.com" aria-label="Follow us">Facebook</a>
<a href="http://twitter.com" aria-label="Follow us">Twitter</a>
```

The purpose of both links can be determined from the link text alone, but in AT both links are voiced as "Follow us"

patrickhlauke:
- tweak the intent of 2.4.4 to make it more heavily lean towards AT users (as non-AT users will not benefit from anything that's purely reliant on "programmatically determinable", and since that is one of the ways to pass 2.4.4, they may be left out by perfectly valid/conformant solutions)
- remove benefits for non-AT users for this reason as well
- clarify that the front-loading of information in a paragraph is also just a suggestion/best practice
- reformatting of some of the existing content for consistency
- removed the odd "html"/"pdf"/"mp3" example as that didn't, at least in its current form, seem to actually address this SC
- removed the example about the image with null alt inside the link, as it seems orthogonal to what the SC tries to address

Closes https://github.com/w3c/wcag/issues/1202


detlevhfischer:
I generally support this clarification - I just wonder whether it might be confusing to talk about the links 'label"'. Following the dfn,
 
> label
text or other component with a text alternative that is presented to a user to identify a component within Web content

...it looks like the label is a separate element labelling a component, whereas the link **text** directly represents the component. So I would prefer avoiding the term 'label' here.

patrickhlauke:
fair point, i'll amend it

dd8:
> /cc @dd8

I thinks that's the best that can be done without changing the normative text. It's unfortunate that 'link text' means different things in two related rules, because this will probably lead to ongoing confusion even with the clarification.

Hopefully this can be fixed in WCAG 3. 

If definitions are changing I'd also suggest changing 'label' to something else to avoid confusion with the `label` element and, especially, `aria-label`. I've seen people waste a lot of time trying to fix WCAG label issues with `aria-label` (because it changes the WCAG name, and not the WCAG label)

bruce-usab:
@dd8 could you clarify your assertion that:
 > It's unfortunate that 'link text' means different things in two related rules...

2.4.4 mentions <q>link text together with its programmatically determined link context</q> whereas both 2.4.4 and 2.4.9 mention <q>link text alone</q>.  I don't agree that 'link text' means different things in two related rules, so I want to be careful that I am not missing something here.

patrickhlauke:
Could this be put to discussion/vote @alastc ?

scottaohara (approved):

bruce-usab:
@scottaohara could you be prepared to wall the TF group through this PR on _next_ Friday's call (8 March)?  As is our pattern, allastair or michael would be screen sharing.

scottaohara (approved):

dav-idc:
On this GitHub issue, there were some questions brought up on the March 8 WCAG 2.X backlog call.

Mainly, a question about whether '2.4.9 Link Purpose (Link Only)' explicitly needs the 'text' to be exclusively visible non-image text, or if it can be a text alternative.

In that meeting I had said that my assumption was that link text could only be visible non-image text. However, on further examination, it appears that 'text alternatives' do satisfy this AAA success criterion.

**Evidence:**
- it links to a sufficient technique that involves adding alt text to images that are used as links ([Technique H24](https://www.w3.org/WAI/WCAG22/Techniques/html/H24))
- it links to a sufficient technique that makes portions of visible text no longer visible ([Technique H2](https://www.w3.org/WAI/WCAG22/Techniques/css/C7))

**Conclusion:**

The intent is not to say that the link purpose needs to be conveyed exclusively in visible non-image text.

Instead, the intent is to say that, in isolation, the link text is all that's needed to interpret the link's purpose.

So basically, in the AAA version, you can no longer rely on "the link text together with its [programmatically determined link context](https://www.w3.org/WAI/WCAG22/Understanding/link-purpose-in-context.html#dfn-programmatically-determined-link-context)", but can only rely on the link text alone, regardless of its form.

scottaohara:
+1 to making sure it's clear that the AAA rule would pass for use of alternative text within a hyperlink, or the hyperlink's accName being updated to give all the necessary context.

patrickhlauke:
Coming back and rereading this nearly 3.5 years later, I'm actually now wondering if I've always had a slight misunderstanding about 2.4.4 Link Purpose (In Context) and 2.4.9 Link Purpose (Link Only).

This may be because, on their face, _these SCs try to serve different audiences_, with different needs and possible solutions - mainly, the split between non-AT and AT users.

For non-AT users, "in context" actually doesn't matter whether it's the same paragraph/table cell/etc. However it's marked up behind the scenes, a sighted non-AT user will "get" the context even if that context isn't "programmatically determinable". It's the AT users that need it to be "programmatically determinable". So this emphasis on "programmatically determinable" always made me think this was more skewed towards serving AT users, and not so much about making both a programmatic determination, AND also verifying if the link and its context were (subjectively) of good quality to represent what the link was about.

Likewise, the way to solve the "link only" situation is much easier for AT users, as you can add lots more to the link - images with `alt` text, `aria-label`, `aria-describedby`, visually hidden but accessible text ... all sorts of extra goodies. Whereas for non-AT users, you really are limited effectively to the visible readable text.

If we *are* saying that these SCs aim to solve the problems for both AT and non-AT users, we'll have to explicitly disallow relying on AT-only methods (like `aria-*`, visually hidden text, even images with `alt`), as these won't work for non-AT users. I somehow doubt that that was the actual intention though? But then I see that for 2.4.9, it already lists [ARIA8: Using aria-label for link purpose](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA8) as a sufficient technique, making me think that no, that was NOT the intention here? Or was the fact that the ARIA solution won't serve the needs of non-AT users just not considered, and this was added by mistake?

patrickhlauke:
I mean, one way out of the dilemma here would be to very clearly repoint 2.4.4 and 2.4.9 to be primarily about AT users. that would actually very clearly mark out what types of solutions are and aren't passes / satisfy the SC. it's the attempt to also cover non-AT users in these SCs that complicates matters and leads to effectively barring AT-specific approaches as solutions, and we're just left simply with "visible readable text" (with, for 2.4.4, and without, for 2.4.9, the context to help out)

patrickhlauke:
Moving this to in-progress rather than drafted for the WCAG 2.x backlog, as clearly this needs more discussion to get to the bottom of what the *actual* intent of 2.4.4 and 2.4.9 is.

stes-acc:
I don't get why this discussion is/are not done on a broader scope.   

Using aria-label / aria-description in concurrency to ANY visible text for a control (may it be a text node, a title attribute exposed in tooltip etc.) is always overriding the speech output for screen reader users for reasons described in [Accessible Name and Description Computation 1.2](https://w3c.github.io/accname/). The link text overriding is JUST ONE example for this. Wouldn't it be better to create a designated separate page within WCAG space for advice and use of best practices/techniques all related requirements can refer to?

understanding/20/link-purpose-link-only.html
```diff
@@ -33,14 +33,22 @@ <h2>Intent of Link Purpose (Link Only)</h2>
...
          web application would be sufficient to describe the purpose of the link (which is
          to take you to the document or web application). Note that it is not required to use
          the name of the document or web application; other things may also describe the purpose
          of the link. 
       </p>
       
+      <div class="note">
+         <p>Link text does not only refer to the visible text of a link. When evaluating this success criterion,
+            consider both the link's visible text that sighted users will perceive, and the link's
+            accessible  <a href="https://www.w3.org/TR/WCAG/#dfn-name">name</a>, which is how the link will be
```
patrickhlauke:
```suggestion
            accessible <a>name</a>, which is how the link will be
```

understanding/20/link-purpose-link-only.html
```diff
@@ -33,14 +33,22 @@ <h2>Intent of Link Purpose (Link Only)</h2>
...
          to take you to the document or web application). Note that it is not required to use
          the name of the document or web application; other things may also describe the purpose
          of the link. 
       </p>
       
+      <div class="note">
+         <p>Link text does not only refer to the visible text of a link. When evaluating this success criterion,
+            consider both the link's visible text that sighted users will perceive, and the link's
+            accessible <a>name</a>, which is how the link will be
+            announced to assistive technology users.
```
patrickhlauke:
```suggestion
            accessible <a>name</a>, which is how the link will be announced to assistive technology users.
```

understanding/20/link-purpose-in-context.html
```diff
@@ -21,12 +21,12 @@ <h2>In brief</h2>
...
-      
-      <p>The intent of this success criterion is to help users understand the purpose of each
-         link so they can decide whether they want to follow the link. Whenever possible, provide
+      <p>The intent of this success criterion is to help users - and specifically users of 
+         assistive technologies - understand the purpose of each link so they can decide
+         whether they want to follow the link. Whenever possible, provide
          link text that identifies the purpose of the link without needing additional context.
          Assistive technology has the ability to provide users with a list of links that are
-         on the web page. Link text that is as meaningful as possible will aid users who want
+         on the Web page. Link text that is as meaningful as possible will aid users who want
```
patrickhlauke:
```suggestion
         on the web page. Link text that is as meaningful as possible will aid users who want
```

understanding/20/link-purpose-link-only.html
```diff
@@ -34,14 +34,21 @@ <h2>Intent of Link Purpose (Link Only)</h2>
          links on a page.
       </p>
       
-      <p>The text in the link is intended to describe the purpose of the link. In cases where
+      <p>The text of the link is intended to describe the purpose of the link. In cases where
```
patrickhlauke, line 37: IF in discussion we determine that no, 2.4.9 can't be satisfied with accessible name, and what matters is only *visible* link text, then this would need to be amended to


```suggestion
      <p>The visible text of the link is intended to describe the purpose of the link. In cases where
```

understanding/20/link-purpose-link-only.html
```diff
@@ -34,14 +34,21 @@ <h2>Intent of Link Purpose (Link Only)</h2>
...
          the link takes one to a document or a web application, the name of the document or
          web application would be sufficient to describe the purpose of the link (which is
          to take you to the document or web application). Note that it is not required to use
          the name of the document or web application; other things may also describe the purpose
          of the link. 
       </p>
       
+      <div class="note">
```
patrickhlauke, line 45: IF in discussion we determine that no, 2.4.9 can't be satisfied with accessible name, and what matters is only visible link text, then this note should be dropped, or changed to clarify that while accName / accDescription are valid for 2.4.4, for 2.4.9 the pure visible link text itself is what matters.

understanding/20/link-purpose-in-context.html
```diff
@@ -40,41 +40,52 @@ <h2>Intent of Link Purpose (In Context)</h2>
...
+         is logically related text that provides the context for the link, and which is usually apparent
+         from the overall layout/structure of the content. In this case, the user should be able to
+         identify the purpose of the link without moving focus from the link.
+         In other words, they can arrive on a link and find out more about it without
          losing their place. This can be achieved by putting the description of the link in
-         the same sentence, paragraph, list item, or table cell as the link, or in the table header cell for a link in a data table, because these are directly associated with the link itself. Alternatively, authors may choose to use an ARIA technique to associate additional
+         the same sentence, paragraph, list item, or table cell as the link, or in the table header cell
+         for a link in a data table, because these are directly associated with the link itself.
+         Alternatively, authors may choose to use an ARIA technique to associate additional
          text on the page with the link.
```
patrickhlauke:
```suggestion
         text on the page with the link – either as part of the link's accessible name, or it's accessible description.
```

understanding/20/link-purpose-link-only.html
patrickhlauke, line 25: IF in discussion we do determine that the intent for 2.4.9 goes beyond AT users, then it's worth making the point here. Otherwise, IF 2.4.9 *is* primarily for the benefit of AT users, then it should have the same change as the one above proposed for 2.4.4 intent

PR #1571, opened 2020-12-29
Example 2 update.

### LINKED ISSUE #1532: Clarity for 1.3.3: Sensory Characteristics needed
https://www.w3.org/WAI/WCAG22/Understanding/sensory-characteristics

First I'm not sure if this SC is about:
an '"**instruction**" to get to a component / place' (a description **in** the web content) **OR** if the component / place on screen itself '**is**' the instruction (or does it apply for both?!)

Example: Is it the instruction to 'go to the right arrow and activate it to proceed' OR 'the right arrow IS the instruction' (indicating to go forward...?!)

First sentence of intent is:

> The intent of this Success Criterion is to ensure that all users can access instructions for using the content

The first example talks about:

> The cell corresponding to the time of a particular event has a specific background color and diamond shaped glyph so it can be identified by color and shape.

A blind user with a screen reader will NOT get the color instruction and also NOT the diamond glyph...!!

This is supported by the last sentence:

> WCAG was designed to apply only to controls that were displayed on a web page. The intent was to avoid describing controls solely via references to visual or auditory cues.

The first example is "describing controls solely via references to visual cues." (although it fits "**do not rely solely on color or shape**"

Then the next example is:

> The arrow is clearly labeled with "Next" and the instructions state, "To move to the next section of the survey, select the green arrow icon labeled 'Next' in the lower right corner below the last survey question." This example uses both positioning, color and labeling to help identify the icon.

Using the word "instructions" here, and not having a definition for instructions, suggests that the SC is all about a accompanying text "instructing on the use of a component, or alike" NOT about the component itself...?!

Is in this case the "green arrow icon labeled 'Next'" not already enough?

-----------------------

In short:

1. what is the definition of instruction here? (describing text of a component OR it can be the component itself (or color to convey meaning))
2. If you only have an icon of an arrow (left/right)  to go to the previous / next page, will this fail? (no label present BUT has an ACC name)

alastc:
Closes #1532

### DIFF
--- a/understanding/20/sensory-characteristics.html
+++ b/understanding/20/sensory-characteristics.html
@@ -107,8 +107,7 @@
                in the lower right hand corner of the content to move from one survey page to the
                next. The arrow is clearly labeled with "Next" and the instructions state, "To move
                to the next section of the survey, select the green arrow icon labeled 'Next' in the
-               lower right corner below the last survey question." This example uses both positioning,
-               color and labeling to help identify the icon. 
+               lower right corner below the last survey question." This example uses positioning and color, with labeling to also identify the icon and pass this criterion. 
             </p>
             
          </li>
--- a/understanding/20/sensory-characteristics.html
+++ b/understanding/20/sensory-characteristics.html
@@ -96,18 +96,20 @@
             <p>
                								       
                <strong>
-                  Example 2: An on-line multi-page survey 
+                  Example 2: An online multi-page survey 
                   
                </strong>
                							     
             </p>
             
             <p>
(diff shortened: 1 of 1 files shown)


understanding/20/sensory-characteristics.html
```diff
@@ -107,8 +107,7 @@ <h2>Examples of Sensory Characteristics</h2>
                in the lower right hand corner of the content to move from one survey page to the
```
awkawk, line 107: Can we also change on-line to "online"?

PR #1578, opened 2021-01-07
Issue1531 focus appearance size calc

### LINKED ISSUE #1531: 2.4.11: How to calculate exactly AND 8 pixels along the shortest side PROBLEM
From: https://broken--deque-workshop.netlify.app/ => the dialogs (2x) for EDIT and RECIPES BLUE BUTTON

**See the following focus styles for the buttons at the footer of the dialogs... ("SAVE" white bracket like thingy on blue background):**

Screen Shot 2020-11-17 at 2 05 16 PM

**And then a wider one (I COOKED IT)**

Screen Shot 2020-11-17 at 2 03 20 PM

- To calculate if they pass I started with the 'save' focus and came to the conclusion to draw a 3x15px box and approximately has two of them to have a full 'bracket thingy' = 150px area.
- I have one left and one right so 300px total. 
- The button itself is 110x36 * 2 = 292px
- So, I say it's a PASS

For the second image though, the width is wider and I do not have to calculate it, it's a FAIL

"a thickness of at least 8 CSS pixels along the shortest side" does not apply here (while we should allow, maybe...) 

This is (still) hard to explain!

alastc:
Closes #1531

### DIFF
--- a/guidelines/sc/22/focus-appearance-minimum.html
+++ b/guidelines/sc/22/focus-appearance-minimum.html
@@ -7,7 +7,12 @@
     
     <p>When user interface components receive keyboard focus, all of the following are true:</p>
     <ul>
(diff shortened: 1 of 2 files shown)


mbgower:
This has been resolved through wording changes in subsequent PRs.

Issue #1449, opened 2020-09-19
[WCAG 2.2 Draft Feedback] Success Criterion 2.4.13 Fixed Reference Points


dshoukry:
“Success Criterion [2.4.13 Fixed Reference Points](https://www.w3.org/TR/WCAG22/#fixed-reference-points) (Level A): When a [web page](https://www.w3.org/WAI/WCAG22/Understanding/fixed-reference-points.html#dfn-web-page) or [set of web pages](https://www.w3.org/WAI/WCAG22/Understanding/fixed-reference-points.html#dfn-set-of-web-pages) is an [electronic publication](https://www.w3.org/WAI/WCAG22/Understanding/fixed-reference-points.html#) with [pagebreak locators](https://www.w3.org/WAI/WCAG22/Understanding/fixed-reference-points.html#), a mechanism is available to navigate to each locator and each locator maintains its place in the flow of content, even when the formatting or platform change.”

We kindly request significant changes to this SC, and if those can't be made we would recommend that the group remove this SC or point readers to [EPUB Requirement 4.4.1 Page Navigation](http://idpf.org/epub/a11y/#sec-page-navhttps://www.w3.org/Submission/epub-a11y/#sec-page-nav). The intent of this SC is excellent, but the wording around it, especially in the opening example and even in the description, feels more like it is specific to something like an EPUB Requirement 4.4.1 Page Navigation. There are many scenarios on the web where there are pagers and other ways of breaking up content, and when referring to sections of the content, a concept of physical location might be used instead of digital context. (Other non-page related examples: to the right, above, below, etc.)

This also seems like an especially important area to have research, a diverse set of use cases, and a specific set of content structures to keep the SC beneficial, appropriately scoped, and attainable. 

Please find detailed specifics covered in our [2.4.13 Fixed Reference Points (Level A) Google Doc](https://docs.google.com/document/d/1oF0ZN5tRZAsCLfB2GGvRMRTISl0VrCP4485a2__fffg/edit#heading=h.23ybyllde921).

dshoukry:
As requested, I've just replaced the Google Doc link with a version with expanded access (didn't make any content changes). Sorry for the access issues with the previous link.

alastc:
Hi @dshoukry,

Regarding your main point:

[...]

PR #1580, opened 2021-01-08
Minor re-arrangement of the intent section.

### LINKED ISSUE #1302: SC 2.4.13: Fixed Reference Points "on the same page"
The paragraph feels out of place
>The term “on the same page” is common in English to describe when two people are talking about the same thing and agree on the content of the conversation. This success criteria makes sure that all users can literally be on the same page.

It sounds likes it's explaining a term that's just been used - when it hasn't. The concept of two people with different content arrangements, knowing they are referring to the same thing hasn't been mentioned yet. 
If it is useful at all as a aide-memoire then I think it belongs lower down in the explanation.

alastc:
Closes #1302

### DIFF
--- a/understanding/22/fixed-reference-points.html
+++ b/understanding/22/fixed-reference-points.html
@@ -45,9 +45,9 @@
 
       <p>The purpose of this success criterion is to allow users with disabilities to find references to content based on the page locators found in the default view or printed version of a publication.</p>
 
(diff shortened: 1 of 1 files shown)


alastc:
Note that this conflicts with #1449. If that passes this paragraph is removed anyway.

alastc:
Closed, as #1619 was merged.

PR #1594, opened 2021-01-19
Wcag22 focus appearance size update

alastc:
Based on survey and discussion: https://www.w3.org/2021/01/19-ag-minutes.html#t02

### DIFF
--- a/guidelines/sc/22/focus-appearance-enhanced.html
+++ b/guidelines/sc/22/focus-appearance-enhanced.html
@@ -7,7 +7,12 @@
     
     <p>For the keyboard focus indicator of each User Interface Component, all of the following are true:</p>
     <ul>
-        <li><strong>Minimum area:</strong> The <a>focus indication area</a> is greater than or equal to a 2 <a>CSS pixel</a> solid border around the control.</li>
+        <li><strong>Minimum area:</strong> The <a>focus indication area</a> is either:
+            <ul>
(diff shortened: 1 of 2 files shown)


detlevhfischer:
I take it that this example would meet the requirement?
http://3needs.org/en/testing/code/focus-visibility.html#bordertest
Would it still meet the requirement if the focus was also vertically offset?

JAWS-test:
> Would it still meet the requirement if the focus was also vertically offset?

I had already addressed this problem: https://github.com/w3c/wcag/pull/936#issuecomment-552729350. Unfortunately without success

alastc:
Hi @detlevhfischer,

So, looking at the buttons with `border-bottom: solid 2px #A9A9A9`, (changing from `#595959`) with a white adjacent background with an internal background of black, and they are around 52 x 33px, so with focus area of 2*52 = 104.

I don't think that meets the min size requirement, which would be either: 
- (52 * 2) + (33 * 2) -4 = 166
- 33*4 = 132

The bottom border changes from `#595959` to `#a9a9a9` which doesn't quite meet the change of contrast (2.98:1), although that's a minor tweak.

It would meet the adjacent contrast bullet.

But the main thing is that the size isn't sufficient.

detlevhfischer:
hi @alastc 
Could it be that in your answer, you are addressing the (dated, and marked as dated) example above the one I linked to? What I was getting at (the example below "Purposefully horrid example...") is that the lack of any requirement regarding the proximity of the focus to the element focused (here, a 4px border offset by 100px) might be a problem.

PR #1596, opened 2021-01-20
more clarity on reasonable objections

michael-n-cooper:
Expand the description of how objections to WG decisions may be raised, how they will be processed, and clarify that they must be on the record.


alastc (changes requested):

decision-policy.php
```diff
@@ -14,27 +14,37 @@
...
 				<ol>
-					<li> Discussion can take place in any <a href="./communication">recognized channel of the Working Group</a> including email on the AG mailing list, comment threads for GitHub issues or pull requests, or on Working Group calls. </li>
+					<li> Discussion can take place in any <a href="./communication">recognized channel of the Working Group</a> including email on the public AG mailing list, AG surveys, comment threads for GitHub issues or pull requests, or on Working Group calls. </li>
 				</ol>
 			</li>
 			<li> When the chairs believe that the group is ready to come to a decision they announce a Call for Consensus by email to the Working Group's mailing list. The Call must remain open for a minimum of two working days. 
 				<ol>
-					<li> The Call is open to responses from all group members. </li>
+					<li> The Call is open to responses from all <a href="https://www.w3.org/groups/wg/ag/participants">group participants</a>. </li>
```
awkawk, line 26: Is there a difference between member and participant? Are you going back to "in good standing"?

decision-policy.php
```diff
@@ -14,27 +14,37 @@
...
+		<p>During discussion on a topic, participants are welcome to raise objections freely to help ensure that all available information can be considered and contribute to the best possible decision. However, when the chairs issue a Call for Consensus (CfC), objections should not be raised unless the individual strongly believes the decision is the wrong one in spite of discussion, and the individual cannot "live with" the decision. Compromise on points that the individual considers suboptimal but can "live with" is an essential part of group decisions that must meet various requirements.  Objections from people who raised their concerns to the group during discussions may be given more consideration than objections from people who had opportunity and chose to not participate in those discussions.</p>
```
awkawk, line 17: This makes it sound like the objections are not evaluated as much on their merits but on whether they have been raised unsuccessfully before.

decision-policy.php
```diff
@@ -14,27 +14,37 @@
...
 				</ol>
 			</li>
+			<li>Responding to the Call for Consensus<ol>
+					<li>Individuals are invited, but not required, to support the decision by replying to the CfC email with “+1”, or to indicate a “can live with” status by replying with a “+0”.</li>
+					<li>Objections can be raised by replying to the CfC email with “-1”. In order to facilitate understanding and discussion, it is recommended to include a rationale which includes:<ol>
+							<li>why the individual objects to the proposed decision;</li>
+							<li>describes how the objection was raised and considered during discussion leading to the CfC, including how reactions to the potential objection were addressed;</li>
+							<li>what alternate decision would remove the objection <b>and</b> be likely to gain consensus in the WG.</li>
+						</ol></li>
+					<li>Objections are subject to examination by the WG and therefore must be filed on the public record.</li>
```
awkawk, line 39: Does this include who raised the objection or just that the content of the objection needs to be public?

decision-policy.php
```diff
@@ -14,27 +14,37 @@
...
-					<li>If working group member(s) continue to disagree and the chairs do not believe it presents substantive new information, or it does not meet the criteria established for adding new normative content, the chairs may decide the draft decision becomes a formal decision of the Working Group despite the objection.</li>
+					<li>If objections are received but the chairs believe the objections have already been considered as far as is possible and reasonable, and the reviewers providing the objections can “live with” the decision, the draft decision becomes a formal decision of the Working Group. </li>
+					<li>If the substance of the objections was raised during discussion leading to the CfC, the chairs may give more consideration to those objections than to objections which were not raised during the discussion leading to the CfC, and reasonably could have been.</li>
```
awkawk, line 45: Same comment as above.

bruce-usab:
Just commenting to express my appreciation of the close review by @awkawk.  Thanks!

decision-policy.php
mbgower: Suggest the following replace the whole paragraph from the second sentence onward (change of "when" to "once" in first sentence; many changes in subsequent):
However, once the chairs issue a Call for Consensus (CfC), objections should not be raised unless the individual strongly believes the decision is the wrong one in spite of discussion, and the individual cannot "live with" the decision. Compromise is an essential part of group decisions. This requires individuals to participate actively in modes of discussion, but "live with" points they consider suboptimal once reaching CfC.

decision-policy.php
mbgower: @awkawk if they are raising objection via an email response, doesn't that automatically provide a 'who'

decision-policy.php
mbgower: Yeah, I think the last new bullet can go. If someone raises an objection that has already been covered and addressed, that should be easier to respond to adequately. I think that is covered by the first new bullet, and I support it.
However, if someone raises a wholly new concern, that _should_ get full attention. I get how you want folks to take part in discussion, but I think the first new bullet addresses that.

decision-policy.php
alastc: Not if they email the chairs only to raise the objection.

PR #1597, opened 2021-01-21
new paragraph in redundant entry

rachaelbradley:
Issue 1345 - For review at the 26 January 2021 AGWG meeting


alastc:
Update from the meeting:
- Drop the understanding update.
- Change the SC text to match AWK's proposal:

> Information previously entered by or provided to the user that is required to be entered again in the same process and in the same user session is either:

PR #1599, opened 2021-01-24
update resource links in G15

scottaohara:
In https://www.w3.org/WAI/WCAG21/Techniques/general/G15.html, the Trace Center link was not resolving.  

This PR updates the link to a URL that works.  Also updated the Harding FPA link to from using http a https


alastc:
Thanks Scott, merging :-)

PR #1608, opened 2021-01-26
Pointer target spacing update

scha14:
* View [SC pointer target spacing](https://raw.githack.com/w3c/wcag/6c6cc164ebb1452c495dc8b4ec394c78bffd54a8/guidelines/sc/22/pointer-target-spacing.html)
*  View [Understanding pointer target spacing](https://raw.githack.com/w3c/wcag/scha14-pointer-target-spacing/understanding/22/pointer-target-spacing.html)
* [Related issues ](https://github.com/w3c/wcag/labels/2.5.8%20Pointer%20Target%20Spacing)


detlevhfischer:
Looks good to me.

alastc:
Thanks for creating this, I did a few updates to match the google doc, but if you get a chance to update the figures & captions, that would be great.

scha14:
Figures and captions are updated. @alastc could you please take a look to make sure they render okay?

detlevhfischer:
I tried to create a link for the[ rendered view of understanding pointer target spacing](https://raw.githack.com/w3c/wcag/6c6cc164ebb1452c495dc8b4ec394c78bffd54a8/understanding/22/pointer-target-spacing.html) via githack but I am not sure these are the updated graphics...

scha14:
They are the latest graphics. They are the same images with the dimensions
updated to match the new SC

alastc:
I think Detlev just meant that the middle would should be marked 'pass'?

scha14:
I can do that. Wasn't sure that this didn't qualify as enclosed and was
trying to avoid the case with these tiny buttons at the corners of images.

alastc:
I think the idea on the middle one is that the icon is on the edge, so overlapping but not enclosed by the picture, therefore it has the spacing top and left.

scha14:
It's updated

detlevhfischer:
OK I have also updated the githack link.
@scha14 @alastc I think we still need to update or replace the third example - the pixel figures underneath the buttons make no sense, it is obvious that all buttons are considerably wider than 24 px (and "Push" has got to be wider than 20 px), and the relative dimensions are not correct, either.
Maybe we can think of another example? Or just drop this example?  A pull-down menu example might be relevant... (my head is empty right now...)

scha14:
@detlevhfischer updated the figure so the relative sizes make more sense. We can replace with another figure as well.

alastc:
As the updates weren't approved, I'll merge this into 'main'.

Issue #1454, opened 2020-09-19
[WCAG 2.2 Draft Feedback] Success Criterion 2.5.7 Dragging


dshoukry:
“Success Criterion 2.5.7 Dragging (Level AA): All functionality that uses a dragging movement for operation can be operated by a single pointer without dragging, unless dragging is essential.”

Adding more examples and clarification to this SC may help to solidify successful implementation of this requirement. We were confused about how this is different from WCAG 2.1 SC 2.5.1 Pointer Gestures and interpreted that most scenarios would likely meet the essential exception. Other comments include: requests to clarifying common essential approved exceptions for dragging besides drawing apps and also providing a few examples of interactions that are incorrectly assumed to be essential approved exceptions, a suggestion to clarify that the exemption covers interactions essential to the task, not essential to a particular interaction design of the task, and a question about a sufficient technique.

Please find detailed specifics covered in our [2.5.7 Dragging (Level AA) Google Doc](https://docs.google.com/document/d/1mOgGI7bdgTqsEYhuOlA9yiWGMU06ONrgL0KbGflYKJw/edit).

dshoukry:
As requested, I just replaced the Google Doc link with a version with expanded access (didn't make any content changes). Sorry for the access issues with the previous link.

alastc:
Hi @dshoukry,

I have a couple of comments and responses to your comments:

> Other than drawing apps, what are some common essential approved exceptions for dragging? ...
> It might also be helpful to explicitly clarify that the exemption covers interactions essential to the task, not essential to a particular interaction design of the task.

We do have a linked definition of essential, which is "if removed, would fundamentally change the information or functionality of the content, and information and functionality cannot be achieved in another way that would conform".
That definition is used quite a lot, I'd rather not add to the SC text for that. Examples in the Understanding doc are fine though, and could include a line about it being essential to the task rather than the interaction.

In a dragging context, needing to enter a handwriting sample is something that requires the information in a particular format and could not be achieved in another way. (You could argue that physical signatures are not needed given that we now have digital signatures now, so I avoid that as an example.)

> We were confused about how this is different from WCAG 2.1 SC 2.5.1 Pointer Gestures and interpreted that most scenarios would likely meet the essential exception.

We have another issue where we are dealing with the overlap. Could you provide any scenarios that you thought would fit the essential exception? 

[...]

PR #1612, opened 2021-02-01
Adding essential examples.

### LINKED ISSUE #1336: 2.5.7: Dragging - essential example
I am trying to work out what an example would be for essential on 2.5.7: Dragging. 

I am aware that the generic essential definition is used across SC. But it's often useful to have examples. For example, pointer cancellation has:

- An activity that emulates a physical on-press trigger, such as when playing an on-screen piano keyboard. Activation on the up-event would significantly alter the desired behaviour.
- A program for shooting skeets where waiting for the "up" event would invalidate the precise timing necessary for the activation.

I am not sure what this would be for dragging, though. 

Maybe a physics-based game where you have to drag and release, like Angry Birds. It would invalid the game if you could just type in the exact angle, velocity, etc.

Another example might be something like https://www.rightmove.co.uk/draw-a-search.html# you can click to put down notes and them drag them around for finer control.

However, in this case, it would be possible to make it work with a single pointer -- e.g. each note would have an up, down, left and right button, or an x and y input field. But I can imagine it being very tricky to make usable.

Although in this case, there are the clear and undo buttons, so the user could get the nodes to the exact position they wanted by using these buttons are trying again. But they would not have access to the "I want to adjust this slightly" functionality.

alastc:
Closes #1336, contributes to #1454

### DIFF
--- a/understanding/22/dragging-movements.html
+++ b/understanding/22/dragging-movements.html
@@ -43,7 +43,7 @@
 
       <p>2.1.1 Keyboard and 2.1.3 Keyboard (No Exception) require dragging features to be keyboard accessible, however, it is possible to create an interface that works with dragging and keyboard controls that does not work using only clicks or taps.</p>
 
-      <p>There may be cases where a dragging movement is essential, for example, writing a signature, or playing a game where the aim is to follow a particular path as accurately as possible. The criterion has an exception for these cases.</p>
+      <p>There may be cases where a dragging movement is essential, for example, writing a signature, or playing a game where the aim is to follow a particular path as accurately as possible. The criterion has an exception for these cases where the task requires dragging.</p>
       
       <h3>Distinguishing dragging movements from path-based pointer gestures</h3>
       <p>A dragging movement is a pointer interaction where only the endpoints matter. Once the pointer engages with a target, the target (or a position mark related to the target) will follow the pointer regardless of the direction of the pointer movement. The target's movement may be constrained to one axis and often, to a particular range on that axis, but within those constraints, the x or y position of the target will mirror the corresponding x or y position of the pointer until the pointer disengages the target.</p> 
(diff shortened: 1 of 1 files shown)


understanding/22/dragging-movements.html
```diff
@@ -41,7 +41,11 @@ <h2>Intent of Dragging Movements</h2>
       
     <p>This Success Criterion applies to dragging motions as opposed to pointer gestures which are covered in Success Criterion 2.5.1 Pointer Gestures. Pointer gestures include directional path-based as well as multi-point gestures. In contrast, for dragging motions, only the start and end point of the motion matters, not the actual path.</p>
 
-		<p>2.1.1 Keyboard and 2.1.3 Keyboard (No Exception) require dragging features to be keyboard accessible, however, it is possible to create an interface that works with dragging and keyboard controls that does not work using only clicks or taps.</p>
+      <p>2.1.1 Keyboard and 2.1.3 Keyboard (No Exception) require dragging features to be keyboard accessible, however, it is possible to create an interface that works with dragging and keyboard controls that does not work using only clicks or taps.</p>
+
+      <p>There may be cases where a dragging movement is essential, for example, drawing a freeform figure, or playing a game where the aim is to follow a particular path as accurately as possible. The criterion has an exception for these cases where the task requires dragging.</p>
```
patrickhlauke, line 46: isn't "follow a particular path" by definition a "path-based gesture", so not a "dragging movement"?

alastc: Hmm, it's a term clash. I'll change that to 'track' to avoid confusion.

dshoukry: +1 to @patrickhlauke, a few of us are also still having trouble understanding how this differs from 2.5.1: Pointer Gestures from 2.1 and just want to make sure we are properly comprehending. If we've understood correctly, the difference is that this dragging SC involves pointer interaction where _only_ the endpoints matter vs. the 2.5.1 PG SC is interaction where _not_ just the endpoints matter.

Would it be possible to further expand on the "follow the path" portion in the game example? Is that meant to illustrate percentage of something being correctly traced/colored (e.g. 96% accuracy), which seems like a path-based gesture? Or rather an in-game character traversing (which could have a single pointer arrow key or maybe even a long press to "move them"), which doesn't seem to be path-dependent?

Thank you for considering!

PR #1626, opened 2021-02-07
Updating the contrast requirement

### LINKED ISSUE #1306: C41 does not fulfil SC 2.4.12
[C41](https://w3c.github.io/wcag/techniques/css/C41) does not fulfil SC 2.4.12  and should therefore not be linked to 2.4.12 or adapted. The size and contrast of the focus indicator are not sufficient for 2.4.12

alastc:
Closes #1306


alastc:
Approved, but need to rename the technique name before merging.

PR #1629, opened 2021-02-08
Correct a citation under Themes from Research

bruce-usab:
- Brajnick -> Brajnik
- added link
- deleted last bullet item: This means that an 80% target for agreement, when audits are conducted without communication between evaluators, is not attainable, even with experienced evaluators.


bruce-usab:
See issue https://github.com/w3c/wcag/issues/1622 for discussion.

PR #1631, opened 2021-02-09
Change STR to Ctrl

detlevhfischer:
Change STR to Ctrl (for modification keys). Not entirely sure if this should be even more generic since different languages will likely have different key labels and I am now replacing the (incorrect) German STR with the English Ctrl. But at least it says "modification keys like..." so I guess as an example it is OK.


bruce-usab (approved):
Thanks for doing the PR!

alastc:
Thanks Detlev, I've merged that.

Issue #1225, opened 2020-07-22
WCAG 2.2 2.5.7 Dragging and its reference to single pointer


guyhickling:
The wording of the SC is contradictory. It says that path-based gestures (I'll use PBGs for brevity) must have an alternative method using "a single pointer without a path-based gesture". But if you look at the glossary definition of those words "single pointer", it actually includes and therefore appears to explicitly allow PBGs as an alternative. It says:- 

> pointer input that operates with one point of contact with the screen, including ... path-based gestures

This is highly confusing to anyone reading it and trying to work out what it all really means. 

I'm raising this now because the new Dragging SC, SC 2.5.7 in WCAG 2.2, uses the same "single pointer" term, and the glossary definition is referenced in the Understanding. Since 2.5.7 is just like 2.5.1 in that it also seeks to exclude PBGs from the allowed alternative methods (established in an earlier GitHub issue), this glossary definition will now confuse two SCs, not just one.

The only other SC to reference this term is 2.5.2, but that is not about PBGs either. 

Could we delete the words "path-based gestures" from the glossary definition of "single pointer"? None of the SCs want it there and it is just confusing.

(NB: I suggest this should have the 2.1 and 2.2 flags added to it as it affects both.)

patrickhlauke:
the definition for "single pointer" is a bit inaccurate/incorrect - see https://github.com/w3c/wcag/pull/809

however, in this specific case, "single pointer" can include all those interactions (like PBGs), but the SC explicitly says "a single pointer *without a path-based gesture*" so it excludes those. from that point of view, i don't think that's incorrect.

guyhickling:
Unfortunately, 2.5.7 doesn't use that phrase "without a path-based gesture", so PBGs aren't excluded in that SC. In other words 2.5.7 expressly suggests PBGs as an alternative to dragging, which contradicts the intention of 2.5.1 which says they cannot be used without an alternative. Or are you saying that PBGs can be used as the alternative for dragging, but that must then be given another alternative that is not a PGB? That would clearly be unreasonable.

So my point in raising this issue was that none of the three SCs that reference "single pointer" intend to allow PBGs as the alternative, so the single pointer definition should not include them.

patrickhlauke:
or conversely...a single pointer interaction simply means "only uses a single pointer" and does include dragging/swiping, and SCs that don't allow it should say so specifically in the SC text.

patrickhlauke:
[...]

PR #1636, opened 2021-02-10
David macdonald patch drag pointer

DavidMacDonald:
This patch addresses Issue #1225 and disambiguates dragging and path based gestures


understanding/21/pointer-gestures.html
```diff
@@ -20,14 +20,16 @@ <h2>Intent of this Success Criterion</h2>
...
+     
+      <h3>Disambiguation of Pointer Gestures and Dragging</h3>
 
-			<p>Dragging motions where the movement between the start and end points can go in any direction and is entirely non-prescriptive are not path-based. Dragging motions are covered in Success Criterion 2.5.7 Dragging.</p>
+      <p>Dragging is a motion between the start and end points without an intermediate point which can go in any direction and is entirely non-prescriptive, and therefore not path-based.  Although a user could drag in the same way as a gesture, the difference is that a gesture requires an intermediate point, dragging does not. Path-based gestures are a communication of a particular command to the system. Dragging involves the displacement of an object from one location to another, gestures do not.  Dragging motions are covered in <a href="https://www.w3.org/WAI/WCAG22/Understanding/dragging.html">Success Criterion 2.5.7: Dragging</a>.</p>
```
detlevhfischer, line 26: For consistency#s sake we should either use "dragging motion" or "dragging movement" in both 2.5.1 and 2.5.7. I leave it to native speakers to determine the better choice.
I believe this text could ans should be shorter (maybe not quire as short as the one it replaces). I can try to rephrase...

understanding/21/pointer-gestures.html
```diff
@@ -20,14 +20,16 @@ <h2>Intent of this Success Criterion</h2>
 			</figure>
 
 			<p>Examples of path-based gestures include swiping, sliders and carousels dependent on the direction of interaction, and other gestures which trace a prescribed path such as drawing a specific shape. Such paths may be drawn with a finger or stylus on a touchscreen, graphics tablet, or trackpad, or with a mouse, joystick, or similar pointer device.</p>
+     
+      <h3>Disambiguation of Pointer Gestures and Dragging</h3>
```
detlevhfischer, line 24: A simpler heading (think COGA) would be "The difference between pointer gestures and dragging"

understanding/22/dragging-movements.html
```diff
@@ -48,8 +48,10 @@ <h2>Intent of Dragging Movements</h2>
...
+      <p>In contrast to dragging movements, path-based gestures involve at least an initial directionality of the pointer movement. This is defined by stating that the path has to go through at least one intermediate point to qualify as pointer gesture. The intermediate point defines the gesture as requiring a specific path, even if the complete path is not defined. So, although a user could drag in the same way as a gesture, the difference is that a gesture requires an intermediate point, dragging does not. For more details, refer to <a href="https://www.w3.org/WAI/WCAG21/Understanding/pointer-gestures.html">Understanding Success Criterion 2.5.1 Pointer Gestures</a>.</p>
```
detlevhfischer, line 52: In my view, the sentences added: "The intermediate point defines the gesture as requiring a specific path, even if the complete path is not defined. So, although a user could drag in the same way as a gesture, the difference is that a gesture requires an intermediate point, dragging does not" make the text more verbose without providing more clarity. I would leave the text as it is.

detlevhfischer (changes requested):
See my PR #1638 for what I believe is a more understandable explanation of the difference between path-based gestures and dragging.

alastc:
Agreed in the meeting, although further updates are likely.

Issue #1453, opened 2020-09-19
[WCAG 2.2 Draft Feedback] Success Criterion 3.3.8 Redundant entry


dshoukry:
“Success Criterion 3.3.8 Redundant entry (Level A): For steps in a process, information previously entered by or provided to the user that is required on subsequent steps is either:

- auto-populated, or
- available for the user to select.
- Exception: When re-entering the information is essential.”

Most of our comments/proposals are requests to: consider updating the SC title to Populate prior data, explicitly clarify that auto-populating security info is different from verification, and consider cross linking to the WCAG 2.2 Accessible Authentication SC3.3.7, consider our suggestions about additional examples of approved exceptions and sufficient techniques, add a section for failure examples, and further expand the reference made to a Turing test.

Please find detailed specifics covered in our [3.3.8 Redundant entry (Level A) Google Doc](https://docs.google.com/document/d/1N9J4FaJFVD00bHxI_aVkrEHEShBcHsH0YTxqxN7Yoeo/edit).

alastc:
Hi @dshoukry,

Tackling the comments in order of the document, and with associate PR #1639:

DRAFT response: 

<hr>

> Can this be renamed to Populate prior data or something else that is more clear and easier to remember?

I'll put that to the group.

> Under Success Criterion 3.3.7 Accessible Authentication, though, auto-populating some security info is recommended. Could it be made clearer that auto-populating security info is different from verification, maybe by cross linking to the earlier success criterion?

The scenario is different, where Accessible Authentication would by nature apply across sessions, and entering a password would be 'essential' to the purpose of that form. There is an extra paragraph saying that in the PR.

> Do repeated entries in search fields also apply here?

In the context of a process, going to a search page would be considered the 'start' of a process, and would not need to pre-fill the previous entries. However, if the second page of a search process (i..e. the results page) would need to pre-fill the results if the input is associated with the previous search. That seems like a reasonable example to add.

dshoukry:
Thank you for all of the clarifications, accepting all of our content changes, and the updated PRs. 

BTW, I've just returned from some time off so I'm catching up on all of the latest updates. In an attempt to be helpful and acknowledge where we are with all of the changes, I'll be summarizing any outstanding questions in this and each of the five other original SC-level issues that are still open. 

[...]

PR #1639, opened 2021-02-11
Redundant entry updates

alastc:
For #1453


alastc:
Agreed in the meeting: https://www.w3.org/2021/03/02-ag-minutes.html#t12

Issue #872, opened 2019-08-25
New SC for keyboard operation?


JAWS-test:
I propose a new success criterion for WCAG 2.2 as a supplement to 2.1.1:
"If a standard operation exists for navigating through the page and for operating the interactive elements, then this should be implemented. Divergences should be documented in the application".

Reason:
* 2.1.1 only requires keyboard operation, but does not specify how this should be done. I.e. according to 2.1.1 it would be ok if the buttons could only be activated via Alt+Shift+X and this is not documented anywhere (see https://github.com/w3c/wcag/issues/857#issuecomment-521319817).
* In 2.1.2 it is specified for keyboard traps that the standard operation must work and that divergences from the standard operation must be documented. 

For WCAG 3.0, it would make sense to combine 2.1.1 and the new criterion, as was done for 2.1.2.

I suggest AA as the level. 

During the test I very rarely encounter keyboard traps, but very often hard to use web applications, because the standard keyboard operation was not implemented and the divergences were not documented. That's why I think this is an important addition.

If necessary, it could be considered to make the new criterion more general and to extend it to standard operation for pointing devices, touch, etc.

jake-abma:
The problem as described seems to be an equally problematic issue for all users. This in contrary to a disproportionate burden for people with different needs and abbilities. This equal disbalance needs to be in place in order for a SC to be considered. The purpose of a SC is to harmonize the disbalance.

jake-abma:
Among other rules, here's the first:

"Success Criteria shall:

1. Address a situation where a user with a disability will be disproportionately disadvantaged (as compared to a user without a disability) if the criteria is not met."

mraccess77:
I'd argue that users with disabilities tend to more use a keyboard interface and that in general mouse operations are intuitive and thus non-documented keyboard interfaces are more of an issue for people with disabilities.  I'd also point to new proposed SC for gesture documentation that is along the same lines as this.

jake-abma:
This may be applicable for Keyboard only users and AT users ... not sure if this is discussed already in the past, but a global standard for keyboard / component functionaly should be part of the reference

jake-abma:
Ah, you beat me Jon... so you know about previous discussions and do we have clear guidance on which key belong to what UIC? There are cases where people do not agree on the patterns.

Like the Tab widget and Tav / arrow keys to navigate...

jake-abma:
B.t.w. the new gesture SC is only about custom gestures, not standard operating. This is a well considered decision.

[...]

Issue #857, opened 2019-08-13
Should role button and input button be a WCAG fail if cannot be activated using space?


LaurenceRLewis:
If a keyboard user accesses a [fake] button implemented using role button or input button, and it can only be activated using the Enter key (not space as may be expected). Should this become a WCAG SC fail. I am proposing it should. 

Why?

When a sighted keyboard user tabs to a button—that visually looks like a button, their first instinct may be to use the Space key. When this actions the browser default and scrolls the page; it could cause confusion, and or disorientation. 

I suggest this could especially affect people with low vision and or cognitive disabilities. 

As I understand it this is not a failure for Success Criterion 2.1.1 Keyboard because the button can be activated using the Enter key. 

Adding as a recommendation is a start, however, in many cases, including my own work requirements for accessibility. Testing is aligned with WCAG Success Criteria. When there is a potential accessibility issue, as described above, I can advise as best practice. But, as it is not a fail of technique, recommendations are often ignored. 

The fact that JAWS and NVDA already include the Space to activate [fake] buttons by default. Implementing Space as well as Enter for non assistive technology keyboard users, could be considered a reason to include this as a requirement for success.

patrickhlauke:
depends how far you want to take this. normatively requiring that controls/widgets always follow the standard keyboard/control interactions (also including, on mobile or touch devices, responding to things like swipe up/down or volume up/down for sliders, which is only possible via AOM or by using real `<input type=“range”>`)? as the standard interactions aren’t normatively defined anywhere (?) this would be tricky to normatively express in a platform agnostic way...

LaurenceRLewis:
How far I’d like to take it is; to not meet the requirements. 😄 However, as you point out—there is more to contemplate  than only button behaviour. Which I confess I had not considered.

LaurenceRLewis:
@patrickhlauke could you please explain the acronym AOM. Thank you.

patrickhlauke:
https://github.com/WICG/aom

JAWS-test:
[...]

PR #1642, opened 2021-02-14
Add two notes/clarifications to 2.1.1/2.1.3 understanding

patrickhlauke:
Add clarifications about 2.1.1 and what it specifically does *not* currently require, normatively.

Cross-reference to the original discussion a while ago https://github.com/w3c/wcag/issues/857

Related discussion https://github.com/w3c/wcag/issues/872


bruce-usab (approved):

patrickhlauke:
Any news on this @alastc ?

understanding/20/keyboard-no-exception.html
```diff
@@ -29,6 +29,28 @@ <h2>Intent of Keyboard (No Exception)</h2>
...
       </p>
+
+      <div class="note">
+         <p>Platforms and user agents usually have conventions for how web content or applications
+            are controlled with a keyboard interface. As a best practice, content should follow
+            the platform/user agent conventions. However, deviating from these conventions does
+            not fail the normative requirement of this success criterion. For instance, buttons
+            that have focus can generally be activated using both the <kbd>Enter</kbd> key and
+            the <kbd>Space</kbd> bar. If a custom button control in a web application instead only
+            reacts to <kbd>Enter</kbd> (or even a completely custom key or key combination), this does <strong>not</strong> fail the success criterion.
```
mbgower, line 40:
```suggestion
            reacts to <kbd>Enter</kbd> (or even a completely custom key or key combination), this does <strong>not</strong> fail the Success Criterion.
```

understanding/20/keyboard-no-exception.html
```diff
@@ -29,6 +29,28 @@ <h2>Intent of Keyboard (No Exception)</h2>
...
+            the platform/user agent conventions. However, deviating from these conventions does
+            not fail the normative requirement of this success criterion. For instance, buttons
+            that have focus can generally be activated using both the <kbd>Enter</kbd> key and
+            the <kbd>Space</kbd> bar. If a custom button control in a web application instead only
+            reacts to <kbd>Enter</kbd> (or even a completely custom key or key combination), this does <strong>not</strong> fail the Success Criterion.
+         </p>
+      </div>
+
+      <div class="note">
+         <p>This success criterion does not require that every visible control that can be activated
```
mbgower, line 45:
```suggestion
         <p>This Success Criterion does not require that every visible control that can be activated
```

understanding/20/keyboard-no-exception.html
```diff
@@ -29,6 +29,28 @@ <h2>Intent of Keyboard (No Exception)</h2>
...
+      </div>
+
+      <div class="note">
+         <p>This Success Criterion does not require that every visible control that can be activated
+            using a mouse or touch screen must also be focusable and actionable using the keyboard.
+            The normative requirement is only that there must be a way for keyboard users to perform
+            the same, or comparable, actions and to operate the content. Generally, the easiest way
+            to achieve this is to provide controls that can be operated with all possible input devices;
+            however, if a web application implements a completely separate mode of operation for keyboard
+            users, it will <strong>not</strong> fail the success criterion.
```
mbgower, line 51:
```suggestion
            users, it will <strong>not</strong> fail the Success Criterion.
```

understanding/20/keyboard.html
```diff
@@ -69,6 +69,27 @@ <h2>Intent of Keyboard</h2>
...
+            the platform/user agent conventions. However, deviating from these conventions does
+            not fail the normative requirement of this success criterion. For instance, buttons
+            that have focus can generally be activated using both the <kbd>Enter</kbd> key and
+            the <kbd>Space</kbd> bar. If a custom button control in a web application instead only
+            reacts to <kbd>Enter</kbd> (or even a completely custom key or key combination), this does <strong>not</strong> fail the success criterion.
+         </p>
+      </div>
+
+      <div class="note">
+         <p>This success criterion does not require that every visible control that can be activated
```
mbgower, line 84:
```suggestion
         <p>This Success Criterion does not require that every visible control that can be activated
```

understanding/20/keyboard-no-exception.html
```diff
@@ -29,6 +29,28 @@ <h2>Intent of Keyboard (No Exception)</h2>
          Rather, it means that content that uses path-dependent input cannot conform to this
          Success Criterion and therefore cannot meet Guideline 2.1 at Level AAA. 
       </p>
+
+      <div class="note">
+         <p>Platforms and user agents usually have conventions for how web content or applications
+            are controlled with a keyboard interface. As a best practice, content should follow
+            the platform/user agent conventions. However, deviating from these conventions does
+            not fail the normative requirement of this Success Criterion. For instance, buttons
```
mbgower, line 37: I think this needs to be softened. I would be very concerned if someone interpreted this to mean that a complete deviation from standard keyboard interaction is never a failure of 2.1.1.
```suggestion
            not necessarily fail the normative requirement of this Success Criterion. For instance, buttons
```

understanding/20/keyboard-no-exception.html
```diff
@@ -29,6 +29,28 @@ <h2>Intent of Keyboard (No Exception)</h2>
...
+            that have focus can generally be activated using both the <kbd>Enter</kbd> key and
+            the <kbd>Space</kbd> bar. If a custom button control in a web application instead only
+            reacts to <kbd>Enter</kbd> (or even a completely custom key or key combination), this does <strong>not</strong> fail the Success Criterion.
```
mbgower, line 40: Again, I'm concerned about the unqualified statement.
508 has a requirement that accessibility features are documented, and IBM has used that to direct that if someone introduces unconventional keyboard interaction, it must be clearly documented, and that documentation must be keyboard accessible and should be easily discovered, preferably at the point of interaction.

Finally, if a 'custom' button failed to react to Space, it would be surprising to a keyboard user who ordinarily expects that behaviour -- particularly where the component is given the role of button. I would definitely flag that in a review, and I would be flag that as suspicious of other likely problems with the implementation.
```suggestion
            reacts to <kbd>Enter</kbd> (or even a completely custom key or key combination), this does <strong>not</strong> necessarily fail the Success Criterion.
```

JAWS-test: @mbgower: I agree with you from an accessibility perspective, but the question is whether the normative text of SC 2.1.1 provides these requirements. I'm afraid not, that's why I created: https://github.com/w3c/wcag/issues/872. And the fact that 508 and IBM go beyond the WCAG requirements is great, but it doesn't change the WCAG.

mbgower: No, just to confirm, you are saying that an item that no one but the author can figure out how to operate by keyboard passes 2.1.1? That is a _very_ tight interpretation of the SC.

mbgower: I get what you're saying. But I do not believe anyone should feel confident using such a tight interpretation of the normative text in a court case or with an indignant client. I do not believe it was the intent of the authors of 2.1.1 that keyboard operation was only a hypothetical achievement where someone may discover how to perform a keyboard operation.
You'll note that I have only added 1 word into my suggested change, to convey that there is a lot of gray area. I'd also very much prefer additional language in the Understanding document on best practices.

understanding/20/keyboard-no-exception.html
patrickhlauke: when is it a failure, normatively? i don't see anything in the normative text that suggests this

understanding/20/keyboard-no-exception.html
patrickhlauke: > That is a very tight interpretation of the SC

and yet, the SC doesn't hint at anything other than that. adding that hint in understanding seems to introduce new normative requirements that simply aren't present in the normative text.

by all means, we can add more "however, as a best practice", but let's not suggest an interpretation that isn't actually in the wording of the SC - and it isn't in the wording of the SC precisely because it's soo wooly/undefined, as there's no authoritative/standardised document anywhere that dictates what the standards for web content are. even 'platform' standards don't necessarily apply, as the platform is not web - and, can't make web content that is platform-agnostic dependent on unknown/unknowable (for authors) platform conventions"

JAWS-test: The problem is that there are people who test accessibility or are dependent on it and therefore interpret SC 2.1.1 in one way, while the software manufacturers, who would then have to fix the problems, often have no desire to do so and argue with the wording of SC 2.1.1, which is the way it is. Because we unfortunately cannot change the wording, I had suggested creating a new SC. Unfortunately, the opportunity was missed. I also have no objection to explaining in the Understanding that SC 2.1.1 does not say how a website must be operated, but that is what it indirectly means. But will all of the WCAG WG agree with this? I'm afraid not.

patrickhlauke: > I do not believe anyone should feel confident using such a tight interpretation of the normative text in a court case or with an indignant client

that would then turn into a "it follows the letter of WCAG, but isn't actually accessible/usable/logical". it's certainly not the first SC that has this problem. luckily, court cases aren't decided *solely* on the absolute baseline of WCAG. as for discussions with clients...

JAWS-test: And weak WCAG requirements are bad for accessibility. But even worse are unclear WCAG requirements that are interpreted differently by everyone. In this respect, I am always in favor of clarifications in the Understanding, even if they weaken the degree of accessibility.

mbgower: > adding that hint in understanding seems to introduce new normative requirements that simply aren't present in the normative text.

I don't agree with that. It is possible to add language in on challenges with custom keyboard interaction and how it is recommended to document unusual keyboard interaction (including making it discoverable in the UI). There is no reason that needs to be written in a way that is prescriptive so as to be read as normative.

> introduce new normative requirements that simply aren't present in the normative text.

There are **many** SCs that are based on underlying assumptions that are unfortunately not stated. For example
> [Audio description](https://www.w3.org/TR/WCAG20/#audiodescdef) is provided for all [prerecorded](https://www.w3.org/TR/WCAG20/#prerecordeddef) [video](https://www.w3.org/TR/WCAG20/#videodef) content in [synchronized media](https://www.w3.org/TR/WCAG20/#synchronizedmediadef). (Level AA)

All criteria that make similar uses of variations on the word "describe" have a clear assumption that the description is _accurate_. But a tight interpretation of such SCs is that it passes WCAG when there is _any_ description, regardless of its quality or veractiy. But I would suggest to you that it is in no one's interest to put the following into the SC:

> a description that does not match the visual content is not a failure of this SC

I see what is being proposed with these additions as being along the same lines. Adding the proposed content to the Understanding document without any additional explanation or caution is going to make it easier for someone anyone to point to the WCAG guidance to defend an interaction that is effectively unusable.

patrickhlauke: adding "necessarily" to this PR does not help clarify the interpretation of the SC for auditors and simply codifies the handwavy status quo of the current SC ... which then makes me wonder what the point would be of doing it in the first place. we either want to bring clarity, or we're happy with different auditors coming to different conclusions here

alastc: How about:
> "Platforms and user agents usually have conventions for how web content or applications
>  are controlled with a keyboard interface. Any method of keyboard will pass the criterion. However, if content does not follow the platform/user agent conventions it
> will be diffiult to use, and testers may assume that it is failing the criterion.
> For instance, buttons that have focus can generally be activated using both the <kbd>Enter</kbd> key and the <kbd>Space</kbd> bar. If a custom button control in a web application instead only reacts to <kbd>Enter</kbd> (or even a completely custom key or key combination), this does <strong>not</strong> fail the Success Criterion."

patrickhlauke: Grafted some of the suggested wording back into my PR here - though I removed the explicit mention of "issues when testing" as they're just as applicable to all regular users as well.

patrickhlauke:
@alastc ready to be re-discussed at the next WCAG 2.x meeting, I think

detlevhfischer (changes requested):

mbgower:
> @detlevhfischer all good feedback. happy to implement all proposed changes, but holding off for now while others review it, so as not to change things from under their feet

@patrickhlauke Let's discuss on today's call. if you're fine with these changes, i think it is better to make at least git suggestions sooner than later, so that others can respond.

understanding/20/keyboard-no-exception.html
```diff
@@ -27,8 +27,35 @@ <h2>Intent of Keyboard (No Exception)</h2>
...
+         </p>
+      </div>
+
+      <div class="note">
+         <p>This Success Criterion does not require that every visible control that can be activated
+            using a mouse or touch screen must also be focusable and actionable using the keyboard.
+            The normative requirement is only that there must be a way for keyboard users to perform
+            the same, or comparable, actions and to operate the content. Generally, the easiest way
+            to achieve this is to provide controls that can be operated with all possible input devices;
+            however, if a web application implements a completely separate mode of operation for keyboard
```
patrickhlauke, line 55:
```suggestion
            however, if a web application implements a separate mode of operation for keyboard
```

understanding/20/keyboard-no-exception.html
```diff
@@ -27,8 +27,35 @@ <h2>Intent of Keyboard (No Exception)</h2>
...
+            <kbd>Enter</kbd> key and the <kbd>Space</kbd> bar. If a custom button control
+            in a web application instead only reacts to <kbd>Enter</kbd>
+            (or even a completely custom key or key combination), this still
+            <strong>passes</strong> the requirements of this Success Criterion.
+         </p>
+      </div>
+
+      <div class="note">
+         <p>This Success Criterion does not require that every visible control that can be activated
+            using a mouse or touch screen must also be focusable and actionable using the keyboard.
```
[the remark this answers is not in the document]
patrickhlauke:
```suggestion
            using a pointer (such as a mouse or touch screen input) must also be focusable and actionable using the keyboard.
```

understanding/20/keyboard.html
```diff
@@ -115,8 +141,12 @@ <h2>Examples of Keyboard</h2>
          <dd>A PDA device that is usually operated via a stylus has an optional keyboard that can
                be attached.  The keyboard allows full Web browsing in standard fashion.  The Web
                content is operable because it was designed to work with keyboard-only access.</dd>
+        <dt>Example 7: Simple search form with mouse-operable submit button</dt>
```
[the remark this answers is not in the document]
patrickhlauke:
```suggestion
        <dt>Example 7: Simple search form with pointer-operable submit button</dt>
```

patrickhlauke:
@detlevhfischer i made the changes you suggested, thank you

bruce-usab (approved):
Changes after today's call look good to me.

alastc:
Moved to "For discussion" as Detlev made some suggested updates (which have been incorporated).

detlevhfischer:
[Rendered view](https://raw.githack.com/w3c/wcag/1c04ca48b52d434ad3207649c8ef4589162dad2b/understanding/20/keyboard-no-exception.html) (for ease of reviewing)

understanding/20/keyboard-no-exception.html
```diff
@@ -27,8 +27,35 @@ <h2>Intent of Keyboard (No Exception)</h2>
...
-         Success Criterion and therefore cannot meet Guideline 2.1 at Level AAA. 
+         Success Criterion and therefore cannot meet Guideline 2.1 at Level AAA.
       </p>
+
+      <div class="note">
+         <p>Platforms and user agents usually have conventions for how web content or
+            applications are controlled with a keyboard interface. If content does not follow
+            the platform/user agent conventions it may be difficult to use, as users will need
+            to learn different interaction methods. As a <em>best practice</em>, content
+            should follow the platform/user agent conventions. However, deviating from these
```
detlevhfischer, line 38: If keyboard operaton works via obscure undocumented keys but operation works fine for mouse users (so a clear disability-related issue) - would it not least be required for conformance that that keyboard operation is documented in context (or in help)?

detlevhfischer (changes requested):

understanding/20/keyboard-no-exception.html
patrickhlauke: if you can normatively and authoritatively define what "obscure" is (or what exactly the "platform conventions" are, or what "non-standard" means, or similar ... even though there is no "standard"), possibly. but because these are malleable and undefined concepts, it gets tricky. also, can't start *requiring* anything here as that would make it a normative change that goes beyond what the normative SC text says? can strongly hint at it / suggest it as a best practice, but not require it.

detlevhfischer: I agree it is tricky - but defining "obscure" (not wedded to the term)  does not seem too difficult. It would be anything that flaunts common expectations ("affordances") of interactive elements (mostly that they can be navigated with TAB and arrow keys and activated with ENTER and/or space bar).
If someone created a scripted custom control looking like a button to respond **only** to to CTR+$ to activate, that would be keyboard accessible **if anyone knew it** - but no one would without documentation. Hence the requirement to document the obscurity. Unless you delve into scripts, an evaluation against 2.1.1 would try to activate that pseudo button with ENTER and come to the same conclusion that we have a FAIL or 2.1.1. Would anyone seriously argue that this judgment would be wrong, tha th thing is keyboard-operable, if the only way to discover such operability would be to delve ito scripts to look for event handlers? I believe not...

patrickhlauke: again, this can be added as a strong suggestion, but this interpretation goes beyond the normative wording of the SC which does not impose any such limits. my PR here simply tries to make this fact (that the normative wording may not be as stringent as folks would like)

> All [functionality](https://www.w3.org/TR/WCAG22/#dfn-functionality) of the content is operable through a [keyboard interface](https://www.w3.org/TR/WCAG22/#dfn-keyboard-interface) without requiring specific timings for individual keystrokes, except where the underlying function requires input that depends on the path of the user's movement and not just the endpoints.

patrickhlauke: > Would anyone seriously argue that this judgment would be wrong, that the thing is keyboard-operable and passes 2.1.1

not sure an argument of "surely nobody would try to argue they're fulfilling the exact wording of the normative SC" is reason enough to try and introduce new tighter normative requirements through the backdoor of an informative document

detlevhfischer: Well - many SCs leave room for interpretation. Understanding texts are meant to guide authors and auditors in arriving at a sound judgment in all those cases of "it depends" (TM @patrickhlauke ?). These informative texts can never be definitive and cover all cases. Other SCs have similar wiggle room, and we can and have to live with that (@mbgower had some good points on the meaning of "descriptive" above). In my view in is not helpful to suggest that having **any kind of theoretic and undocumented keyboard operability** would be enough to meet this SC. Explaining what 'keyboard operability' is commonly supposed to mean based user expectations, and offering that as a help for authors/auditors to determine whether they meet the intention of the SC seems more useful than a bare-bones contraction to the point of 'anything will do'. It is certainly helpful to explain that normatively it is be sufficient to support just ENTER and not space bar (one of two expected behaviors), but weird undocumented keyboard commands that no one has a chance of guessing are a different kettle of fish. I strongly object to the change as it is now proposed (for whatever that's worth from an invited expert).

patrickhlauke: > many SCs leave room for interpretation

sure, but they have wiggle room in the normative language. in this case, i don't see that ambiguity/flexibility in the normative SC text that would then warrant defining a *requirement* in the understanding part. i.e. "this fails 2.1.1 because you're using an unusual keyboard shortcut/combination without documenting it" ... where does 2.1.1's normative text hint at this "if you don't document it, it fails this SC"?

detlevhfischer: > where does 2.1.1's normative text hint at this "if you don't document it, it fails this SC"?

I think it is implicit in the term 'operable', which to in my understanding tacitly implies adherence to common keyboard conventions. So if you completely take away the conventions, you'd have to provide some clue how on earth to operate that thing. It boils down to a philosophical argument whether naked words can ever have definitive meaning, or whether implicit context is alway present and needs to be acknowledged. I could dig out some Wittgenstein on that....

understanding/20/keyboard-no-exception.html
```diff
@@ -27,8 +27,35 @@ <h2>Intent of Keyboard (No Exception)</h2>
...
+            conventions does <em>not</em> fail the normative requirement of this Success Criterion.</p>
+         <p>
+            For instance, buttons that have focus can generally be activated using both the
+            <kbd>Enter</kbd> key and the <kbd>Space</kbd> bar. If a custom button control
+            in a web application instead only reacts to <kbd>Enter</kbd>
+            (or even a completely custom key or key combination), this still
+            <strong>passes</strong> the requirements of this Success Criterion.
```
WilcoFiers, line 45: Nitpick; Success criteria aren't passed, they are satisfied.

```suggestion
            <strong>satisfy</strong> the requirements of this Success Criterion.
```

understanding/20/keyboard.html
```diff
@@ -69,6 +69,32 @@ <h2>Intent of Keyboard</h2>
...
+            the platform/user agent conventions it may be difficult to use, as users will need
+            to learn different interaction methods. As a <em>best practice</em>, content
+            should follow the platform/user agent conventions. However, deviating from these
+            conventions does <em>not</em> fail the normative requirement of this Success Criterion.</p>
+         <p>
+            For instance, buttons that have focus can generally be activated using both the
+            <kbd>Enter</kbd> key and the <kbd>Space</kbd> bar. If a custom button control
+            in a web application instead only reacts to <kbd>Enter</kbd>
+            (or even a completely custom key or key combination), this still
+            <strong>passes</strong> the requirements of this Success Criterion.
```
WilcoFiers, line 84:
```suggestion
            <strong>satisfied</strong> the requirements of this Success Criterion.
```

understanding/20/keyboard.html
```diff
@@ -115,8 +141,12 @@ <h2>Examples of Keyboard</h2>
          <dd>A PDA device that is usually operated via a stylus has an optional keyboard that can
                be attached.  The keyboard allows full Web browsing in standard fashion.  The Web
                content is operable because it was designed to work with keyboard-only access.</dd>
+        <dt>Example 7: Simple search form with pointer-operable submit button</dt>
+        <dd>A search form includes a text input field followed by a submit button. The submit button itself
+               has been coded so that it does <em>not</em> receive focus, and can only be activated using a pointer input.
+               However, since keyboard users can submit the search by pressing <kbd>Enter</kbd> in the text input
+               after typing their search terms, the form passes this Success Criterion.</dd>
```
WilcoFiers, line 148:
```suggestion
               after typing their search terms, the form satisfies this Success Criterion.</dd>
```

WilcoFiers (changes requested):
Nitpick

understanding/20/keyboard-no-exception.html
scottaohara: I don't disagree that Instructions should be provided for custom controls / common controls that have modified behaviors.  But I agree that requiring documentation under 2.1.1 is a stretch of what the SC normative text states.    

It would make a lot more sense, in my opinion, to require this under 3.3.2 Labels and Instructions.  Same outcome, unorthodox keyboard functionality is required to be documented for people, but just under an SC that already calls out that instructions need to be provided to people...

understanding/20/keyboard-no-exception.html
patrickhlauke: weirdly (but not surprisingly) WCAG's various understandings and techniques seem to use both "passes" and "satisfies" interchangeably (i.e. there are quite a few cases of "passes"/"fails")

patrickhlauke: the correct replacement here would then be "satisfies" (rather than "satisfy")

understanding/20/keyboard.html
patrickhlauke: the correct replacement here would then be "satisfies" (rather than "satisfied")

understanding/20/keyboard-no-exception.html
detlevhfischer: The concept  of "operable through a keyboard interface" was always meant to support assistive technology like switch input that uses the keyboard interface to activate interactive items by a simple pressing of the switch. If the narrowest reading of 2.1.1 would have to accept  **any keyboard input** to activate the common controls like links and buttons  and their custom-built lookalikes instead of ENTER/space bar, that group of AT users would be excluded (if I get it right) - another point to support the argument that the idea of "keyboard operability" implies supporting the common expected activation methods for the common elements and their lookalikes.

@scottaohara Your point is valid - lack of documentation would fail 3.3.2, but I see this as a separate issue from ensuring keyboard operability in line with common user expectations. .In my view, to satisfy 2.1.1, there would have to be some commonly expected keyboard activation behavior present, and I think this is the case inn the useful examples given (activate with ENTER if space bar fails, activate with enter on the input field when the element after the search field only accepts pointe input). So just to be clear, I do not argue against including the examples in the two notes and think they are helpful.

patrickhlauke:
> @patrickhlauke I made change suggestion which I wanted to submit as separate PR - unfortunately pressed the wrong button so it is a [direct commit](https://github.com/w3c/wcag/pull/1642/commits/9602cbdc22961ba807adb79fbd451a9ec3d8f38d)

so this change partially removes the point of my PR it seems (because I'm still not sold on the idea that "of course, operable implies that they don't use 'non-standard' - whatever that means - key combinations"). I may revert this for now, but feel free to suggest the change again.

understanding/20/keyboard-no-exception.html
patrickhlauke: resolved via https://github.com/w3c/wcag/pull/1642/commits/38cce8aada87e526bc2fe0e1449b073828672819

understanding/20/keyboard.html
patrickhlauke: we have a few other cases in the spec where we use the phrase "passes this Success Criterion", so leaving this as is. however, edited the ones that originally said "passes the requirements", as there it's consistently "satisfies the requirements"

understanding/20/keyboard.html
patrickhlauke: resolved in https://github.com/w3c/wcag/pull/1642/commits/38cce8aada87e526bc2fe0e1449b073828672819

detlevhfischer:
Created two PRs #3670 and #3669 to amend this PR to clarify that normatively accepting any arbitrary key or key combination to satisfy 2.1.1 and 2.1.3 would fail assistive technology users (like switch users) relying on the keyboard interface.

mraccess77:
> Created two PRs #3670 and #3669 to amend this PR to clarify that normatively accepting any custom key or custom key combination to satisfy 2.1.1 and 2.1.3 would fail assistive technology users (like switch users) relying on the keyboard interface.

In the past I thought the group had discussed failing lack of support for "click" events when keyboard interface support was provided might be an SC 4.1.2 issue.

patrickhlauke:
> accepting any custom key or custom key combination to satisfy 2.1.1 and 2.1.3 would fail assistive technology users (like switch users) relying on the keyboard interface.

would it though? do we know for sure? and conversely then, is there a list of "safe" key combinations that DO work for AT users? how do switch etc users operate things that require cursor key interactions, pressing `Esc` to close things, etc? seems a fairly broad statement to say "if you don't follow standards (which we can't actually document/tell you about) then you fail"

patrickhlauke:
> In the past I thought the group had discussed failing lack of support for "click" events when keyboard interface support was provided might be an SC 4.1.2 issue.

i think that discussion was part of some early WCAG 3 musings about what gaps there are in terms of input support (in general) in WCAG 2.x ... I may be wrong though

detlevhfischer:
> do switch etc users operate things that require cursor key interactions, pressing Esc to close things, etc?

Would be interesting to research. But even when some elements should prove problematic in general, that does not mean that 99% (?) of user interface controls out there (links, buttons, checkboxes etc) should not be required to be built in a way that they support the common forms of activation.

understanding/20/keyboard-no-exception.html
scottaohara: common user expectations may also be contingent on the platform.  at a very basic level, macOS and Windows have differences in what keyboard accessibility looks like and users have to adjust expectations if going between platforms.  So what's 'common' can potentially be different degrees of variable.

But then there's the specially made interfaces for specific users that may have been purposefully implemented with non-standard keyboard commands, because they make sense for their audience.  For instance, a game might be created that uses specific keyboard keys to activate different "buttons", or an educational application might have similarly mapped certain keyboard keys to different buttons because it is necessary/appropriate for what they are trying to teach. The "piano" UI that is mentioned in the Orientation rule.

But taking unique interfaces out of the question, I am truly curious if have you found many instances of what it is you seem to be trying to make sure is failed under 2.1.1?  A website/product taking the time to actually make their interface with custom/customized interactive elements keyboard accessible BUT doing so in a way that bucks common expectations for the platform they're on, AND they are not making these obscure keyboard behaviors known to their users, AND they need to be told by WCAG what they've done is wrong, otherwise they won't document said information or change to more common keyboard expectations?

i know i'm laying it on a bit thick there with that last bit, but I'm really struggling with understanding the scope of the potential issue that i think you're seeing.

patrickhlauke:
normatively define "common forms of interaction", then normatively define "operable" to mean "must use those defined common forms of interaction"

bruce-usab:
On call 2/3 (among other discussion) @detlevhfischer suggesting borrowing from 1.1.1 test exception.
> If non-text content is a test or exercise that would be invalid if presented in [text](https://www.w3.org/TR/WCAG22/#dfn-text), then text alternatives at least provide descriptive identification of the non-text content.

alastc:
Summary:

- The new note is correct to the normative language.
- The is concern that it appears to open a loophole, or draw attention to not-good implementations.
- It might be possible to add something to Labels & Instructions to require/encourage documentation for odd keyboard short-cuts.
- It doesn't seem likely that people put the effort into making something keyboard accessible, and then use weird commands, and not provide documentation.
- Note that 2.1.2 does specifically say that instructions are provided, but 2.1.1 does not.

mraccess77:
> > In the past I thought the group had discussed failing lack of support for "click" events when keyboard interface support was provided might be an SC 4.1.2 issue.
> 
> i think that discussion was part of some early WCAG 3 musings about what gaps there are in terms of input support (in general) in WCAG 2.x ... I may be wrong though (biting my tongue hard not to blurt out "accessibility supported", as that's just deferring to another handwavy concept)

To be clear - I'm not referring to pointer input - but programmatic activation as 4.1.2 indicates "and values that can be set by the user can be programmatically set; and notification of changes to these items is available to user agents, including assistive technologies."

I relate this to the switch control discussion as screen readers may use programmatic methods to activate controls in virtual cursor/browse mode.

patrickhlauke:
> To be clear - I'm not referring to pointer input 

I understand, and I believe that (going beyond "it's either a pointer or a keyboard?") was part of the discussion for inputs in general we had a while ago in the WCAG 3 context

understanding/20/keyboard.html
```diff
@@ -50,8 +50,7 @@ <h2>Intent of Keyboard</h2>
          example, clicking, selecting, moving, sizing). However, there is a small class of
          input that is done with a pointing device that cannot be done from the keyboard in
          any known fashion without requiring an inordinate number of keystrokes. Free hand
-         drawing, watercolor painting, and flying a helicopter through an obstacle course are
-         all examples of functions that require path dependent input. Drawing straight lines,
+         drawing, or watercolor painting require path dependent input. Drawing straight lines,
```
patrickhlauke, line 53: thank you @chaals for removing the peculiar helicopter example there...hadn't even clocked it at the time

alastc:
Next step - Detlev agreed to continue with this one, effectively dropping the objection, so we can merge once the review period is over.

understanding/20/keyboard.html
bruce-usab: @patrickhlauke the helicopter example was meant to be a gaming reference.

patrickhlauke: sure, but not really representative, since there's a myriad of games where you can control a helicopter with keyboard, joystick, etc.

mbgower:
> Next step - Detlev agreed to continue with this one, effectively dropping the objection, so we can merge once the review period is over.

@detlevhfischer I'd like to see us also include some wording around documenting unconventional keyboard interactions, so please incorporate into your next changes.

mbgower:
@detlevhfischer and @WilcoFiers although it appears to me that all your comments have been dealt with, the system still thinks there is something unresolved from your reviews, so I've triggered a rereview so that I can merge.
LMK if you have questions!

WilcoFiers (approved):
"Keyboard interface" seems a little odd. An on screen keyboard is an interface, but I wouldn't call the physical device an interface. I can live with it though.

GreggVan:
> "Keyboard interface" seems a little odd. An on screen keyboard is an interface, but I wouldn't call the physical device an interface. I can live with it though.

Hi Wilco -  the keyboard is not  the "interface" referred to.   What is referred to is that the software does not get input from the keyboard -- it gets its "keyboard input" from the software API/ interface to the software.   The actual 'keyboard input' it gets from that "keyboard interface" may come from a physical keyboard,  and onscreen keyboard,  a morse code program,  a speech input system etc.    Hence we don't require that software work with keyboards -- but with the  "keyboard interface (API) that the software gets "keystrokes" from. 

Make sense now?

CityMouse:
recommendation: 
Change "Keyboard interface" to "Keyboard input"

reason for recommendation: for consistency (w/ pointer)  and to reduce confusion
It would would address display rendering of keyboard & hardware keyboard]

mbgower:
@CityMouse This PR was merged before your comment. If you'd like to see this become "Keyboard interface", please open a new issue so that it gets tracked/discussed. Feel free to cross reference this PR.

detlevhfischer (commented):
This looks like what was areed in the Friday call. While I would personally prefer to leave out the bit "(or even a completely custom key or key combination)" because I am not a fan of broadcasting loopholes, I accept the text.

patrickhlauke:
wasn't "keyboard interface" chosen because it matches the actual normative wording of the SC? i originally just had "keyboard" and "keyboard users" etc, but we tweaked it to be more closely aligned with the normative text....

PR #1643, opened 2021-02-15
Add notes to 2.4.3 understanding to try and address static content receiving focus and what does/doesn't constitute a focus order failure

### LINKED ISSUE #1001: 2.4.7 as it applies to static content and programmatic focus

### LINKED ISSUE #2655: 2.4.3 focus order with non interactive element with tabindex="0"
This is the example about ARIA tabpannel that element has tabindex="0".
[ARIA Authoring Practices Guide](https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-automatic.html)

The example where non-interactive elements(tabpannel) would have a tabindex="0" attribute you invalidate 2.4.3?
(In this case, non-interactive elements(tabpannel) would have not a scroll)

### LINKED ISSUE #4634: Static content in tab order - failure of 2.4.3 or not?
A discussion that seems to keep bubbling up (most recently on WebAIM https://webaim.org/discussion/mail_thread?thread=10037) is whether or not static content that receives focus (e.g. because it has been marked with `tabindex="0"`) is a failure of 2.4.3 or not.

My take for the longest time has been that, unless the *order* that these static elements receive focus in makes stuff illogical/affects meaning of the overall content, that this is fine. I don't think static content receiving focus violates the normative wording of the SC

> If a Web page can be navigated sequentially and the navigation sequences affect meaning or operation, focusable components receive focus in an order that preserves meaning and operability.

I also don't see anything in the understanding document that suggests anything about static content being focusable as being a failure.

To be clear, I would generally flag inappropriately focusable static content as something that a developer should address, but I make the point of clarifying that it doesn't strictly (to me at least) fail 2.4.3 (or any other SC). That it's more of a usability problem.

Long story short: it would be good to get a consensus position on this, and to add an example/clarification/note to the understanding document about this.

patrickhlauke:
Closes https://github.com/w3c/wcag/issues/4634
Closes https://github.com/w3c/wcag/issues/2655
Closes #1001 

In short, this touches on aspects of focus order that have always been a bit... "sketchy"/handwaved:

- static/non-interactive content receiving focus (addresses w3c/wcag#4634 and w3c/wcag#2655)
- focus doesn't necessarily need to follow visual layout (as long as this doesn't affect meaning/break any hierarchy/relationship implied by the layout) (no related issue for this, but as I was already proposing updates to 2.4.3 understanding, added this constant doubt/bugbear here)

Deploy preview for [Understanding SC 2.4.3 Focus Order](https://deploy-preview-1643--wcag2.netlify.app/understanding/focus-order) and [diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FUnderstanding%2Ffocus-order&doc2=https%3A%2F%2Fdeploy-preview-1643--wcag2.netlify.app%2Funderstanding%2Ffocus-order).


understanding/20/focus-order.html
```diff
@@ -54,6 +54,31 @@ <h2>Intent of Focus Order</h2>
...
+         Again, as a best practice, care should be taken to make sure the focus order does not
+         appear random and illogical.
+      </p>
+
+      <div class="note">
+         <p>Normatively, it is not a failure of this success criterion if static/non-interactive content
+            receives focus. Making an element that does not act as an actionable user interface component
+            focusable, or programmatically moving focus to such an element, does not necessarily constitute
+            a failure of this success criterion. Similarly, redundant focusable elements – for instance,
+            having two or more "focus stops" for what may visually appear to be a single user interface
```
scottaohara, line 73: might be useful for a clarifying example here, especially since the example of nested focusable elements is used later as a possible mean of failure.

regarding an example that I would say falls into the "not a failure but annoying" category would be a "card" like pattern.  where an image and heading text are both separate links, but could/should have been a single link.

patrickhlauke: expanded this a little bit. danger is adding too much HTML in understanding, as it's supposed to be slightly more "tech agnostic" ...

scottaohara: understood, but i think the addition is helpful.  thank yoiu

mbgower:
I'm a lot more concerned with redundant focus stops (i.e., the focus does not appear to move) than I am on say focus moving to a tab panel. To the point where I'm a little concerned with this: "Redundant focus stops...are allowed..." If something is truly redundant, as in there is no change in focus  (which is different than a link within a card you mention, IMO), I think that should not pass.

Whereas, there are several situations where it may be necessary to put focus on a non-interactive component. Some may be needed to put the reading order there (think of a 'go to top' button), even a skip to main may require that (focus goes to region, for example). Think of a tablist; you want to put focus in a tab panel to allow screen reader consumption, even though there is no operable control in the tab panel.. The same thing can happen on a modal overlay in which nothing is operational (strictly informational). If focus isn't placed there, it may prevent a keyboard user from scrolling up and down in the content.

patrickhlauke:
> I'm a lot more concerned with redundant focus stops (i.e., press Tab and the focus does not appear to move) [...] If something is truly redundant, as in there is no change in focus (which is different than a link within a card you mention, IMO), I think that should not pass.

Not sure I follow what you mean exactly. Assuming it's something like it *appears* not to change, but practically does? Something silly like

`<div tabindex="0"><button>Test</button></div>`

perhaps?

Would you say those fail 2.4.3? And If so, how would you disallow this sort of thing / define it in a way that still allows for the non-problematic cases where a non-interactive element receives focus?

patrickhlauke:
Even removing the wording around redundant, wouldn't something like

```
<div tabindex="0"><button>Test</button></div>
```

still not be understood as ok based on the rest of the note as well, based on the loose wording of

> Normatively, it is not a failure of this success criterion if static/non-interactive content
> receives focus. Making an element that does not act as an actionable user interface component
> focusable, or programmatically moving focus to such an element, does not necessarily constitute
> a failure of this success criterion.

mbgower:
I think the example you provide should be considered a failure of 2.4.3:
> If a [Web page](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-focus-order.html#webpagedef) can be [navigated sequentially](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-focus-order.html#nav-seqdef) and the navigation sequences affect meaning or operation, focusable components receive focus in an order that preserves meaning and operability. (Level A)

It's not a serious error, but it _is_ confusing and disruptive to press a tab and have nothing happen. If I was testing a page and 2 tabstops occur on the same item, I'd ask that to be remediated, and if the answer came back "it's intentional", we'd be having a discussion on why.

if we start putting language in saying such redundancy is 'okay', I think we need to be specific about what contexts it is okay. In the tile/link on tile example, if I understand what you are saying, two different focus indicators take place. Visually the focus has progressed as the user presses tab. So I wouldn't consider that a failure here. (I've seen examples where activating both may result in the same action, which I wouldn't call great design, but I think is orthogonal to Focus Order).

detlevhfischer (commented):
static content: "does not necessarily constitute a failure" and then: "These situations are only failures if they significantly impede the operation of the content". With focus stops on static content (e.g. all explanatory texts in a form) is remains a judgment call at what point this becomes a significant impediment. Which I believe is unavoidable - I'd call it not a "subjective" but a "context-dependent" aspect of conformance evaluation.
Would a few negative examples help to clarify what cases would be seen to fall under "significantly impede" and in turn, be cases to FAIL?

understanding/20/focus-order.html
```diff
@@ -54,6 +54,32 @@ <h2>Intent of Focus Order</h2>
...
+         of content implied by the visual presentation is preserved. For instance, in a two-column
+         layout, visually the assumption in western left-to-right systems may be that elements in the
+         left-hand column would receive focus first. However, if meaning and operation are not affected,
+         it is not a normative failure if elements in the right-hand column receive focus first.
+         Again, as a best practice, care should be taken to make sure the focus order does not
+         appear random and illogical.
+      </p>
+
+      <div class="note">
+         <p>Normatively, it is not a failure of this success criterion if static/non-interactive content
```
alastc, line 69: Most people generally don't know what "Normatively, " means, I think it can be removed without affecting the meaning.

alastc:
Historically we've been quite strict on redundant tab-stops, on the basis that stopping twice on the same thing is confusing because you expect to go to the next thing. A common (ish) example is a listing on an ecommerce site and you stop once on each item which isn't active, press tab again and find the active tab-stop. 

It fits in with failing invisible/off-screen items under focus-visible.

I'm not sure where I'd drawn the line if you had a couple of instances vs 20?

understanding/20/focus-order.html
alastc: I disagree in this context. We agree that some people won't understand it, and (more importantly) no meaning is lost by removing it. We are still saying that "This fails".

mbgower:
> I'm not sure where I'd drawn the line if you had a couple of instances vs 20?

I would argue that "a couple" is a still a failure of Focus Order -- it is just the degree of impact/severity that is affected. If I press on the same item many times with no apparent change in focus, that is more severe. Some users will just abandon the site. If I have to press Tab twice on every single item to advance it, that is more severe, due to time on task, etc. 
If it happens once on the page, it may just be confusing -- especially if the user just presses again. But the user has NO idea if something important was intended to happen and it didn't.

patrickhlauke:
I've removed the explicit mention of "redundant" in this PR

understanding/20/focus-order.html
patrickhlauke: Remove the various "normatively"/"normative" mentions. Agree it sounded a bit "spec-head-y"

mbgower:
I like how it's ended up with @patrickhlauke's latest changes

 but just to counter your comment, @detlevhfischer...
> @mbgower I would argue that (most?) screen reader users will have a good idea since if it is just a p with tabindex="0" they won't get an interactive role like link or button exposed. So I feel it would be overly strict to fail content on 2.4.3 for a couple of tab stops on non-interactive content. If we are unable to apply tolerances for minor issues, we set the bar too high, iMO.

1. Focus Order is not just for screen reader users. The majority of keyboard users don't get info on the role. 
2. I was not arguing that stops on non-interactive content should fail (there are a few good cases for them [I mentioned earlier](https://github.com/w3c/wcag/pull/1643#issuecomment-1112546310)). I was arguing that tab stops where nothing seems to change/happen (focus doesn't move perceptibly) should fail.

bruce-usab:
> @bruce-usab - What was the aim when the SSA(?) developers made a paragraph tab-able? 
> In that scenario, if it is clear where you are and it isn't a duplicate I don't think I'd fail that, but I'm curious about the scenario.

At the time, their very strong opinion was that, for public facing forms, someone using screen reading software should not have to go into and out of <q>forms mode</q> (or even know about forms mode) in order to navigate and fill the web form.

alastc:
Hmm, that would likely be the first time a SR user encountered this approach, which would be confusing. For complex questions and a well-tested approach, I'd refer to the [text-input GDS component](https://design-system.service.gov.uk/patterns/question-pages/#asking-complex-questions-without-using-hint-text).

patrickhlauke:
> that would likely be the first time a SR user encountered this approach,

often see this practice (again, in more naive intranet type big applications) where developers misunderstood how SR users navigate and just test by tabbing...

detlevhfischer:
> * Focus Order is not just for screen reader users. The majority of keyboard users don't get info on the role.
> * I was not arguing that stops on non-interactive content should fail (there are a few good cases for them [I mentioned earlier](https://github.com/w3c/wcag/pull/1643#issuecomment-1112546310)). I was arguing that tab stops where nothing seems to change/happen (focus doesn't move perceptibly) should fail.

@mbgower OK, I think we agree. If tab stops have visble focus on text (such as instructions in forms) it would also be clear to sighted users that the stop is non-interactive. Not arguing here that this is a good practice, though.

bruce-usab:
Thanks @alastc for that example!  In my experience, seasoned SR users would not be too confused since, while unexpected, it small potatoes compared to the daily obstacles and surprises.  (Things used to be much worse.  E.g., Early SR ignored CSS hidden property -- so <q>well meaning</q> developers used that bug to provide SR-specific instructions.)

Yes @detlevhfischer the practice stems from developers who misunderstand how SR users typically navigate.

mbgower:
A couple examples of where focus on non-interactive components occurs.
First, in the context of a tabpanel or overlay that contains no items that take focus but a substantial amount of text exists, to the point where it cannot all be seen in the viewport, and so the focus needs to be put somewhere in order for the keyboard user to scroll content.
Second, in the case of a tabpanel without a focusable element. When the user presses Tab to move from the tablist to the tabpanel, the entire contents of the tabpanel are bypassed, resulting in a screen reader potentially being unaware of the existence of the content. A focus indicator is placed around the whole panel, providing the ability to scroll and orienting the screenreader to the content. All keyboard users experience the same focus order context. It's tough to say that should be a failure.
The wording of the [APG](https://www.w3.org/TR/wai-aria-practices-1.2/#tabpanel) even suggests this is expected behaviour:
> When the tab list contains the focus, moves focus to the next element in the page tab sequence outside the tablist, which is the tabpanel unless the first element containing meaningful content inside the tabpanel is focusable.

Both of the APG [examples](https://www.w3.org/TR/wai-aria-practices-1.2/examples/tabs/tabs-1/tabs.html) implement that interaction

understanding/20/focus-order.html
```diff
@@ -54,6 +54,28 @@ <h2>Intent of Focus Order</h2>
...
+         Again, as a best practice, care should be taken to make sure the focus order does not
+         appear random and illogical.
+      </p>
+
+      <div class="note">
+         <p>It is not a failure of this success criterion if static/non-interactive content
+         receives focus. Making an element that does not act as an actionable user interface component
+         focusable, or programmatically moving focus to such an element, does not necessarily constitute
+         a failure of this success criterion. This situation is only failures if these additional focus stops
+         significantly impede the operation of the content, or create illogical focus orders - for example, a control
```
alastc, line 73: I don't think "significantly impede the operation of the content" is really covered or implied by the SC text, can we remove that and just have it as "create an illogical focus order"?

understanding/20/focus-order.html
```diff
@@ -54,6 +54,28 @@ <h2>Intent of Focus Order</h2>
...
+         appearing to receive focus multiple times due to the use of nested focusable elements.</p>
+         <pre><code>&lt;div tabindex="0"&gt;&lt;button&gt;...&lt;/button&gt;&lt;/div&gt;</code></pre>
+         <p>As a best practice, authors should avoid excessive use of focusable elements which cannot be operated
```
alastc, line 76: We get dinged on use of "should", how about:
"As a best practice avoid excessive use of..."

understanding/20/focus-order.html
```diff
@@ -54,6 +54,28 @@ <h2>Intent of Focus Order</h2>
...
          sets of users and does not appear to either of them to jump around randomly. 
       </p>
+
+      <p>Focus order also does not need to follow the visual presentation of the web page,
+         as long as the order in which elements receive focus is logical, and the hierarchy and relationship
+         of content implied by the visual presentation is preserved. For instance, in a two-column
+         layout, visually the assumption in western left-to-right systems may be that elements in the
+         left-hand column would receive focus first. However, if meaning and operation are not affected,
+         it is not a failure if elements in the right-hand column receive focus first.
+         Again, as a best practice, care should be taken to make sure the focus order does not
```
alastc, line 64: Suggest: "As a best practice, make sure the focus order does not..."

understanding/20/focus-order.html
mbgower: I would remove the word "excessive" here. People _should_ avoid use of focus on inoperable elements. Having to do so implies, to me, some failure in the design. If we're saying it's a best practice, I don't think we need to quantify it.

mbgower:
@patrickhlauke sorry to contribute to what has become messy. I'm aware I'm vacillating back and forth on this one. The reality to me is sometimes it's useful to have an SC like this that's a little malleable. Yes, it creates differences of opinion, but it also gives a place to potentially fail something (or at least call it out) where one knows there's a problem with the keyboard experience and it doesn't really fit anywhere else.

So when we try to create clarity, even in a non-normative document, if our interpretation tightens down scope, in some wacky way it potentially disenfranchises people trying to improve a site via WCAG. I'm not advocating we _not_ update guidance, but I do think that the longer an SC has been around for, the more we have to be careful with our work.

understanding/20/focus-order.html
bruce-usab: I agree with removing the word <q>excessive</q>.  I am glad not have tab stop on `P` be a failure, but agree that developers should avoid use of focus on inoperable elements.

understanding/20/focus-order.html
patrickhlauke: Done

understanding/20/focus-order.html
patrickhlauke: Changed to @mbgower's suggestion

understanding/20/focus-order.html
patrickhlauke: Tweaked

patrickhlauke:
probably splitting hairs, but: what if it *appears* random or illogical, but doesn't affect meaning / operation? taking a silly example, something like

```
<a href="..." tabindex="2">something</a>
<a href="..." tabindex="1">something else</a>
<a href="..." tabindex="3">and another</a>
```

say there's no real reason for those links to visually appear in that order, and no meaning is lost by focusing on them in that order. is that a failure of 2.4.3?

mbgower:
Note that my suggested wording didn't say 'it's a failure'.

Your random order in a sequence of tool buttons is a good example of something I probably wouldn't fail if they were all together in a toolbar but would question (if it was an internal product I was testing). If it was jumping between different parts of the screen, it gets closer to something I might consider failing. It's ultimately going to be a matter of degree, I suspect, which is why I get a little nervous with just 'best practice'

I'm probably splitting hairs here. I'm comfortable with something you're comfortable with for an Understanding doc. Thanks!

patrickhlauke:
> It's ultimately going to be a matter of degree, I suspect, which is why I get a little nervous with just 'best practice'

likewise, I'm probably nervous from the other end that an unqualified "make sure that it doesn't appear illogical or random" (even if it doesn't affect meaning) seems to lean more towards "otherwise it's a failure".

bruce-usab:
I am less comfortable with characterizing logical/not-random as mere best practice than I am with an unqualified should appear logical/not-random.  But do I agree that implying <q>otherwise it's a failure</q> is a concern as well.  How about:
> It is important that focus order not appear random or illogical.

Just how important can be left as an exercise for the reader...

patrickhlauke:
it needs some "glue" in front though, as it's part of a whole bit about how in that case, it's not a failure, etc.

so, how's about this, which is slightly stronger, without trying to appear too prescriptive:

changing

```
However, if meaning and operation are not affected, in this case it is not a failure if elements in the right-hand
column receive focus first. As a best practice, though, make sure the focus order does not appear random or illogical.
```

to

```
However, if meaning and operation are not affected, in this case it is not a failure if elements in the right-hand
column receive focus first. **In general**, though, make sure the focus order does not appear random or illogical.
```

bruce-usab:
That is very good!  Just to nit pick, why the <q>, though,</q>?  The stand-alone sentence seems fine:

> In general, make sure the focus order does not appear random or illogical.

patrickhlauke:
because without it, to me it feels like it's a contradiction of the sentence that comes immediately before it

patrickhlauke:
any movement on this?

mbgower:
I have just made a commitment to address the last few back and forths from [May 19](https://github.com/w3c/wcag/pull/1643#issuecomment-1132022932) and 20.
i have removed "illogical" since the text earlier states that focus order needs to be logical. Since you have basically discussed how L to R languages read, I have just stated that baldly here. I debated adding it as well in the the preceding wording, and that remains an option, but I think this reads fine.

mbgower:
I was a bit more aggressive and removed "random" too. 
Feel free to reject those two commits if you feel strongly, but the old 'when in doubt, take it out' seems to be relevant here.

patrickhlauke:
thanks @mbgower ... happy enough with those edits (at this stage, just keen to see the bulk of this advice make it live somewhere, to be honest, as it's something that i keep having to discuss with folks :) )

bruce-usab:
> I have just made a commitment to address the last few back and forths...

Thanks Mike!

patrickhlauke:
Any chance this could be looked at/merged soon? @alastc

bruce-usab (approved):

patrickhlauke:
ping...

patrickhlauke:
any chance this could be looked at/progressed at some point? only asking because i've had a few more questions from other auditors specifically about this point

mbgower (approved):
I think this is an improvement. Thanks for ushering it through.

patrickhlauke:
related to this, would be interested in other WG members' thoughts on https://github.com/w3c/wcag/issues/3029 ...

understanding/20/focus-order.html
```diff
@@ -55,6 +53,28 @@ <h2>Intent of Focus Order</h2>
          sets of users and does not appear to either of them to jump around randomly. 
       </p>
       
+      <p>Focus order also does not need to follow the visual presentation of the web page,
+         as long as the order in which elements receive focus is logical, and the hierarchy and relationship
+         of content implied by the visual presentation is preserved. For instance, in a two-column
+         layout, visually the assumption in western left-to-right systems may be that elements in the
```
mbgower, line 59: Given the preceding sentence, I don't think "visually" is needed here, and it's making the sentence a bit harder to read
```suggestion
         layout, the assumption in western left-to-right systems may be that elements in the
```

patrickhlauke: fairy snuff

understanding/20/focus-order.html
```diff
@@ -55,6 +53,28 @@ <h2>Intent of Focus Order</h2>
...
+         layout, the assumption in western left-to-right systems may be that elements in the
+         left-hand column would receive focus first – and, more generally, that focus order proceeds across lines,
+         from components in the top-left corner of the viewport to the bottom-right corner.
+         However, if meaning and operation are not affected, in this case it is not a failure if elements in the right-hand
+         column receive focus first. As a best practice, though, make sure the focus order reinforces the reading order.
+      </p>
+      
+      <div class="note">
+         <p>It is not a failure of this success criterion if static/non-interactive content
+         receives focus. Making an element that does not act as an actionable user interface component
```
mbgower, line 68:
```suggestion
         receives focus. Making an element focusable that does not act as an actionable user interface component
```

understanding/20/focus-order.html
```diff
@@ -55,6 +53,28 @@ <h2>Intent of Focus Order</h2>
...
+         focusable, or programmatically moving focus to such an element, does not necessarily constitute
```
mbgower, line 69:
```suggestion
         or programmatically moving focus to such an element, does not necessarily constitute
```

understanding/20/focus-order.html
patrickhlauke: to me, that reads much clunkier

mbgower: K, no problem. Feel free to reject it.

understanding/20/focus-order.html
```diff
@@ -55,6 +53,28 @@ <h2>Intent of Focus Order</h2>
...
+         a failure of this success criterion. This situation is a failure if these additional focus stops
+         impede the operation of the content, or create illogical focus orders - for example, a control
+         appearing to receive focus multiple times due to the use of nested focusable elements.</p>
+         <pre><code>&lt;div tabindex="0"&gt;&lt;button&gt;...&lt;/button&gt;&lt;/div&gt;</code></pre>
+         <p>As a best practice, avoid having focusable elements which cannot be operated
+         or actioned, as these are likely to make operation unnecessarily tedious for keyboard users.</p>
```
awkawk, line 75: Recommend removing "unnecessarily" - is tedious ever a design decision borne out of necessity?

understanding/20/focus-order.html
```diff
@@ -63,6 +83,8 @@ <h2>Intent of Focus Order</h2>
             operability only when navigation sequences affect meaning and operability.
          </li>
          
+         <li>Focus order does not need to follow visual presentation or layout.</li>
```
awkawk, line 86: I don't think that I agree with this statement. The issue is that the interpretation of the visual presentation is subjective.

patrickhlauke: does not *necessarily* need to follow visual presentation, perhaps? i've seen extreme cases where auditors gleefully fail things if they're not exactly in the same order as the visual, with things like three little icon-based buttons that receive focus right-to-left instead of left-to-right (because the author used CSS floats that effectively reversed the order) ... but where the meaning was not impacted and it was not really a problem.

there must be some way we can somehow make that clear...

understanding/20/focus-order.html
patrickhlauke: yup, good catch

understanding/20/focus-order.html
patrickhlauke: wonder if it's too handwavy though. maybe best to expand this, or add some more words that reflect what @awkawk rightly says...that it's subjective?

alastc, line 86:
```suggestion
         <li>Focus order does not necessarily need to follow visual presentation or layout.</li>
```

patrickhlauke: "Who is deciding what is the visual presentation order?" not sure i follow...is this rhetorical? the author/designer will work on the visual/layout, but for various reasons the focus order may be slightly different. the point is that as long as meaning/operation aren't affected (and it's...subjectively...not too haphazard jumping all over the place) it's not necessarily a failure if the focus order doesn't strictly follow the western top-to-bottom/left-to-right (in left-to-right writing cultures)

alastc:
We'll need to come back to this, some concern over what the 'visual order' would be, theres some assumptions in there. Can we adjust to avoid that? (See the minutes.)

patrickhlauke:
@alastc 

> (See the minutes.)

have they been published yet?

awkawk:
@patrickhlauke part of the discussion was that this specific part of the fix doesn't seem to be related to the issues that were raised, which was about non-interactive items in the tab order. There was discussion that there was an additional issue that was related, but no one knew what it was.

patrickhlauke:
> part of the discussion was that this specific part of the fix doesn't seem to be related to the issues that were raised

quite possibly, but since I was already working on the 2.4.3 understanding, I opted to also add wording to cover that aspect (which I encounter constantly in discussions) without first filing an issue.

so what's the idea? make this only about the two issues, and then file another issue for this other aspect and wait another 3+ years until we get around to addressing THAT?

patrickhlauke:
digging through the issue, I believe that originally (back in Feb 2021) the PR was more focused on just those two referenced issues. as we started wordsmithing and expanding things here, the additional more general "doesn't necessarily have to match layout" part came in as well (around the time of https://github.com/w3c/wcag/pull/1643#issuecomment-1132028605 perhaps)

awkawk:
Suggestions for the PR:
Lines 56-64 suggestion:
Focus order needs to allow the user to navigate focusable elements in a logical order, and that order needs to preserve any meaning or operation that the page is conveying. For instance, focus order is often considered to generally run left-to-right on sites with left-to-right text, but that is subject to the specific situation. For example, a video player with a row of button controls along the bottom of the player could implement a tab sequence that is left-to-right, but there might also be an important control such as a play/pause control which is visually centered and precedes the other controls in the tab order. In this case, meaning and operation are not affected, so it is not a failure if elements in the right-hand column receive focus first. As a best practice, though, make sure the focus order reinforces the reading order.

Line 70-76: I agree with the text content, but I’m not sure about the example. I am certainly not going to advocate for links in buttons (or the inverse) but I’m also not sure how to rationalize failing it. The strongest argument is that users may focus an item that can’t be actioned by the user, but of course that is what we are allowing with the setting of focus on non-interactive elements also… I think that we need a different example. 

Line 86:
This addition on line 86 is covered by the bullet items before it and after it: “Focus order does not necessarily need to follow visual presentation or layout.” I think that we can remove it.

patrickhlauke:
Made a few tweaks/changes to try and disambiguate things a bit more:

* while @awkawk's example of video player controls is an interesting one, I do quite like the two-column layout example as well. As the original example was a bit muddled, I tried to write it out a bit more explicitly (removing the confusing "across lines" part). we could add a further example here for cases like the video player buttons, but then it feels like we're probably overloading that one paragraph and get more into "this needs a whole new section, ideally with diagrams" territory, which I don't mind doing, but not before we get some agreement that that's what we want to do
* the "nested focusable elements where only one is actionable" example is still a good one I feel. I added the more subjective "too many" to the best practice sentence to try and square the circle of telling authors that yes, it's ok if non-interactive elements receive focus, just don't overdo it
* in the "For clarity" bullet list, I removed the "only when navigation sequences affect meaning and operability" part on the first bullet as it seems like a meaningless truism "make sure elements receive focus in an order that preserves meaning/operability only in situations where the focus order affects meaning/operability". well duh. Also remove the bullet "In those cases where it is required, there may be more than one order that will preserve meaning and operability" as, unless I'm missing something, this makes no sense? and I think it's covered better by the following bullet again about multiple orders"

mbgower (approved):

understanding/20/focus-order.html
```diff
@@ -53,25 +53,27 @@ <h2>Intent of Focus Order</h2>
...
          focusable, or programmatically moving focus to such an element, does not necessarily constitute
-         a failure of this success criterion. This situation is a failure if these additional focus stops
-         impede the operation of the content, or create illogical focus orders - for example, a control
-         appearing to receive focus multiple times due to the use of nested focusable elements.</p>
+         a failure of this Success Criterion. This situation is a failure if these additional focus stops
+         impede the operation of the content, or create confusing and/or illogical focus orders - for example,
+         a control appearing to receive focus multiple times due to the use of nested focusable elements.</p>
          <pre><code>&lt;div tabindex="0"&gt;&lt;button&gt;...&lt;/button&gt;&lt;/div&gt;</code></pre>
-         <p>As a best practice, avoid having focusable elements which cannot be operated
+         <p>As a best practice, avoid having too many focusable elements which cannot be operated
```
mbgower, line 76: I wonder if instead of "too many" this could be something like "unwarranted"? 
I think "too many" is overly generous.

mbgower:
@patrickhlauke 
> while @awkawk's example of video player controls is an interesting one, I do quite like the two-column layout example as well. As the original example was a bit muddled, I tried to write it out a bit more explicitly (removing the confusing "across lines" part). 

A bit more explanation of what I heard in the discussion. There was some confusion over the lack of division between reading order and focus order, and the 2-column example seemed to enforce that.
One idea I had... If you made that 2-column example more precise, it might help. So, how about talk about a page with a left navigation layout where, instead of the focus moving through the items in the left nav and then proceeding to the main content, the focus goes across by line, so that the focus jumps between the left nav and a link that happens to fall on the same line in the main content.

patrickhlauke:
@mbgower 

> instead of the focus moving through the items in the left nav and then proceeding to the main content, the focus goes across by line, so that the focus jumps between the left nav and a link that happens to fall on the same line in the main content.

that would be a clear failure. my point was that there are cases where it's fine to go right-hand column first, the left-hand column. for instance, left-hand is navigation, right-hand is content, but the author decided that focus goes through the content first, and THEN the navigation. would that example work?

patrickhlauke:
it would be nice to get this considered at some point ...

giacomo-petri:
During my work on the "little tweak," I've come to realize that I have more concerns than initially anticipated.

In my perspective, the current proposal's allowance for including static elements in the tab sequence is overly permissive, leading to subjective judgments. For example, if we envision a scenario with a thousand sibling nodes on a page, each assigned a tabindex="0" attribute, the proposal views this as a best practice rather than a violation of WCAG. 
Discussions within the Silver WG have revolved around this topic, addressing user needs such as:
- As a user of a keyboard, I should be able to complete actions in a comparable amount of time to other users.
- As a user of a keyboard, the complexity and number of input commands required to complete a task should be comparable with those of a user of other methods of input.

Considering that the normative area of SC 2.4.3 emphasizes how "navigation sequences affect meaning or operation," it's important to recognize that making static elements tabbable also impacts meaning and operation. **This is because reliance on focusability to ascertain operability becomes compromised**. Indeed, encountering tabbable elements that are inoperable depending on context can be frustrating for users.

Take, for example, this image displaying a collection of static logos: 

collection of static logos
- By using a mouse and hovering over these elements, it's apparent they are not operable.
- When navigating with a keyboard and skipping past them during tabbing, their inoperability becomes evident.

Yet, if the author makes these elements tabbable, keyboard users could mistakenly presume they are functional. 
This circumstance raises questions about whether confusion is being generated or not, since it's not fully covered by examples.

The web is replete with such scenarios. Consider another instance: non-operable cards showcasing various sizes of luggage below:

static cards containing different baggages options for an airline company

While we might concur that the layout is suboptimal, if these elements were made "tabbable", keyboard users would undoubtedly anticipate being able to interact with them.

Rather than solely highlighting failing examples and deeming everything else as passing, it could be beneficial to adopt the opposite approach by explaining when having a focusable, tabbable element in the tab order is deemed acceptable. Alternatively, including both successful and unsuccessful examples, with a focus on the circumstances under which adding a static element to the tab order could be advantageous, and conversely, when it should be considered a failure, could provide a more comprehensive understanding. Additionally, integrating visual examples with explanations may further enhance comprehension.

Lastly, it's common to observe developers adding tabindex="0" because they've noted that "while using a screen reader, I'm unable to audibly perceive it while tabbing."

I firmly believe that maintaining a robust and consistent method for presenting elements is advantageous for all users. 
Therefore, I advocate for limiting tabbing accessibility to only user interface components, with the reasonable exceptions mentioned within the issue comments such as:
- The author facilitates users to bypass a set of operable controls and move directly to the next area, such as newly activated tab panel content accessed via a tab within a tab list.
- The author enables users to focus on a scrollable container that exclusively contains static elements.
- The author allows users to focus on crucial instructions within a form, ensuring that screen reader users navigating the page in "forms mode" can readily access these instructions.
- The author allows users to tab through a toolbar container, facilitating navigation of elements using arrow keys.


Note: Upon revisiting my feedback and examining the current proposal last time, I've read it from a different perspective (concerned that I might have misunderstood the concept entirely) regarding the term "focusable." As outlined in the issue description, the aim is to address "focusable and tabbable" elements, yet the proposal only references "focusable." It's important to clarify that being focusable does not automatically mean inclusion in the tab order. Nevertheless, I believe it could be valuable to allocate some words to clarify the intent within the proposal, particularly for individuals who may not have access to the issue history.
Indeed, understanding the proposal with "focusable" referring to elements that can receive focus but not necessarily be tabbable holds a distinct meaning compared to "focusable and included in the tab order."

patrickhlauke:
> In my perspective, the current proposal's allowance for including static elements in the tab sequence is overly permissive, leading to subjective judgments.

> explaining when having a focusable, tabbable element in the tab order is deemed acceptable

then we want to make damned sure we cover all possible acceptable cases, because otherwise by implication we're saying that anything not covered there will automatically FAIL the criterion

> It's important to clarify that being focusable does not automatically mean inclusion in the tab order.

in terms of focusable and receiving programmatic focus, I would hope that you're at least suggesting that it's kosher to set programmatic focus anywhere the author deems necessary, at least

understanding/20/focus-order.html
```diff
@@ -64,30 +63,54 @@ <h2>Intent of Focus Order</h2>
...
+         a failure if elements in the right-hand column receive focus first, followed by the elements in the left-hand column.
+         As a best practice, though, make sure the focus order reinforces the reading order implied by the visual layout.
+      </p>
+      
+      <div class="note">
+         <p>It is not a failure of this Success Criterion if static/non-interactive content
+         receives focus. Making an element that does not act as an actionable user interface component
+         focusable, or programmatically moving focus to such an element, does not necessarily constitute
+         a failure of this Success Criterion. This situation is a failure if these additional focus stops
+         impede the operation of the content, or create confusing and/or illogical focus orders - for example,
```
mfairchild365, line 83: I whole heartedly agree with this proposed language, however I've found that testers have trouble with determining when additional focus stops impede the operation of the content. For example, is just one unnecessary tab stop enough to say that it impedes the operation? 3? 6? Does it depend on context? It may be worthwhile to establish some sort of bar/test for which to determine this consistently. I don't have an immediate suggestion, but I'll think on it.

mfairchild365: Perhaps I should rea the convo in the pull request before posting a comment directly in the file. I see that there is already discussion on this. 😅

understanding/20/focus-order.html
```diff
@@ -62,30 +62,54 @@ <h2>Intent of Focus Order</h2>
...
```
mbgower, line 82:
```suggestion
         <p>This requirement does not specify what should or should not receive focus, but rather addresses the order in which elements receive focus. This Success Criterion does not prohibit making non-operable content (e.g., static text) focusable, or programmatically moving focus to such content. However, it is a failure of Focus Order if items receive focus in an order that impedes the operation of content, or creates confusing or illogical focus orders — for example,
```

understanding/20/focus-order.html
```diff
@@ -62,30 +62,54 @@ <h2>Intent of Focus Order</h2>
...
-         <li>In those cases where it is required, there may be more than one order that will preserve
-            meaning and operability.
-         </li>
+         <li>Focus order does not necessarily need to follow the visual presentation or layout.</li>
          
          <li>If there is more than one order that preserves meaning and operability, only one of
             them needs to be provided.
          </li>
          
+         <li>Static/non-interactive elements can receive focus, as long as they don't significantly
```
mbgower, line 103: We believe removing "significantly" is prudent. "Impede" already has a concept of hindering.
```suggestion
         <li>Static/non-interactive elements can receive focus, as long as they don't
```

understanding/20/focus-order.html
```diff
@@ -62,30 +62,54 @@ <h2>Intent of Focus Order</h2>
...
+            impede operation of the content.
+         </li>
```
mbgower, line 105: We believe removing "significantly" is prudent. "Impede" already has a concept of hindering.
```suggestion
         <li>Static/non-interactive elements can receive focus, as long as they don't
            impede operation of the content.
         </li>
```

giacomo-petri: I recommend using "impede or mislead" to ensure we also address scenarios like those mentioned in my [previous comment](https://github.com/w3c/wcag/pull/1643#issuecomment-2027644498).

Additionally, I believe we should discourage readers from implementing solutions like:

```
document.querySelectorAll("*").forEach(node => {
    node.setAttribute("tabindex", "0");
});
```

as the current wording might imply such practices are acceptable.

patrickhlauke:
@giacomo-petri 

> if we envision a scenario with a thousand sibling nodes on a page, each assigned a tabindex="0" attribute, the proposal views this as a best practice rather than a violation of WCAG

late reply, but: WCAG never failed that scenario, there IS no WCAG failure there. it's always been a best practice/usability problem, rather than a hard WCAG failure, and the normative wording doesn't say anything there either.

we can try and discourage that practice, but not suggest it's a hard failure, i think

jamieherrera:
After reading many of the comments here and the original posts, I'm curious where this stands for a comment on static content receiving focus for 2.4.3 in the Understanding document? Inquiring minds want to know :)  @patrickhlauke since you assigned yourself maybe you have an updated timeline?

mbgower:
I believe the current wording in the PR is valid, and am moving this PR to "Drafted".

There are a number of points raised in this thread which cannot be addressed with the normative wording of WCAG 2. They may be possible considerations for WCAG 3.0.

understanding/20/focus-order.html
```diff
@@ -62,30 +62,50 @@ <h2>Intent of Focus Order</h2>
...
-            meaning and operability.
-         </li>
+         <li>Focus order does not necessarily need to follow the visual presentation or layout.</li>
          
          <li>If there is more than one order that preserves meaning and operability, only one of
             them needs to be provided.
          </li>
          
+         <li>Static/non-interactive elements can receive focus, as long as they don't
+            impede operation of the content.
```
giacomo-petri:
```suggestion
         <li>Static/non-interactive elements can receive focus, as long as they don't
            impede operation of the content, or result in confusing or illogical focus order.
```

giacomo-petri (approved):

bruce-usab (approved):
Mentioned on TF call 9/19/2025.

patrickhlauke:
there's been recent activity on https://github.com/w3c/wcag/issues/4634#issuecomment-3323259545

understanding/20/focus-order.html
```diff
@@ -62,30 +62,50 @@ <h2>Intent of Focus Order</h2>
...
+         left-hand column first, from the top to the bottom of the column, and then proceed to the right-hand column.
+         However, if the two columns are independent of each other, and meaning/operation are not affected, it is not
+         a failure if elements in the right-hand column receive focus first, followed by the elements in the left-hand column.
+         As a best practice, though, make sure the focus order reinforces the reading order implied by the visual layout.
+      </p>
+      
+      <div class="note">
+        <p>This requirement does not specify what should or should not receive focus, but rather assesses the order in which elements receive focus. This Success Criterion does not prohibit making non-operable content (e.g., static text) focusable, or programmatically moving focus to such content. However, it is a failure of Focus Order if items receive focus in an order that impedes the meaning or operation of content, or creates confusing or illogical focus orders — for example,
+         a control appearing to receive focus multiple times due to the use of nested focusable elements.</p>
```
mitchellevan:
```suggestion
        <p>This requirement does not specify what should or should not receive focus, but rather assesses the order in which elements receive focus. This Success Criterion does not prohibit making non-operable content (e.g., static text) focusable, or programmatically moving focus to such content. However, it is a failure of Focus Order if items receive focus in an order that impedes the meaning or operation of content, or creates confusing or illogical focus orders — for example:</p>
        <ul>
          <li>a pattern of static and interactive elements receiving focus that sacrifices the meaning of which elements are controls</li>
          <li>a single control appearing to receive focus multiple times due to the use of nested tab-focusable elements</li>
        </ul>
```

I can accept passing 2.4.3 for unnecessary tab focus where it does not change the meaning. However, we also need to acknowledge that some kinds of overuse could result in losing the meaning of what's a control and what's not. For example, if a page contains static icons and button icons, and they all receive focus, then I would still fail 2.4.3 for this.

patrickhlauke:
Note recent additional discussion happened on https://github.com/w3c/wcag/issues/4634

understanding/20/focus-order.html
```diff
@@ -62,30 +62,50 @@ <h2>Intent of Focus Order</h2>
...
+         <pre><code>&lt;div tabindex="0"&gt;&lt;button&gt;...&lt;/button&gt;&lt;/div&gt;</code></pre>
+         <p>As a best practice, avoid having unwarranted focusable elements which cannot be operated
```
patrickhlauke:
```suggestion
         <p>As a best practice, avoid having focusable elements which cannot be operated
```

not sure where "unwarranted" came from, but it sounds a bit too convoluted. wonder if just removing that still works, particularly since this is only framed as a best practice

understanding/20/focus-order.html
patrickhlauke: not sure the "a pattern of static and interactive elements receiving focus" is quite clear though. is this intended to signify something that is done quite a lot/excessively? might be better to spell it out, rather than introducing "pattern" and such

understanding/20/focus-order.html
```diff
@@ -62,30 +62,50 @@ <h2>Intent of Focus Order</h2>
...
+         layout, the assumption in western left-to-right systems may be that focus moves through the elements in the
+         left-hand column first, from the top to the bottom of the column, and then proceed to the right-hand column.
+         However, if the two columns are independent of each other, and meaning/operation are not affected, it is not
+         a failure if elements in the right-hand column receive focus first, followed by the elements in the left-hand column.
+         As a best practice, though, make sure the focus order reinforces the reading order implied by the visual layout.
+      </p>
+      
+      <div class="note">
+        <p>This requirement does not specify what should or should not receive focus, but rather assesses the order in which elements receive focus. This Success Criterion does not prohibit making non-operable content (e.g., static text) focusable, or programmatically moving focus to such content. However, it is a failure of Focus Order if items receive focus in an order that impedes the meaning or operation of content, or creates confusing or illogical focus orders — for example,
```
mbgower:
```suggestion
        <p>This requirement does not specify what should or should not receive focus, but rather assesses the order in which elements receive focus. This success criterion does not prohibit making non-operable content (e.g., static text) focusable, or programmatically moving focus to such content. However, it is a failure of Focus Order if items receive focus in an order that impedes the meaning or operation of content, or creates confusing or illogical focus orders — for example,
```

understanding/20/focus-order.html
mbgower: oops, @mitchellevan my quick fix to put "success criterion" in lower case caused github to mark this thread as resolved. Not my intent, but since it's happened...

 I'm inclined to agree with Patrick on this one. We've given one example which is demonstrative and (hopefully) clear. I don't think including a second example is necessary or helps. If you still want it to be reconsidered, please just re-add the suggestion.

understanding/20/focus-order.html
mbgower: Agreed, this change took place after it was set to 'ready' but not yet sent. Removing for now.

understanding/20/focus-order.html
```diff
@@ -62,30 +62,50 @@ <h2>Intent of Focus Order</h2>
...
+         left-hand column first, from the top to the bottom of the column, and then proceed to the right-hand column.
+         However, if the two columns are independent of each other, and meaning/operation are not affected, it is not
+         a failure if elements in the right-hand column receive focus first, followed by the elements in the left-hand column.
+         As a best practice, though, make sure the focus order reinforces the reading order implied by the visual layout.
+      </p>
+      
+      <div class="note">
+        <p>This requirement does not specify what should or should not receive focus, but rather assesses the order in which elements receive focus. This success criterion does not prohibit making non-operable content (e.g., static text) focusable, or programmatically moving focus to such content. However, it is a failure of Focus Order if items receive focus in an order that impedes the meaning or operation of content, or creates confusing or illogical focus orders — for example,
+         a control appearing to receive focus multiple times due to the use of nested focusable elements.</p>
```
mitchellevan, line 79:
```suggestion
        <p>This requirement does not specify what should or should not receive focus, but rather assesses the order in which elements receive focus. This success criterion does not prohibit making non-operable content (e.g., static text) focusable, or programmatically moving focus to such content. However, it is a failure of Focus Order if items receive focus in an order that impedes the meaning or operation of content, or creates confusing or illogical focus orders — for example:</p>
        <ul>
          <li>a mixture of operable and non-operable elements receiving focus with no visual indication of which elements are operable</li>
          <li>a single control appearing to receive focus multiple times due to the use of nested tab-focusable elements</li>
        </ul>
```

Here's a modified suggestion in response to https://github.com/w3c/wcag/pull/1643#discussion_r2440256712. Without this additional example, "making non-operable content (e.g., static text) focusable" could be misunderstood as being absolutely allowed. It should be allowed only the extent that meaning is preserved.

mbgower: @mitchellevan we are going to open a new issue to deal with the scenario you raise, since we agree we would like to address, but feel it is difficult to fit into a single list item in this Understanding document.

mbgower: with issue #4706 created, I am marking this resolved in this PR. Please mark "Unresoved" and provide more follow up if you disagree, otherwise, the conversation can be carried on at the new issue.

PR #1645, opened 2021-02-16
Pointer target spacing - Wilco update

alastc:
Suggestions from survey


detlevhfischer (approved):
+1 Looks like a sensible change emphasizing a much simpler wording.

Issue #977, opened 2019-11-28
Understanding 3.3.1: Error Identification slightly unfocused/mixes up various concepts


patrickhlauke:
https://www.w3.org/WAI/WCAG21/Understanding/error-identification.html seems to cover various different ideas, but in the end still makes it quite unclear what it's really asking/requiring.

one common interpretation of this SC is that if there's an error, you need to have a textual error message next to the relevant form field that is in error. so, before/after/next to an erroneous form field, have actual text that explains that there's an error.

however, at the start of the intent section, it reads

> Screen reader users, for example, will not know there was an error until they encounter one of the indicators. They may abandon the form altogether before encountering the error indicator, thinking that the page simply is not functional.

this would also be true for a textual indicator next to each invalid/erroneous form field, i assume? so is the requirement that there be a further error message/text at the start of the form, to say generically that there were errors? like a separate boxout that lists any errors, before the form? and not having that, and instead showing textual errors in context next to each form control, would fail this?

under benefits, it lists

> This Success Criterion may help people with cognitive, language, and learning disabilities who have difficulty understanding the meaning represented by icons and other visual cues.

so, the use of just icons is not sufficient, it needs to be text. Is the use of icons (or color, or something non-textual) what is meant by "indicating the fields" here? or does "indicating the fields" also mean "even if the indication happened in text, next to the field"

>  In the case of an unsuccessful form submission, re-displaying the form and indicating the fields in error is insufficient for some users to perceive that an error has occurred.

patrickhlauke:
apart from possibly a bit of a rewording/focusing of the understanding, this could also really do with a few good representative examples (in case there are multiple scenarios here that are deemend appropriate, like a form with a list of errors at the start, and form fields then just identified with an icon/color; and, if acceptable as a solution, a form where each form field that is in error has an inline/contextual text-based error message showing, but NO list of errors at the start; etc)

patrickhlauke:
to clarify the question, here's an example of what i mean by "in context" text-based error messages. a user tries to register an account and leaves out fields. fields are denoted both with an icon AND with text adjacent to the relevant form field, indicating there's an error with a particular field. there is no overarching "there were errors" / "here's a list of errors" or similar type separate listing/notification.

![screenshot of an account registration form - all fields were left blank, so all required fields show an error below the input in text](https://user-images.githubusercontent.com/895831/69822194-a7553a00-11fd-11ea-81ef-4f8affcfebf6.png)

[...]

PR #1651, opened 2021-02-23
Rewrite for 3.3.1 Error Identification understanding document

patrickhlauke:
* remove the misleading statement about screen reader users needing to know that an error occurred before hitting one of the indicators/fields. this implicitly suggests that the intent is for an error message/list to be shown to (screen reader) users before the actual form, which is not in fact the intention nor the requirement of this SC
* tweak the benefit for users with cognitive/language/learning disabilities
* refocus on the benefit to ALL users, not just screen reader users
* ~add allowance for situations where an error indication is already self-explanatory/obvious from context (e.g. a form where fields have already explicitly been identified as mandatory/required - not necessary for compliance to now ALSO explicitly say "as we already told you, this field is mandatory")~ [potentially removed by mbgower suggestion]

Closes https://github.com/w3c/wcag/issues/977


mraccess77:
I like the updates!  I agree for a login form that additional error for an empty field is not required as in a simple form it is assumed these two fields are required.

mraccess77:
For the last bullet where a field is already marked required - if the field is using an asterisk and the asterisk is described at the top of the form as indicated a required field are we now saying nothing else needs to be said/done - so when the user submits the form with a missing required field the focus could just stay on submit and no visual changes need to take place?  I just want to make sure we have thought of all situations.

patrickhlauke:
For the last bullet it's saying that error still needs to be *indicated*, but a redundant additional explanation in text may not be needed.

> There may also be situations in which it is clear from the context what the specific error
         is - for instance, in a simple login form with a username and password field, if a user
         leaves one of the fields completely blank, simply indicating that the empty field is in
         error using an image or icon (with appropriate alternative text) would be sufficient.

i.e. no need to have a field that was marked as required, after failing has now an additional indication (like a warning triangle)...i'd argue there's no point now also requiring a repetitious and not very informative "You must fill in this required field" just to satisfy the "error is described in text".

bruce-usab:
FWIW, I have always understood that <q>described to the user in text</q> is a very low bar.  It excludes the most egregious barriers, for example, adding (only) a red border (around the error) or (only) a pointing-hand graphic.  But it can be helpful to be able to fail situations similar to those two not-uncommon error indicators (even when they they pass 1.4.1 and 1.1.1), and 3.3.1 provides that.

bruce-usab (approved):

mraccess77:
I believe it's reasonable that login forms should not have to have required field indicators as by nature of authentication you have to enter a username and password.       However, upon submitting without entering one or the other or invalid content then I'd expect some sort of error message to then be required.

patrickhlauke:
> I'd expect some sort of error message to then be required

but for those situations where it's quite clear from context (e.g. of course you can't log in with only a username and not providing a password), does the error absolutely need to be conveyed in text, or is it then sufficient to provide an error *indication* (which could be just an icon or similar - provided it passes other SCs like 1.1.1, 1.4.11, etc) without the need for words? that's the circle i was trying to square here with some nuance/human judgement.

mraccess77:
I personally think an icon that indicates missing field with alt text is sufficient when combined with the implicit or explicit required state.

patrickhlauke:
Made some amendments based on the comments from yesterday's AGWG meeting:

- @gundulaniemann wanted to keep "The error message should be as specific as possible". I thought this was already sufficiently covered in "The error message should be descriptive", but I've now expanded that to be "The error message should be as descriptive and specific as possible."
- various good spots that @mbgower had for missing words, too many commas, and one sentence that made no sense

There were further comments that I'm not quite sure of at this stage. For instance, asking that the understanding should demand errors to be inline next to their fields, and something about source? Not sure these ideas are currently part of the understanding, and they seem to limit the possible ways in which errors can be implemented...and would, in my view, require more thorough discussion (and they're not present in the current wording in the understanding).

Oliver Kiem had a comment: "The sample uses a login form. As login forms sometimes have strange behavior (such not allowing the user to understand that a userID is available but the password does not match - for security reasons) the use of a app like "Create Account" may have less inherent issues."

Note that the example I have here is about *login* forms, *not* about "create an account" forms. Login forms generally just ask you for a username/password, and if one or the other fails, it's fairly self-explanatory why (because one or both are wrong). agree that account *creation* forms will have more nuanced error states, which of course can't just get away with indicating a basic "nope, fails" with a simple icon or something.

patrickhlauke:
Randomly, I know it's not "the process", but particularly for small specific requests for changes...it would be much more helpful and straightforward to use github's comment functionality (which lets you comment even down to individual line level), rather than then sending an email referencing the questionnaire comments and minutes...

gundulaniemann:
As a sentence (or novel) can be descriptive without being specific, I appreciate the change.

Please let me also turn to the login screen example (remark from Oliver Keim):
During password creation, the sever usually responds whether the ID is already taken, invalid or whatever, and it usually also gives feedback whether the password is valid.
Yet specifically login forms do not respond with this level of detail, but only tell the login was unsuccessful. They will not tell whether the user name was unknown or the password was wrong for security reasons. This collides with the level of detail we request in general for Accessibility reasons.
Therefore I support the suggestion to take a different example than a login form.
An account creation form might be suitable, or  a 'change password' form.

patrickhlauke:
but "login" is not "account creation". those are distinct things.

gundulaniemann:
'Login' and 'Account Creation' definitely are different things.
The error messages given for an unsuccessful login often is not helpful to the end user. And this is intentional for security reasons.

Therefore we (Oliver as well as myself) suggest to take a different example.
My idea is 'Change password', as it is very simple, and it is allowed to give precise error messages which are as short as you would like to have them.

patrickhlauke:
but the login example can't just be replaced, because it's used in my proposed change here to illustrate the scenarios where extensive error messages are NOT needed.

```
There may also be situations in which it is clear from the context what the specific error
         is - for instance, in a simple login form with a username and password field, if a user
         leaves one of the fields completely blank, simply indicating that the empty field is in
         error using an image or icon (with appropriate alternative text) would be sufficient.
         However, if the user did enter data in both of those fields, and login was unsuccessful, further
         information and an actual specific error message will be required (e.g. "no account found with this
         username", or "the username or password are incorrect"). Likewise, in a form that explicitly indicates
         that all fields are mandatory/required, it will be sufficient to indicate empty fields as being in error,
         without the need to reiterate that the user must fill in the field. 
```

ripping out the example/replacing it would leave this without an example that shows this, which is one of the points of this PR - to show possible situations where the requirement for lengthy error descriptions is not necessarily needed. so unless your comment is in fact "remove that whole section that exempts this" (which is a different proposition from saying the example should be changed), then not sure how to square that.

patrickhlauke:
or if your comment is to ALSO add an additional example for account creation, sure... but the login one is kind of needed to prove my point about scenarios where it's not needed

patrickhlauke:
to put it yet another way: is the whole point

```
* add allowance for situations where an error indication is already self-explanatory/obvious from context (e.g. a form where fields have already explicitly been identified as mandatory/required - not necessary for compliance to now ALSO explicitly say "as we already told you, this field is mandatory")
```

(per the first comment here for the PR)

contentious (probably) and you're proposing to remove the allowance/exemption altogether, or just asking for an additional example (like account creation, as opposed to login) where descriptive errors ARE necessary (in contrast to the example for login)

patrickhlauke:
visualising the proposed common sense exemptions:

login form with a warning error triangle next to the label for the password field which was left empty
In a login field, where it's clear from context that both username and password are required, is it sufficient to just have an icon (assume it has appropriate alt text etc) rather than requiring explicit text error message that identifies the problem?

---
For a field that is already clearly identified as required, and that was left empty, is it sufficient to have an icon (as above)?

'City (required)' label for an empty field, with a red error warning triangle next to it

or are we really demanding that even in those cases there must be explicit text that essentially restates what the label already says ("Please fill in this required field") otherwise it fails the SC?

'City (required)' label for an empty field, with a red error warning triangle next to it, but also an explicit text error message below the field

jake-abma:
> or are we really demanding that even in those cases there must be explicit text that essentially restates what the label already says ("Please fill in this required field") otherwise it fails the SC?

Good question, but "unless we can be sure icons are understood (think we can not / never)" and "the icons are described by WCAG SWG how they should look like(?!)" the text is probably needed for all user to understand the input is in error.

patrickhlauke:
there's no guarantee that text will be understood by all users unless it's also defined HOW to write good error messages etc...

in my view, this just makes it all very reductive and restrictive / doesn't offer nuance. demanding that there MUST be text, no matter how self-explanatory the author thinks it is, is a blunt instrument.

understanding/20/error-identification.html
```diff
@@ -99,7 +97,7 @@ <h2>Intent of Error Identification</h2>
             In other cases, it may be more appropriate to show errors inline, with error messages next the specific
             fields that are in error. Errors could also be listed in an alert, or dialog. This criterion does not
             cover which of these methods should be used - the only requirement is for errors (which are not already
-            clear from the specific nature/context of the input fields) to be presented to users in text.
+            clear from the specific nature/context of the input fields) to be presented to users in text or alt text.
```
awkawk, line 100: replace "alt text" with "a text alternative for non-text content"

patrickhlauke: done (with a bit more wordsmithing to make it flow better)

understanding/20/error-identification.html
```diff
@@ -70,11 +69,7 @@ <h2>Intent of Error Identification</h2>
          error information to the user. For example, certain technologies can specify that
          the user's input must not fall outside a specific range, or that a form field is required.
          Currently, few technologies support this kind of programmatic information, but this
-         Success Criterion does not require, nor prevent it. However, other Success Criteria may apply - for
-         instance, the programmatic identification of form controls/inputs that are in error is covered
-         more specifically in <a href="name=role-value">4.1.2 Name, Role, Value</a>, and whether or not
-         error messages are programmatically linked to their respective form controls/inputs is covered
-         by <a href="info-and-relationships">1.3.1 Info and Relationships</a>.
+         Success Criterion does not require, nor prevent it. However, other Success Criteria may apply such as <a href="name=role-value">4.1.2 Name, Role, Value</a>, and <a href="info-and-relationships">1.3.1 Info and Relationships</a>.
```
patrickhlauke, line 72: would be nice to hint at how those other SCs apply, directly here, rather than just handwaving...

alastc: We could either continue quite a long discussion about that (@WilcoFiers disagreed on the face of it), or get the current changes in and deal with that separately.

patrickhlauke: maybe worth adding a tiny bit of context on these sorts of changes to PRs then, as they otherwise seem quite arbitrary

alastc: Yep, adding that to a new survey item.

mraccess77: We should make it clear that if an image is used the image itself must communicate the errors also communicate in alt text.  We don't want people to stuff extra needed details into alt text with only a visual image that doesn't communicate the error and what is wrong.  This could be misused if we don't clarify this.

patrickhlauke:
I would still like to float the idea that self-evident situations should be exempted (where it's clear what would the error itself is, and only identification is required), which was the whole context of the idea of allowing something like a warning triangle image.

> add allowance for situations where an error indication is already self-explanatory/obvious from context (e.g. a form where fields have already explicitly been identified as mandatory/required - not necessary for compliance to now ALSO explicitly say "as we already told you, this field is mandatory")

this was ripped out in the change 12 days ago. and yes, without that part, i would not be happy with saying the error description can just be in the alt. the whole point of the original as written was that for those cases where it is clear what the nature of a failure is, the "in text" part is already evident/clear from the existing labels, or even (in very limited cases) context (like the login form example, where it's obvious why leaving one or the other out would be a failure)

understanding/20/error-identification.html
mraccess77: I agree that a self evident error would be ok with regards to required fields.  For example, if a required field is empty and has an error icon then it's ok as it's required and an icon is sufficient in that case to communicate the fact that a required field was not filled out.  For other errors though like incorrect data format it's a different story.

patrickhlauke:
i.e. are we really saying that unless there's explicit text (as in the last example of https://github.com/w3c/wcag/pull/1651#issuecomment-856000866) it's a failure of the SC even when it's arguably self-evident WHAT the error is?

understanding/20/error-identification.html
patrickhlauke: (a small side note here @mraccess77 - you're replying to a sub-thread about the change @alastc made to take our explicit references to other SCs that may also apply, based on @WilcoFiers's feedback at the time. your response, I believe, relate to the main discussion thread here)

patrickhlauke:
noting also that this is why fundamentally changing the nature of a PR (ripping out whole sections) this way gets confusing, as part of the discussion/point (and any previous discussion along the thread referring to that) ends up making no sense later on (and it's not completely straightforward to go and dig back into the original PR's text)

bruce-usab:
Just sharing my take-away impression of the conversation in todays AG WG call.

> I would still like to float the idea that self-evident situations should be exempted (where it's clear what would the error itself is, and only identification is required), which was the whole context of the idea of allowing something like a warning triangle image.

I do not disagree that this might not be a bad idea.  But IMHO, this sort of exemption needs to wait until/unless we decide to consider normative changes that are less significant than this one.  Also, I am not sure how much I trust page authors to assert <q>self-evident situations</q>.  Seems like a slippery slope to me!  I can live with the consequence that a plain reading of SC 3.3.1 implies that we might end up with stuff like <q>this required field is required</q>.

> i.e. are we really saying that unless there's explicit text (as in the last example of #1651 (comment)) it's a failure of the SC even when it's arguably self-evident WHAT the error is?

Yes.  It is not so bad.

patrickhlauke:
>  I am not sure how much I trust page authors to assert self-evident situations. Seems like a slippery slope to me! 

We trust authors/auditors to decide when headings/labels are descriptive enough, when text alternatives are appropriate, etc. But sure...

> Yes. It is not so bad.

Having SCs that require redundant busywork won't make authors any more enthusiastic in following them...

bruce-usab:
@patrickhlauke I am hoping your <q>You must fill in this required field</q> illustrated example makes it into Understanding.

It clarifies that strict conformance might require something that seems redundant.  I am sure you and I can both come up with other examples.

And to be honest, I would not expect the average end-user to think twice about a pop-up message that reads <q>this required field is required</q>.

Your concern for developers being tasked with SCs that they (understandably) view as requiring redundant busywork is fair.  OTOH developers have a long history of not communicating as well as they think they are with the end user.

We are talking a about a situation where (1) the field was marked as required, and yet (2) the end-user missed the initial prompt.  Please consider that just a pop-up icon might not be as effective as you assume it is.  (After all, they missed the word <q>required</q> next to <q>City</q> the first time.  Why would they necessarily cognitively process the parenthetical any better the second time?  The new information, the scary red triangle, might actually make paying attention to the parenthetical _harder_!)  So yes, adding <q>You must fill in this required field</q> improves accessibility (usability too) and, even with this example, the requirement is that <q>the error is described to the user in text</q>.

patrickhlauke:
obviously, adding the redundant error message text will require enough space for it, which will have to impact design decisions too. things like login forms are often added as very minimalist little additions to things like menus and toolbars, not always in a nice spacious page with extra real estate.

as for all the "developers have historically been bad at communicating things" and "the user missed it, so it's right that we require the text / just having an icon might make it even less likely for users to see why they failed" ... there's a lot of underlying author-hostility here, and quite a lot of assumption/speculation about users (who, conversely, may have just been skimming over the form before without much care, but are then prompted to more carefully look at the existing labels for fields that have been identified as in error).

in any case, finding it a bit odd that this aspect - at least based on the minutes from the last calls where this was discussed - was hardly touched on in the actual meeting, and we're now backfilling rationales after it was already ripped out of the PR. but hey ho...tired of arguing this further. just make sure to correctly update the commit message if this gets merged...

patrickhlauke:
Any news on this @alastc ?

mbgower:
@patrickhlauke one thing I tweaked on is that we're inconsistent in WCAG on style (and use of) the term "color blind".
I see it as:
- "color-blindness"([use of color](https://www.w3.org/WAI/WCAG21/Understanding/use-of-color)) (which also has it spelled a bunch of ways by outside references
- "colorblind" (error identification)

I also see usage of the term "color deficiency" ([contrast minimum](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum) and [contrast enhanced](https://www.w3.org/WAI/WCAG21/Understanding/contrast-enhanced), , which I had always understood to be the preferred term.

I'd advocate for it to be 2 words AND used parenthetically to color deficiency, e.g., "color deficiency (color blind)"

mbgower (approved):

mraccess77:
Color vision was also discussed in the LVTF previously: https://github.com/w3c/low-vision-a11y-tf/discussions/121

understanding/20/error-identification.html
```diff
@@ -14,14 +14,11 @@ <h2>Intent of Error Identification</h2>
...
-         one of the indicators. They may abandon the form altogether before encountering the
-         error indicator, thinking that the page simply is not functional. Per the definition in WCAG 2.0, an "input error" is information provided by the user
-         that is not accepted. This includes: 
+         has occurred and can determine what is wrong. In the case of an unsuccessful form submission,
+         it is not sufficient to only re-display the form without providing any hint that the submission
+         failed. It also won't be sufficient to merely indicate that fields are in error without providing users with sufficient information about the specific nature of the error. The error messages must be descriptive to help users understand what the problem is. The error can be indicated in <a>text</a>, which does allow for visually apparent images or icons with suitable alternative text. <a href="error-suggestion">3.3.3 Error Suggestion</a> covers what the error messages should contain in certain circumstances.
+      </p>
+      <p>An "input error" is information provided by the user that is not accepted. This includes:
```
WilcoFiers, line 21:
```suggestion
      <p>An "input error" occurs when information provided by the user that is not accepted. This includes: 
```

understanding/20/error-identification.html
```diff
@@ -72,20 +68,21 @@ <h2>Intent of Error Identification</h2>
          that user agents or assistive technologies can use to identify an error and provide
          error information to the user. For example, certain technologies can specify that
          the user's input must not fall outside a specific range, or that a form field is required.
-         Currently, few technologies support this kind of programmatic information, but the
-         Success Criterion does not require, nor prevent it.
-         
+         Currently, few technologies support this kind of programmatic information, but this
+         Success Criterion does not require, nor prevent it. However, other Success Criteria may apply such as <a href="name=role-value">4.1.2 Name, Role, Value</a>, and <a href="info-and-relationships">1.3.1 Info and Relationships</a>.
```
WilcoFiers, line 72: This change suggests to me that 4.1.2 or 1.3.1 might require `aria-valid`, `aria-required`, `min`, `max`, or `pattern` might be required. I don't think they do, or even that it could. Those attributes didn't exist when the SC was written. It couldn't possibly have required them.

```suggestion
         Success Criterion does not require, nor prevent it.
```

WilcoFiers (changes requested):

understanding/20/error-identification.html
patrickhlauke: I'd say that's for 1.3.1 and 4.1.2 to then clarify at their end. The statement that they *may* apply though is correct, I'd say.

patrickhlauke:
I see this has a `SurveyAdded` tag. what was the outcome of the survey? lost track of this... @alastc

understanding/20/error-identification.html
```diff
@@ -22,16 +22,17 @@ <h2>In brief</h2>
...
-         Screen reader users, for example, will not know there was an error until they encounter
-         one of the indicators. They may abandon the form altogether before encountering the
-         error indicator, thinking that the page simply is not functional. Per the definition,
-         an "input error" is information provided by the user that is not accepted. This includes: 
+      <p>The intent of this success criterion is to ensure that users are aware that an error
+         has occurred and can determine what is wrong. In the case of an unsuccessful form submission,
+         it is not sufficient to only re-display the form without providing any hint that the submission
+         failed. It also won't be sufficient to merely indicate that fields are in error without providing
+         users with sufficient information about the specific nature of the error.
+         The error messages must be descriptive to help users understand what the problem is.
```
alastc, line 30: I think this could be read as adding to the SC requirement, which I don't think includes "help users understand what the problem is", which is more what "error suggestion" is for. Also, we should probably avoid "must/should/may" language where possible. Suggest removing this line.

dbjorge:
@gundulaniemann  @patrickhlauke I think your discussion about whether the understanding doc should include an example of whether a login form needs to clarify whether username vs password was incorrect is a good discussion and something that it'd be helpful to add an example for, but probably out of scope for this PR. I think existing issue #2157 covers this already (the title says 3.3.3 but the discussion/labels clarify that it's scoped to both 3.3.3 and 3.3.1)

understanding/20/error-identification.html
patrickhlauke: good points @alastc ... I'll tweak these two aspects

patrickhlauke: Tweaked (moving the text about how descriptive the error should be to refer to 3.3.3)

understanding/20/error-identification.html
```diff
@@ -1,124 +1,104 @@
...
-         one of the indicators. They may abandon the form altogether before encountering the
-         error indicator, thinking that the page simply is not functional. Per the definition,
-         an "input error" is information provided by the user that is not accepted. This includes: 
+
+      <p>The intent of this success criterion is to ensure that users are aware that an error
+         has occurred and can determine what is wrong. In the case of an unsuccessful form submission,
+         it is not sufficient to only re-display the form without providing any hint that the submission
+         failed.
+         The error can be indicated in <a>text</a>, which does allow for visually apparent images or
+         icons with suitable alternative text.
```
mbgower, line 29: I don't think I agree with this. Error identification specifically requires that "the error is described to the user in text." An image or an icon is not text.
IMO the error must be described in text to pass the SC.

understanding/20/error-identification.html
```diff
@@ -1,124 +1,104 @@
...
       <p>For example:</p>
-      
+
       <ul>
-         
-         <li>the user fails to enter the proper abbreviation in to state, province, region, etc.
-            field; 
-         </li>
-         
+         <li>the user fails to enter the proper abbreviation in to state, province, region, etc. field;</li>
```
mbgower, line 44: "to" seems wrong, and I've pulled out the "etc" as superfluous
```suggestion
         <li>the user fails to enter the proper abbreviation in a state, province, or region field;</li>
```

understanding/20/error-identification.html
```diff
@@ -1,124 +1,104 @@
...
-         Screen reader users, for example, will not know there was an error until they encounter
-         one of the indicators. They may abandon the form altogether before encountering the
-         error indicator, thinking that the page simply is not functional. Per the definition,
-         an "input error" is information provided by the user that is not accepted. This includes: 
+
+      <p>The intent of this success criterion is to ensure that users are aware that an error
+         has occurred and can determine what is wrong. In the case of an unsuccessful form submission,
+         it is not sufficient to only re-display the form without providing any hint that the submission
+         failed.
+         The error can be indicated in <a>text</a>, which does allow for visually apparent images or
```
mbgower, line 28:
```suggestion
         The error can be indicated in <a>text</a>, which does allow for visually apparent images or
         icons with suitable alternative text.
```

understanding/20/error-identification.html
patrickhlauke: i'll admit that, since it's been over 3 years since i wrote most of this...i can't quite recall where that phrasing i suggested came from. i'm fairly sure it bubbled up from related discussions mentioned here. (that's the joy of revisiting really old PRs)

not particularly wedded to any of these wordings at the moment, just trying to address some of the other more dramatic issues listed in the description bullets here

understanding/20/error-identification.html
mbgower, line 28:
```suggestion
         The error must be indicated in <a>text</a>.
```

understanding/20/error-identification.html
mbgower, line 29:
```suggestion

```

mbgower: @patrickhlauke I have made a suggested change which makes the need for text explicit. I'd like to revert to this, then open up a separate issue to directly address whether there is support to allow only a symbol with alt in specific circumstances.
If that works for this PR, please incorporate.

patrickhlauke:
accepted the changes that removed the suggestion that images with appropriate `alt` would be sufficient. would still love to qualify somewhere/somehow that in *some* cases, existing text (like already having "required" clearly stated in the label) coupled with an additional indicator (which could just be an extra icon or similar) could pass ... as there seems to be at least partial agreement that these situations do exist and that this is indeed acceptable, rather than having to have very explicit extra text. at least in practice this is what we see a lot out there. and then the classic case of a form that, contextually, of course requires both fields ... where just having an icon like a warning triangle (e.g. on a username/password login, and you leave one blank) on the empty one would probably be sufficient hint. but i'll leave this for another SC. see if this now makes sense, with that aspect left out for now...

bruce-usab (approved):

mbgower:
> accepted the changes that removed the suggestion that images with appropriate `alt` would be sufficient. would still love to qualify somewhere/somehow that in _some_ cases, existing text (like already having "required" clearly stated in the label) coupled with an additional indicator (which could just be an extra icon or similar) could pass ... as there seems to be at least partial agreement that these situations do exist and that this is indeed acceptable, rather than having to have very explicit extra text. at least in practice this is what we see a lot out there. and then the classic case of a form that, contextually, of course requires both fields ... where just having an icon like a warning triangle (e.g. on a username/password login, and you leave one blank) on the empty one would probably be sufficient hint. but i'll leave this for another SC. see if this now makes sense, with that aspect left out for now...

I have added this as a consideration for WCAG 3 https://github.com/w3c/wcag/wiki/WCAG-3-parking-lot-considerations-from-2.x

PR #1659, opened 2021-03-02
Redundent entry update from @mbgower

### LINKED ISSUE #1660: Alter note for Redundant Entry
The current wording of the note for Redundant Entry  is:
> Security verification, such as repeating a password, is considered essential.

I believe that this wording could be misinterpreted (as talking about authentication method), and is overly general. No one has mentioned another 'security verification' where this note fits, except re-entering a new password to reduce user error (from mistyping their new password when setting it). Therefore, I recommend altering this to make the exception more clear:
> Requiring re-entry of a new password is considered essential.

This change is contained in https://github.com/w3c/wcag/pull/1659

alastc:
Closes #1660


alastc:
Replaced by #1751

PR #1673, opened 2021-03-08
Target Size (Minimum): updates

detlevhfischer:
Reflecting the additional exceptions and the proposed re-naming of the SC.


alastc:
Hi Detlev,

We had the renaming and (not-agreed) exceptions in other PRs, and the files have been renamed as well, I recommend closing this one and adding the understanding doc updates to a separate branch/PR.

Issue #975, opened 2019-11-26
Requires 1.4.12 compatibility with one or all methods of adjusting text spacing?


JAWS-test:
[Understanding for SC 1.4.12](https://www.w3.org/WAI/WCAG21/Understanding/text-spacing.html) says

> Such changes may be achieved via user stylesheet, bookmarklet, extension, or application. 

In practice, it is often the case that customization does not cause any problems with user styles, but with the bookmarklet. Reason is:
* Adjustment with User Styles takes place when rendering the page, so that the rest of CSS can react to it.
* Adjustment with bookmarklet is done after rendering the page, so the text often flows out of the blocks.

Is SC 1.4.12 met if user styles work but the bookmarklet doesn't?

alastc:
The bookmarlet is not the ultimate arbiter, for example, it cannot work on sites with a strict 'content security policy'. My backup then is the stylish plugin, which is also applied at the end of the cascade.

However, I'm struggling to see a way that a site would adapt to a user style sheet that would not also work with bookmarklet / plugin approach? The site has no means to know there is a user stylesheet.

JAWS-test:
I'm afraid that doesn't answer my question. My question was whether one or all of the methods mentioned in the Understanding have to work or what happens if one of the methods leads to a bad result and the others do not.

> The site has no means to know there is a user stylesheet

That may be, but I've already found several pages in the test where the bookmarklet adjusts the texts so that the texts overlap, while with userstyles everything was ok. I didn't find out the exact cause. It could be that it's not the CSS, but the JS. This means that when the page is loaded, the JS determines how much space the text containers need and and these are then positioned. If I adjust the distances afterwards, the JS function is not triggerd again.

JAWS-test:
If the bookmarklet is only to be used for testing purposes, but is not used by impaired people, I am in favor of deleting it in the Understanding. Otherwise, I am in favor of describing in the Understanding whether all or only one method must work.

patrickhlauke:
what the SC requires is that IF a user manages to change the spacing metrics, the site doesn't become unusable (cut off/overlapped content etc). it does not specify how users would change those metrics, nor does it require that the site somehow makes provision for/makes it easy to actually change those in some way by the user.

mraccess77:
[...]

PR #1687, opened 2021-03-17
Understanding 1.4.12: clarify the intent and author responsibility

patrickhlauke:
As discussed in https://github.com/w3c/wcag/issues/975 and other places, there is confusion about what the exact responsibility of authors is with 1.4.12.

questions like "if i use the bookmarklet and nothing happens, do I fail this SC?" or "how can a user change text metrics on mobile/tablet/some other environment that doesn't allow bookmarklets/extensions to run?" are very common when discussing 1.4.12 and how to interpret it.

my understanding at least is that 1.4.12 is not about "this method must work", but rather "IF a user manages to inject/change the metrics, THEN content should not overlap/become unreadable/stop working".

to try and clarify this a bit more, I've made some tweaks to both the intent, and the author responsibility section.

/cc @mraccess77 @alastc as I *think* we've said roughly the same/agreed on this in the past


JAWS-test:
I think the change is very good. Thank you! But I do not understand what the following means:

> provided that the content is not actively and purposely preventing the properties from being added

patrickhlauke:
> I do not understand what the following means

it would be something like "the page runs a script that constantly checks if the user made some CSS changes/injections, and strips them out again" or similar. basically adversarial coding that tries to prevent any changes to layout. unusual perhaps, but possible...

patrickhlauke:
(and to clarify: if a *browser*/*user agent* prevents things from happening - for instance because of CSP policies - then that's *not*, to my mind, a problem of the content, but a decision of the browser, so not a failure per se of the SC - but in an audit, i'd note this as 'while it passes, note that...' perhaps)

JAWS-test:
@patrickhlauke I understand your explanation, but the proposed text does not make it clear enough, I think

JelisaCooper (approved):

alastc (approved):

patrickhlauke:
From the survey/today's discussion, I see that the comment about wanting to keep the "that people *can* override". Again, I'll reiterate that this is not actually what the SC normatively says. The SC doesn't say that the user must be able to change the text metrics (compare with, say, the resize text SC which normatively says "text can be resized"). It says normatively that "no loss of content or functionality occurs by setting all of the following and by changing no other style property" - it does not stipulate that users *can* set these style properties (or how), just that if they do manage (somehow) to "set" them, then content/functionality is not broken. That, to me, is the nuance given by using the "*when* people override..." wording. it's more in the sense of "*if people manage to override*".

patrickhlauke:
From [today's minutes](https://www.w3.org/2021/06/01-ag-minutes.html#t08), I see that the use of "when" is still causing confusion. Happy to change this to "if" ... does that make it clearer? (basically still making sure that the wording here doesn't imply that the user's ABILITY to change text metrics is what determines pass/fail, but rather that when/if users to manage to change the metrics, whether or not content starts breaking/becoming non-functional is the actual point of the SC).

so, imagine the first sentence there swapping out the "when" with "if" ... does that make it clearer? /cc @gundulaniemann 

> The intent of this Success Criterion (SC) is to ensure that *if* people override author specified text spacing

gundulaniemann:
In fact, specifically when looking at Patrick's comment from '13 days ago', I prefer the wording with 'when'. Yet I can live with each  of the two 'when' and 'if'.
For clarification, the new paragraph which I cite below should prevent any misunderstanding (no change from my side, just to clarify which statement I rely on).

"Further, this SC is not concerned with <em>how</em> users change the line height and spacing metrics. It does not require that content implement its own mechanisms to allow users to do this. It is not a failure of the content if a user agent or platform does not provide a way for users to do this. Content does not fail this SC if the method chosen by the user - for instance, the use of an extension or bookmarklet - fails to correctly set the line height and spacing text properties on the content (provided that the content is not actively and purposely preventing the properties from being added)."

alastc:
Approved: https://www.w3.org/2021/06/01-ag-minutes.html#t08

PR #1699, opened 2021-03-25
Aria22 modification for aria-atomic

### LINKED ISSUE #1137: Update needed for Technique ARIA22 example "Using role=status to Provide Status Message on Shopping Cart"?
Hi,
It seems that the example referenced in Technique [ARIA22 Using role=status to Provide Status Message on Shopping Cart](https://www.w3.org/WAI/WCAG21/working-examples/aria-role-status-shoppingcart/) does not work, maybe because of changes to the claimed default value `aria-atomic` of `role=status`.
According to the Technique, the output should be "Shopping Cart # items" or "# items".

```
<p role="status"><img src="shopping-cart.png" alt="Shopping Cart">
<span id="cart">1</span> items</p>
```
Current output however is:
---> NVDA/FF: exposes number only
---> JAWS/FF: no output
---> JAWS/Chrome: output "2 items"
---> VO/iOS: output "shopping cart 2 items"

Adding `aria-atomic="true"` after `role="status"`  in the devtools achieves the desired output "Shopping cart 2 items" (only tested in NVDA/FF).

detlevhfischer:
Addressing changes in response to #1137 
Adding an explicit `aria-atomic="true"` in Technique [ARIA22: Using role=status to present status messages](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA22) and the linked examples (search results, shopping cart)  since the default behaviour of `role="status"` (reading the entire content of a changed container marked with `role="status"`) is currently not well supported.

The change to search results example is actually unnecessary since the entire content is written into the container and thus exposed. So not sure whether it is better to skip that change, or change both examples for consistency's sake.

Closes #1137


bruce-usab:
Discussed on backlog call 10/25, concern if syntax is still correct.

Issue #1681, opened 2021-03-10
Add the use of aria-current as a sufficient technique for Info And Relationships?


fstrr:
The `aria-current` attribute would seem to be a good candidate as a sufficient technique for programmatically determining the current item in a collection. Examples could be:

- current page in a nav bar
- current step in breadcrumb navigation
- current time in list of conference sessions.

If there is positive feedback, I'm happy to work on this.

JAWS-test:
And also for SC 2.4.8 Location

alastc:
Sounds good, the next step would be to create a branch, e.g. "tech-aria-current", and create a draft.  I suggest starting with another technique as a template, e.g. https://github.com/w3c/wcag/blob/main/techniques/aria/ARIA13.html

fstrr:
Thanks for the pointer to that other template. Will get started ASAP.

patrickhlauke:
Closed by #1707

PR #1707, opened 2021-03-29
New ARIA technique for the aria-current attribute

fstrr:
This PR is in relation to issue #1681

### DIFF
--- a/techniques/aria/ARIA25.html
+++ b/techniques/aria/ARIA25.html
@@ -1,0 +1,178 @@
+<!DOCTYPE html>
+  <html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml">
+  <head>
+    <title>Using aria-current to identify the current item in a set</title>
+    <link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove">    
+    <link rel="stylesheet" type="text/css" href="../../css/editors.css" class="remove"/>
+  </head>
+  <body>
+    <h1>Using <code class="att">aria-current</code> to identify the current item in a set</h1>
+ 
+    <section class="meta">
+      <p class="id">ID: ARIA25</p>
+      <p class="technology">Technology: aria</p>
+      <p class="type">Type: Technique</p>
+    </section>
+
+    <section id="applicability">
+      <h2>When to Use</h2>
+      <p>Technologies that support <a href="https://www.w3.org/TR/wai-aria/">Accessible Rich Internet Applications (WAI-ARIA)</a>.</p>
+    </section>
+
+    <section id="description">
+      <h2>Description</h2>
+      <p>The purpose of this technique is to programmatically indicate the current item in a user interface component that contains a set of related items. The <code class="att">aria-current</code> attribute provides a way to programmatically indicate the element that is usually only visually highlighted, for example: the current page in a navigation bar.
+      </p>
+
+    </section>
+
+    <section id="examples">
+      <h2>Examples</h2>
+    
+      <section class="example">
+        <h3>Identify the current page in a navigation bar</h3>
+     
+        <p>This is an example of <code class="att">aria-current</code> used on a navigation bar. It uses the attribute's <code>page</code> value to programmatically identify the current page that the user is on.</p>
+      <pre xml:space="preserve">
+&lt;nav&gt;
+  &lt;ul&gt;
+    &lt;li&gt;&lt;a href=&quot;/&quot;&gt;Home&lt;/a&gt;&lt;/li&gt;
+    &lt;li&gt;&lt;a href=&quot;/food&quot; aria-current=&quot;page&quot;&gt;Food&lt;/a&gt;&lt;/li&gt;
+    &lt;li&gt;&lt;a href=&quot;/drinks&quot;&gt;Drinks&lt;/a&gt;&lt;/li&gt;
+    &lt;li&gt;&lt;a href=&quot;/contact&quot;&gt;Contact&lt;/a&gt;&lt;/li&gt;
+  &lt;/ul&gt;
+&lt;/nav&gt;
+</pre>
+
+      <p>
+        <a href="../../working-examples/aria-current/#example1">A working example of identifying the current page in a navigation bar</a>.
+      </p>
+    </section>
+      
+    <section class="example">
+      <h3>Identify the current step in a process</h3>
+     
+      <p>This is an example of <code class="att">aria-current</code> used to mark up the steps in a recipe. It uses the attribute's <code>step</code> value to programmatically identify the current step:</p>
+<pre xml:space="preserve">
+  &lt;h2&gt;Meringue Recipe Steps&lt;/h2&gt;
+  &lt;ol&gt;
+    &lt;li&gt;&lt;a href=&quot;#oven&quot;&gt;Heat your oven&lt;/a&gt;&lt;/li&gt;
+    &lt;li&gt;&lt;a aria-current=&quot;step&quot; href=&quot;#separate&quot;&gt;Separate eggs&lt;/a&gt;&lt;/li&gt;
+    &lt;li&gt;&lt;a href=&quot;#whisk-eggs&quot;&gt;Whisk egg whites&lt;/a&gt;&lt;/li&gt;
+    &lt;li&gt;&lt;a href=&quot;#whisk-sugar&quot;&gt;Whisk in the sugar&lt;/a&gt;&lt;/li&gt;
+    &lt;li&gt;&lt;a href=&quot;#pipe&quot;&gt;Pipe mixture into shapes&lt;/a&gt;&lt;/li&gt;
+    &lt;li&gt;&lt;a href=&quot;#bake&quot;&gt;Bake&lt;/a&gt;&lt;/li&gt;
+    &lt;li&gt;&lt;a href=&quot;#cool&quot;&gt;Allow to cool&lt;/a&gt;&lt;/li&gt;
+  &lt;/ol&gt;</pre>
+
+      <p>
+        <a href="../../working-examples/aria-current/#example2">A working example of identifying the current step in a process</a>.
+      </p>
+    </section>
+
+    <section class="example">
+      <h3>Identify the current time in a timetable</h3>
+     
+      <p>This is an example of <code class="att">aria-current</code> used to mark up the current time in a timetable. It uses the attribute's <code>time</code> value to programmatically identify the current lesson in a class schedule:</p>
+<pre xml:space="preserve">
+  &lt;h2&gt;Morning Class Schedule&lt;/h2&gt;
+  &lt;table&gt;
+  &lt;tr&gt;
+    &lt;th&gt;Time&lt;/th&gt;
+    &lt;th&gt;Subject&lt;/th&gt;
+  &lt;/tr&gt;
+  &lt;tr&gt;
+    &lt;td&gt;
+      &lt;time datetime=&quot;2021-03-25T08:30&quot;&gt;8:30-9:00am&lt;/time&gt;
+    &lt;/td&gt;
+    &lt;td&gt;Morning Meeting&lt;/td&gt;
+  &lt;/tr&gt;
+  &lt;tr&gt;
+    &lt;td&gt;
+      &lt;time datetime=&quot;2021-03-25T09:00&quot;&gt;9:00-9:30am&lt;/time&gt;
+    &lt;/td&gt;
+    &lt;td&gt;Math&lt;/td&gt;
+  &lt;/tr&gt;
+  &lt;tr&gt;
+    &lt;td&gt;
+      &lt;time aria-current=&quot;time&quot; datetime=&quot;2021-03-25T09:30&quot;&gt;9:30-10:30am&lt;/time&gt;
+    &lt;/td&gt;
+    &lt;td&gt;Reading&lt;/td&gt;
+  &lt;/tr&gt;
+  &lt;tr&gt;
+    &lt;td&gt;
+      &lt;time datetime=&quot;2021-03-25T10:30&quot;&gt;10:30-10:45am&lt;/time&gt;
+    &lt;/td&gt;
+    &lt;td&gt;Recess&lt;/td&gt;
+  &lt;/tr&gt;
+  &lt;tr&gt;
+    &lt;td&gt;
+      &lt;time datetime=&quot;2021-03-25T10:45&quot;&gt;10:45-11:30am&lt;/time&gt;
+    &lt;/td&gt;
+    &lt;td&gt;Music&lt;/td&gt;
+  &lt;/tr&gt;
+  &lt;tr&gt;
+    &lt;td&gt;
+      &lt;time datetime=&quot;2021-03-25T11:30&quot;&gt;11:30-12:00pm&lt;/time&gt;
+    &lt;/td&gt;
+    &lt;td&gt;Independent Work&lt;/td&gt;
+  &lt;/tr&gt;
+  &lt;tr&gt;
+    &lt;td&gt;
+      &lt;time datetime=&quot;2021-03-25T12:00&quot;&gt;12:00-12:45pm&lt;/time&gt;
+    &lt;/td&gt;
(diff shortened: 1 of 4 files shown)


detlevhfischer:
View Technique [Using aria-current to identify the current item in a set](https://raw.githack.com/w3c/wcag/6d8be58d4555cfad6370907ba50d0c243158603f/techniques/aria/ARIA25.html)

detlevhfischer:
Looks good. The only thing I wondered about is if the current link should actually link to the current page or better get rid of the `href`-Attribute - not sure.

fstrr:
I go back and forth on removing the `href` attribute from current links, but have pretty much settled on keeping it. My thinking is that having the `aria-current` attribute gives a screen reader user the information that this is the current page, which hopefully clarifies things.

detlevhfischer:
> having the aria-current attribute gives a screen reader user the information that this is the current page, which hopefully clarifies things.

@fstrr There is the visual aspect as well, the link with aria-current will likely still appear active (underline, other focus indication) even though is does nothing but reload the page (and some people say that can be a useful thing). I can certainly live with both options, I just wanted views on what would be the preferable implementation (as Techniques will be seen as best practice).

JAWS-test:
> The only thing I wondered about is if the current (recursive) link should actually link to the current page or better get rid of the href-Attribute

It can also be a main menu and the current page was located under the same main menu item, but is not identical to the page accessed via the menu item in the main menu. That's why I think href makes sense

OwenEdwards:
> I go back and forth on removing the href attribute from current links, but have pretty much settled on keeping it. My thinking is that having the aria-current attribute gives a screen reader user the information that this is the current page, which hopefully clarifies things.

The problem with removing the `href` from the current link is that it is then just semantically a `<span>`  element, which is no different from static text, so it isn't very clear to screen reader users. To some extent, what the developer may be trying to achieve is closer to `<a href="..." disabled>`, so that it is still semantically a link but is no longer focusable or actionable. It's not exactly `disabled` though, because a disabled button doesn't need to meet contrast requirements, whereas the link to the current page _probably_ shouldn't be grayed out (maybe it should in some cases?).

But I think at least JAWS treats a link with aria-current on it as disabled, so it conveys the correct semantic intent, right?

JAWS-test:
> But I think at least JAWS treats a link with aria-current on it as disabled, so it conveys the correct semantic intent, right?

In HTML a link cannot be disabled (the attribute is not allowed). aria-disabled is allowed on links. Disabled and aria-current are different things and are also output differently by screen readers. But a link with aria-current is not automatically disabled (because e.g. with this link the current page can be reloaded), so the different output is correct.

OwenEdwards:
@JAWS-test sorry, I should have been clearer - I didn't mean that `<a href="..." disabled>` was valid HTML.

I meant that in some ways what developers mean when they remove the `href` on a link that refers to the current page is similar to what they mean when using `<button disabled>`, in that it is still semantically a button. But removing an `href` from a link has a more significant effect, and it seems like a lot of developers don't realize that.

I agree that `disabled` and `aria-current` aren't the same thing, though. I just didn't express it very well!

I guess I should ask the question - how do JAWS (and other screen readers) announce each of these:

1. `<a href="..." aria-current="page">name of current page</a>`
2. `<a aria-current="page">name of current page</a>` (where the link doesn't have an `href`)
3. `<span aria-current="page">name of current page</span>`

JAWS-test:
@OwenEdwards 

> how do JAWS (and other screen readers) announce each of these

With JAWS and NVDA, "current page" is correctly output as the status in all three examples. The only difference between example 1 and 2+3 is that in example 1 a link is output and in 2+3 only text is output. Only example 1 is reached via quick navigation for links and is displayed in the link overview. Only example 1 gets focus with tab key

kfranqueiro:
Looks like this lost momentum a few years ago. I've updated the branch to resolve conflicts since the way relationships between Techniques and Understanding are stored changed, and fixed pre/code tags to match the more recent cleanup work on that.

New previews:

- [ARIA25 technique](https://deploy-preview-1707--wcag2.netlify.app/techniques/aria/aria25)
- [1.3.1: Info and Relationships](https://deploy-preview-1707--wcag2.netlify.app/understanding/info-and-relationships)
- [2.4.8: Location](https://deploy-preview-1707--wcag2.netlify.app/understanding/location)

understanding/20/info-and-relationships.html
```diff
@@ -33,7 +33,7 @@ <h2>Intent of Info and Relationships</h2>
       
       <p>
          						Sighted users perceive structure and relationships through various visual cues
-         — headings are often in a larger, bold font separated from paragraphs by blank lines;
+         &mdash; headings are often in a larger, bold font separated from paragraphs by blank lines;
```
patrickhlauke:
```suggestion
         – headings are often in a larger, bold font separated from paragraphs by blank lines;
```

seems unnecessary to change valid utf-8 to html entity

working-examples/aria-current/index.html
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    </style>
+
+  </head>
+
+  <body>
+    
+    <svg hidden>
+      <symbol id="carat" viewbox="0 0 16 16">
```
patrickhlauke: is this symbol even used? (and it would be "caret" not "carat", unless we're talking gold)

bruce-usab:
Looks like a great addition!  I think it needs closer technical review than I can provide.  Reviewed on 13 Feb TF call.  Left in drafted to give folks additional time to review.

understanding/20/info-and-relationships.html
kfranqueiro: FYI what you put in this suggestion is an en dash (–), not an em dash (—), so it would still be changing what was there before

patrickhlauke: ah, good catch @kfranqueiro ... got my windows ALT codes confused
```suggestion
         — headings are often in a larger, bold font separated from paragraphs by blank lines;
```

working-examples/aria-current/index.html
patrickhlauke: removed the symbol

patrickhlauke:
note that this will need to be renumbered if the progress bar + live region technique from https://github.com/w3c/wcag/pull/4686 gets merged first

bruce-usab:
Reviewed on TF call 20 Feb and moved to Ready for approval.

working-examples/aria-current/index.html
```diff
@@ -0,0 +1,267 @@
...
+      }
+
+      #recipe{
+        background:var(--redish);
+        color:var(--white);
+        margin-top:0;
+        text-align:center;
+      }
+
+      .current--step, .current--step a{
```
dbjorge: nit: consider instead dropping the use of the `current--step` class and using this CSS; it'd be nice for the example to encourage the use of the aria attribute as the CSS condition, it makes it harder for authors that copy this as a starting point to create bugs with CSS and ARIA that are out of sync.

```suggestion
      li:has([aria-current=step]), [aria-current=step]{
```

working-examples/aria-current/index.html
```diff
@@ -0,0 +1,267 @@
...
+
+      tr td{
+        border-bottom:1px solid hsl(0,0%,90%);
+      }
+
+      th, td{
+        padding:0.25em;
+      }
+
+      tr.current-time td{
```
dbjorge: same nit as above:


```suggestion
      tr:has([aria-current=time]){
```

working-examples/aria-current/index.html
```diff
@@ -0,0 +1,267 @@
...
+                </td>
+                <td>Morning Meeting</td>
+              </tr>
+              <tr>
+                <td>
+                  <time datetime="2021-03-25T09:00">9:00-9:30am</time>
+                </td>
+                <td>Math</td>
+              </tr>
+              <tr class="current-time">
```
dbjorge: class is unnecessary if you accept the above suggestion

```suggestion
              <tr>
```

working-examples/aria-current/index.html
```diff
@@ -0,0 +1,267 @@
...
+      </section>
+
+      <section id="example2">
+        <h2>Example 2: Using <code class="att">aria-current</code> to mark up the steps in a recipe</h2>
+
+        <div aria-labelledby="recipe" role="group">
+          <h3 id="recipe">Meringue Recipe Steps</h3>
+          <ol>
+            <li><a href="#oven">Heat your oven</a></li>
+            <li class="current--step">
```
dbjorge: unnecessary if you accept the above suggestion

```suggestion
            <li>
```

working-examples/aria-current/index.html
```diff
@@ -0,0 +1,267 @@
...
+      #recipe{
+        background:var(--redish);
+        color:var(--white);
+        margin-top:0;
+        text-align:center;
+      }
+
+      .current--step, .current--step a{
+      color:var(--redish);
+      font-weight:bold;
```
dbjorge:
```suggestion
        color:var(--redish);
        font-weight:bold;
```

working-examples/aria-current/index.html
```diff
@@ -0,0 +1,267 @@
...
+        color:var(--blue-dark);
+      }
+
+    .time-icon{
+      color:var(--blue-mid);
+      fill:var(--white);
+      height:1.5em;
+      vertical-align:middle;
+      width:0.8em;
+    }
```
dbjorge:
```suggestion
      .time-icon{
        color:var(--blue-mid);
        fill:var(--white);
        height:1.5em;
        vertical-align:middle;
        width:0.8em;
      }
```

working-examples/aria-current/index.html
```diff
@@ -0,0 +1,267 @@
...
+      font-weight:bold;
+      }
+
+      [aria-current=step] svg{
+        color:var(--redish);
+        fill:currentColor;
+        height:0.75em;
+        width:0.75em;
+      }
+
```
dbjorge: This appears unused; only the `time` example is using an svg, not the step one

```suggestion
```

techniques/aria/ARIA25.html
```diff
@@ -0,0 +1,174 @@
+<!DOCTYPE html>
+  <html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml">
```
dbjorge: The xml lang stuff isn't necessary and we don't use it in most newer examples

```suggestion
  <html lang="en">
```

working-examples/aria-current/index.html
```diff
@@ -0,0 +1,267 @@
...
+    <meta charset="UTF-8">
+    <title>Using aria-current to identify the current item in a set examples</title>
+    <style>
+
+    :root {
+        font-size: 100%;
+        box-sizing: border-box;
+        scroll-behavior: smooth;
+        -moz-osx-font-smoothing: grayscale;
+        -webkit-font-smoothing: antialiased;
```
dbjorge: Are these necessary for the example? I think it'd probably be better to keep the example focused. We should especially avoid showing an example of smooth scrolling without a `prefers-reduced-motion` guard (but for this example, I think better to just remove it rather than add a guard)


```suggestion
```

working-examples/aria-current/index.html
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    </style>
+
+  </head>
+
+  <body>
+    
+    <svg hidden>
+      <symbol id="clock" viewbox="0 0 16 16">
```
dbjorge: svg `viewBox` is technically case sensitive; HTML5 includes a case adjustment step to correct it, so browsers will render it correctly as-is, but better to be standards compliant in the example.

```suggestion
      <symbol id="clock" viewBox="0 0 16 16">
```

working-examples/aria-current/index.html
```diff
@@ -0,0 +1,267 @@
...
+        <circle cx="8" cy="8" r="8" color="class"></circle>
```
dbjorge: `color="class"` doesn't do anything here

```suggestion
        <circle cx="8" cy="8" r="8"></circle>
```

fstrr:
Thanks for that review, @dbjorge.

dbjorge:
Thanks @fstrr, confirmed that you got all the comments.

My only remaining concern is more general: I just tested the current behavior in TalkBalk v16.2 on my phone (OnePlus 12, version CPH2583_16.0.0.202(EX01)), and it doesn't appear that it reads `aria-current` state at all (either in any of these examples or in any of the example cases in [a11ysupport.io's test file](https://a11ysupport.io/tests/html/aria/aria-current.html). Are we really comfortable with marking something that isn't supported in TalkBalk as a "Sufficient" technique? I know we couch our general [Understanding Techniques](https://www.w3.org/WAI/WCAG22/Understanding/understanding-techniques) doc's Sufficient Techniques section in "...and it is accessibility-supported for your users", but I think at absolute minimum we should probably include a warning note about it, and I'm not so sure we should mark it a sufficient technique at all (even with a warning) if a current version of a top-2 mobile screen reader doesn't support it.

~~(:-1: on the PR description is to mark this as needing more discussion)~~

Edit: see later comments; this appears to be an issue specific to some version in play on my particular phone, and doesn't repro in a reference Pixel.

alastc:
> My only remaining concern is more general: I just tested the current behavior in TalkBalk v16.2 on my phone (OnePlus 12, version CPH2583_16.0.0.202(EX01)), and it doesn't appear that it reads `aria-current` state at all ... Are we really comfortable with marking something that isn't supported in TalkBalk as a "Sufficient" technique? 

Assuming you would fail a navigation link that highlights the current page that didn't use something to convey that, there are a few options: `aria-current` is one way, but you could hide text, or make it a non-link, there are a few ways to achieve the same thing. 

The way [WCAG2 defines accessibility support](https://www.w3.org/TR/WCAG22/#dfn-accessibility-supported) talks about things being widely distributed. I read that as "if X is supported by widely available AT then it's ok". If it's just talkback that doesn't support it, then I think it satisfies the general accessibility supported concept. (This makes me happier with the approach we're taking in WCAG3 and actually specifying what we test with).

As such, I think having this as a sufficient technique, with a warning, is reasonable.

techniques/aria/ARIA26.html
```diff
@@ -0,0 +1,172 @@
...
+
+    <section id="applicability">
+      <h2>When to Use</h2>
+      <p>Technologies that support <a href="https://www.w3.org/TR/wai-aria/">Accessible Rich Internet Applications (WAI-ARIA)</a>.</p>
+    </section>
+
+    <section id="description">
+      <h2>Description</h2>
+      <p>The purpose of this technique is to programmatically indicate the current item in a user interface component that contains a set of related items. The <code>aria-current</code> attribute provides a way to programmatically indicate the element that is usually only visually highlighted, for example: the current page in a navigation bar.
+      </p>
```
alastc, line 23:
```suggestion
      </p>
      <p class="note">This technique has wide support across current assistive technologies. As of publication a mobile screen reader does not announce the <code>aria-current</code> attribute. It is still a useful technique, and an issue has been raised with that screen reader.</p>
```

patrickhlauke:
@dbjorge maybe i'm holding it wrong, but ... chrome/android/talkback 16.2.xxx (maybe not the absolute latest model phone/android), `aria-current` seems to be supported just fine...

https://github.com/user-attachments/assets/809325f7-d451-47db-a3b5-3bbac6d6cab5

so ... it may be a bug/shortcoming/regression/problem only on your particular phone?

dbjorge:
> it may be a bug/shortcoming/regression/problem only on your particular phone?

@patrickhlauke  Good call out. I just tested it on my husband's phone (Pixel 7a, Android 16, Talkback 16.2, Chrome 145) and got similar results to your video (though, interestingly, on [a11ysupport.io's aria-current test page](https://a11ysupport.io/tests/html/aria/aria-current.html), I do repro the behavior they recorded in their 2021 test where `<li aria-current="time">` wasn't announced - very bizarre that it's announced in this new example but not when it's on an `<li>`).

That's enough for me to retract my concern. Thanks for double checking me!

patrickhlauke:
confirming that my TB does indeed not announce `aria-current` when on the `<li>` ... which sounds like a bug in TB

PR #1724, opened 2021-04-07
typo

mitchellevan:


alastc:
I had to [look this up](https://www.britannica.com/story/lay-lie-lied-lain-when-do-we-use-which), but I think you're right!

PR #1780, opened 2021-05-04
Updating the contrast value

### LINKED ISSUE #360: Non-sRGB color spaces, outdated sRGB threshold, and other issues in the "relative luminance" definition
_From @peteroupc on March 24, 2018 8:47_

Also found at [this issue](https://github.com/w3c/wcag/issues/308).  Posting here in case I get a reply sooner.

The value 0.03928 (as used in what was formerly a note to the "relative luminance" definition but what is now a "normative" part of that definition) comes from the sRGB proposal. This is outdated and <s>either</s> 0.04045 <s>or 12.92 * 0.0031308 = 0.040449936 (I don't know which is used in the IEC standard)</s> should appear in its place.

I understand, however, that all three thresholds don't make a difference in the results if only 8-bit linear sRGB values are involved. On the other hand, if no change to the 0.03928 threshold is planned soon, a note on the origin of that threshold should be added to the document (in particular, why that threshold was chosen rather than the one given or implied in the IEC standard).

EDIT (Mar. 27): Clarification.
EDIT (Apr. 9): Struck out part of the text.  Since so few source code files on GitHub use 0.040449936 compared to 0.04045, I reasonably believe that the IEC standard uses 0.04045 as the relevant threshold, assuming it even mentions an inverse transfer function.

_Copied from original issue: w3c/wcag21#815_

alastc:
Closes #360, #1213

### DIFF
--- a/guidelines/terms/20/relative-luminance.html
+++ b/guidelines/terms/20/relative-luminance.html
@@ -12,13 +12,13 @@
    
    <ul>
       
-      <li>if RsRGB &lt;= 0.03928 then <strong>R</strong> = RsRGB/12.92 else <strong>R</strong> = ((RsRGB+0.055)/1.055) ^ 2.4
+      <li>if RsRGB &lt;= 0.04045 then <strong>R</strong> = RsRGB/12.92 else <strong>R</strong> = ((RsRGB+0.055)/1.055) ^ 2.4
       </li>
       
-      <li>if GsRGB &lt;= 0.03928 then <strong>G</strong> = GsRGB/12.92 else <strong>G</strong> = ((GsRGB+0.055)/1.055) ^ 2.4
+      <li>if GsRGB &lt;= 0.04045 then <strong>G</strong> = GsRGB/12.92 else <strong>G</strong> = ((GsRGB+0.055)/1.055) ^ 2.4
       </li>
       
-      <li>if BsRGB &lt;= 0.03928 then <strong>B</strong> = BsRGB/12.92 else <strong>B</strong> = ((BsRGB+0.055)/1.055) ^ 2.4
+      <li>if BsRGB &lt;= 0.04045 then <strong>B</strong> = BsRGB/12.92 else <strong>B</strong> = ((BsRGB+0.055)/1.055) ^ 2.4
       </li>
       
    </ul>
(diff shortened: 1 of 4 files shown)


guidelines/terms/20/relative-luminance.html
```diff
@@ -40,6 +40,8 @@
    </p>
 
   </div>
+
+  <p class="note">Before May 2021 the value of 0.04045 in the definition was different. It was taken from an older version of the specification and has been updated. It should have no practical effect on the calculations.</p>
```
patrickhlauke, line 44: worth mentioning the old value, to provide a historical record (and to have a reference for any tools that may have used this value internally)

`was different (0.03928).`

techniques/general/G18.html
```diff
@@ -152,6 +152,7 @@
             <li>#4 is true.</li>
          </ul>
       </section>
+      <p class="note">Before May 2021 the value of 0.04045 in the definition was different. It was taken from an older version of the specification and has been updated. It should have no practical effect on the calculations.</p>
```
patrickhlauke, line 155: ditto

mraccess77:
I'd assert that it has no impact on the calculation.   The value falls between 10/255 and 11/255 which falls outside of the values that are set below and above each version and thus the calculated values that could exist are not impacted by this change at all.

patrickhlauke:
> I'd assert that it has no impact on the calculation. The value falls between 10/255 and 11/255 which falls outside of the values that are set below and above each version and thus the calculated values that could exist are not impacted by this change at all.

not worried about end results being different. more in terms of somebody digging into the source code of a tool and being left baffled why that tool may have used that other value / why that particular value.

bruce-usab:
I agree with the logic of mentioning the old value.  But please delete <q>should</q> from the explaination.

mraccess77:
I agree on removing "should".  Both values will work and either can be used to meet the requirement as they provide the same result.

alastc:
All the changes were included and agreed today. 

@michael-n-cooper I've removed the references to the old sRGB standard. Anywhere we used it was in conjunction with the other one, which is the more up to date version. Does removing it from the json file and removing the instances of `[[srgb]]` do the job?

Issue #1775, opened 2021-04-30
Does the adjacent color clause of SC 1.4.11 apply when focus changes swap colors but also are adjacent to the background ?


mraccess77:
I originally wrote this as a discussion question 21 days ago in this github repository but it has gotten no attention.  So I'm posting here as an issue.   
---

In the screenshot below the button background and page background are the same color.  The focus indicator is communicated by a color change of the circular button to become darker.  SC 1.4.11 doesn't require the focused and unfocused states to have contrast - but must the new focused color contrast with adjacent page background colors as well because the focus indicator takes up the whole button area?

![image](https://user-images.githubusercontent.com/5640755/114189634-68b6b000-9918-11eb-9505-6b7432c39c17.png)

For a circular button - what is considered the component rather than page background - is it the bounding box or the visible component?

patrickhlauke:
if i can't see/perceive the darker circle against the overall page background, I can't perceive the focus state. so to me yes the dark circle must contrast with 3:1 against the background. yes, we did say previously that we can't simply compare "focused state vs uncofused state", but the individual states themselves, when in effect, also still need to pass adjacent colour criterion for the bit that matters...which in this case, is the focus indicator.

patrickhlauke:
the bone of contention in the past was that you can't compare "focused vs unfocused" because unless you have both focused and unfocused components butting up right against each other, there's no "adjacent" colours to compare things to.

patrickhlauke:
in the case you're showing in your screenshot, no, it wouldn't make a difference if the dark circle got larger (larger than what? the unfocused state has no circle?)

there's no "inset" here, as there's no outer boundary. or am i missing something?

patrickhlauke:
randomly, if somebody tried a "ah, but even when not focused, there's a blue circle already there, it's just that it happens to be the same colour as the background so you can't see it" argument here, i'd say fine, this then fails 1.4.1 use of color, since you're changing this "preexisting circle" that isn't actually visible from blue to dark blue only...and the difference is less than 3:1

mraccess77:
@patrickhlauke I think this should fail because the swapped colors touch the adjacent page background - and I think it sounds like you agree - but I wanted to get others take on this because I'm not sure that's how others read it.  I may have mis-read your 2nd comment when I replied above.    

Do you agree with interpretation in the understanding that if the control had an inset indicator that swapped colors but did not reach the border then it would not be covered under SC 1.4.11?
[...]

Issue #1781, opened 2021-05-05
1.4.1 also relevant if hue is the same but brightness is different?


JAWS-test:
The Understanding of 1.4.1 is exclusively about the fact that people who are color blind cannot distinguish color values (hue). It is not about contrasts when the color is the same. Nevertheless, in the discussions here (e.g. https://github.com/w3c/wcag/issues/1775) 1.4.1 is also often applied when the hue is the same and the brightness is different. Color-blind people should probably have no problem with this (even though I am not an expert in this field). Hence my question:
 
* Is 1.4.1 only about hue differences? If so, do we need a new SC for sufficient contrasts (which currently only exists when it comes to adjacent colors (SC 1.4.3, 1.4.6 and SC 1.4.11) or focus (SC 2.4.11 and 2.4.12))?
* or is 1.4.1 in the end about contrast differences (see 3:1 exception). If yes: then shouldn't this be explicitly mentioned in Understanding (i.e. fixation on color blindness removed).

patrickhlauke:
> fixation on color blindness removed

color blindness is not "not being able to perceive hue". a user with different colour perception will differentiate colours (hue+brightness+saturation) if they have a certain contrast delta. that delta can come from particular strong shifts in hue, or from the other two factors.

alastc:
I would think of it as contrast being a mechanism to distinguish colours that doesn't rely on the hue, and there are other mechanisms as well (e.g. shape). I'm not seeing a reason to update the understanding doc.

JAWS-test:
> color blindness is not "not being able to perceive hue".

@patrickhlauke Someone who is e.g. red-green-blind can therefore distinguish a light red from a dark red worse than someone who is not red-green-blind?

JAWS-test:
@alastc I'm afraid I don't understand your answer. Why do we need contrast requirements for the focus (SC 2.4.11) if they are already included in 1.4.1? Apart from the fact that 1.4.1 does not contain the 3:1 exception in the normative SC text.

alastc:
I don't think the other SCs are covered by 1.4.1, and it would be very confusing to try and do it that way.

> Someone who is e.g. red-green-blind can therefore distinguish a light red from a dark red (same hue, different brightness) worse than someone who is not red-green-blind?

My understanding is that in red-green colourblindness you are missing some/all of the 'red cones', meaning that you don't see the red hue and it is darker than people who have the red cones.

So it isn't just about misisng a hue, but the brightness can be affected as well. 

[...]

PR #1788, opened 2021-05-09
Changes to 1.4.1 Use of Color understanding

patrickhlauke:
Various tweaks/additions to the 1.4.1 understanding doc, as this seems to still cause some confusion (mainly due to its overlap with other SCs present and future - see #1775)

* Remove benefit mentioning monochrome displays - this situation may not be helped by the "using a difference in contrast/brightness, not just *hue*" way of addressing this SC.
* Change the last benefit, as it only applies if something else like extra visible/announced text is added. And the "listen" part implies that this is intended to help AT users, which for this SC is not the case (that's a 1.3.1 / 4.1.2 issue)
* Add two more examples - toggles whose state (on or off) is only distinguished by a color change, and controls whose focused state is only conveyed by a color change.
* Reword first examples for information conveyed by color differences so they don't sound like instructions/descriptions - currently, these examples sound far more like 1.3.3 Sensory Characteristics related instructions
* restructure the examples as a more readable/scannable list
* add an explicit example of links versus static text 

Closes https://github.com/w3c/wcag/issues/1775

### DIFF
--- a/understanding/20/use-of-color.html
+++ b/understanding/20/use-of-color.html
@@ -29,11 +29,12 @@
          only in color.
       </p>
       
-      <p>Examples of information conveyed by color differences: “required fields are red",
-         “error is shown in red", and “Mary's sales are in red, Tom's are in blue". Examples
-         of indications of an action include: using color to indicate that a link will open
-         in a new window or that a database entry has been updated successfully. An example
-         of prompting a response would be: using highlighting on form fields to indicate that
+      <p>Examples of information conveyed by color differences: required fields are only identified
+         by changing their border to a red color; sales figures for different employees are only
+         visually distinguished by using different foreground colors( e.g. Mary's sales are in red,
+         Tom's are in blue). Examples of indications of an action include: using color to indicate
+         that a link will open in a new window or that a database entry has been updated successfully.
+         An example of prompting a response would be: using highlighting on form fields to indicate that
          a required field had been left blank. Examples of distinguishing a visual element:
          for a series of toggle buttons, the pressed/activated buttons are only visually indicated
          by a difference in color; the interactive element that currently has focus is only
--- a/understanding/20/use-of-color.html
+++ b/understanding/20/use-of-color.html
@@ -29,20 +29,26 @@
          only in color.
       </p>
       
-      <p>Examples of information conveyed by color differences: required fields are only identified
+      <p>Examples of common failures include:</p>
+      <ul>
+        <li><strong>Information conveyed by color differences:</strong> required fields are only identified
          by changing their border to a red color; sales figures for different employees are only
          visually distinguished by using different foreground colors( e.g. Mary's sales are in red,
-         Tom's are in blue). Examples of indications of an action include: using color to indicate
-         that a link will open in a new window or that a database entry has been updated successfully.
-         An example of prompting a response would be: using highlighting on form fields to indicate that
-         a required field had been left blank. Examples of distinguishing a visual element:
-         for a series of toggle buttons, the pressed/activated buttons are only visually indicated
-         by a difference in color; the interactive element that currently has focus is only
-         distinguished by changing the color of the element's <em>existing</em> background or
-         border - however, if focus indication introduces a new visual element that was not previously
-         present, this does not count as a change in color, since the indicator was not presentation
-         to begin with.
-      </p>
+         Tom's are in blue).</li>
+        <li><strong>Indicating an action:</strong> using color to indicate
+         that a link will open in a new window or that a database entry has been updated successfully.</li>
+        <li><strong>Prompting a response:</strong> using highlighting on form fields to indicate that
+         a required field had been left blank.</li>
+        <li><strong>Distinguishing a visual element:</strong> links are only distinguished from surrounding text
+         using a subtle color difference (with a contrast difference that is less than 3:1); for a series of toggle
+         buttons, the pressed/activated buttons are only visually indicated by a difference in color;
+         the interactive element that currently has focus is only distinguished by changing the color
+         of the element's existing background or border. However, particularly in the case of state/focus
+         indication, note that if the visual indicator was not previously present, this does not count as
+         a change in color. For instance, if a focused element has a new border, or a new solid background
+         color, that was not there in its unfocused state, this does not count as a change of color alone,
+         since the indicator was not presentation to begin with.</li>
+      </ul>
 
       <div class="note">
          
--- a/understanding/20/use-of-color.html
+++ b/understanding/20/use-of-color.html
@@ -98,13 +98,7 @@
             in other visual ways.
          </li>
          
-         <li>People using limited color monochrome displays may be unable to access
-            color-dependent information.
-         </li>
-         
-         <li>Users who have problems distinguishing between colors can look or listen for text
-            cues.
-         </li>
+         <li>Users who have problems distinguishing between colors can look for other visual cues.</li>
          
       </ul>
       
--- a/understanding/20/use-of-color.html
+++ b/understanding/20/use-of-color.html
@@ -34,7 +34,13 @@
          of indications of an action include: using color to indicate that a link will open
          in a new window or that a database entry has been updated successfully. An example
          of prompting a response would be: using highlighting on form fields to indicate that
-         a required field had been left blank.
+         a required field had been left blank. Examples of distinguishing a visual element:
+         for a series of toggle buttons, the pressed/activated buttons are only visually indicated
+         by a difference in color; the interactive element that currently has focus is only
+         distinguished by changing the color of the element's <em>existing</em> background or
+         border - however, if focus indication introduces a new visual element that was not previously
+         present, this does not count as a change in color, since the indicator was not presentation
+         to begin with.
       </p>
 
       <div class="note">
@@ -131,7 +137,7 @@
             
             <p>
                								       
-               <strong> An examination.</strong>
+               <strong>An examination</strong>
                							     
             </p>
             
@@ -142,6 +148,33 @@
                the numbers.
             </p>
             
+         </li>
+
+         <li>
+            <p>
+               <strong>A series of toggle buttons</strong>
+            </p>
+            <p>Users are presented with a series of toggle buttons that can be turned on or off. Visually,
+               when a toggle button is turned on, it not only has a different background color, but is
+               also styled to appear "pressed", with a distinctive inner shadow. Alternatively, toggles that
+               are turned on could have an extra "ticked" icon to distinguish them from toggles that are
+               turned off.
+            </p>
+         </li>
+
+         <li>
+            <p>
+               <strong>Focus indication</strong>
+            </p>
+            <p>When an interactive control (like a button, toggle, or checkbox) receives focus, its background color
+               changes. In addition to this, an additional outline (which was not present in the control's unfocused state)
+               appears around the control.
+            </p>
+            <p>Note that when it comes to focus indication, there are additional criteria
+               that apply. For instance, in this case, the additional outline will pass this criterion - as it provides
+               another visual cue to distinguish the focused element - but the outline itself will still need to pass 
+               <a href="#non-text-contrast">1.4.11 Non-text Contrast</a>. 
+            </p>
          </li>
          
       </ul>


JAWS-test:
@patrickhlauke 

Thank you for your PR. I still have 3 comments

Is it possible to include a reference that color means not only difference in hue, but also in saturation or lightness, so that e.g. using light gray and dark gray or white and light gray is also color coding? See: https://github.com/w3c/wcag/issues/1781

> However, particularly in the case of state/focus  indication, note that if the visual indicator was not previously present, this does not count as a change in color. For instance, if a focused element has a new border, or a new solid background  color, that was not there in its unfocused state, this does not count as a change of color alone, since the indicator was not presentation to begin with

I'm not sure that's sufficiently understandable: Every element has a background. What is a new background that wasn't there before? Probably you mean when the background of the element is identical to the background of the page/surrounding area and there is no border arount the element. In this case, however, insufficient contrast of the new background would automatically be a violation of SC 1.4.11 - this could be mentioned here already and not below where you mention it, but only in connection with focus.

> Note that when it comes to focus indication, there are additional criteria that apply. For instance, in this case, the additional outline will pass this criterion - as it provides another visual cue to distinguish the focused element - but the outline itself will still need to pass 1.4.11 Non-text Contrast

I find this section a bit misleading because it sounds like the additional requirements only apply to the focus indicator. But they apply to all information transmitted via color. For the focus indicator there are only specific requirements from WCAG 2.2 with SC 2.4.11 and 2.4.12.

patrickhlauke:
Any news on this @alastc ?

wardav:
Forgive my ignorance (and please signpost me to another ticket if there is a better place for this), but I am still struggling to wrap my head around how the SC applies to focus indication. 

At time of writing, the live wording does not seem to mention focus indication explicitly. As well, focus visible for not seem to talk about it. There is some wording in Non-Text Contrast, though.

These updates talk about checking focus indication, but I am not sure I understand it, still. 

* When I focus in something, I'll check that the touching colours pass **Non-Text Contrast** (including any focus indication), but I won't check the focus and unfocused states against each other, though. 
* In the future, I may use **Focus Appearance** to check the before and after contrast, thickness of the focus, and as on.

But for use of colour, my understanding was I am looking for one of two things:

1. The use of colour relies on my ability to perceive an exact colour (e.g. green).
2. The use of colour relies on my ability to perceive that two colours are different (e.g. when I focus on something I don't perceive the colour of the new border, only that it is there).

For example, with informative text: green means good and red means bad. No amount of contrast matters here. I need to have access to the same information some other way, such as an up arrow and a down arrow. 

But with the case of focus, do users need to visually perceive the exact colour, or do they just need to perceive that the colour is different? 

For example, if a link has a border and the border changes colour radically (i.e. the difference in the colours used for the border contrast very strongly), based on the wording in the success criterion, would I not be able to perceive that...

1. the border has visually changed
2. the border is now different to the other borders on the page.

This seems to be the case in this wording:

> If content is conveyed through the use of colors that differ not only in their hue, but that also have a significant difference in lightness, then this counts as an additional visual distinction, as long as the difference in relative luminance between the colors leads to a contrast ratio of 3:1 or greater. For example, a light green and a dark red differ both by color (hue) and by lightness, so they would pass if the contrast ratio is at least 3:1. 

> However, if content relies on the user's ability to accurately perceive or differentiate a particular color an additional visual indicator will be required regardless of the contrast ratio between those colors. For example, knowing whether an outline is green for valid or red for invalid.

However, there is then the following as an example failure

> the interactive element that currently has focus is only distinguished by changing the color of the element's existing background or border

patrickhlauke:
it's not necessarily being able to perceive *exact* colour. just that colour is used to signify something, identify something, etc.

say you have buttons with gray backgrounds. the focused one becomes subtly greenish/gray to differentiate it. the difference is too subtle to count as a change of luminosity (the difference between the gray and the greenish/gray is less than 3:1). in that case, the focus indication relies purely on the ability to perceive colour/colour differences (not which *exact* colour is used)

wardav:
If in that case, the difference was more than 3:1, would you still fail it? For example, the border changes from blue to orange, and otherwise passes the non-text contrast checks.

patrickhlauke:
if the change is 3:1 or higher, it doesn't count as colour alone, as it also significantly changes in lightness/luminosity. as per the second note in https://www.w3.org/WAI/WCAG21/Understanding/use-of-color.html

patrickhlauke:
@alastc @mbgower any chance this could be looked at?

understanding/20/use-of-color.html
```diff
@@ -38,13 +38,26 @@ <h2>Intent of Use of Color</h2>
...
-         in a new window or that a database entry has been updated successfully. An example
-         of prompting a response would be: using highlighting on form fields to indicate that
-         a required field had been left blank.
-      </p>
+      <p>Examples of common failures include:</p>
+      <ul>
+        <li><strong>Information conveyed by color differences:</strong> required fields are only identified
+         by changing their border to a red color; sales figures for different employees are only
+         visually distinguished by using different foreground colors( e.g. Mary's sales are in red,
+         Tom's are in blue).</li>
```
giacomo-petri, line 46:
```suggestion
        <li>
            <strong>Information conveyed by color differences:</strong> 
            <ul>
                <li>required fields are only identified by changing their border to a red color;</li>
                <li>the company transactions are only visually distinguished by using different foreground colors: red for debits and green for credits;</li>
                <li>subway lines are visually distinguished solely by employing various filled colors.</li>
            </ul>
        </li>
```

Not essential, but I've proposed a slightly different example, replacing the scenario involving Mary and Tom with a more common one, where credit and debit transactions are distinguished solely by colors. 

Additionally, I've included another common example, where subway lines are identified only by colors. 

As for the WCAG policy on nested lists, I'm uncertain. When there's only one nested list, it often feels more organized to me, but I'm open to using multiple sentences if needed and change my suggestion.

understanding/20/use-of-color.html
```diff
@@ -38,13 +38,26 @@ <h2>Intent of Use of Color</h2>
...
+        <li><strong>Distinguishing a visual element:</strong> links are only distinguished from surrounding text
+         using a subtle color difference (with a contrast difference that is less than 3:1); for a series of toggle
+         buttons, the pressed/activated buttons are only visually indicated by a difference in color;
+         the interactive element that currently has focus is only distinguished by changing the color
+         of the element's existing background or border. However, particularly in the case of state/focus
+         indication, note that if the visual indicator was not previously present, this does not count as
+         a change in color. For instance, if a focused element has a new border, or a new solid background
+         shape, that was not there in its unfocused state, this does not count as a change of color alone,
+         since the indicator was not presentation to begin with.</li>
+      </ul>
```
giacomo-petri, line 60:
```suggestion
        <li><strong>Distinguishing a visual element:</strong>
            <ul>
                <li>links are only distinguished from surrounding text using a subtle color difference (with a contrast difference that is less than 3:1);</li>
                <li>for a series of toggle buttons, the pressed/activated buttons are only visually indicated by a difference in color;</li>
                <li>the interactive element that currently has focus is only distinguished by changing the color of the element's existing background or border.</li>
         </ul>
     </li>
     <p>However, particularly in the case of state/focus indication, note that if the visual indicator was not previously present, this does not count as a change in color. For instance, if a focused element has a new border, or a new solid background shape, that was not there in its unfocused state, this does not count as a change of color alone, since the indicator was not presentation to begin with.</p>
```

Consistent with the change above

understanding/20/use-of-color.html
```diff
@@ -139,13 +156,7 @@ <h2>Benefits of Use of Color</h2>
...
          </li>
          
-         <li>People using limited color monochrome displays may be unable to access
-            color-dependent information.
-         </li>
-         
-         <li>Users who have problems distinguishing between colors can look or listen for text
-            cues.
-         </li>
+         <li>Users who have problems distinguishing between colors can look for other visual cues.</li>
```
giacomo-petri, line 159: Not a problem, but in points 1 and 2, we delineate the beneficiaries of a proper implementation. In points 3 and 4, alongside identifying the beneficiaries, we also clarify that they will reap the benefits if other visual cues are present (although this may not be necessary). Should we stay consistent?

understanding/20/use-of-color.html
```diff
@@ -157,18 +168,27 @@ <h2>Examples of Use of Color</h2>
...
-               cannot perceive all the color differences can still understand the image by relying on
-               the numbers.</dd>
+             present based <strong>both</strong> on the colors used, as well as numbers next to each
+             element. A legend shows the color and number for each type of element. Sighted users who
+             cannot perceive all the color differences can still understand the image by relying on
+             the numbers.</dd>
+         <dt>A series of toggle buttons</dt>
+         <dd>Users are presented with a series of toggle buttons that can be turned on or off. Visually,
+             when a toggle button is turned on, it not only has a different background color, but is
+             also styled to appear "pressed", with a distinctive inner shadow. Alternatively, toggles that
```
giacomo-petri, line 184: I suggest substituting the term "shadow" with an alternative. The word "shadow" may evoke associations with CSS features like box-shadow, which might not always be visible, especially in high contrast modes.

giacomo-petri:
I've suggested a few minor adjustments, but I already approve this PR as my edits do not alter the intended meaning.

giacomo-petri (approved):

understanding/20/use-of-color.html
patrickhlauke: as the current thinking is that HCM is still not directly covered by requirements of the SC (as it's effectively a customisation/change to author styles that is not standardised, so can be unpredictable in some cases for authors), I may leave this as is for now.

understanding/20/use-of-color.html
patrickhlauke: i think with the first two the only thing you could then say is a platitude/truism a la ", so not relying on color alone will help them". in a sense, the first three bullets all get to the same point and could even be combined.

detlevhfischer:
So it seems the illustratons are gone? Were they more confusing than helpful?

giacomo-petri:
I can't recall any illustration for 1.4.1.
@detlevhfischer, were you referring to 1.4.11 perhaps?

MrFirthy (approved):

detlevhfischer:
I had prepared are two illustrations for the examples (row of buttons in unfocused and focused state, inline link in unfocused and focused state) which are not present in the rendered view and would help understanding how 1.4.1 can apply to focus state changes using a change of color. I trust these will be included in a separate PR - correct?

@giacomo-petri
> were you referring to 1.4.11 perhaps?

 No, not 1.4.11 - these illustrations were specifically prepared for 1.4.1 (and @patrickhlauke made some editorial changes to the fig captions and the second illustration following misunderstandings regarding the underline in the second inline text link example).

mbgower:
> rendered view of this PR

@patrickhlauke have you seen @detlevhfischer comment on images? Can you incorporate?

fstrr:
This is really confusing. There's a separate PR, #3717 ("1.4.1 Use of color: adding examples to understanding text") that has Detlev's images in.

bruce-usab:
@patrickhlauke would you be available able to pull the illustrations/example bit from this PR into https://github.com/w3c/wcag/pull/3717 ?

bruce-usab:
@mbgower and I discussed, and we think the phrasing with #3717 is more mature and better socialized.  The discussion threads on both PRs has gotten unwieldly.

mbgower:
@bruce-usab
> would you be available able to pull the illustrations/example bit from this PR into https://github.com/w3c/wcag/pull/3717 ?

 I think you may have your recent comments backward? The illustrations are in [3717](https://github.com/w3c/wcag/pull/3717), not this PR.

understanding/20/use-of-color.html
```diff
@@ -38,13 +38,28 @@ <h2>Intent of Use of Color</h2>
...
-         of prompting a response would be: using highlighting on form fields to indicate that
-         a required field had been left blank.
-      </p>
+      <p>Examples of common failures include:</p>
+      <ul>
+        <li>
+            <strong>Information conveyed by color differences:</strong> 
+            <ul>
+                <li>required fields are only identified by changing their border to a red color;</li>
+                <li>the company transactions are only visually distinguished by using different foreground colors: red for debits and green for credits;</li>
```
mbgower, line 47:
```suggestion
                <li>the nature of transactions are only visually distinguished by using different foreground colors: red for debits and green for credits;</li>
```

understanding/20/use-of-color.html
```diff
@@ -38,13 +38,28 @@ <h2>Intent of Use of Color</h2>
...
-         of prompting a response would be: using highlighting on form fields to indicate that
-         a required field had been left blank.
-      </p>
+      <p>Examples of common failures include:</p>
+      <ul>
+        <li>
+            <strong>Information conveyed by color differences:</strong> 
+            <ul>
+                <li>required fields are only identified by changing their border to a red color;</li>
+                <li>the nature of transactions are only visually distinguished by using different foreground colors: red for debits and green for credits;</li>
```
mbgower, line 47:
```suggestion
                <li>the type of transaction is only visually distinguished by using different foreground colors: red for debits and green for credits;</li>
```

understanding/20/use-of-color.html
```diff
@@ -38,13 +38,28 @@ <h2>Intent of Use of Color</h2>
...
+      <ul>
+        <li>
+            <strong>Information conveyed by color differences:</strong> 
+            <ul>
+                <li>required fields are only identified by changing their border to a red color;</li>
+                <li>the type of transaction is only visually distinguished by using different foreground colors: red for debits and green for credits;</li>
+                <li>subway lines are visually distinguished solely by employing various filled colors.</li>
+            </ul>
+        </li>
+        <li><strong>Indicating an action:</strong> using color to indicate
```
mbgower, line 51:
```suggestion
        <li>using color alone to indicate
```

understanding/20/use-of-color.html
```diff
@@ -38,13 +38,28 @@ <h2>Intent of Use of Color</h2>
...
+            <strong>Information conveyed by color differences:</strong> 
+            <ul>
+                <li>required fields are only identified by changing their border to a red color;</li>
+                <li>the type of transaction is only visually distinguished by using different foreground colors: red for debits and green for credits;</li>
+                <li>subway lines are visually distinguished solely by employing various filled colors.</li>
+            </ul>
+        </li>
+        <li>using color alone to indicate
+         that a link will open in a new window or that a database entry has been updated successfully.</li>
+        <li><strong>Prompting a response:</strong> using highlighting on form fields to indicate that
```
mbgower, line 53:
```suggestion
        <li>using subtle color highlighting on form fields to prompt for a response on 
```

understanding/20/use-of-color.html
```diff
@@ -38,13 +38,28 @@ <h2>Intent of Use of Color</h2>
...
+            <ul>
+                <li>required fields are only identified by changing their border to a red color;</li>
+                <li>the type of transaction is only visually distinguished by using different foreground colors: red for debits and green for credits;</li>
+                <li>subway lines are visually distinguished solely by employing various filled colors.</li>
+            </ul>
+        </li>
+        <li>using color alone to indicate
+         that a link will open in a new window or that a database entry has been updated successfully.</li>
+        <li>using subtle color highlighting on form fields to prompt for a response on 
+         a required field had been left blank.</li>
```
mbgower, line 54:
```suggestion
         required fields that have been left blank.</li>
```

understanding/20/use-of-color.html
```diff
@@ -38,13 +38,28 @@ <h2>Intent of Use of Color</h2>
...
+            </ul>
+        </li>
+        <li>using color alone to indicate
+         that a link will open in a new window or that a database entry has been updated successfully.</li>
+        <li>using subtle color highlighting on form fields to prompt for a response on 
+         required fields that have been left blank.</li>
+        <li><strong>Distinguishing a visual element:</strong>
+            <ul>
+                <li>links are only distinguished from surrounding text using a subtle color difference (with a contrast difference that is less than 3:1);</li>
+                <li>for a series of toggle buttons, the pressed/activated buttons are only visually indicated by a difference in color;</li>
```
mbgower, line 58:
```suggestion
                <li>for a series of toggle buttons, the pressed/activated buttons are only visually indicated by a subtle difference in color;</li>
```

understanding/20/use-of-color.html
```diff
@@ -38,13 +38,28 @@ <h2>Intent of Use of Color</h2>
...
+        <li>using subtle color highlighting on form fields to prompt for a response on 
+         required fields that have been left blank.</li>
+        <li><strong>Distinguishing a visual element:</strong>
+            <ul>
+                <li>links are only distinguished from surrounding text using a subtle color difference (with a contrast difference that is less than 3:1);</li>
+                <li>for a series of toggle buttons, the pressed/activated buttons are only visually indicated by a subtle difference in color;</li>
+                <li>the interactive element that currently has focus is only distinguished by changing the color of the element's existing background or border.</li>
+         </ul>
+     </li>
+     <p>However, particularly in the case of state/focus indication, note that if the visual indicator was not previously present, this does not count as a change in color. For instance, if a focused element has a new border, or a new solid background shape, that was not there in its unfocused state, this does not count as a change of color alone, since the indicator was not presentation to begin with.</p>
```
mbgower, line 62:
```suggestion
     <p>However, particularly in the case of state/focus indication, note that if the visual indicator was not previously present, this does not count as a change in color. For instance, if a focused element has a new border, or a new solid background shape, that was not there in its unfocused state, this does not count as a change of color alone, since the indicator was not present to begin with.</p>
```

understanding/20/use-of-color.html
```diff
@@ -38,13 +38,28 @@ <h2>Intent of Use of Color</h2>
...
+            </ul>
+        </li>
+        <li>using color alone to indicate
+         that a link will open in a new window or that a database entry has been updated successfully.</li>
+        <li>using subtle color highlighting on form fields to prompt for a response on 
+         required fields that have been left blank.</li>
+        <li><strong>Distinguishing a visual element:</strong>
+            <ul>
+                <li>links are only distinguished from surrounding text using a subtle color difference (with a contrast difference that is less than 3:1);</li>
+                <li>for a series of toggle buttons, the pressed/activated buttons are only visually indicated by a subtle difference in color;</li>
```
mbgower, line 58:
```suggestion
                <li>for a series of toggle buttons, the pressed and unpressed buttons are only visually indicated by a subtle difference in color;</li>
```

understanding/20/use-of-color.html
```diff
@@ -38,13 +38,28 @@ <h2>Intent of Use of Color</h2>
...
+        </li>
+        <li>using color alone to indicate
+         that a link will open in a new window or that a database entry has been updated successfully.</li>
+        <li>using subtle color highlighting on form fields to prompt for a response on 
+         required fields that have been left blank.</li>
+        <li><strong>Distinguishing a visual element:</strong>
+            <ul>
+                <li>links are only distinguished from surrounding text using a subtle color difference (with a contrast difference that is less than 3:1);</li>
+                <li>for a series of toggle buttons, the pressed and unpressed buttons are only visually indicated by a subtle difference in color;</li>
+                <li>the interactive element that currently has focus is only distinguished by changing the color of the element's existing background or border.</li>
```
mbgower, line 59: I am removing this example. I believe that including indicators of focus in the Use of Color discussion for anything other than text color changes is likely to result in confusion. Focus Visible does not define a minimum contrast variation, and Use of Color does not list state or focus in its wording. Further, some users are definitely going to be confused about whether shading and other additive visual effects are uses of color. These short comings are specifically why Focus Appearance was introduced in 2.2
```suggestion
```

understanding/20/use-of-color.html
```diff
@@ -38,13 +38,28 @@ <h2>Intent of Use of Color</h2>
...
+                <li>subway lines are visually distinguished solely by employing various filled colors.</li>
+            </ul>
+        </li>
+        <li>using color alone to indicate
+         that a link will open in a new window or that a database entry has been updated successfully.</li>
+        <li>using subtle color highlighting on form fields to prompt for a response on 
+         required fields that have been left blank.</li>
+        <li><strong>Distinguishing a visual element:</strong>
+            <ul>
+                <li>links are only distinguished from surrounding text using a subtle color difference (with a contrast difference that is less than 3:1);</li>
```
mbgower, line 57:
```suggestion
                <li>text links are only distinguished from surrounding text using a subtle color text difference (with a contrast difference that is less than 3:1);</li>
```

understanding/20/use-of-color.html
```diff
@@ -38,13 +38,28 @@ <h2>Intent of Use of Color</h2>
...
+                <li>subway lines are visually distinguished solely by employing various filled colors.</li>
+            </ul>
+        </li>
+        <li>using color alone to indicate
+         that a link will open in a new window or that a database entry has been updated successfully.</li>
+        <li>using subtle color highlighting on form fields to prompt for a response on 
+         required fields that have been left blank.</li>
+        <li><strong>Distinguishing a visual element:</strong>
+            <ul>
+                <li>text links are only distinguished from surrounding text using a subtle color text difference (with a contrast difference that is less than 3:1);</li>
```
mbgower, line 57:
```suggestion
                <li>text links are only distinguished from surrounding text using a subtle text color difference (with a contrast difference that is less than 3:1);</li>
```

understanding/20/use-of-color.html
```diff
@@ -38,13 +38,28 @@ <h2>Intent of Use of Color</h2>
...
-         in a new window or that a database entry has been updated successfully. An example
-         of prompting a response would be: using highlighting on form fields to indicate that
-         a required field had been left blank.
-      </p>
+      <p>Examples of common failures include:</p>
+      <ul>
+        <li>
+            <strong>Information conveyed by color differences:</strong> 
+            <ul>
+                <li>required fields are only identified by changing their border to a red color;</li>
```
mbgower, line 46:
```suggestion
                <li>required fields are only identified by changing their existing border to a red color;</li>
```

understanding/20/use-of-color.html
```diff
@@ -38,13 +38,28 @@ <h2>Intent of Use of Color</h2>
...
+                <li>the type of transaction is only visually distinguished by using different foreground colors: red for debits and green for credits;</li>
+                <li>subway lines are visually distinguished solely by employing various filled colors.</li>
+            </ul>
+        </li>
+        <li>using color alone to indicate
+         that a link will open in a new window or that a database entry has been updated successfully.</li>
+        <li>using subtle color highlighting on form fields to prompt for a response on
```
mbgower, line 53: I know this was a pre-existing example, but I really question if this is a Use of Color failure, and I think we should remove it. If someone used grey to do that prompting (i.e., the background was white by default and it had a grey visual effect to prompt, I don't think that would be considered a Use of Color. A subtle grey highlight may not be a *great* prompt; I'm just saying I don't think it fails this SC.
You somewhat touch on this concept in the Note (which I also think may confuse some people). It is clear to me that the working group did NOT regard the appearance of new content as 'use of color'. Assuming a white background, ANY visually distinguished content (i.e., non-white) is theoretically a color, but clearly Use of Color is talking about the _change_ of color in pre-existing information (i.e., going from an absence of color to the appearance of color is NOT a Use of Color).  Until 2.1, we did not even have a measure for non-text contrast, so I also feel that Use of Color is biased to text -- and to visual elements that are considered part of text elements (such as the outline around a text input).

This entire area should really be more emphatically discussed in the Understanding document, and we should attempt to see if we can get working group consensus on it.

understanding/20/use-of-color.html
```diff
@@ -99,30 +114,34 @@ <h2>Intent of Use of Color</h2>
...
-              Authors can <em>only</em> use color to modify the <code>:visited</code> CSS pseudoclass style. The technology constrains any non-color styling. Due to palette limitations, an author cannot use color alone to achieve 3:1 contrast between link and non-link text as well as between visited and unvisited links while also achieving 4.5:1 contrast for all link and non-link text.
-           </li>
-           <li>
-              Authors also cannot set the visited state of links. The anchor element does not include a "visited" attribute; therefore the author has no ability to alter the state through an attribute setting. As such, authors cannot achieve <a href="info-and-relationships">1.3.1 Info and Relationships</a> or
-              <a href="name-role-value">4.1.2 Name, Role, Value</a> in regard to visited links.
-           </li>
-        </ol>
+     </div>
 
+     <div class="note">
```
mbgower, line 119: It's not really clear what someone is supposed to do with this new note.

understanding/20/use-of-color.html
```diff
@@ -157,18 +170,27 @@ <h2>Examples of Use of Color</h2>
...
+             element. A legend shows the color and number for each type of element. Sighted users who
+             cannot perceive all the color differences can still understand the image by relying on
+             the numbers.</dd>
+         <dt>A series of toggle buttons</dt>
+         <dd>Users are presented with a series of toggle buttons that can be turned on or off. Visually,
+             when a toggle button is turned on, it not only has a different background color, but is
+             also styled to appear "pressed", with a distinctive inner shadow. Alternatively, toggles that
+             are turned on could have an extra "ticked" icon to distinguish them from toggles that are
+             turned off.</dd>
+         <dt>Focus indication</dt>
```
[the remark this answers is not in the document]
patrickhlauke: but unarguably you're just relying on color here to convey something. so while non-text contrast can't compare two different states that are not adjacent, use of color can (because if the change is less than 3:1 it counts as only a change in color, which is a use of color)

understanding/20/use-of-color.html
```diff
@@ -157,18 +170,27 @@ <h2>Examples of Use of Color</h2>
...
          <dd>Students view an SVG image of a chemical compound and identify the chemical elements
-               present based <strong>both</strong> on the colors used, as well as numbers next to each
-               element. A legend shows the color and number for each type of element. Sighted users who
-               cannot perceive all the color differences can still understand the image by relying on
-               the numbers.</dd>
+             present based <strong>both</strong> on the colors used, as well as numbers next to each
+             element. A legend shows the color and number for each type of element. Sighted users who
+             cannot perceive all the color differences can still understand the image by relying on
+             the numbers.</dd>
+         <dt>A series of toggle buttons</dt>
```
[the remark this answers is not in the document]
patrickhlauke: again, non-text contrast can't compare things that aren't visually adjacent. use of color, on the other hand, can

understanding/20/use-of-color.html
```diff
@@ -157,18 +170,27 @@ <h2>Examples of Use of Color</h2>
...
+             Users who cannot perceive the difference between the optional field labels and the
+             red labels for the required fields will still be able to see the icon next to the
+             red labels.</dd>             
          <dt>An examination</dt>
          <dd>Students view an SVG image of a chemical compound and identify the chemical elements
-               present based <strong>both</strong> on the colors used, as well as numbers next to each
-               element. A legend shows the color and number for each type of element. Sighted users who
-               cannot perceive all the color differences can still understand the image by relying on
-               the numbers.</dd>
+             present based <strong>both</strong> on the colors used, as well as numbers next to each
```
mbgower, line 179:
```suggestion
             present based <strong>both</strong> on the colors used, as well as oaasdas next to each
```

understanding/20/use-of-color.html
```diff
@@ -38,13 +38,28 @@ <h2>Intent of Use of Color</h2>
...
-         a required field had been left blank.
-      </p>
+      <p>Examples of common failures include:</p>
+      <ul>
+        <li>
+            <strong>Information conveyed by color differences:</strong> 
+            <ul>
+                <li>required fields are only identified by changing their existing border to a red color;</li>
+                <li>the type of transaction is only visually distinguished by using different foreground colors: red for debits and green for credits;</li>
+                <li>subway lines are visually distinguished solely by employing various filled colors.</li>
```
mbgower, line 48:
```suggestion
                <li>subway lines are visually distinguished solely by employing various filled colors;</li>
```

understanding/20/use-of-color.html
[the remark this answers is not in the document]
mbgower: > but unarguably you're just relying on color here to convey something

Not unarguable, since that's exactly what I'm challenging! I think it is clear that a transition from nothing (blank white page) to _any_ content was not intended to be caught by Use of Color. By that definition, EVERYTHING added to or removed from a page is a failure of use of color, since the text, the symbols, the shadows, etc., is simply a use of color. Color alone is forming the letters. Color alone is forming the shapes. Color alone is forming the fills.
Clearly, Contrast (Minimum) was intended to specifically assess the contrast between text and its background. And Non-text Contrast was intended to assess contrast between any non-text visually present feature and its adjacent color (typically the page background).
I'm not trying to split hairs here. Your final note identifies the exact same problem.
> However, particularly in the case of state/focus indication, note that if the visual indicator was not previously present, this does not count as a change in color. For instance, if a focused element has a new border, or a new solid background shape, that was not there in its unfocused state, this does not count as a change of color alone, since the indicator was not present to begin with.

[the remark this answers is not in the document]
patrickhlauke: getting lost between resolved comments, your myriad of tiny changes (could really do with doing them with proper git and pushing up a single commit, rather than piecemeal) etc. thought this comment was still about toggles, but i see it's the next one. and in this case, this was an example where it's not JUST a use of color change, and it's the additional outline that absolves the example from being a FAIL.

understanding/20/use-of-color.html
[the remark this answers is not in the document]
patrickhlauke: the point of this example was the "also styled to appear pressed" and the alternative use of icon, just to make sure it's more than just a change of background colour alone (below 3:1) that distinguishes toggles that are on and toggles that are off. without those, you're relying purely on the user's ability to be able to tell the difference between the two shades, which fails use of color (and *can't* fail non-text contrast because they're not adjacent, unless the toggles all butt up with each other with no dividing line or anything)

understanding/20/use-of-color.html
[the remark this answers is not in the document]
patrickhlauke: if anything, it's the whole angle of these (existing) examples (which i then added to) ... even the existing ones aren't examples of failures, but examples of "this would fail, were it not for..."

understanding/20/use-of-color.html
patrickhlauke: > Use of Color does not list state or focus in its wording

doesn't mean, in my view, that it excludes it. adding extra visuals is not a use of color. only changing an existing color to another color (with a change that's less than 3:1) is a failure of use of color, I'd argue, as it exactly fits the definition of "Color is not used as the only visual means of conveying information" (where the information here is the state of the component ... not just focus, but also pressed/toggled/etc)

detlevhfischer:
[Rendered view ](https://raw.githack.com/w3c/wcag/de207432ca453a7fbe5089dfba900691a378a51d/understanding/20/use-of-color.html)
Wasn't there the idea to have illustrations to give examples of focussing situations where a mere color change is enough (row of buttons where bg color difference on focus is above 3:1) and where it is not, i.e. not alone (underlined link where difference color alone doesn't have enough contrast when also meeting 1.4.3)? These are still missing, it seems.

Or are there some real differences of opinion in the WG as to what 1.4.1 means and whether it also applies to focus states? I thought that was clear and uncontroversial? Have no time right now to delve deep into this...

patrickhlauke:
As there's been some back and forth in code comments etc, wanted to take a step back to reset the conversation here with my rationale - as it seems some folks are vehemently opposed to making use of color apply to anything relating to state...

See https://codepen.io/patrickhlauke/pen/YzbEdPL/10c3d5ebeb8bc713ebf1637ca3c74ae0 for examples, but in short my thinking is, and always has been:

* if you rely purely on color (meaning the change/difference is below 3:1 contrast, as per the note in the understanding for use of color that talks about contrast https://www.w3.org/WAI/WCAG22/Understanding/use-of-color.html#intent) to convey information - which does include "conveying the state of something" - then you fail use of color
* if this state happens to be focus indication, and you do evaluate against AAA, then yes, something may fail *both* use of color *and* focus appearance
* can't evaluate against non-text contrast, because the different colours applied to with/without that state are not visually adjacent (can't compare a change, only two pixels in proximity to each other at the same time)

patrickhlauke:
> Wasn't there the idea to have illustrations

probably...i'll be honest, there's so many places where WCAG could do with illustrations, and time is so limited...

but if we can get behind the principle (as outlined in my codepen there https://github.com/w3c/wcag/pull/1788#issuecomment-2167446058) then I'd be happy to polish up some better/cleaner illustrations here too.

patrickhlauke:
x-ref https://github.com/w3c/wcag/pull/3717 which I still need to see where the overlaps are with this older PR of mine

Myndex:
I saw something in the meeting minutes that I believe needs to be addressed, hopefully this is the place to do so.

## Regarding lightness, color, and visual function
<details><summary><b>Click for CIE definitions and links</b></summary>

.
Definitions: The CIE is the authoritative reference for color terminology. 

The most recently revised glossary provides two definitions for the word "color" by itself, and a number of compound definitions:

> ### [**color, \<psychophysical>** ](https://cie.co.at/eilvterm/17-23-001)   
> specification of a [colour stimulus](https://cie.co.at/eilvterm/17-23-002) in terms of operationally defined values, such as three [tristimulus values](https://cie.co.at/eilvterm/17-23-038)

And

> ### [**color, \<perceptual>, perceived color**](https://cie.co.at/eilvterm/17-22-040)   
> characteristic of [visual perception](https://cie.co.at/eilvterm/17-22-082) that can be described by attributes of [hue](https://cie.co.at/eilvterm/17-22-067), [brightness](https://cie.co.at/eilvterm/17-22-059) (or [lightness](https://cie.co.at/eilvterm/17-22-063)) and [colourfulness](https://cie.co.at/eilvterm/17-22-072) (or [saturation](https://cie.co.at/eilvterm/17-22-073) or [chroma](https://cie.co.at/eilvterm/17-22-074)) --- **_Note 1 to entry:_** Perceived colour depends on the [spectral distribution](https://cie.co.at/eilvterm/17-21-029) of the [colour stimulus](https://cie.co.at/eilvterm/17-23-002), on the size, shape, structure and surround of the stimulus area, on the [state of adaptation](https://cie.co.at/eilvterm/17-22-015) of the observer's visual system, and on the observer's experience of the prevailing and similar situations of observation.     

But an important definition of color is the specific achromatic color:

> ### achromatic colour, <in the perceptual sense>
> perceived [colour](https://cie.co.at/eilvterm/17-22-040) devoid of [hue](https://cie.co.at/eilvterm/17-22-067) --- _**Note 1 to entry:**_ The colour names white, grey and black or, for transmitting objects, colourless and neutral are commonly used in relation to achromatic colour.

**In other words, the _broadest_ term "color" can have "no colorfulness."**

_SIDE NOTE: I recognize that the terminology surrounding light and color is confusing and a big part of the problem towards understanding._

- - - - - 
</details>


**The _broadest_ use of the term "color" includes a color with "no colorfulness."**

Is the broadest definition of color as "lightness, hue, and colorfulness" the same "color" that is mentioned in 1.4.1? There is a reductio ad absurdum here: If 1.4.1 means not only hue and saturation, but _also_ _lightness_, then nearly all websites fail. Text is most often displayed as _only_ a difference in lightness/darkness, and even black and white line drawings would fail, or greyscale photos.

### _Therefore this cannot be the meaning of 1.4.1_

Instead, let's examine the three aspects of color:

1. [Lightness](https://cie.co.at/eilvterm/17-22-063)/darkness/[brightness](https://cie.co.at/eilvterm/17-22-059) (perception of luminance)
2. Uniqueness (perception of [hue](https://cie.co.at/eilvterm/17-22-067))
3. [Colorfullness](https://cie.co.at/eilvterm/17-22-072) (perception of saturation/chroma)

And what is important to us, is how the human vision system processes each. After the three LMS cone types absorb light, a matrix of cells converts the three cone signals to a red/green opponent (a), a yellow/blue opponent (b), and a **_separate_** lightness/darkness opponent (L) which holds fine details, and is critical for text.

Cones and Opponent_colors, starting with representations of the retinal cones on the left, and those signals being processed into the opponents mentioned in the text



Lightness (our _perception_ of luminance) is processed separately from hue and chroma. Variations in lightness are fundamental to vision. Individuals with color vision deficiency are frequently _more capable_ at detecting lightness contrasts than standard vision.

In terms of accessibility, it is hue that is the biggest concern, followed by chroma (the amount that a hue is different than neutral grey). Reading 1.4.1, it's hard to see how to apply it to anything other than hue/chroma. 

<details><summary><b>Click for More on Why 1.4.1 is About Hue/Chroma and not Lightness</b></summary>

.

Standard trichromatic human vision, with LMS cone types perceives four unique hues of red, yellow, green, blue, with secondaries purple, magenta, orange, cyan. 

Typical dichromatic vision (LS or MS) effectively perceives blue and yellow, some studies indicate a heightened sensation of saturation, but otherwise a significant limitation distinguishing the standard four unique hues. Dichromats tend to use the _luminance differences of colors_ to aid in differentiation.

In the example below, we see that the natural lightness differences of red and green mean the difference is distinguishable to all in this case. But notice that the lack of lightness difference between green and yellow (for the protan) makes green and yellow indistinguishable.

RedGreenYellowProcessedCVD for example's upper left is standard version followed by various versions of CVD

And Lightness/darkness is distinctly different from hue/chroma as far as our visual function is concerned. "Saturation" is chroma divided by lightness, so yes you can say that lightness is an aspect of color per se. But as far as visual function, lightness and hue are very separate.


- - - - - 
</details>

----
Hi @patrickhlauke 

> * _if you rely purely on color (meaning the change/difference is below 3:1 contrast, as per the note in the understanding for use of color that talks about contrast https://www.w3.org/WAI/WCAG22/Understanding/use-of-color.html#intent) to convey information - which does include "conveying the state of something" - then you fail use of color_

I've been given the notion that what is written in the understanding document is not "normative" — or is it? The understanding document seems to change a lot after the SCs become recommendations.

That said, there is no science to back up the assertion that 3:1 is needed to define a state change. At best, the 3:1 defined in WCAG 2.x is a reference to a 1980s era standard for text on monochrome displays, but otherwise lacking in evidence—[Dr. Arditi's 2017 paper ](https://jov.arvojournals.org/article.aspx?articleid=2628138) discussed the problems with using contrast math as a guideline.

More to the point, sticking the 'ol 3:1 into SCs because it happens to by lying around, does not make it useful nor correct.

### What's missing in 1.4.1

What is actually important, and what should be focused on in 1.4.1, is the ability to _discriminate_ between two items of information.

For items that are adjacent or near adjacent, a lightness difference creates a luminance contrast sensation. The actual amount of lightness contrast needed is dependent on the spatial characteristics of the element, which is generally not accounted for in WCAG 2.

But for items that are spaced apart, or completely independent, a lightness change without an adjacent reference, must be **_substantial_** to be easily recognized. This is true for all vision types, because by itself, a grey is neutral and not unique.

If 1.4.1 is to be helped, then some discussion of adjacent/near adjacent vs distant or isolated items would be a good step in the right direction. 

### Focus Indication

The nature of the indicator: spatial, luminance, difference from the surrounding, and near by comparisons to similar but un-focused elements, all together dictate what is needed here. As a multi-axis set of conditions, a specific _normative_ statement is elusive.

- a 1px outline close to a button needs to be nearly maximum luminance contrast to be reliably readable.
- a 4px outline, especially with a 2 px space from a button, works with less than 3:1 contrast.
- But a 4px outline around every button, where the only change is from green to yellow, can be a non starter, and needs evaluation for CVD issues. 

<details><summary><b>Click for additional examples and visual simulations of CVD</b></summary>
.

- For another example, red `#ff2200` and dark green `#004400` meets the 3:1 criteria, but for someone with protanopia, the difference in invisible even with adjacency as in the below example:

red or green focus ... upper left is standard version followed by various versions of CVD

----
- In the following, the same colors but increasing the focus border width 50% to 6px:

red or green focus ... upper left is standard version followed by various versions of CVD - this version with a thicker border for focus

The 50% thicker border is helpful for the white buttons, as the border itself registers are 50% thicker. But the dark buttons may still be challenging, because the button is close to the luminance of the border, and so the 2px increase in border thickness is a much smaller percentage of change to the total dark area of the button.

- - - - - 
</details>

----

## TL;DR

I came here initially just to say that lightness or luminance contrast does not belong in 1.4.1, which should be referencing hue and chroma. Lightness plays a part here only insofar as it affects saturation or the perceived colorfulness.

Otherwise, lightness (luminance) is processed in our brain separately from hue and chroma, and for accessibility reasons should also be handled separately. The place to handle lightness differences or contrast is in a guideline with a perceptually uniform contrast metric.


### References
- The [CVD simulator](https://www.myndex.com/CVD) as used above.
- [Overview of perceptual contrast and color](https://www.smashingmagazine.com/2022/09/realities-myths-contrast-color/), with further references.
- [Linktree of related resources](https://linktr.ee/Myndex)

patrickhlauke:
@myndex while I'll be the first to acknowledge that the proposed changes here don't go far enough, and should really be expanded much more to distinguish the different aspects of vision ... that would go far beyond what we can do in a non-normative document. this PR tries to patch/cover how people have been interpreting "use of color" right now. while the `3:1` contrast ratio is a very blunt and simplified measure, it's what we have. it's the "for this exercise, we assume the cow is spherical" oversimplification.

> For items that are adjacent or near adjacent, a lightness difference creates a luminance contrast sensation. The actual amount of lightness contrast needed is dependent on the spatial characteristics of the element, which is generally not accounted for in WCAG 2.
>
> But for items that are spaced apart, or completely independent, a lightness change without an adjacent reference, must be substantial to be easily recognized. This is true for all vision types, because by itself, a grey is neutral and not unique.

As the SC aims to provide a simple pass/fail that authors can evaluate, this sort of *correct* thinking is too complex. Particularly as relative position of items can change dramatically depending on viewport size, reflow, etc. Nobody's going to be able to evaluate things if they depend on a complex analysis. So yes, as with many other things, this is an oversimplification, for the purpose of allowing auditors to give a quick and hopefully consistent pass/fail.

> The place to handle lightness differences or contrast is in a guideline with a perceptually uniform contrast metric.

as we're not adding any new guidelines/criteria, and have to make do with what we currently have, that's not going to happen. I recommend pushing for this in WCAG3 discussions.

patrickhlauke:
in short...yes, WCAG 2.1 introduced (in the understanding doc, rather than hidden away as it was in a technique) the idea that two colours (defined as both hue and chroma, together) need to have a contrast difference of at least `3:1` to be considered "different" enough. this was always was a brutal simplification, an "assume the cow is spherical" approach.

but now, *given* that simplification, this PR applies the thinking to common cases where we see things being differentiated visually by different colours (different hue and/or different chroma). and again, the aim here is not perfection in modelling vision (noting also that the elephant in the room here is that authors have no control over a user's settings, badly calibrated monitors, challenging ambient lighting around their monitor/screen, etc), but to give a very simplified line in the sand to be able to make a binary pass/fail judgement.

Myndex:
Hi @patrickhlauke 

And thank you for the comments, and as you know I am aware of these nuances.

The principal reason for my comments relates to statements I read in the minutes to the related meeting, which presented a common misunderstanding of the importance of separating hue/chroma from lightness. It is best practice to recognize their difference as visual functions that are processed separately and independently therein. **And also in pertinent aspect, that 1.4.1 is best described as a matter of hue/chroma and** the disregarding of luminance or lightness.

This is well supported by nearly a century of vision and color science. And in fact, the claim in 1.4.1 understanding, though unsupported, that one can assert a 3:1 contrast difference (which applies to luminance only) does by definition indicate the separation of hue/chroma vs lightness.

But again, the reality is not well described by such a simple claim, outside of some given context.

> _"...As the SC aims to provide a simple pass/fail that authors can evaluate, this sort of correct thinking is too complex. Particularly as relative position of items can change dramatically depending on viewport size, reflow, etc..."_

If _"thinking correctly is too complex"_ for an SC, then arguably, it should not be an SC. Instead it may be useful in a best practices document. 

This is one of several fundamental failings inside WCAG. And if WCAG were nothing but a voluntary guideline, such assertions could be disregarded. **Unfortunately, some legislatures are elevating WCAG 2.x into laws or regulations, with legal effect and purpose**. 

This is not going to end well, especially where AGWG and the resultant WCAG guidelines continue to develop claims or make assertions that are unsupported by the vast corpus of peer-reviewed science. 

Ultimately, this is a suggestion that some consideration be made, as to where the continued severance from actual science and factual evidence-based knowledge might be softened a bit. This is to the benefit of the document. Instead of continuing unsupportable claims with terminology such as "shall" or "must", perhaps using non-normative terms, such as "suggested" or "in some cases", might help salvage the various untenable problems inherent throughout WCAG. Problems which I hasten to add, will not stand per the legal contexts that are being promoted. 

WCAG is a fertile container wherein pandas might find a plum-tree of possibilities.


Thank you for reading.

patrickhlauke:
> If "thinking correctly is too complex" for an SC, then arguably, it should not be an SC. Instead it may be useful in a best practices document.
> 
> This is one of several fundamental failings inside WCAG. And if WCAG were nothing but a voluntary guideline, such assertions could be disregarded. Unfortunately, some legislatures are elevating WCAG 2.x into laws or regulations, with legal effect and purpose.

and again, for 2.x, you've missed the boat on these things by ... a good two decades. so yes, by all means, take these concerns to WCAG 3. for 2.x, we're fairly bound by what has been set up already, and just filling in some of the gaps and bits around the edges.

Myndex:
Hi @patrickhlauke 

> _..for 2.x, you've missed the boat on these things by ... a good two decades._

Yes, the compelling reason it is now a matter for discussion is the recent mad-dash to convert this, a voluntary guideline, into statute law or regulations with legal consequence.

That was not intended, as was so stated back in 2007.



> _…so yes, by all means, take these concerns to WCAG 3…_

As far as anyone knows, WCAG_3 is as much as ten years away. As such, it is not relevant to the present problem.



> _…for 2.x, we're fairly bound by what has been set up already…_

Yes we have discussed this at length. If WCAG 2.x were not being promoted as a stable standard suitable for, and defensible in, a court of law, then there would be no concern or controversy.

But the fact remains that it has been so promoted.

I know you feel I am constantly missing your point Patrick, just as I recognize that my posts are rarely taken as intended.

But the point I’ve wanted to impress herein, is how liability attaches to _WCAG the document_ when it is specifically referenced as the basis for statute law.

My previous posts in this thread were to indicate ways to limit that.

patrickhlauke:
> Yes, the compelling reason it is now a matter for discussion is the recent mad-dash to convert this, a voluntary guideline, into statute law or regulations with legal consequence.

And again, this has already been the case for aeons in many parts of the world, not something new...

alastc:
From the meeting: Pat to review and check that the pass/fail aspect is clear.

understanding/20/use-of-color.html
```diff
@@ -38,13 +38,28 @@ <h2>Intent of Use of Color</h2>
...
+        <li>
+            <strong>Information conveyed by color differences:</strong> 
+            <ul>
+                <li>required fields are only identified by changing their existing border to a red color;</li>
+                <li>the type of transaction is only visually distinguished by using different foreground colors: red for debits and green for credits;</li>
+                <li>subway lines are visually distinguished solely by employing various filled colors;</li>
+            </ul>
+        </li>
+        <li>using color alone to indicate
+         that a link will open in a new window or that a database entry has been updated successfully.</li>
```
detlevhfischer: > using color alone to indicate that a link will open in a new window

Has anyone ever seen this implemented? The example seems quite artificial.

patrickhlauke: happy to take it out. i did come across something like this on a site ages ago, but yeah not very common

patrickhlauke: removed

understanding/20/use-of-color.html
```diff
@@ -91,25 +102,51 @@ <h2>Intent of Use of Color</h2>
...
+              Authors also cannot set the visited state of links. The anchor element does not include a "visited" attribute; therefore the author has no ability to alter the state through an attribute setting. As such, authors cannot achieve <a href="info-and-relationships">1.3.1 Info and Relationships</a> or
+              <a href="name-role-value">4.1.2 Name, Role, Value</a> in regard to visited links.
+           </li>
+        </ul>
+     </div>
+        
+     <div class="note">
+         <p>Note that particularly in the case of focus indication, this criterion is likely to overlap with
+            other, more specific criteria, such as
+            <a href="focus-appearance">2.4.11 Focus Appearance</a> (new in WCAG 2.2).
```
patrickhlauke:
```suggestion
            <a href="focus-appearance">2.4.11 Focus Appearance</a> (new in WCAG 2.2), which will have further requirements.
```

understanding/20/use-of-color.html
patrickhlauke: expanded the note to make it clearer why it's here

PR #1803, opened 2021-05-11
Update/correct 3.3.1 understanding

### LINKED ISSUE #1797: Change 3.3.1: Error Suggestion to 3.3.3: Error Suggestion, 3.3.1: Error Identification
https://www.w3.org/WAI/WCAG22/Understanding/error-identification

Last sentence before the Benefit says:

> See also 3.3.1: Error Suggestion.

This of course should be:

> See also 3.3.3: Error Suggestion.

### LINKED ISSUE #1802: Referral from ARIA2 to 3.3.3 should be 3.3.1
https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA2

> Applicability
> Technologies that support Accessible Rich Internet Applications (WAI-ARIA).
> 
> This technique relates to:
> 
> Success Criterion 1.3.1: Info and Relationships (Advisory)
> Success Criterion 3.3.3: Error Suggestion (Sufficient)

"Success Criterion 3.3.3: Error Suggestion" (Sufficient) should be "Success Criterion 3.3.1: Error Identification"

patrickhlauke:
* remove the unnecessary reference to the "definition in WCAG 2.0"
* remove the now outdated "Currently, few technologies support this kind of programmatic information", as that's not true anymore
* clarify that programmatic information is not covered by this SC, but that others like 4.1.2 apply
* corrects link to error suggestion from "3.3.1" to "3.3.3"
* add ARIA2 as sufficient technique

Closes https://github.com/w3c/wcag/issues/1797, closes https://github.com/w3c/wcag/issues/1802


patrickhlauke:
Any news on this @alastc ?

bruce-usab (approved):
Looks fine, but I am still not straight on cross ref syntax (line 76), and looking to be educated on (1/19) call.
> `...by other criteria such as <a href="#name-role-value">4.1.2 Name, Role, Value</a>.`

Presumably that links back to WCAG.  But should it link to Understanding doc for 4.1.2 ?

understanding/20/error-identification.html
```diff
@@ -72,21 +72,18 @@ <h2>Intent of Error Identification</h2>
          that user agents or assistive technologies can use to identify an error and provide
          error information to the user. For example, certain technologies can specify that
          the user's input must not fall outside a specific range, or that a form field is required.
-         Currently, few technologies support this kind of programmatic information, but the
-         Success Criterion does not require, nor prevent it.
-         
+         This type of programmatic information is not required for this success criterion, but is covered
+         by other criteria such as <a href="#name-role-value">4.1.2 Name, Role, Value</a>.
```
WilcoFiers, line 76: I think this is a little too strong. Some stuff could be argued to fall under 4.1.2, but I'm not so sure that 4.1.2 required that for example an input range has to be expressed through a component's programmatic properties.

```suggestion
         This type of programmatic information is not required for this success criterion, but may be covered
         by other criteria such as <a href="#name-role-value">4.1.2 Name, Role, Value</a>.
```

WilcoFiers:
👎 as is, 👍 with my suggested edit (or something along those lines anyway)

understanding/20/error-identification.html
patrickhlauke: makes sense, yes

mraccess77:
SC 1.3.1 can be solved with programmatic information or with text - so while the SC may not require programmatic - I think we need to be careful in how we describe programmatic information such as required fields to SC !.3.1 if that information is available in presentation but not programmatically or as text.  I think the key is if it is presented or not.  If it's not presented visually then I would agree it would go to 4.1.2.

patrickhlauke:
@mraccess77 I can see your concern here, but the added reference there that says this falls under 4.1.2 was in the context of the whole paragraph starting with "The identification and description of an error can be combined with programmatic information". i.e. that being *in addition to* other identification and description. Wonder if I should make that clearer somehow...any thoughts?

understanding/20/error-identification.html
```diff
@@ -191,6 +188,12 @@ <h4>Situation A: If a form contains fields for which information from the user i
                   <a href="https://www.w3.org/WAI/WCAG21/Techniques/general/G83" class="general">Providing a text description that identifies the field as mandatory</a>
                   									         
                </li>
+
+               <li>
+                  										           
+                  <a href="https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA2" class="aria">ARIA2: Identifying required fields with the "required" property</a>
+                  									         
+               </li>
```
detlevhfischer, line 196: * In subsection "Sufficient Techniques for Error Identification" there are several empty list items / links that should probably be removed.
* There are two empty sections, "Resources for Error Identification" and, under "Techniques for Error Identification", "Failures for Error Identification". I guess these should be removed?

understanding/20/error-identification.html
```diff
@@ -72,21 +72,18 @@ <h2>Intent of Error Identification</h2>
          that user agents or assistive technologies can use to identify an error and provide
          error information to the user. For example, certain technologies can specify that
          the user's input must not fall outside a specific range, or that a form field is required.
-         Currently, few technologies support this kind of programmatic information, but the
-         Success Criterion does not require, nor prevent it.
-         
+         This type of programmatic information is not required for this success criterion, but may be covered
```
detlevhfischer, line 75: While we are at it, would it be worth covering the frequent case of inline error handling, e.g. mentioning the recommended position of error messages (if placed above the field these would easiliy be missed in SR reading mode)? Information could also indicate that these messages should be programmatically available (e.g. by inserting them in a programmatically linked `label` or linking them with `aria-describedby`). If I get the gist of the reference to 4.1.2 you think that failure to programmatically link error messages should be reported elsewhere, like the existence of the required attribute when required is visually indicated?

understanding/20/error-identification.html
```diff
@@ -72,21 +72,18 @@ <h2>Intent of Error Identification</h2>
...
+         by other criteria such as <a href="#name-role-value">4.1.2 Name, Role, Value</a>.
```
detlevhfischer, line 76: The idea of 3.3.1 is: **ALL** users should know that some error has occured. We have always thought it appropriate to cover missing programmatic linking of error messages, and programmatic identification of required state where indiated visually, under 3.3.1 since it is an aspect of the use case of realising that there has been an error. Is the added reference to 4.1.2 meant to indicate that that kind of error should be reported in 4.1.2 and not here? If so, that is not explicit now. I guess programmatic identification of errors might possibly be done in several ways (programmatically moving focus to the message?) that would not **necessarily** fall under 4.1.2. Another way would be 4.1.3, which is not referenced here - I guess it should be, by analogy.

I can see the advantage of covering aspects of bad errror handling under other SCs: 4.1.2 (messages not programmatically referenced) 1.3.1 ('required' indicated only visually), 1.3.2 (error message appears above the field), 4.1.3 (error status message not available) - BUT from the point of use of an audit report, this means the issues around error handling are widely dispersed across the report. I'm OK with that (because what it changes is nothing substantial, but only the allocation of errors to SCs), but I wonder if at least the programmatic availability of error messages is something that can usefully be reported under 3.3.1 (because it does not concern so much **the value / state of a UIC** but the availability / identification of the result of a user interaction).

detlevhfischer (dismissed):

detlevhfischer:
[Rendered view (19. Feb 2024, 11:10)](https://raw.githack.com/w3c/wcag/2495b864efaa74fbd94d3ecfd65a6ec810dbfa0d/understanding/20/error-identification.html)

understanding/20/error-identification.html
mbgower: > While we are at it, would it be worth covering the frequent case of inline error handling, e.g. mentioning the recommended position of error messages (if placed above the field these would easiliy be missed in SR reading mode)?

If you want to open a new issue to include that, it's welcome, but I think it's important that we maintain a process of trying not to increase scope in PRs that have been put in front of the working group.

understanding/20/error-identification.html
detlevhfischer: > Finally it is fairly clearly established that we do not create a catalog of ways something could fail other SCs in the Understanding documents. I think the balance created here is reasonable.

Yes, I see the point in that. But my point was that there are reasons to fail unreferenced error messages **here**, in 3.3.1, as this cannot easily be captured under 4.1.2 (it is not narrowly a name, role or value of an UIC that is at fault) and arguably 1.3.1 is met if output is available in browse mode. And it is an accessibility issue that should be captured somewhere - why not here? 

I am not in the way of progressing this PR without such addition, though, and to create a separate PR for it.

understanding/20/error-identification.html
patrickhlauke: those were parts of the file i hadn't touched, and looking at the live version, it seems the xslt/publish script silently removes there. happy to rip them out of the file itself ...

understanding/20/error-identification.html
mbgower: Thank you, please open a new issue/PR to address your concerns.

understanding/20/error-identification.html
dbjorge: Agree with patrick, I think this is actually just a case where the preview rendering is inaccurate. This happens because some of the links here are written in the source like this:

```html
               <li>

                  <a href="https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA21" class="aria"></a>
                  
               </li>
```

...which looks like empty list items as rendered by the preview, but when processed to build the live version, actually results in the `<a>` elements getting filled in using the title of the page (you can see an "ARIA21" link like the above example correctly in the rendered live page)

detlevhfischer:
@patrickhlauke @alastc Is this going anywhere, any time soon?

patrickhlauke:
hmmm, it seems that this was merged into main, but a subsequently merged PR https://github.com/w3c/wcag/pull/1651 overwrote at least part of it (the change from *this* PR that removes the inaccurate "Currently few technologies support this...") https://github.com/w3c/wcag/commit/3cf31a01cb119998e92479e6d333687d6a5438f9#diff-7ab9dda99450b42951e62a63fb7f6fff1783fcb67381d09992ff71ad5275d1d4L84

I think this was the result of two slightly competing/stale PRs that sat there far too long, and then got some weird merge conflicts that were awkwardly resolved. I would actually be in favour of re-applying the change to that sentence, so it goes back to what this PR wanted to do:

```
This type of programmatic information is not required for this success criterion, but may be covered
by other criteria such as <a href="#name-role-value">4.1.2 Name, Role, Value</a>.
```

As for your additional comments/suggestions you made @detlevhfischer (looking further up here), @mbgower suggested that you file follow-up PRs for those

patrickhlauke:
Filed https://github.com/w3c/wcag/pull/4355 to re-change the incorrect/outdated statement about "few technologies" that support extra programmatic information.

Issue #1533, opened 2020-11-18
Mechanism provided by the platform AND stopping 1.4.2: Audio Control


jake-abma:
> 1.4.2 Audio Control (Level A): If any audio on a Web page plays automatically for more than 3 seconds, either a mechanism is available to pause or stop the audio...

Intent is clear, also what is needed, BUT:

According to the wording/text and the definition of "mechanism" can I conclude that:

> mechanism
> process or technique for achieving a result
> 
> NOTE
> The mechanism may be explicitly provided in the content, or may be relied upon to be provided by either the platform or by user agents, including assistive technologies.

A mechanism provided by the platform is a 'turn off audio' functionality (all devices have one) and so IF sound plays on a web application, you have the opportunity to turn it off, basically escaping this SC all along? (I can stop the sound)

ps. should this 'platform provision / ability' be ruled out and made clear in the Understanding doc?

patrickhlauke:
> A mechanism provided by the platform is a 'turn off audio' functionality (all devices have one) and so IF sound plays on a web application, you have the opportunity to turn it off, basically escaping this SC all along? (I can stop the sound)

maybe splitting hairs, but "pause or stop *the* audio" (on the particular page) is different from "pause and stop *all* audio" (for everything on the device/system). The intention is clearly the first one, reinforced by the "independently from the overall system volume level" part of the normative wording of the SC later on when talking of volume levels.

now, many browsers on desktop do offer things like audio controls for individual tabs, and some operating systems do give you access (though not very easily) to individual volume controls on a per application basis, but i don't think those are quite pervasive enough yet (and certainly not on mobile/tablet OSs)

mraccess77:
If a browser supports a mute tab feature then use of that browser likely would meet the criteria although we still tend to flag these issues as most situation users have varied browsers on mobile and PC platforms.

jake-abma:
@patrickhlauke 

> maybe splitting hairs, but "pause or stop the audio" (on the particular page) is different from "pause and stop all audio" (for everything on the device/system). The intention is clearly the first one, reinforced by the "independently from the overall system volume level" part 

But this is exactly where the question comes from, the intent was clear for me all along, not for people I have in training.
'stop the audio on the particular page' is just not what it says, and the 'reinforced' is not really 'reinforcing' as there is an 'OR'.

Your comment lack the word 'OR' which plays a big part here, separating that part from the first.

patrickhlauke:
[...]

PR #1824, opened 2021-05-21
Rewording of normative text for 1.4.2 Audio Control

patrickhlauke:
Taking @bruce-usab's suggestion from https://github.com/w3c/wcag/issues/1533#issuecomment-733025076

Closes https://github.com/w3c/wcag/issues/1533

### DIFF
--- a/guidelines/sc/20/audio-control.html
+++ b/guidelines/sc/20/audio-control.html
@@ -4,9 +4,11 @@
    
    <p class="conformance-level">A</p>
    
-   <p>If any audio on a Web page plays automatically for more than 3 seconds, either a <a>mechanism</a> is available to pause or stop the audio, or a mechanism is available to control audio
-      volume independently from the overall system volume level.
-   </p>
+   <p>For any audio on a Web page that (1) plays automatically, and (2) last for more than 3 seconds, at least one of the following is true:</p>
+   <dl>
+      <dt>Pause or Stop</dt><dd>The Web page provides a mechanism to pause or stop the audio.</dd>
+      <dt>Volume Control</dt><dd>The Web page provides a mechanism to control audio volume independently from the overall system volume level.</dd>
+   </dl>
    
    <p class="note">Since any content that does not meet this success criterion can interfere with a user's
       ability to use the whole page, all content on the Web page (whether or not it is used


detlevhfischer:
I would read this version as clearly implying that the **author** is responsible for implementing the control to stop / change level of / audio - which might be a good thing, given that availability and place of UA/OS level controls will vary and many users will struggle reaching them.

patrickhlauke:
but the normative wording already currently talks about "mechanisms" being available, which can come from UA/OS. whether those mechanism are then pervasive enough is the question that people evaluating content need to assess (as they do with other "mechanism is available" SCs)

patrickhlauke:
https://www.w3.org/TR/WCAG/#dfn-mechanism

> The mechanism may be explicitly provided in the content, or may be relied upon to be provided by either the platform or by user agents, including assistive technologies.

specifying that the mechanism must be done by the Web page itself as in this original suggestion seems to go against this idea. which is why #1825 is my preference, since it keeps the original wording/spirit

detlevhfischer:
>  #1825 is my preference, since it keeps the original wording/spirit

It is certainly more consistent and in line with the normative text. Still, for something as disruptive unstoppable audio blaring out of a page, I would prefer a requirement that it can be stopped by content controls (admittedly this is more then the normative ask - but for the sake of consistency I think #1825 is preferable). Lacking an author-provided control, any PASS/FAIL decision then has to take account the accessibility baseline of any conformance claim - I would imagine on several UA / platforms there won't be a control to mute audio independent from system audio, but that would need some digging...

patrickhlauke:
>  I would imagine on several UA / platforms there won't be a control to mute audio independent from system audio, but that would need some digging...

which is fine, so in practice nothing changes here. that's what's the SC currently says. and we'd still fail these cases because we know that the UA/OS based solutions aren't pervasive enough yet to count. But then, in audits that are very specifically scoped to a platform (e.g. a site that only targets a particular UA/OS, which does provide this option easily), it can be passed. that nuance/decision making ("does this mechanism count as being an acceptable mechanism to pass this SC") has already been there from the start.

mraccess77:
I agree that we should make it clear that in some very specific situations where the scope of the audit can be limited to certain browser combinations that you can rely on the browser feature to mute audio otherwise today this is generally solved by the author providing a mechanism in the page to pause/stop audio.

Of note Trusted Tester indicates that auto play should be disabled in the browser for this test: https://section508coordinators.github.io/TrustedTester/auto.html

Note:
A browser’s ability to disable auto-playing media and/or mute a specific tab are acceptable mechanisms to stop or control the volume of auto-playing audio content. However, not all browsers have the capability to disable auto-playing media or mute specific windows or tabs.

patrickhlauke:
Could this be taken for discussion/vote @alastc ?

patrickhlauke:
To simplify matters, I'm going to close this and say it's superseded by https://github.com/w3c/wcag/pull/1825

PR #1825, opened 2021-05-21
Alternative rewording of the normative text for 1.4.2 Audio control

patrickhlauke:
In contrast to https://github.com/w3c/wcag/pull/1824 this rewording accepts that the mechanism does not necessarily have to be implemented by the web page itself, but - per the definition of mechanism - can also be provided by the user agent/system. What this change does implement though is the clarification for the pause/stop clause that the pausing/stopping must be *independent* from any other audio output (which does address @jake-abma's original concern that a system wide general mute/audio off would automatically satisfy the SC, which is not the intended outcome https://github.com/w3c/wcag/issues/1533

while a normative change, I don't believe this actually changes the meaning as intended at the time.

Closes https://github.com/w3c/wcag/issues/1533

### DIFF
--- a/guidelines/sc/20/audio-control.html
+++ b/guidelines/sc/20/audio-control.html
@@ -5,10 +5,12 @@
    <p class="conformance-level">A</p>
    
    <p>If any audio on a Web page plays automatically for more than 3 seconds, at least one of the following is true:</p>
-   <ul>
-     <li>a <a>mechanism</a> is available to pause or stop the audio</li>
-     <li>a <a>mechanism</a> is available to control the audio volume, independently from any other audio output of the user agent and the overall system.</li>
-   </ul>
+   <dl>
+     <dt>Pause or Stop</dt>
+     <dd>A <a>mechanism</a> is available to pause or stop the audio.</dd>
+     <dt>Volume Control</dt>
+     <dd>A <a>mechanism</a> is available to control the audio volume, independently from any other audio output of the user agent and the overall system.</dd>
+   </dl>
    
    <p class="note">Since any content that does not meet this success criterion can interfere with a user's
       ability to use the whole page, all content on the Web page (whether or not it is used
--- a/guidelines/sc/20/audio-control.html
+++ b/guidelines/sc/20/audio-control.html
@@ -4,8 +4,11 @@
    
    <p class="conformance-level">A</p>
    
-   <p>If any audio on a Web page plays automatically for more than 3 seconds, either a <a>mechanism</a> is available to pause or stop the audio, or a
-      mechanism is available to control the audio volume, independently from any other audio output of the user agent and the overall system.</p>
+   <p>If any audio on a Web page plays automatically for more than 3 seconds, at least one of the following is true:</p>
+   <ul>
+     <li>a <a>mechanism</a> is available to pause or stop the audio</li>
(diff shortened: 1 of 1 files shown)


patrickhlauke:
Changed this to be less radical, but still cover the clarification for @jake-abma's concern in #1533, while also incorporating the note along the lines of trusted tester as pointed out by @mraccess77 https://github.com/w3c/wcag/pull/1824#issuecomment-847070698

patrickhlauke:
Any news on this @alastc ?

bruce-usab (changes requested):
I am glad to see this PR.  As I noted in the related issue thread, please add short handles to the choices.  This is consistent with other SC using the "one of the following" pattern.  2.1.4, 2.2.1, 2.5.2, 3.3.4, 3.3.6.

patrickhlauke:
> please add short handles to the choices

ah sorry, meant to do it and then promptly forgot when I made the update to the PR. done now.

awkawk:
@patrickhlauke If this hasn't been approved as an errata it should have "Erratum Raised" as the label and only gets the "Errata" label once finalized.

patrickhlauke:
@awkawk i'll leave the tagging up to you and your process ;)

awkawk:
I think this is a more substantial normative change and will need more time to evaluate implications.

bruce-usab (approved):
I think the first bullet should end <q>; or</q> rather than a period.

Unfortunately, 2.0/2.1 is not 100% consistent with its <q>at least one of the following is true</q> pattern.

bruce-usab (changes requested):
The <q>any other audio output of the user agent and</q> is too much of a change for my tastes.

patrickhlauke:
> The addition of any other audio output of the user agent and is too much of a change for my tastes.

without that, i could claim that going into windows' volume mixer and just muting/changing the volume for Chrome satisfies this, as it's independent of the OS overall

bruce-usab:
> without that, i could claim that going into windows' volume mixer...

That is beyond the UA, as is asserting PC mute is an option for audio pause/stop.  (Something which, until today, I never would have thought anyone brazen enough to claim.)

The audio (which plays automatically) is on the web page, it obviously follows that the mechanism for controlling the audio must be on the web page.

But we have a few SC which authors can claim to meet by relying upon browser features.  So it is not as tidy as I would like.

Asserting use of platform/hardware to meet web page SC is not a loophole just for this SC.  I do not agree that the idea warrants any air at all.  But if it does, maybe the Conformance Requires need a tweeking?

patrickhlauke:
> I do not agree that the idea warrants any air at all

well, you do you. seem like this PR https://github.com/w3c/wcag/pull/2621 is superseding this one anyway

alastc:
I'm afraid this wasn't seen as a sufficient problem to overcome the friction of an SC update: https://www.w3.org/2022/08/16-ag-minutes.html#t10

I'll close this, but there is an understanding update on the cards.

GreggVan:
+1 to this

And we should do this uniformly as an editorial fix for both clarity and uniformity.  
Suggest this be added to survey 

Proposal 
That throughout WCAG 2.2 (as an editorial fix):
All lists that are "any of the following" have each choice end with ;  and the last choice end with    ; or    (or   ; OR)
All lists that are "all of the following"  have each choice end with ; and the last choice end with    ; and    (or   ; AND)

(The caps MIGHT be justified since it is a logical AND  or  OR …. )



Gregg Vanderheiden
***@***.***

Issue #1543, opened 2020-11-24
Failure F23 is NOT a failure if you look at the Procedure


jake-abma:
> Failure of 1.4.2 due to playing a sound longer than 3 seconds where there is no mechanism to turn it off

Agreed this is a failure, but the Procedure states:

> Check that there is a way in a Web page to turn off any sound that plays automatically for more than three seconds.

https://www.w3.org/WAI/WCAG22/Techniques/failures/F23

It does not have to be "in a Web Page" and should be something like "for the Web Page"

see also "if a mechanism is available" which may be options 'out of the Web Page' itself (OS, UA, AT)

> mechanism
> process or technique for achieving a result
> 
> NOTE
> The mechanism may be explicitly provided in the content, or may be relied upon to be provided by either the platform or by user agents, including assistive technologies.

awkawk:
@jake-abma would you be satisfied with:

Check that there is a way @@ to turn off any sound that plays automatically @@in a Web page@@ for more than three seconds.

jake-abma:
the problem is we get stuck at the same problem as mentioned in https://github.com/w3c/wcag/issues/1533

is a 'device setting' to turn of ALL sound not also @@ to turn off any sound that plays automatically @@in a Web page@@ ?

We need to talk specifically about the sound that plays due to something added to the Web Page, not sound in general

bruce-usab:
> is a 'device setting' to turn of ALL sound not also ... 

It is explicitly not sufficient for satisfying the SC text.

Can this issue be resolved simply by updating the F23 procedure?

patrickhlauke:
i'd say the normative SC wording should be patched first, otherwise the technique may appear to contradict the normative text, which is never a good thing... https://github.com/w3c/wcag/issues/1533

bruce-usab:
I agree that the techniques appears to have some conflict with the normative text.

But, I think it is the technique which is misleading and that the normative text is sub-optimal but okay as-is.  Plus there seems to be a lot of resistance to even the most modest edits to the normative text.

jake-abma:
> That is explicitly not sufficient for satisfying the SC text.
> 
> Can this issue be resolved simply by updating the F23 procedure?

Yes I think so, change...

> Procedure
> Check that there is a way in a Web page to turn off any sound that plays automatically for more than three seconds.

...to something like...

> Procedure
> Check that there is a way for the Web page only to turn off sound that plays automatically for more than three seconds.

jake-abma:
[...]

PR #1826, opened 2021-05-21
Reword F23

patrickhlauke:
in line with the proposed clarification for 1.4.2 https://github.com/w3c/wcag/pull/1825 this clarifies that a *mechanism* needs to be available. the mechanism can actually be provided by the UA/system, not necessarily by the Web page. Key here is that it must be possible to turn off the sound *independently* from any other audio output.

Closes https://github.com/w3c/wcag/issues/1543

~Only makes sense to merge if https://github.com/w3c/wcag/pull/1825 (or an equivalent/modified version of the PR) is also merged~ (Completed)

### DIFF
--- a/techniques/failures/F23.html
+++ b/techniques/failures/F23.html
@@ -1,40 +1,64 @@
-<!DOCTYPE html><html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml"><head><title>Failure of  1.4.2 due to playing a sound longer than 3 seconds where
-                    there is no mechanism to turn it off</title><link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove"></link></head><body><h1>Failure of  1.4.2 due to playing a sound longer than 3 seconds where
-                    there is no mechanism to turn it off</h1><section class="meta"><p class="id">ID: F23</p><p class="technology">Technology: failures</p><p class="type">Type: Failure</p></section><section id="applicability"><h2>When to Use</h2>
+<!DOCTYPE html>
+<html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml">
+<head>
+   <title>Failure of  1.4.2 due to playing a sound longer than 3 seconds where there is no mechanism to turn it off</title>
+   <link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove"></link>
+</head>
+<body>
+   <h1>Failure of 1.4.2 due to playing a sound longer than 3 seconds where there is no mechanism to turn it off</h1>
+   <section class="meta">
+      <p class="id">ID: F23</p><p class="technology">Technology: failures</p><p class="type">Type: Failure</p>
+   </section>
+   <section id="applicability">
+      <h2>When to Use</h2>
       <p> Applies to all technologies except those for voice interaction. </p>
-   </section><section id="description"><h2>Description</h2>
-      <p> This describes a failure condition for Success Criteria involving sound. If sound does not
-                        turn off automatically within 3 seconds and there is no way to turn the
-                        sound off then Success Criterion 1.4.2 would not be met. Sounds that play over 3 seconds
-                        when there is no mechanism to turn off the sound included in the content
-                        would fall within this failure condition. </p>
-   </section><section id="examples"><h2>Examples</h2>
+   </section>
+   <section id="description">
+      <h2>Description</h2>
+      <p>This describes a failure condition for Success Criteria involving sound. If sound does not
+         turn off automatically within 3 seconds and there is no way to turn the
+         sound off, independently from any other audio output, then Success Criterion 1.4.2 would not be met.
+         The sound would fall within this failure condition. </p>
+   </section>
+   <section id="examples">
+      <h2>Examples</h2>
       <section class="example">
          
             <ul>
-               <li> A site that plays continuous background music </li>
+               <li>A site that plays continuous background music </li>
             </ul>
          
       </section>
       <section class="example">
          
             <ul>
-               <li> A site with a narrator that lasts more than 3 seconds before
-                  	stopping, and there is no mechanism to stop it. </li>
+               <li>A site with a narrator that lasts more than 3 seconds before
+                  stopping, and there is no mechanism to stop it. </li>
             </ul>
          
       </section>
-   </section><section id="tests"><h2>Tests</h2>
+   </section
+   ><section id="tests">
+      <h2>Tests</h2>
       <section class="procedure"><h3>Procedure</h3>
          <ol>
-            <li> Check that there is a way in a Web page to turn off any sound that
-                                    plays automatically for more than three seconds.</li>
+            <li>Check that there is a mechanism to turn off any sound that
+               plays automatically for more than three seconds, independently from any other audio output
+               of the user agent or any other applications on the system.</li>
          </ol>
       </section>
-      <section class="results"><h3>Expected Results</h3>
+      <section class="results">
+         <h3>Expected Results</h3>
          <ul>
-            <li> If step #1 is not true then content fails Success Criterion 1.4.2
-                                </li>
+            <li> If step #1 is not true then content fails Success Criterion 1.4.2</li>
          </ul>
       </section>
-   </section><section id="related"><h2>Related Techniques</h2></section><section id="resources"><h2>Resources</h2></section></body></html>
+   </section>
+   <section id="related">
+      <h2>Related Techniques</h2>
+   </section>
+   <section id="resources">
+      <h2>Resources</h2>
+   </section>
+</body>
+</html>
--- a/techniques/failures/F23.html
+++ b/techniques/failures/F23.html
@@ -39,7 +39,7 @@
       <section class="procedure"><h3>Procedure</h3>
          <ol>
             <li>Check that there is a mechanism, independent from the overall system volume control, to turn off any sound that
-               plays automatically for more than three seconds, independently from the overall system volume level.
+               plays automatically for more than three seconds.
             </li>
          </ol>
       </section>
--- a/techniques/failures/F23.html
+++ b/techniques/failures/F23.html
@@ -38,7 +38,7 @@
       <h2>Tests</h2>
       <section class="procedure"><h3>Procedure</h3>
          <ol>
-            <li>Check that there is a mechanism to turn off any sound that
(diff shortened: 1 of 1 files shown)


patrickhlauke:
Any news on this @alastc ?

patrickhlauke:
bump

mbgower:
@patrickhlauke you stated
> Only makes sense to merge if https://github.com/w3c/wcag/pull/1825 (or an equivalent/modified version of the PR) is also merged

That was closed due to lack of support to change the normative text. However, there was a minor tweak to the Understanding document. Do you consider that sufficient to support the incorporation of this?

techniques/failures/F23.html
```diff
@@ -1,40 +1,64 @@
...
-                        sound off then Success Criterion 1.4.2 would not be met. Sounds that play over 3 seconds
-                        when there is no mechanism to turn off the sound included in the content
-                        would fall within this failure condition. </p>
-   </section><section id="examples"><h2>Examples</h2>
+   </section>
+   <section id="description">
+      <h2>Description</h2>
+      <p>This describes a failure condition for Success Criteria involving sound. If sound does not
+         turn off automatically within 3 seconds and there is no way to turn the
+         sound off, independently from any other audio output, then Success Criterion 1.4.2 would not be met.
```
mbgower, line 20:
```suggestion
         sound off, independently from the overall system volume level, then Success Criterion 1.4.2 would not be met.
```

alastc:
@mbgower you wanted to quickly scan this one before putting out to the group.

techniques/failures/F23.html
```diff
@@ -1,40 +1,60 @@
...
-   </section><section id="tests"><h2>Tests</h2>
+   </section
+   ><section id="tests">
+      <h2>Tests</h2>
       <section class="procedure"><h3>Procedure</h3>
          <ol>
-            <li> Check that there is a way in a Web page to turn off any sound that
-                                    plays automatically for more than three seconds.</li>
+            <li>Check that there is a mechanism to turn off any sound that
+               plays automatically for more than three seconds, independently from the overall system volume level.
```
mgifford, line 42: So if the volume is the same as the system volume does the length of an auto-played audio file matter? I would hope that anything automatically played you could shut off after 3 minutes (regardless of the volume level).

patrickhlauke: i think you're missing the point here - the "independently..." refers to the mechanism to turn off any sound. i.e the mechanism isn't "just turn off the audio of your entire system ... there ... now you satisfy the SC".

however, if this is confusing (that the "independently" refers to the mechanism), can be turned around to something like

`Check that there is a mechanism, independent from the overall system volume level, to turn off any sound that plays automatically for more than three seconds`

mraccess77:
Does this now mean that if checking conformance with a browser like Chrome this SC would be always met?

techniques/failures/F23.html
```diff
@@ -1,40 +1,60 @@
...
       </section>
-   </section><section id="tests"><h2>Tests</h2>
+   </section
+   ><section id="tests">
+      <h2>Tests</h2>
       <section class="procedure"><h3>Procedure</h3>
          <ol>
-            <li> Check that there is a way in a Web page to turn off any sound that
-                                    plays automatically for more than three seconds.</li>
+            <li>Check that there is a mechanism to turn off any sound that
```
mbgower, line 41:
```suggestion
            <li>Check that there is a mechanism, independent from the overall system volume control, to turn off any sound that
```

techniques/failures/F23.html
mbgower, line 42:
```suggestion
               plays automatically for more than three seconds.
```

techniques/failures/F23.html
```diff
@@ -1,40 +1,60 @@
...
-                    there is no mechanism to turn it off</title><link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove"></link></head><body><h1>Failure of  1.4.2 due to playing a sound longer than 3 seconds where
-                    there is no mechanism to turn it off</h1><section class="meta"><p class="id">ID: F23</p><p class="technology">Technology: failures</p><p class="type">Type: Failure</p></section><section id="applicability"><h2>When to Use</h2>
+<!DOCTYPE html>
+<html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml">
+<head>
+   <title>Failure of  1.4.2 due to playing a sound longer than 3 seconds where there is no mechanism to turn it off</title>
+   <link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove"></link>
+</head>
+<body>
+   <h1>Failure of 1.4.2 due to playing a sound longer than 3 seconds where there is no mechanism to turn it off</h1>
```
mbgower, line 8:
```suggestion
   <h1>Failure of 1.4.2 due to playing a sound longer than 3 seconds where there is no mechanism to turn it off or down</h1>
```

techniques/failures/F23.html
```diff
@@ -1,40 +1,60 @@
...
-                        sound off then Success Criterion 1.4.2 would not be met. Sounds that play over 3 seconds
-                        when there is no mechanism to turn off the sound included in the content
-                        would fall within this failure condition. </p>
-   </section><section id="examples"><h2>Examples</h2>
+   </section>
+   <section id="description">
+      <h2>Description</h2>
+      <p>This describes a failure condition for Success Criteria involving sound. If sound does not
+         turn off automatically within 3 seconds and there is no way to turn the
+         sound off, independently from the overall system volume level, then Success Criterion 1.4.2 would not be met.
```
mbgower, line 20:
```suggestion
         sound off, or control the volume independently from the overall system volume level, then Success Criterion 1.4.2 would not be met.
```

techniques/failures/F23.html
```diff
@@ -1,40 +1,60 @@
...
       </section>
-   </section><section id="tests"><h2>Tests</h2>
+   </section
+   ><section id="tests">
+      <h2>Tests</h2>
       <section class="procedure"><h3>Procedure</h3>
          <ol>
-            <li> Check that there is a way in a Web page to turn off any sound that
-                                    plays automatically for more than three seconds.</li>
+            <li>Check that there is a mechanism, independent from the overall system volume control, to turn off any sound that
```
mbgower, line 41:
```suggestion
            <li>Check that there is a mechanism, independent from the overall system volume control, to turn off  or turn down any sound that
```

mbgower:
> Does this now mean that if checking conformance with a browser like Chrome this SC would be always met?

@mraccess77 Is it your sense that changing the failure technique has altered the SC requirement? Or is it that as a result of re-assessing 1.4.2 during this review, are you identifying a possible way to meet it? I'm just trying to decide if you are identifying a concern with this correction.

patrickhlauke:
@mraccess77 i believe that was the long discussion that was had which then led to this PR (though i admit that because this has been in limbo so long, I might only be half remembering it).

I believe the discussion started with https://github.com/w3c/wcag/issues/1533 (though I also see https://github.com/w3c/wcag/issues/1543)

we then had various split off/parallel attempts to address this like https://github.com/w3c/wcag/pull/1825
there was discussions here https://www.w3.org/2022/08/16-ag-minutes.html#item10

I *think* this PR here was the final outcome after all of that, in my mind.

mraccess77:
I believe this does not change the SC but the practice of how/when people fail this issue. That is - it changes the interpretation that most people had.  This is fine -but I want to make sure that the group is alignment that this interpretation is indeed correct and the result should be for many to no longer fail this SC most of the time when audio is playing automatically without other page controls.  I believe this reading is in line with Trusted Tester.

techniques/failures/F23.html
```diff
@@ -1,40 +1,60 @@
-<!DOCTYPE html><html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml"><head><title>Failure of  1.4.2 due to playing a sound longer than 3 seconds where
-                    there is no mechanism to turn it off</title><link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove"></link></head><body><h1>Failure of  1.4.2 due to playing a sound longer than 3 seconds where
-                    there is no mechanism to turn it off</h1><section class="meta"><p class="id">ID: F23</p><p class="technology">Technology: failures</p><p class="type">Type: Failure</p></section><section id="applicability"><h2>When to Use</h2>
+<!DOCTYPE html>
+<html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml">
+<head>
+   <title>Failure of  1.4.2 due to playing a sound longer than 3 seconds where there is no mechanism to turn it off</title>
```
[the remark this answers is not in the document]
WilcoFiers: "down" suggests to me lower volume is okay. If you want it more precise "muted" is probably the word to use.

techniques/failures/F23.html
```diff
@@ -1,40 +1,60 @@
...
-   </section><section id="description"><h2>Description</h2>
-      <p> This describes a failure condition for Success Criteria involving sound. If sound does not
-                        turn off automatically within 3 seconds and there is no way to turn the
-                        sound off then Success Criterion 1.4.2 would not be met. Sounds that play over 3 seconds
-                        when there is no mechanism to turn off the sound included in the content
-                        would fall within this failure condition. </p>
-   </section><section id="examples"><h2>Examples</h2>
+   </section>
+   <section id="description">
+      <h2>Description</h2>
```
WilcoFiers, line 17: I think it's important to mention that most browsers use heuristics to mute automatically playing audio / video now. That's probably not sufficient as a mechanism, at least that's what ACT concluded when we looked at it. Similarly, several browser support muting individual tabs, we didn't think that qualifies as a mechanism either. I think its important to make that clear.

techniques/failures/F23.html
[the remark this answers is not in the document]
mraccess77: I agree with Wilco that turn down isn't specific enough as the SC requires stopping not just lowering the volume.  Turn down doesn't say by how much.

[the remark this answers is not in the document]
mbgower: In that case, I suggest we leave as is, since there is little practical difference between muting and turning off. The SC does allow lowering the volume as a mechanism, provided lowering can go all the way to zero, but for concision, I'm going to just resolve this without changing.

mbgower:
IMO, there are enough comments about what is considered an acceptable mechanism that that should be opened as a separate issue/discussion. I believe that discussion can take place independent of accepting this PR, and have opened https://github.com/w3c/wcag/discussions/3623

Issue #704, opened 2019-04-28
Failure technique F94 (1.4.4 resize text): remove "1280 pixels wide" step in test procedure


patrickhlauke:
trying to summarise a lengthy stream of consciousness on twitter https://twitter.com/patrick_h_lauke/status/1122478448986001408 ...

in short:

- the intention of the SC is that users should be able to resize text to 200% of its original default size (as defined by the author) 
- normatively, the SC does not scope the applicability to any particular size of viewport/browser width. users should be able to do this, regardless of whatever the starting size of their browser window/viewport is
- having "Set window size to 1280 pixels wide." as the second step in the testing procedure can lead to testers missing out potential failures (if a site defines a particular "mobile" set of styles, for instance, where `vh`/`vw` sizing for text is used - per the failure technique)

The fear seems to be that if a site uses responsive web approaches, it would be impossible to test as a site could continuously change text sizes for smaller viewport sizes. But this appears to be irrelevant, as the original intent is: from whatever the author defined as the starting size for their text (the "100%" state), can a user get the text to be twice as big without the need for AT. if they use browser zoom, and the site switches layout and makes the text smaller again, then that's a clear failure of the SC.

And per the "Full pages" note https://www.w3.org/TR/WCAG21/#cc2

> A full page includes each variation of the page that is automatically presented by the page for various screen sizes (e.g. variations in a responsive Web page). Each of these variations needs to conform (or needs to have a conforming alternate version) in order for the entire page to conform.

So, IF a site uses responsive breakpoints to swap out layouts, then testing should happen taking each breakpoint *as a starting point*. From each of those styles/layouts, can a user enlarge text to twice the size defined there. And if by trying to reach that 200% size they happen to trigger another breakpoint, that's mostly irrelevant to being able to ascertain if, from the starting size tested, the user is able to get to twice the size of text that was presented at the starting size.

patrickhlauke:
taking the example failure code of the technique, if a site (for whatever reason...e.g. under the assumption of "anything below 600 CSS px width is a mobile, and we want it to be app-like and not scale") did the following, the test procedure wouldn't catch it if testing just using browser full page zoom starting at 1280 as suggested in step 2 of the procedure:

```
.callout {
    font-size:1em;
}

@media (max-width: 600px) {
    .callout {
      font-size: 1vh;
    }
}
```

[...]

Issue #1522, opened 2020-11-13
Stacked SC: SC 1.4.4: Resize text AND 1.4.3: Contrast Do they have a interrelationship?


jake-abma:
Resizing text to 200% is possible for : https://dequeuniversity.com/demo/mars/ 

When I do a resize 200% text only in Firefox the contrast fails though.

Now I have followed discussions in the past about stacking SC, but not sure if this is a fail...
(I don't see a note in Contrast or Resize about the interrelationship)

![image](https://user-images.githubusercontent.com/6384002/99062818-d8ac4e80-25a3-11eb-900a-0e916ecb2560.png)

jake-abma:
refer https://github.com/w3c/wcag/issues/1524

patrickhlauke:
if i resize, and anything breaks some SC once resized, i would fail it under that relevant SC. however, in this case, the resize SC asks for having at least one way for users to resize text. they can use full-page zoom, not necessarily have to use text-only. so in this case, it feels a bit less clear-cut (as with #1524 this feels more like a "user is forcibly changing the styles defined by the author, so do we penalise the author for not foreseeing this?")

jake-abma:
@patrickhlauke  wrote:

> they can use full-page zoom, not necessarily have to use text-only. so in this case, it feels a bit less clear-cut

@alastc wrote:

My interpretation is that: If you can resize with zoom (which is the more common/available method) and get 200% increase in text size without loss of content, then it's a pass.

I think the question then is

**Do we want to fix it for a person only using text zoom (might be all he/she knows / wants) or is the 'page zoom' option the one this person should adopt / switch to, if text zoom does not provide the desired design?** 

(or should both 'text' and 'page' zoom be applicable for the SC, placing the bourdon on content authors, not the users preferrences)

patrickhlauke:
> Do we want to fix it for a person only using text zoom (might be all he/she knows / wants) or is the 'page zoom' option the one this person should adopt / switch to, if text zoom does not provide the desired design?

my personal view here is that there are many things a user would prefer to do. but placing the burden on authors to accommodate to all possible permutations of what a user may prefer to do is perhaps not realistic. currently the SC asks that " text can be resized" - and the general reading of that is that there's at least one way in which the text can be resized. and if there's at least one way a user can achieve this without introducing breakages (which would result in failures of other SCs), then the text resize SC itself is passed.

[...]

Issue #730, opened 2019-05-13
1.4.4 - Clarification needed for Success Criterion for mobile browsers


clshortfuse:
I've been in the process of porting some applications from Native to Web and have hit a roadblock a with Google Chrome. I've been trying, without any success, for over a year to get Google to make changes for text-based accessibility issues. There is no text-scaling support, nor is there zoom (as defined by WCAG) support for Chrome on Android. The only thing available is "pinch-to-zoom" which is basically screen magnification. I've seen too many comments and user-created articles and such stating that this is acceptable, but in actual practice, it's not, and from the reading the guidelines it's not either

I'll begin with the text of 1.4.4:

>Success Criterion 1.4.4 Resize text (Level AA): Except for captions and images of text, text can be resized without assistive technology up to 200 percent without loss of content or functionality.

Now, from the Intent section of:

https://www.w3.org/WAI/WCAG21/Understanding/resize-text.html

>The intent of this Success Criterion is to ensure that visually rendered text, including text-based controls (text characters that have been displayed so that they can be seen [vs. text characters that are still in data form such as ASCII]) can be scaled successfully so that it can be read directly by people with mild visual disabilities, without requiring the use of assistive technology such as a screen magnifier. Users may benefit from scaling all content on the Web page, but text is most critical.

Now, let's look at the **Sufficient Techniques**:

>### 1. G142: Using a technology that has commonly-available user agents that support zoom

According to [G142](https://www.w3.org/WAI/WCAG21/Techniques/general/G142)

>The objective of this technique is to ensure content can be scaled uniformly by using a Web technology supported by user agents that change text size via a Zoom tool.

The examples posted include

>Internet Explorer 7 and Opera 9 provide a zoom function that scales HTML/CSS page content uniformly.

To my understanding, *zoom* is using Ctrl + and Ctrl -, on the keyboard. This means the layout essentially changes and the content reflows. This is similar to change the CSS `zoom` property. 
  
The problem is, despite it's name, pinch-to-zoom does not "zoom" as explained by WCAG. Pinch-to-zoom magnifies instead. That means nothing changes in layout or size. No reflow happens. And according to 1.4.4, screen magnification CANNOT satisfy Success Criterion, even if it's provided by the user-agent. If the browser does not have actual page zoom (not magnify) function, then we cannot use this technique. Therefore, both Safari for iOS and Chrome for Android cannot use G142.

>### **2. L22: Supporting Browser Zoom in Silverlight**
[...]

PR #1838, opened 2021-06-01
Remove initial window/browser size from F94 test

patrickhlauke:
it's irrelevant, as the SC applies at any size, so doesn't need to appear to be limited to any particular one.

This addresses initial concern of https://github.com/w3c/wcag/issues/704 (but that issue has more discussion about the potential need for a normative change for 1.4.4, or addition of note(s) in the understanding, to set some lower bound)


patrickhlauke:
Any news on this @alastc ?

patrickhlauke:
Expanded slightly to add an extra note

techniques/failures/F94.html
```diff
@@ -45,17 +45,19 @@ <h3>Procedure</h3>
 				<p>In a desktop browser with a resizable window:</p>
 				<ol>
 					<li>Set zoom level to 100%.</li>
-                    <li>Set window size to 1280 pixels wide.</li>
-                    <li>Visit the page to be tested.</li>
-                    <li>Use any of the following methods to resize text when available:
+					<li>Visit the page to be tested.</li>
+					<li>Use any of the following methods to resize text when available:
 						<ul>
 							<li>the zoom feature of the browser</li>
```
mbgower, line 51:
```suggestion
							<li>the zoom feature of the browser,</li>
```

techniques/failures/F94.html
```diff
@@ -45,17 +45,19 @@ <h3>Procedure</h3>
...
+							<li>the zoom feature of the browser,</li>
 							<li>the text-sizing feature of the browser,</li>
 							<li>on-page controls for resizing text.</li>
 						</ul>
 					</li>
-                    <li>Check that the text resizes by one of the methods above, and can be resized to at least 200% of the default.</li>
+					<li>Check that the text resizes by one of the methods above, and can be resized to at least 200% of the default.</li>
 				</ol>
+				<div class="note">
+					<p>If the page has variations that are automatically presented for various screen sizes (e.g. media query breakpoints), you should run the test procedure for each variation.</p>
```
mbgower, line 59:
```suggestion
					<p>If the page has variations that are automatically presented for various screen sizes (for instance, media query breakpoints), you should run the test procedure for each variation.</p>
```

mbgower (approved):

mbgower:
@patrickhlauke Suggest taking the starting size of the test and incorporating into the meat of the F technique's content. Also if possible would like some guidance on tackling what is reasonable for lower resolutions. @scottaohara would be a good person to run your draft text by

bruce-usab:
@scottaohara requests that WCAG2ICT TF gives this issue some attention.

scottaohara:
thanks to @dbjorge for calling out #2101 as a directly related issue for a minimum testing resolution/viewport.

bruce-usab:
On call 12/8, @dbjorge also notes intersection with #1522 #730 and https://github.com/w3c/wcag2ict/issues/4

patrickhlauke:
this one *may* actually be superseded by https://github.com/w3c/wcag/pull/3739 - need to double-check once/if that one is merged if this is still needed. certainly the note

```
<div class="note">
	<p>If the page has variations that are automatically presented for various screen sizes (for instance, media query breakpoints), you should run the test procedure for each variation.</p>
</div>
```

appears new here compared to #3739 so this may be a refocused/separate PR just about the note

patrickhlauke:
As F94 uses `vw` as font size, it's irrelevant to point out in the note that this needs to be repeated at different sizes. closing

Issue #1855, opened 2021-06-10
WCAG 2.2 3.3.7: Accessible Authentication - confusion about "paste"


patrickhlauke:
https://www.w3.org/WAI/WCAG22/Understanding/accessible-authentication.html gives the impression that password managers/browsers rely on pasting (actual CTRL+V) passwords into fields. 

```
does not block copy-paste then browser features or password managers can
save the user’s information and refill the login.
```

```
Copy and paste can be relied on to avoid transcription. Examples of this are password
managers automatically filling in username and passwords,
```

```
A web site uses a properly marked up username (or email) and password fields as the login
authentication (...). Copy & paste is not blocked on the inputs, so the user's browser or extension
can identify the purpose of the inputs and automatically fill in the username and password.
```

this is not the case - they generally inject the `value` straight into the DOM. this can be observed if you do something like `addEventListener('paste', function() { console.log('paste triggered'); })` on a login form field.

The "can the user paste something they copied from an email or manually from a 3rd party, non-integrated password manager on their local system" aspect, and the "it works with password managers" aspect, are two separate things and should not be conflated/mushed together like this.

alastc:
That's probably be from me, I've seen a couple of password entries which prevented both pasting and password managers. Correlation not causation!

PR needed.

patrickhlauke:
Parking this here for now - quick and dirty test page https://codepen.io/patrickhlauke/full/jOBepRE

patrickhlauke:
If anybody uses some other password manager, like LastPass, it'd be interesting to see if they get different results on the above test page.

patrickhlauke:
PR submitted https://github.com/w3c/wcag/pull/1898

PR #1898, opened 2021-06-13
WCAG 2.2 3.3.7: Accessible Authentication - disambiguate "paste" and "password managers"

patrickhlauke:
- make the points separately about using properly marked-up 1.3.5 compliant form fields (nothing to do with paste). removed the mention of 1.3.1 inforel, as this seems unrelated (and more to the point, if we're saying password managers rely on the accessible name of an input, then it's not 1.3.1 but 4.1.2 that counts). added reference to 4.1.2 instead.
- tweak the copy/paste paragraph to be purely about ability/inability to copy from a third-party password manager and to then paste it into the form/webpage.
- add an expansion (with example) about the "different format between copied and pasted text" - a classic one, loved by banks, that ask you to enter specific digits/characters from a much longer password (which then 
requires manual transcription)
- add an example of accessible authentication about the copy/paste in isolation

This now more explicitly allows copy/paste-ability as a way to pass this criterion. This seems in line with the intent of https://github.com/w3c/wcag/pull/1419/files at the time? This also seems to tally with the result of the lengthy discussion in https://github.com/w3c/wcag/issues/1359 ?

(to be clear, the end result is basically: in general, as long as a page doesn't block pasting into the login fields - no other requirements beyond not requiring some "only enter the 3rd, 4th, etc" - it passes. this seems the end result of all the discussions above?)

Closes https://github.com/w3c/wcag/issues/1855


patrickhlauke:
I note that in one of the previous discussions, there was a mention of "copy/paste is fine, but not if it has additional restrictions like some time limit ... e.g. 'we emailed you, but the password is only valid for like 10 seconds'". this aspect doesn't seem to be present in the final understanding now. this is also a common thing with one-time-password solutions (where a user can copy the OTP but has a limited amount of time to paste it into a page). so is this still a valid/ok way to go?

understanding/22/accessible-authentication.html
```diff
@@ -40,19 +40,20 @@ <h2>Cognitive function test definition</h2>
...
  
         <p>Remembering a password is a <a>cognitive function test</a>. Such tests are known to be problematic for many people with cognitive disabilities. Whether it is remembering random strings of characters, a pattern gesture to perform on a touch screen, or identifying which images include a particular object, cognitive function tests will exclude some people. When a cognitive function test is used, at least one other authentication method must be available which is not a cognitive function test.</p>
             
-        <p class="note">Web sites can employ username (or email) and password inputs as an authentication method if it is properly marked-up and does not block copy-paste functionality. If the login form meets <a href="identify-input-purpose">Success Criterion 1.3.5 Input Purpose</a>,  <a href="info-and-relationships">Success Criterion 1.3.1: Info and Relationships</a>, and does not block copy-paste then browser features or password managers can save the user’s information and refill the login.</p>
```
alastc, line 47: I think it would be better to alter the "does not block" bit, we do need to say that websites can employ username & passwords, we got a ton of comments about that last time.

alastc: Nevermind, saw it moved further down.

alastc (approved):

patrickhlauke:
As this came up in a discussion with somebody again today, I just want to be absolutely crystal: with this change, which has now been merged, the absolute minimum requirement to pass this SC is "don't stop paste operations into login fields". *no* other requirement, not even requirement to "properly" mark up / identify input purpose (though of course that would still fail the separate SC 1.3.5).

PR #1909, opened 2021-06-16
Adding site-specific to understanding

### LINKED ISSUE #1900: 3.3.7 - Are system-level tests out of scope? Are PINs and Passwords synonymous?
The [success criterion](https://www.w3.org/WAI/WCAG22/Understanding/accessible-authentication.html) notes that 

> "Remembering a password is a cognitive function test.". 

Later on, it talks about 

> The user's device could use any available modality. Common methods on laptops and phones are facial-scan, fingerprint, and pin number. 

My understanding in this instance is that the technique is referring to a website agnostic PIN. For example, the PIN of their iPhone, rather than the PIN of the website.

Just that if I am right about the PIN, presumably remembering your device password is out of scope. For example, Safari might ask you to type into your device password after trying to use the OS FaceID or Touch ID. 

As well, it's common for banking website to have their own PIN numbers. So I think the line to draw here is the distinction between domain passwords and PINs (a website) and operating-system passwords and PINs.

Personally, I would consider defining "password" in the 'key terms" section. 

> In this context of this success criterion, a password is any input required from the user by the author. 
>
> This could be a password with alphanumeric characters and symbols, numerical PIN, file upload, API token, memorable phrases, and so on. 
> 
> When the author requires inputs form the user, such that they can authenticate themselves, it is a password. 
 
I would also consider adding an exception for OS passwords. For example:

>  For each step in an authentication process that relies on a cognitive function test, at least one other authentication method is available that does not rely on a cognitive function test, or a mechanism is available to assist the user in completing the cognitive function test.
>
> **Except when the cognitive test is required by the operating system ands is unmodified by the author**.

alastc:
Closes #1900


alastc:
Approved during the meeting, need to un-pick the conflict.

patrickhlauke:
unpicked it for you (in essence, taking in your change, but removing the `<p class="note">Web sites can employ username (or email) and password inputs...` para which was removed as part of the recently merged https://github.com/w3c/wcag/pull/1898)

alastc:
Ah, great, thanks @patrickhlauke

PR #1946, opened 2021-07-06
Rewrite h69 d macdonald

DavidMacDonald:
To begin to address Issue #1712 
https://github.com/w3c/wcag/issues/1712

### DIFF
--- a/techniques/html/H69.html
+++ b/techniques/html/H69.html
@@ -9,7 +9,7 @@
          <li>to demarcate different navigational sections like top or main navigation, left or
         				secondary navigation and footer navigation</li>
          <li>to mark up images of text that are used as headings</li>
-		 <li>to allow users the ability to navigate a page by sections or skip repeated blocks of information</li>
+		 <li>to allow users the ability to navigate a page by sections or skip repeated blocks of information, if they have assistive technology that can navigate to headings</li>
       </ul>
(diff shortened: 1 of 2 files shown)


patrickhlauke:
> The intent of this Success Criterion is to allow people who navigate sequentially through content more direct access to the primary content of the Web page.

If, per today's minutes, "navigate sequentially" doesn't include sighted keyboard users (despite the fact that they do *navigate* sequentially, moving focus with `Tab`/`Shift+Tab` primarily), then this should be made clear. For years at least my, and many other author/auditor's understanding, has been that keyboard *is* a classic sequential navigation mechanism. Is the meaning here "reading/consuming" sequentially then, rather than "navigating"?

kengdoj:
Removing this benefit for keyboard users from the Understanding article is significant. Is it being removed just so H69 will fit?

> People who use only the keyboard or a keyboard interface can reach content with fewer keystrokes. Otherwise, they might have to make dozens of keystrokes before reaching a link in the main content area. This can take a long time and may cause severe physical pain for some users.

Agree with @patrickhlauke that sighted keyboard users are navigating sequentially so an explanation of why they are not included would be needed.

patrickhlauke:
Closing this, as the PR is now heavily conflicted with changes that have happened since it was originally filed. The original issue remains open as a to-do item.

PR #1947, opened 2021-07-06
removed benefits that don't rely on AT

DavidMacDonald:
Addresses issue #1712
https://github.com/w3c/wcag/issues/1712

### DIFF
--- a/understanding/20/bypass-blocks.html
+++ b/understanding/20/bypass-blocks.html
@@ -78,17 +78,7 @@
             content is spoken.
          </li>
          
-         <li>People who use only the keyboard or a keyboard interface can reach content with fewer
-            keystrokes. Otherwise, they might have to make dozens of keystrokes before reaching
-            a link in the main content area. This can take a long time and may cause severe physical
-            pain for some users.
-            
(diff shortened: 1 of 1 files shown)


patrickhlauke:
this looks/looked concerning. the comments from today's call about "there's no pure keyboard users, they can use a mouse or mouse keys" would, to me, start to also put into question other SCs like ... 2.1.1 Keyboard? (which explicitly talks about mouse keys not being a suitable escape clause)

mraccess77:
I agree mousekeys have never been acceptable as a solution.   Mousekeys require use of site and they also can't provide all of the same functionality needed and definitely are not equivalent in terms of control and dexterity (and time) required to use them

patrickhlauke:
I understand of course there's nuance in this case as it is primarily about sighted keyboard users (since AT users are assumed to have better tools, such as navigating by heading, which is then why H69 would be a valid technique to pass), but yes even in that scenario, I wouldn't say mouse keys should be a get-out-of-jail-free card to say it's fine for keyboard users. anyway, maybe just me...

Issue #1927, opened 2021-06-23
Page Break Navigation uses undefined term to define a term, and needs to clarify


mbgower:
2.4.13 Page Break Navigation has the defined term "page break locators" in the normative text, which reads:
> programmatic markers that are arranged in a meaningful sequence to determine the location of a page in relation to others in the set.

I find it a little disconcerting that “page break locators” is defined as “programmatic markers”, both because it's defining a term with another undefined term and also because the definition infers any page numbers that are just text would be ignored in this requirement. Sorta like saying “where you’ve put in an anchor for a page number, provide a link to it, otherwise, don’t worry about page numbers”.

At the least I think it should read "[programmatically determinable](https://www.w3.org/TR/WCAG21/#dfn-programmatically-determinable) anchors or destinations that are arranged..."

If my inference is correct that page numbers that are not programmatically determinable do not need to have page break navigation, this should be stated clearly in the Understanding (hopefully along with language encouraging authors to make the page numbers programmatically determinable).

alastc:
> Sorta like saying “where you’ve put in an anchor for a page number, provide a link to it, otherwise, don’t worry about page numbers”.

Sorta, but more like: "where you’ve put in an anchor for a page number, provide a link to it, otherwise, don’t worry about **the links**".
It is quite a narrow scope, but intentionally so because we don't want to catch lots of different things like headings, HRs, etc.

Regarding "programmatic markers", I find it helped to work back from what is the primary technique:
`<span id="pg1" role="doc-pagebreak" aria-label="page 1">`

Anchor is a specifically HTML (technology) term, not the end of the world, but I'm not sure whether anchor or destination would be better than marker?

aubbink1:
How about "**Indicators** which are arranged in a meaningful sequence to determine the location of a page in relation to others in the set."

Indicator can mean programmatic, identifier, locator and anchor, all of which point to the purpose of 'page break locators'.

alastc:
> How about "Indicators which are arranged...

To me an "indicator" implies something visual, particularly when we use that term with focus-indicators.

A marker (dictionary definition) is "an object used to indicate a position, place, or route.", which seems right. I could live with "anchor or destination", we can put that to the group.

Otherwise, the request was for "If my inference is correct that page numbers that are not programmatically determinable do not need to have page break navigation, this should be stated clearly in the Understanding", which needs someone to work on it.

alastc:
Change approved in the meeting.

PR #1952, opened 2021-07-09
Page Break Nav terminology clarity

aubbink1:
Updated content to better define the term 'programmatic markers'
<hr>
Related to Issue: Page Break Navigation uses undefined term to define a term, and needs to clarify #1927

### DIFF
--- a/understanding/22/page-break-navigation.html
+++ b/understanding/22/page-break-navigation.html
@@ -24,7 +24,7 @@
       </blockquote>
 
       <h3>Definitions:</h3>
-      <p><strong>page break locators</strong> programmatic markers that are arranged in a meaningful sequence to determine the location of a page in relation to others in the set.</p>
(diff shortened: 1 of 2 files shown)


understanding/22/page-break-navigation.html
```diff
@@ -24,7 +24,7 @@ <h2>Page Break Navigation Normative Text</h2>
       </blockquote>
 
       <h3>Definitions:</h3>
-      <p><strong>page break locators</strong> programmatic markers that are arranged in a meaningful sequence to determine the location of a page in relation to others in the set.</p>
+      <p><strong>page break locators</strong> programmatically determinable anchors or destinations that are arranged in a meaningful sequence to determine the location of a page in relation to others in the set.</p>
```
patrickhlauke, line 27: main WCAG spec document uses `<dl>`/`<dt>`/``<dd>` for definitions - may be worth adopting the same structure here?

alastc: This part is only in the understanding-doc previews, when it's published properly it is discarded.

PR #1964, opened 2021-07-16
Target size minimum: conforming alternate version

detlevhfischer:
Adding a paragraph about meeting Target Size (Minimum) via conforming alternate versions, taking input type=number as an example (following discussion in the WCAG issues meeting today). I do not feel particularly happy about it, partly because it uses input type=number as an example. This feels odd for two reasons: 

1. native input type=number would fall under the user agent exception, so not need a conforming alternate version for that reason;
2. input and the arrows generated in some UA are essentially one control so it feels odd treating this control as two separate things, the field being treated as an alterate version to the arrows. Alternate versions would usually be distinct (like a number field below a custom slider that is not keyboard accessible).

Since the in-page conforming alternate version is potentially an option for many different things anyway, and one we would not usually want to encourage, it stands to reason that it sends the wrong signal to explicitly include this Understanding text.

### DIFF
--- a/understanding/22/target-size-minimum.html
+++ b/understanding/22/target-size-minimum.html
@@ -63,7 +63,7 @@
       
       <h3>Alternatives for targets on the same page</h3>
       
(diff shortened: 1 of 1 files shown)


scottaohara:
could not a custom styled stepper be used instead of the native `input type=number`'s buttons?  That would get around the native element's UA styling being exempt, and would call out a more likely scenario where custom built or adjusted controls would the author's responsibility .

For instance, this [codepen](https://codepen.io/prasanthmj/pen/EWQPrK) (not mine) still uses the native `input type=number`, but has been adjusted to remove the UA buttons and new custom up/down arrows are added instead.  This would be a good candidate for this, no?

detlevhfischer:
@scottaohara sounds good, and would remove that dissonance with native control exception

PR #1975, opened 2021-07-27
Update G219.html

mbgower:
Fixed typo and reduced wording. If we have to spend so much time telling people what a Kanban board is, I suggest it's not a great example. I circumvented by just adding "planning".


detlevhfischer (approved):
Looks good to me!

PR #1991, opened 2021-08-12
Keep session, scoping 3rd parties out

### LINKED ISSUE #1880: 3.3.8: Ensuring single transactions (from the user's perspective) are covered
There are several sites on which, in order to complete a transaction such as buying something, the user is redirected to a payment provider before being sent back to the original site. E.g. when buying something from eBay, some of the screens in the middle of this process will be provided by PayPal.

We understand that such screens _would_ be covered here, as from the user's perspective they form part of one single transaction.

This was the case with the initial version of the wording, and we gather it would be the case with the latest wording, which includes the term "user-session". The term is not defined, and could have many different meanings based on who is interpreting the SC, so we suggest it be removed as it appears to be redundant.

This issue has been raised on behalf of [TPGi](https://www.tpgi.com), based on an internal consultation process. [An article containing all of our comments](https://www.tpgi.com/wcag-22-working-draft-feedback/) can be found on our blog, along with an introduction and acknowledgments.

alastc:
Closes #1880

### DIFF
--- a/understanding/22/redundant-entry.html
+++ b/understanding/22/redundant-entry.html
@@ -41,7 +41,7 @@
         
         <p>Requiring people to recall information already entered in the previous steps can cause them to give up or enter incorrect information. The autocomplete feature of browsers is not considered sufficient because it is the content (the web site) that needs to provide the stored information for a redundant entry, or avoid asking for the same information again.</p>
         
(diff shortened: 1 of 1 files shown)


alastc:
Not accepted.

Issue #1976, opened 2021-07-27
2.5.8 Target size (minimum) lacks 'Alternative' exception


mbgower:
During the conversation for allowing a user agent exception for [target size](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html) as per https://github.com/w3c/wcag/issues/1904, ShawnT gave the date picker (and color picker) as an example of a native html element that have target sizes that fail the criterion. @awkawk also mentioned the spin button.

Looking at an example of an input type="date" [implementation](https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_input_type_date),
![image](https://user-images.githubusercontent.com/14298245/127198672-1472dcd6-fbe7-4e7c-8521-8186b823401c.png)
 I realized I've always ignored these because I simply type in the data. Which made me realize that there is no wording here for an alternative method exception.

Such wording exists in the AAA Target Size (Enhanced) requirement:
> Equivalent
The target is available through an equivalent link or control on the same page that is at least 44 by 44 CSS pixels;

I propose modifying that slightly for the AA to read:
> Equivalent
The function can be achieved through an equivalent control that is at least 24 by 24 CSS pixels;

---
On the related topic of user agent exception, I thought I'd mention that there is also an exception in our AAA requirement:
> User Agent Control
The size of the target is determined by the user agent and is not modified by the author;

so it seems odd that we do not have at least this equivalent in the AA criterion

---

Part of me wonders if we need _both_ exceptions. I believe the user agent exception may suffice?

mbgower:
@rachaelbradley ^

mraccess77:
Typing in a date in not equivalent to having access to the calendar to provide context of dates by day of week, month, looking in future, or and allows selection without typing.  We wanted all of the pointer related criteria to apply regardless of keyboard alternative.

mbgower:
There are a few interesting things to parse there, @mraccess77 

### User agent exception
![image](https://user-images.githubusercontent.com/14298245/127326324-b3bafade-bb9a-48a4-902c-3f8e17dfc5fd.png)
I just pasted in a snippet of the same calendar widget, this time taken from Chrome (in my original post, it's Safari). Interestingly, all the calendar grid components pass in the Chrome example, while all of them appear to fail in the Safari example.

[...]

Issue #1904, opened 2021-06-14
Target size and user-agent components that can't be styled


JamesCatt:
It might be good to include something in this SC explicitly stating that it does not apply to user-agent components that cannot be adequately styled by the author. For example:

- the options list of `<select>` dropdowns
- date picker for the HTML date input
- color picker for the HTML color input

This may seem obvious/implied for people who understand the HTML/CSS and the limitations around these, but I think it would be helpful to make it explicit for those who don't.

patrickhlauke:
in essence, an exemption similar to the one in 2.5.5, but reworded from

> User Agent Control: The size of the target is determined by the user agent and is not modified by the author;

to something like

> User Agent Control: The size of the target is determined by the user agent and *cannot be modified* by the author;

mraccess77:
Controlled by the user agent and not modified and "cannot be modified by the author" are two very different things.  Which one was the intent of the working group?

jake-abma:
My memory says "not modified" and this already / also covers "cannot be modified by the author"

jake-abma:
> "Controlled by the user agent and not modified "and "cannot be modified by the author"

All target sizes are in the basis / default controlled (determined) by the UA, and most can be modified.

BUT, all who are capable of being adjusted must be fixed, so the "Controlled by the user agent and not modified" must be modified if possible. This was the intent.

Leaving the only option open for the suggestion:

> User Agent Control: The size of the target is determined by the user agent and cannot be modified by the author;

detlevhfischer:
@jake-abma 

> BUT, all who are capable of being adjusted must be fixed, so the "Controlled by the user agent and not modified" must be modified if possible.

I don't think the intention was to exempt only those elements that at a given point in time **cannot** be styled. Since this can change any time in new browser versions, int would be unclear whether content that could not be modified yesterday is suddenly found to fail today because now, it **can** be modified.
I think (but I am not positively sure) that the WG understanding was that if authors leave elements alone (i.e. apply no styling to select, checkboxes, radio inputs, etc.) then the target size requirement would be exempt.

alastc:
Noting the last discussion on this is at: https://www.w3.org/2021/11/16-ag-minutes#item13

alastc:
We've got this exception:
> User agent control: The size of the target is determined by the user agent and is not modified by the author;

And it is explained in the [understanding document](https://w3c.github.io/wcag/understanding/target-size-minimum.html).

PR #1993, opened 2021-08-13
Target Size addition of user agent and equivalent exceptions

mbgower:
This incorporates what I believe was the group decision on incorporating legal into the Essential bullet.
As well I have added exceptions for user agent and equivalent, which are drawn from the wording that exists in the Target Size (Enhanced) SC, as discussed in https://github.com/w3c/wcag/issues/1976

Addresses #1976, #1904.


mbgower:
@alastc Time permitting, we can possibly discuss this on tomorrow's call? I will try to get the Dragging understanding document updated before that meeting as well.

mraccess77:
Thanks for making these updates.  The exception for alternatives is "The function can be achieved through a different control that has an area of at least 24 by 24 CSS pixels" but the SC itself allows for further exceptions such as no other targets in a 24x24 area.   Was this on purpose to make the alternative exception stricter than the SC itself?  I'm fine with that - I just wanted to confirm the intent.

patrickhlauke:
logically, it should really be that the different control satisfies the SC (including its exception), rather than just a more limited aspect of it, in my opinion.

detlevhfischer:
Agree that it might be more straightforward to say 
> **Equivalent:** The function can be achieved through a different control that meets the requiremens of this SC 

...or similar - but since this exception does not spell out that the equivalent control has to be on the same page, it could technically lead to a recursive "passing fo the bucket" even across several pages (the control has an equivalent which itself does not meet the SC but has an equivalent elsewhere etc.) - so this exception appears to be diluting the requirement too much. How can a formulation prevent this?

patrickhlauke:
i'd rather have the risk of recursion (which I would say is low, because you'd then essentially "short-circuit" the passing the buck as long as there's one control at the end of the chain that passes the SC, and THAT is then the alternative for the first one in the chain) than an unintended tightening of scope/requirements.

alastc:
This was approved here: https://www.w3.org/2021/11/16-ag-minutes#item13 With the caveat that the understanding doc needs additional work.

PR #2003, opened 2021-08-18
Update dragging-movements.html

### LINKED ISSUE #1979: SC 2.5.7 Dragging exceptions seems like example of free form paths rather than dragging
The 2 examples given for exceptions seem more like free form rather than dragging motions which we define to be only concerned with the beginning and endpoint.  There are some types of drawing tools that are that way - such as when you draw a line and you set a target at the beginning and then move to wherever and set the end target and a line is drawn.   It might be worth clarifying this.   Although in these cases it seems like it could be done without a drag.  Simply move the pointer to the start and press a button and then locate the end point and press another button.  If we state that a particular track needs to be followed then it's by virtual not a dragging motion and thus isn't an exception - it's just not applicable.   A better example in a game might be a dragging motion where the endpoint is moving - but even then it seems like it could be met another way.

Current text
"drawing a freeform figure, or playing a game where the aim is to follow a particular track as accurately as possible. The criterion has an exception for these cases where the task requires dragging."

detlevhfischer:
Changing the essential exception example to separate it from dragging requiring a specific path
Closes #1979


understanding/22/dragging-movements.html
```diff
@@ -41,7 +41,7 @@ <h2>Intent of Dragging Movements</h2>
 
     <p>This Success Criterion applies to dragging motions as opposed to pointer gestures, which are covered in Success Criterion 2.5.1 Pointer Gestures. Pointer gestures include directional path-based as well as multi-point gestures. In contrast, for dragging motions, only the start and end point of the motion matters, not the actual path.</p>
 
-      <p>There may be cases where a dragging movement is essential, for example, drawing a freeform figure, or playing a game where the aim is to follow a particular track as accurately as possible. The criterion has an exception for these cases where the task requires dragging.</p>
+      <p>There may be cases where a dragging movement is essential, for example, when playing a game where the aim is to quickly drag an element over a moving, jumping, or suddenly emerging dragging target. The criterion has an exception for these cases where the task requires dragging.</p>
```
patrickhlauke, line 44: `dragging target` maybe better `drop target` (to keep with the drag-n-drop nomenclature for things)

PR #2004, opened 2021-08-18
Updating G195, and separating a technique for focus-app-enh

### LINKED ISSUE #1931: Update Technique G195 procedure
https://w3c.github.io/wcag/techniques/general/G195

I believe the first two steps of the test procedure need to be updated, with an added preamble, to read:
> For each user interface element on the page that should receive keyboard focus: 
>1. Navigate to the element and check that it has a visible focus indicator.
>1. Check that the focus indicator area…

Step 3 in the old version becomes step 2 here with the remaining 2 steps becoming steps 3 and 4. Revised version to read:
> For each user interface element on the page that should receive keyboard focus: 
>1. Navigate to the element and check that it has a visible focus indicator.
>1. Check that the focus indicator area is at least the size of a 1 CSS px border around the component.
>1. Check that the change of contrast of the indicator between focused and unfocused states has a ratio of 3:1 for the minimum focus indicator area.
>1. If the focus indicator does not have 3:1 contrast ratio with its adjacent colors, check that it is at least 2px thick.

I think step 1 covers ‘partially visible' as well (and Focus Visible too, hmmm). However, there is a second challenge with this test, in that it doesn’t check that the focus indicator achieves 3:1 against the background (which is required for 1.4.11 Non-text Contrast). Not sure whether we should even go there… Kinda underscores how overlapped these SCs are. It may be a case of making 2 or 3 separate techniques plus the Focus Visible one, making them combine to achieve this. Similar to how [H49](https://www.w3.org/WAI/WCAG21/Techniques/html/H49) is dealt with in 1.3.1
> This technique relates to Success Criterion 1.3.1: Info and Relationships (Sufficient, together with G115: Using semantic elements to mark up structure).

alastc:
Closes #1931


techniques/general/G195.html
```diff
@@ -1,7 +1,7 @@
...
       <p>Generally applicable.</p>
    </section><section id="description"><h2>Description</h2>
-      <p>The objective of this technique is enhance the focus indicator in the browser, by creating a highly visible one in the content. The default focus indicator in some browsers is a thin, dotted, black line. It can be difficult to see the line when it is around a form element which already has an outline, when the focused element is inside a table cell, when the focused element is very small, or when the background of the page is a dark color. Some browsers use a pale blue outline, which can be difficult to see on some backgrounds.</p>
+      <p>The objective of this technique is enhance the focus indicator in the browser, by creating a  visible one in the content. The default focus indicator in some browsers is a thin, dotted, black line. It can be difficult to see the line when it is around a form element which already has an outline, when the focused element is inside a table cell, when the focused element is very small, or when the background of the page is a dark color. Some browsers use a pale blue outline, which can be difficult to see on some backgrounds.</p>
```
patrickhlauke, line 4: stray double space now where `highly` was removed

techniques/general/G195.html
```diff
@@ -24,11 +24,12 @@ <h3>Menus</h3>
       </section>
    </section><section id="tests"><h2>Tests</h2>
       <section class="procedure"><h3>Procedure</h3>
+         <p>For each user interface element on the page that should receive keyboard focus:</p>
```
patrickhlauke, line 27: `user interface component` rather than element? https://www.w3.org/TR/WCAG22/#dfn-user-interface-components

techniques/general/G195.html
```diff
@@ -24,11 +24,12 @@ <h3>Menus</h3>
...
          <ol>
-            <li>Place focus on each focusable user interface element on the page using the keyboard.</li>
-            <li>Check that there is a highly visible focus indicator.</li>
+            <li>Navigate to the element and check that it has a visible focus indicator.</li>
```
patrickhlauke, line 29: matching change from `element` to `component` ?

techniques/general/Gxxx-highly-visible-focus-indicator.html
```diff
@@ -0,0 +1,43 @@
...
+    </section>
+    <section class="example">
+       <h3>Menus</h3>
+       
+          <p>A Web page includes an interactive menu with sub-menus. A user can move focus in the menu using the arrow keys. As focus moves, the currently focused menu item changes its background to a different color, which has a 15:1 contrast ratio with the surrounding items (whilst maintaining text contrast).</p>
+       
+    </section>
+ </section><section id="tests"><h2>Tests</h2>
+    <section class="procedure"><h3>Procedure</h3>
+       <p>For each user interface element on the page that should receive keyboard focus:</p>
```
patrickhlauke, line 26: same suggested change here and in the next `<li>` about `component` rather than `element`

patrickhlauke (approved):

PR #2020, opened 2021-08-27
Issue 1932 Gramatical rewrite of visible controls

### LINKED ISSUE #1932: Suggested rewording to normative text of Visible Controls
A hard SC to parse, especially the preamble:
>Where receiving pointer hover or keyboard focus triggers user interface components to be visible, information needed to identify that user interface components are available is visible, except when:

Always have to reread just to sort out where the verb is. Wondering if it would be better with "Where UICs become visible as a result of pointer hover or keyboard focus, information needed to identify…"

Would become:
> Where UICs become visible as a result of pointer hover or keyboard focus, information needed to identify that user interface components are available is visible, except when:

DavidMacDonald:
We also have a few options instead of "such"

- use "these components"
- use the full term "user interface components"
- after the first appearance of "user interface components (UIC)" and then use UIC the second time.

NOTE: I still feel a bit awkward with "identify that ..." but I"m not sure there is a fix.

I also added to the Understanding doc with an FAQ section about when the SC does not apply.

Closes #1932

### DIFF
--- a/guidelines/sc/22/visible-controls.html
+++ b/guidelines/sc/22/visible-controls.html
@@ -5,7 +5,7 @@
     <p class="conformance-level">AA</p>
 	<p class="change">New</p>
     
-    <p>Where receiving pointer hover or keyboard focus triggers <a>user interface components</a> to be visible, information needed to identify that user interface components are available is visible, except when:</p>
+  <p>Where <a>user interface components</a> become visible as a result of pointer hover or keyboard focus, information needed to identify that such components are available is provided visually, except when:</p>
     <ul>
         <li>The information needed to identify the user interface components is available through an equivalent component that is visible on the same page or on a different step in a multi-step process without requiring pointer hover or keyboard focus;</li>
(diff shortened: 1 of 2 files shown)


mbgower:
@DavidMacDonald I think rather than have the Q&A section, you should just create sections. The first could be "Where controls become visible on hover or keyboard focus"
For the ones that _do_ apply, you could use actual examples for those.

mbgower:
For now, I'm going to create a branch that only captures the normative changes, and leave the understanding doc changes for later.

mbgower:
@DavidMacDonald please have a look at this and merge if friendly. Objective would be to get normative changes accepted now, and follow up with Understanding doc non-normative changes later.
https://github.com/w3c/wcag/pull/2029

bruce-usab (commented):
Have we implemented a Q&A approach with any other Understanding docs?

Q&A docs are (generally) nice, I just don't know if this SC warrants a new pattern.

alastc:
Visible controls has been removed from 2.2, so closing.

Issue #1872, opened 2021-06-10
2.5.8: Further Examples Required


benja11y:
To fully understand the impact this SC will have, further examples should be provided that include complex widgets that contain multiple, but separate, discrete targets.  Examples could include [material design style navigation drawers](https://material.io/components/navigation-drawer), date pickers, compact tab controls, ARIA `<select>`-style controls or interactive toast messages that partially obscure content underneath. 

This issue has been raised on behalf of [TPGi](https://www.tpgi.com), based on an internal consultation process. [An article containing all of our comments](https://www.tpgi.com/wcag-22-working-draft-feedback/) can be found on our blog, along with an introduction and acknowledgments.

detlevhfischer:
@benja11y I had a look at the various material design components and nothing strikes me as potenial fail here. Elements may be smaller than 24 x 24 px but there seems to be always ample spacing. So I am just curious where you see the benefit of adding examples from that design vocabulary. I would imagine edge cases ould be more important to cover.

detlevhfischer:
I just checked a few of the ARIA Authoring practices examples:
[Date picker aspin button example:](https://www.w3.org/TR/wai-aria-practices-1.1/examples/spinbutton/datepicker-spinbuttons.html)
Spin buttons 20 x 20 px but sufficient space above and plenty sidewards

Combobox examples:
The example below shows line height is set to 24 px so it just meets the SC:
![combobox-listbox](https://user-images.githubusercontent.com/9989439/121702715-69afbd80-cad2-11eb-877f-e29d5dcadbae.gif)

The same seems to be the case on the other combobox examples. If there are any examples you'd like to call out, let me know. So far I do not see what adding further examples from either source will add.

becka11y:
@benja11y I believe the above comment is for you :-)

alastc:
Closing with #2024

PR #2024, opened 2021-09-02
target size of obscured targets

detlevhfischer:
Addressing #1872 asking for clarification regarding dynamic content that obscures other targets
- Requirement does not apply to targets while they are fully or partly obsured by dynamic content
- Requirement does however apply to additional targets brought up by interaction

An issue potentially still to be discussed is if this exception equally holds for content that appears not as a result of user interaction but as a result of scripted behaviour of the content (think "Take a survey now" dialogs or cookie banners). Such content will often violate other SCs already (such as 2.4.3 Focus Order and 2.2.1 Timing Ajustable). My hunch is that both types of dynamic content should be treated the same in terms of the target size requirement.

### DIFF
--- a/understanding/22/target-size-minimum.html
+++ b/understanding/22/target-size-minimum.html
@@ -64,8 +64,7 @@
         <li>activating a button displays a modal dialog</li>
         <li>The page displayas a cookie banner after page load</li>
         <li>The page displays a "Take a survey" dialog after some period of user inactivity</li>
-      </ul>
-          
+      </ul>    
       <p>The requirement does however apply to targets in any new content that appears on top of other content.</p>
       
       <p>While the Success Criterion primarily helps touch users by providing target sizing to prevent accidental triggering of adjacent targets, it is also useful for mouse or pen users. It reduces the chances of erroneous activation due to either a tremor or reduced precision, whether because of reduced fine motor control or input imprecision.</p>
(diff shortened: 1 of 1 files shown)


understanding/22/target-size-minimum.html
```diff
@@ -57,8 +57,17 @@ <h2>Intent of Target Size (Minimum)</h2>
         </ol>
 
         <p>The requirement is independent of the zoom factor of the page; when users zoom in the CSS pixel size of elements does not change. This means that authors cannot meet it by claiming that the target will have enough spacing or sufficient size if the user zooms into the page.</p>
-
-        <p>While the Success Criterion primarily helps touch users by providing target sizing to prevent accidental triggering of adjacent targets, it is also useful for mouse or pen users. It reduces the chances of erroneous activation due to either a tremor or reduced precision, whether because of reduced fine motor control or input imprecision.</p>
+      
+      <p>The requirement does not apply to targets while they are fully or partly obsured by content displayed as a result of a user interaction or scripted behaviour of content, for example:</p>
```
patrickhlauke, line 61: `partly` perhaps better `partially`

patrickhlauke (approved):

mraccess77:
So it's clear we are now staying we don't need to test obscured items - but do those obscured items come into calculation when testing the non-obscured items overlapping them?

PR #2029, opened 2021-09-07
Update visible-controls.html

mbgower:
Updated the normative text in the Understanding document to match that used in the SC text, and removed the Q&A section, which is in an unusual format, and which can be incorporated later as a separate non-normative change to the SC.

### DIFF
--- a/understanding/22/visible-controls.html
+++ b/understanding/22/visible-controls.html
@@ -21,7 +21,7 @@
          <p class="conformance-level">AA</p>
         <p class="change">New</p>
          
-         <p>Where receiving pointer hover or keyboard focus triggers <a>user interface components</a> to be visible, information needed to identify that user interface components are available is visible, except when:</p>
+         <p>Where <a>user interface components</a> become visible as a result of pointer hover or keyboard focus, information needed to identify that such components are available is provided visually, except when:</p>
          <ul>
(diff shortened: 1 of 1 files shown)


detlevhfischer:
Adding a link to a rendered version of the Understanding doc:
https://raw.githack.com/w3c/wcag/bc3b52b54f67c6ae57d7defae99c08a4a02985ac/understanding/22/visible-controls.html

detlevhfischer:
Not sure whether this has been discussed before - but in the section "Embedded controls", I am not sure whether in the case of content with controls that dissapear a few seconds after activation, a requirement is intended that there should be a permanently visible control to display the video controls when they become hidden. The common situation is that such controls may re-appear on hover or focus (sometimes only on hover). I take it that this would not be sufficient to meet the SC? 

I can turn this into a new issue if it is better dealt with separately.

mraccess77:
Hi @detlevhfischer I think this was covered in the understanding and that the presence of the video player was enough to communicate the additional controls.

mbgower:
This PR appears to be redundant, since other changes have been made to the Understanding document.
Although I will say @alastc that on relooking at this, I do think this preamble scans better than what we have (with or without "keyboard" and "pointer")
> Where [user interface components](https://w3c.github.io/wcag/understanding/visible-controls.html#dfn-user-interface-component) become visible as a result of pointer hover or keyboard focus

Existing
> When [user interface components](https://w3c.github.io/wcag/understanding/visible-controls.html#dfn-user-interface-component) are invisible until hover or focus makes them visible,

Issue #2033, opened 2021-09-10
Understanding for Contrast does not call out Color Blindness


bruce-usab:
The [Understanding document](https://www.w3.org/WAI/WCAG22/Understanding/contrast-minimum.html) for minimum Contrast (SC 1.4.3) does not explicitly mention that the 4.5:1 metric well addresses the needs of people with (the four types of) red-green color blindness and (the two types) of blue-yellow color blindness.

I believe this to be a problem, as the normative requirement lacks obvious face validity.  My own personal experience is that I have had to explain this feature dozens of times.  I think is would be better for our authoritative reference to include this detail.

I will submit a pull request with a minimal edit.

bruce-usab:
_Originally posted by @GreggVan in https://github.com/w3c/wcag/issues/695#issuecomment-916568064_

> But remember that in the end you need 
> 1. to account for color blindness as well as low vision 
> 2. account for 3 layer contrast  (A contrasts with B that contrasts with A and C) 
> 3. that you need enough colors left after you do 1 and 2 to be acceptable to designers.

For (3) is it worth mentioning that of the 2.8x10^14 RGB pairs, _X_ provide contrast of 4.5:1 or higher?

Myndex:
Hi Bruce @bruce-usab and @mbgower 

I also responded BRIEFLY in the pull request with specific change requests...  #2034 

### The Ableist Term "Color Blind"
Commenting here, includes some of what I mention in the pull request.

Before I get into "technical stuff" I want to being on an agenda item I want to discuss for on the LVTF, namely the term "color blind" ... I am hoping we can start normalizing something that is not so ableist, as 99.98% of CVD types are **not color blind,** they are color limited. But they still see colors.

_AS SUCH_, I object to adding that line in the PR as is, can instead it say something like

_"sometimes, though incorrectly, referred to as color blindness"_ 

or perhaps (my preference but longer) 

_**"formerly referred to as 'color blind', though in fact the vast majority of color-limited vision types do differentiate at least some colors"**_

## Some Additional Technical Comments:

For the first line (18) in the PR, can we add "_...so long as luminance contrast is high enough to support rapid reading._" the suggested edit of the first line then reads:

**For people without color deficiencies, hue and saturation have minimal or no effect on legibility so long as luminance contrast is high enough to support rapid reading** _(Knoblauch et al., 1991)._

[...]

PR #2034, opened 2021-09-10
Update contrast-minimum.html

bruce-usab:
* Light editorial to opening intent.
* Add paragraph about color blindness being addressed.
* Add sentence about 4.5:1 permitting black-on-grey-on-white.
* Add link to NEI explainer on color blindness.

### DIFF
--- a/understanding/20/contrast-minimum.html
+++ b/understanding/20/contrast-minimum.html
@@ -1,130 +1,506 @@
-<h1>Understanding Contrast (Minimum)</h1>
-<section>
-	<h2>Intent of Contrast (Minimum)</h2>
-	<p> 
-    The intent of this Success Criterion is to provide enough contrast between text and its background so that it can be read by people with moderately low vision.
+<!DOCTYPE html>
+<html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml">
+<head>
+   <meta charset="UTF-8"></meta>
+   <title>Understanding Contrast (Minimum)</title>
+   <link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove"></link>
+</head>
+<body>
+   <h1>Understanding Contrast (Minimum)</h1>
+   
+   
+   <section id="intent">
+      <h2>Intent of Contrast (Minimum)</h2>
+      
+      
+	<p>The intent of this Success Criterion is to provide enough contrast between text and its background so that it can be read by people with moderately low vision.
     This user group does not typically use contrast-enhancing assistive technology. 
     For people without color deficiencies, hue and saturation have minimal or no effect on legibility as assessed by reading performance (Knoblauch et al., 1991).
     Color deficiencies can affect luminance contrast somewhat. 
     Therefore the contrast is calculated in such a way that color (hue) is not a key factor. 
-    People who have a color vision deficit will also have adequate contrast between the text and the background.
-  </p> 
-  <p> 
-    The contrast ratios selected address the needs of people with the four types of red-green color blindness (deuteranomaly, protanomaly, protanopia, and deuteranopia) and the two types of blue-yellow color blindness (tritanomaly and tritanopia). 
-    Monochromacy is also addressed. 
-  </p> 
-  <p> 
-    While end-user needs are the primary consideration behind the contrast ratios used, a ratio of 4.5:1 provides content authors a wide range of color options for foreground and background colors. 
-    A contrast ratio of 4.5:1 also allows content authors some limited options for three colors that each contrast with each other. 
-    (For example, a web page could use use a middle gray background color while featuring white and black body text.) 
-  </p>
-	<p>Text that is decorative and conveys no information is excluded. For example, if random words are used to create a background and the words could be rearranged or substituted without changing meaning, then it would be decorative and would not need to meet this criterion.</p>
-	<p>Text that is larger and has wider character strokes is easier to read at lower contrast. The contrast requirement for larger text is therefore lower. This allows authors to use a wider range of color choices for large text, which is helpful for design of pages, particularly titles. 18 point text or 14 point bold text is judged to be large enough to require a lower contrast ratio. (See The American Printing House for the Blind Guidelines for Large Printing and The Library of Congress Guidelines for Large Print under <a href="#resources">Resources</a>). "18 point" and "bold" can both have different meanings in different fonts but, except for very thin or unusual fonts, they should be sufficient. Since there are so many different fonts, the general measures are used and a note regarding fancy or thin fonts is included.</p>
-	<div>
-		<p>When evaluating this success criterion, the font size in points should be obtained from the user agent or calculated on font metrics in the way that user agents do. Point sizes are based on the CSS pt size as defined in <a href="https://www.w3.org/TR/css-values-3/#reference-pixel">CSS3 Values</a>. The ratio between sizes in points and CSS pixels is 1pt = 1.333px, therefore 14pt and 18pt are equivalent to approximately 18.5px and 24px.</p>
-		<p>Because different image editing applications default to different pixel densities (e.g. 72 PPI or 96 PPI), specifying point sizes for fonts from within an image editing application can be unreliable when it comes to presenting text at a specific size. When creating images of large-scale text, authors should ensure that the text in the resulting image is roughly equivalent to 1.2 and 1.5 em or to 120% or 150% of the default size for body text. For example, for a 72 PPI image, an author would need to use approximately 19 pt and 24 pt font sizes in order to successfully present images of large-scale text to a user.</p>
-	</div>
-	<p>The previously-mentioned contrast requirements for text also apply to images of text (text that has been rendered into pixels and then stored in an image format) as stated in Success Criterion 1.4.3.</p>
-	<p>This requirement applies to situations in which images of text were intended to be understood as text. Incidental text, such as in photographs that happen to include a street sign, are not included. Nor is text that for some reason is designed to be invisible to all viewers. Stylized text, such as in corporate logos, should be treated in terms of its function on the page, which may or may not warrant including the content in the text alternative. Corporate visual guidelines beyond logo and logotype are not included in the exception.</p>
-	<p>In this provision there is an exception that reads "that are part of a picture that contains significant other visual content,". This exception is intended to separate pictures that have text in them from images of text that are done to replace text in order to get a particular look.</p>
-	<div>
-		<p>Images of text do not scale as well as text because they tend to pixelate. It is also harder to change foreground and background contrast and color combinations for images of text, which is necessary for some users. Therefore, we suggest using text wherever possible, and when not, consider supplying an image of higher resolution.</p>
-	</div>
-	<p>The minimum contrast success criterion (1.4.3) applies to text in the page, including placeholder text and text that is shown when a pointer is hovering over an object or when an object has keyboard focus. If any of these are used in a page, the text needs to provide sufficient contrast.</p>
-	<p>Although this Success Criterion only applies to text, similar issues occur for content presented in charts, graphs, diagrams, and other non-text-based information. Content presented in this manner should also have good contrast to ensure that more users can access the information.</p>
-	<p>See also <a href="contrast-enhanced">1.4.6: Contrast (Enhanced)</a>.</p>
-	<section>
-		<h3>Rationale for the Ratios Chosen</h3>
-		<p>Text that is larger and has wider character strokes is easier to read at lower contrast. The contrast requirement for larger text is therefore lower. This allows authors to use a wider range of color choices for large text, which is helpful for design of pages, particularly titles. 18 point text or 14 point bold text is judged to be large enough to require a lower contrast ratio. (See The American Printing House for the Blind Guidelines for Large Printing and The Library of Congress Guidelines for Large Print under <a href="#resources">Resources</a>). "18 point" and "bold" can both have different meanings in different fonts but, except for very thin or unusual fonts, they should be sufficient. Since there are so many different fonts, the general measures are used and a note regarding fancy or thin fonts is included.</p>
-		<div>
-			<p>When evaluating this success criterion, the font size in points should be obtained from the user agent or calculated on font metrics in the way that user agents do. Point sizes are based on the CSS pt size as defined in <a>CSS3 Values</a>. The ratio between sizes in points and CSS pixels is 1pt = 1.333px, therefore 14pt and 18pt are equivalent to approximately 18.5px and 24px.</p>
-			<p>Because different image editing applications default to different pixel densities (e.g. 72 PPI or 96 PPI), specifying point sizes for fonts from within an image editing application can be unreliable when it comes to presenting text at a specific size. When creating images of large-scale text, authors should ensure that the text in the resulting image is roughly equivalent to 1.2 and 1.5 em or to 120% or 150% of the default size for body text as rendered by the browser.</p>
-		</div>
-		<p>Hues are perceived differently by users with color vision deficiencies (both congenital and acquired) resulting in different colors and relative luminance contrasts than for normally sighted users. Because of this, effective contrast and readability are different for this population. However, color deficiencies are so diverse that prescribing effective general use color pairs (for contrast) based on quantitative data is not feasible. Requiring good luminance contrast accommodates this by requiring contrast that is independent of color perception. Fortunately, most of the luminance contribution is from the mid and long wave receptors which largely overlap in their spectral responses. The result is that effective luminance contrast can generally be computed without regard to specific color deficiency, except for the use of predominantly long wavelength colors against darker colors (generally appearing black) for those who have protanopia. (We provide an advisory technique on avoiding red on black for that reason). For more information see [[ARDITI-KNOBLAUCH-1994]] [[ARDITI-KNOBLAUCH-1996]] [[ARDITI]].</p>
-		<div>
-			<p>Some people with cognitive disabilities require color combinations or hues that have low contrast, and therefore we allow and encourage authors to provide mechanisms to adjust the foreground and background colors of the content. Some of the combinations that could be chosen may have contrast levels that will be lower than those found in the Success Criteria. This is not a violation of this Success Criterion provided there is a mechanism that will return to the default values set out in the Success Criteria.</p>
-		</div>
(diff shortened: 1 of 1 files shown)


bruce-usab:
This PR closes issue 2033.

patrickhlauke (approved):

mbgower:
Looks good @bruce-usab. I [committed a couple of trivial edits](https://github.com/w3c/wcag/pull/2034/commits/a5c87af406a09bfa1783d0ad057be51e6a191778) straight into your changes. Please take a look.

bruce-usab:
Thanks @mbgower -- those are edits are very good.

Myndex:
I have a longer response with deeper dives in the issue thread request. #2033 

### _Ableist Term "Color Blind"_
I am hoping we can start normalizing something that is not so ableist, as 99.98% of CVD types **are not color blind,** they are color limited. But they still see colors.

**AS SUCH, I object to adding the line using the term "color blind" without qualification** as is added here, can instead it say something like

_"sometimes, though incorrectly, referred to as color blindness"_ 

or perhaps (my preference but longer) 

_**"formerly referred to as 'color blind', though in fact the vast majority of color-limited vision types do differentiate at least some colors"**_

## _A Couple Other Pull Issues_
Additional technical discussion in the issues thread #2033 , but if possible can we:

### FIRST LINE(18):
For the first line in the PR, can we add "_...so long as luminance contrast is high enough to support rapid reading._" the suggested edit of the first line then should read:

**For people without color deficiencies, hue and saturation have minimal or no effect on legibility so long as luminance contrast is high enough to support rapid reading** _(Knoblauch et al., 1991)._

### Second Line (19)
It reads "Color deficiencies can affect luminance contrast somewhat." 

Can we change to: 
**"Color deficiencies can affect luminance contrast in some cases, as can other impairment types."** 

### CVD Line (22)
I suggest removing any claims of assisting Monochromacy because that is unsupportable.

### Other
In the issue thread I outline my other objections, most of which I have maintained since opening issue #695 in April 2019. 


Thank you!

Andy

Myndex:
### Regarding the FIRST LINE (18)
While I suggested a minor edit in the previous post, I wanted to comment further as that line seemed a bit askew and out of context relative to some more recent research. I discuss in [detail in this post in #2033](https://github.com/w3c/wcag/issues/2033#issuecomment-923684961)

### _Suggested New Line 18:_
**For people with normal vision and reading very large letters, hue and saturation have minimal effect on legibility. (Knoblauch et al., 1991) However for people with low vision, for people with color limited vision, luminance contrast is the most important for readability (Legge, Psychophysics of Reading XI). This also holds true for normal vision when using small letters such as in body text. (Barten 2003, CSF)**


Thank you,

Andy

mraccess77:
I'd suggest we not say "normal vision" - what is normal?  Perhaps typical or just stating without visual impairment.

Myndex:
> I'd suggest we not say "normal vision" - what is normal? Perhaps typical or just stating without visual impairment.

Hi @mraccess77

Hi Jonathan,

### "Normal" vision is a specific definition, and a clinical definition:

- Snellen acuity of 20/20 or lower (20/16 is "perfect" acuity, 20/200 is SSA disabled) 
- Peli-Robson contrast sensitivity of 1.95 or higher (2.25 is "best")
- Farnsworth Munsell Hue Color TES of 60 or less (TES 0 is perfect)
- 100% Visual field with MD not lower than -2db (0 is best, -20db is SSA disabled)

Depending on the study, "normal" can include "with refractive correction" if the correction can result in the above scores. An example: "normal with correction needed for presbyopia". Presbyopia is "normal" for over age 40 for instance.

### "Normal" is not "most perfect"

Because normal is specifically defined, it also provides a useful baseline relating to user needs. Normal vision has natural limitations that define minimum requirements for readability. From this foundation of normal vision needs, we can then indicate the degree of additional accommodation needed for varying degrees of impairment(s).

patrickhlauke:
> "Normal" vision is a specific definition, and a clinical definition

doesn't the same also apply to "Color Blind"?

alastc:
It looks like most US places default to color blindness, but I quite like how the NHS in the UK approach it:
https://www.nhs.uk/conditions/colour-vision-deficiency/

Defaulting to "Colour vision deficiency (colour blindness)".

Myndex:
Hi @patrickhlauke 

> > "Normal" vision is a specific definition, and a clinical definition
> doesn't the same also apply to "Color Blind"?

No, the clinical term for "color blind" is **Color Vision Deficiency (CVD)**, and in every clinical or academic example I've seen, where they may use to colloquial term colorblind early in the articly, they also immediately point out that people are not actually blind of color except the very rare cases of schromatopsia.

I would like to see the public adopt the term "Color Limited."

And it is certainly possible to do so. The easy example is that relating to terms for intellectual disability. In the mid 1970, the **_clinical term_** for someone that was cognitively challenged was _"moron"_ Circa 1975 they changed it because it had become derogatory... they changed it to "mental retardation"  then in 2010, that had becomd derogatory, so Obama officially changed it again, this time to "intellectual disability." 

IMO it's a matter of pointing out that "Color Blind" is ableist.

Myndex:
> I'd suggest we not say "normal vision" - what is normal? Perhaps typical or just stating without visual impairment.

Hi Jonathan @mraccess77 

I just wanted to follow up on this — it looks like the term "standard vision" is used  often in research, and I think it is more neutral than "normal vision" as such I intend to start using "standard vision" in my writing.

Thank you for pointing this out,

Andy

mbgower:
@alastc I suspect this is another situation where an outstanding PR was missed because the issue was never marked for survey, so I just added it.

bruce-usab:
It looks to me like these edits got out of sync with other proposed changes.  I am all for closing this PR and working from the most resent version, since that might be the easiest way to resolve the conflicts.

> No, the clinical term for "color blind" is Color Vision Deficiency (CVD)...

I agree with Understanding using more precise and correct terminology.  Please note that in the U.S. (at least) CVD is fairly widely understood (among laypeople) to mean cardiovascular disease.

> ...where they may use to colloquial term colorblind early in the articly, they also immediately point out that people are not actually blind of color .except the very rare cases of schromatopsia.

I am of the opinion that 1.4.3 Understanding should fit that pattern.

> I would like to see the public adopt the term "Color Limited."

That term is not very catchy!  Yes, people's use of words changes over time.  I am big proponent of people-first language.

> IMO it's a matter of pointing out that "Color Blind" is ableist.

I have been in (live) conversations where "ableist" was used to effectively argue a point.  Regardless, it is not a term that I (personally) feel comfortable using.  I am skeptical that it would help this Understanding page.

understanding/20/contrast-minimum.html
```diff
@@ -13,15 +13,14 @@ <h1>Understanding Contrast (Minimum)</h1>
...
-         hue and saturation have minimal or no effect on legibility as assessed by reading
-         performance (Knoblauch et al., 1991). Color deficiencies can affect luminance contrast
-         somewhat. Therefore, in the recommendation, the contrast is calculated in such a way
-         that color is not a key factor so that people who have a color vision deficit will
-         also have adequate contrast between the text and the background.
-      </p>
+	<p>The intent of this Success Criterion is to provide enough contrast between text and its background so that it can be read by people with moderately low vision.
+    This user group does not typically use contrast-enhancing assistive technology. 
+    For people without color deficiencies, hue and saturation have minimal or no effect on legibility as assessed by reading performance (Knoblauch et al., 1991).
+    Color deficiencies can affect luminance contrast somewhat.
```
patrickhlauke, line 19: reads a bit odd/informal with the "somewhat" at the end. Maybe "Color deficiencies can have a slight impact on the perception of luminance contrast"

understanding/20/contrast-minimum.html
```diff
@@ -13,15 +13,14 @@ <h1>Understanding Contrast (Minimum)</h1>
...
-         somewhat. Therefore, in the recommendation, the contrast is calculated in such a way
-         that color is not a key factor so that people who have a color vision deficit will
-         also have adequate contrast between the text and the background.
-      </p>
+	<p>The intent of this Success Criterion is to provide enough contrast between text and its background so that it can be read by people with moderately low vision.
+    This user group does not typically use contrast-enhancing assistive technology. 
+    For people without color deficiencies, hue and saturation have minimal or no effect on legibility as assessed by reading performance (Knoblauch et al., 1991).
+    Color deficiencies can affect luminance contrast somewhat. 
+    Therefore the contrast is calculated in such a way that color (hue) is not a key factor. 
+    People who have a color vision deficit (often referred to as "color blindness") will also have adequate contrast between the text and the background.</p>
```
patrickhlauke, line 21: maybe add a "do" to make it a bit clearer ("For people without color deficiencies ... People who do have a color vision deficit..."

"will also have adequate contrast between the text and the background." maybe at the end, add - to make it a more complete sentence - "if the requirements of this Success Criterion are followed", as currently it's left a bit hanging...

understanding/20/contrast-minimum.html
bruce-usab: I agree about the "somewhat" at the end.  How about using parallel of previous sentence?  So instead of:

> Color deficiencies can affect luminance contrast somewhat. 

We would have:

> For many people with color vision deficiencies, luminance contrast has a significant impact on legibility.

understanding/20/contrast-minimum.html
bruce-usab: Agree that this needs improvement.

understanding/20/contrast-minimum.html
gundulaniemann: I feel the impact of luminance difference is on all people and not specific to color vision deficiencies. To the contrary, people with color vision impairments often have a very big luminance sensitivity.

How do you like the following:
The intent of this Success Criterion is to provide enough contrast between text and its background so that it can be read by people with moderately low vision.
    This user group does not typically use contrast-enhancing assistive technology. 
For many people, luminance contrast has a significant impact on legibility.
    For people without color deficiencies, hue and saturation have minimal or no effect on legibility as assessed by reading performance (Knoblauch et al., 1991).
    Yet color deficiencies can also affect luminance contrast.

Myndex: Hi @gundulaniemann 

> _I feel the impact of luminance difference is on all people and not specific to color vision deficiencies._

As far as **reading and readability** is concerned, it's all about **luminance** contrast, for all sighted users.

The visual word form area (VWFA) is the pre-lexical processing that recognizes whole words in the letter pears, and it functions on the luminance channel exclusively without consideration of color as a hue or colorfulness. MRI studies using the Stroop test confirm this (as does the Stroop test itself for that matter).

**_All vision types_** require ample luminance contrast for best readability. By "ample" I mean what Ian Bailey/Jan Lovi-Kitchin call the "contrast reserve", for best fluent readability, they suggest 20 times the threshold (JND).


> _To the contrary, people with color vision impairments often have a very big luminance sensitivity._

Yes, the deutan/protan/tritan forms of color vision deficiency (CVD) have as good or better contrast sensitivity as standard vision. The exception is protan and red/orange against black.

I am going to respond to this more in depth in the later post.

mraccess77:
I think what the text may be trying to say is that contrast while important to all - low contrast may still allow for reading of text because of the difference in hue can augment it - for example red text on green background with typical color vision.  However, in this example - for people with limited color perception the aid of hue is not available in some cases and thus sufficient contrast is more important in these specific situations.

Myndex:
## _Short Recap_

For readability, **all sighted users** need ample luminance contrast. 

For **_all sighted users_**, saturated blue should be the darkest of a color pair, as blue has nearly no luminance. This goes for saturated reds and oranges, but reds and oranges are _critical_ because they are seen even darker by the _protan_. Reds, oranges, and purples should generally not be paired with a darker color like black. 

In other words, per the most recent peer reviewed science, luminance contrast is not negatively affected among color vision deficient,²⁻³ with the exception of reds paired with darker colors like black.⁴


### A Gentle Objection

**These lines lack accuracy:**

> ~Therefore the contrast is calculated in such a way that color (hue) is not a key factor.~

Despite summing the colors to a _calculated_ achromatic luminance, **hue is still a key factor**, because the WCAG 2.x math is not perceptually uniform and does not consider the luminance deficit experienced by certain CVD types, namely protan who see reds and oranges **_darker_**. See the "Minutiae" twisty below.


> ~People who have a color vision deficit (often referred to as "color blindness") will also have adequate contrast between the text and the background.~

Not necessarily. The WCAG 2.x contrast math can not deliver on this promise, and therefore this is not a statement that should be made IMO. WCAG 2.x contrast does nothing special nor particularly helpful for color vision deficiency. See the "Minutiae" twisty below.

--------

## _Please Consider This Alternate Wording:_

The intent of this Success Criterion is to provide enough luminance contrast between text and its background so that it can be read by people with moderately impaired contrast perception, without the use of contrast-enhancing assistive technology. 

Hue and saturation have minimal to no effect on readability, as assessed by reading performance.¹ Further, Color Vision Deficiencies (CVD) per se, do not negatively affect luminance contrast perception.²⁻³

However, because some CVD types see colors containing a substantial amount of red as darker (less luminance), colors such as saturated reds, oranges, and purples, should usually be the _darkest_ of a color pair.

### _Refs:_
1) [Effect of chromatic and luminance contrast on reading Knoblauch, Arditi, Szlyk, 1991. DOI: 10.1364/JOSAA.8.000428](http://dx.doi.org/10.1364/JOSAA.8.000428)
2) [Luminance contrast sensitivity in color-deficient observers](https://pubmed.ncbi.nlm.nih.gov/24103453/)
3) [Contrast sensitivity of patients with CVD](https://metrovision.fr/2019/2019_Ilhan_Contrast_sensitivity_in_congenital_color_deficiency.pdf)

-------
-------

## The Minutiae

<details><summary>CLICK for foundation and details</summary>


The deutan/protan/tritan forms of color vision deficiency (CVD) have as good or better contrast sensitivity as standard vision²⁻³. The exception is protan and red/orange against black⁴.

In fact dichromat CVD such as deuteranopia and protanopia have slightly **better** visual function overall than standard vision, a current theory being due to a lower neural noise.

As far as readability is concerned, and luminance contrast is concerned, the **one issue  is with protanopia**. As protanopia is missing L cones, they are insensitive to red and colors composed mostly of red, such as orange. They see **reds and oranges _darker_**, and therefore, **when reds or oranges are against black, the protan has significantly reduced luminance contrast perception.**

**Generally speaking, reds, and some oranges, should be the darkest of a color pair** and this does also apply to standard vision, but **it is critical for certain CVD types**. In other words saturated reds and many oranges & purples should be paired with white.

WCAG 2.x contrast **_does not_** calculate to accommodate this, and instead frequently rejects white yet allows black, the opposite of what is needed for color vision deficiencies. _(this is partly because the WCAG2 math is not divided around contrast center for the top and the bottom 4.5:1)_. (Performance claims, if they exist, should be removed). 

Aside from protanopia, the _actual color-blind_ types, i.e. the rare blue cone monochromat or rod-only achromatopsia, also do not "see red". But these types require AT as they are co-morbid with low vision and severe photophobia⁵.

Tritans are not affected as far as luminance contrast is concerned, as S cones (blue) are not a part of _luminance_ in the human vision system. 

Deuteranopia, having all L cones and no M cones have essentially standard red sensitivity (but only luminance and a 'saturation' for distinguishability from green), and as the L cone response overlaps M, there is little change and studies have confirmed that deutans have equal or better luminance **_contrast_** sensitivity compared to standard vision²⁻³.


REFS
2] [Luminance contrast sensitivity in color-deficient observers](https://pubmed.ncbi.nlm.nih.gov/24103453/)

3] [Contrast sensitivity of patients with CVD](https://metrovision.fr/2019/2019_Ilhan_Contrast_sensitivity_in_congenital_color_deficiency.pdf) </details>

4] [What's Red and Black and Also Not Read](https://tangledweb.xyz/whats-red-black-also-not-read-573b9c0a97ed)

5] [CVD Simulator with experimental achromatopsia sim](https://www.myndex.com/CVD/)

Myndex:
Hi Jon @mraccess77 

> ..._low contrast may still allow for reading of text because of the difference in hue can augment it_...

At the higher spatial frequencies of _most_ text, color contrast becomes insignificant to readability. There are some studies that have shown that color contrast can be readable... but they were also using _**gigantic**_ letters (over 560px !!).

Here's an example, the first three are maximum color (hue) contrast, with little to no luminance contrast. The last one is luminance contrast and related saturation contrast, but no hue contrast.

bad design choice color contrasts, there are three buttons where the text is contrasting only with hue but not luminance, and then the fourth button the contrast of the Text is luminance

It's important that the majority of text, especially text that is intended to be read, rely on achromatic luminance contrast.

bruce-usab:
@Myndex I very much like your updated intent:

> The intent of this Success Criterion is to provide enough luminance contrast between text and its background so that it can be read by people with moderately impaired contrast perception, without the use of contrast-enhancing assistive technology.

Your next two paragraphs are educational, but I do not think they on topic for Understanding 1.4.3.

> Hue and saturation have minimal to no effect on readability, as assessed by reading performance.¹ Further, Color Vision Deficiencies (CVD) per se, do not negatively affect luminance contrast perception.²⁻³

> However, because some CVD types see colors containing a substantial amount of red as darker (less luminance), colors such as saturated reds, oranges, and purples, should usually be the darkest of a color pair.

Myndex:
Hi @bruce-usab 

> _Your next two paragraphs are educational, but I do not think they on topic for Understanding 1.4.3._

### I agree... 

Rather than delete these three sentences, I was rewording/correcting some of what was above, i.e. could be deleted, here's why I reworded:

> _Color deficiencies can affect luminance contrast somewhat._ 

No, except for protan with reds on black. _**Otherwise**_, CVD have equal or better visual function, including contrast sensitivity, as standard vision.


> _Therefore the contrast is calculated in such a way that color (hue) is not a key factor._ 

MMMmmm... This is an unusual statement—reducing the color to luminance does not disregard hue, hues are weighted per expected luminance contribution... But more importantly we don't want to make people think that hue is not important, because it still is for some things, just not readability.


> _People who have a color vision deficit (often referred to as "color blindness") will also have adequate contrast between the text and the background._

No, not with WCAG2 math they won't. This is a performance claim which shouldn't be made.


References supporting these supra (in the above posts)

bruce-usab:
I am taking a fresh stab at this, with a third PR.

alastc:
Closing as superseded by #3240

PR #2040, opened 2021-09-15
Re-doing the border aspect.

### LINKED ISSUE #1328: How should 2.4.11 / 2.4.12 be applied to inline elements that wrap over multiple lines?

alastc:
Closes #1328

### DIFF
--- a/understanding/22/focus-appearance-minimum.html
+++ b/understanding/22/focus-appearance-minimum.html
@@ -224,10 +224,10 @@
                     <figcaption> The focused button with the non-contrasting areas removed, showing that it is a thick indicator that meets the requirements.</figcaption>
                 </figure>
 
-                <p>If an inline link is broken over two or more lines, the indicator can be split between them. In this case the link element is still treated as a single control.</p>
+                <p>If an inline link is broken over two lines, some methods of outline are treated differently by browsers. Using CSS, an outline will either wrap each part of the link separately, or wrap the whole link as one perimeter. Each part of the link will pass or fail as though they were separate links. If a border is used the indicator can be split between them. When using a border the thickness may need to be increased to compensate.</p>
                 <figure id="focus-indicator-inline-link">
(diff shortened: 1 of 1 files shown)


understanding/22/focus-appearance-minimum.html
```diff
@@ -224,10 +224,10 @@ <h4>Unusual shapes and gradients</h4>
                     <figcaption> The focused button with the non-contrasting areas removed, showing that it is a thick indicator that meets the requirements.</figcaption>
                 </figure>
 
-                <p>If an inline link is broken over two or more lines, the indicator can be split between them. In this case the link element is still treated as a single control.</p>
+                <p>If an inline link is broken over two lines, some methods of outline are treated differently by browsers. Using CSS, an outline will either wrap each part of the link separately, or wrap the whole link as one perimeter. Each part of the link will pass or fail as though they were separate links. If a border is used the indicator can be split between them. When using a border the thickness may need to be increased to compensate.</p>
```
patrickhlauke, line 227: suggest adding commas for better readability

`If a border is used, the indicator can be split between them. When using a border, the thickness may need to be increased to compensate.`

alastc: done.

mraccess77:
Saying multiple lines rather than two is better if links wrap on more than 3 lines.

PR #2042, opened 2021-09-17
Updates to resolve #1902

### LINKED ISSUE #1902: 3.3.7 - Defining common objects
My apologies if this is re-opening issues such as https://github.com/w3c/wcag/issues/1256, but I can see in the updated draft that the success criterion now includes...

> If the test is based on something the website has set such as remembering or transcribing a word, or **recognising a picture the website provided**, that would be a cognitive functional test. **Recognising common objects**, or a picture the user has provided, would not be a cognitive functional test. 

In this context, what is the definition of a common object? It means sense to me that "which of this pictures matches the layout of the ACME corp secure website" **would** be a cognitive test and "select the black cabs" **would not** be.

However, how about this example from LinkedIn where users have to recognise the spiral galaxy. 

screenshot from LinkedIn login flow requiring that users select a picture of a spiral galaxy out of other galaxy formations

It is common in the broad sense ... there are plenty of spiral galaxies. But not common in what I think is meant here. I assume that the meaning is "common, everyday objects". I think that this will always be fuzzy and so may need to be expanded further. For example: 

> Recognizing common objects, or a picture the user has provided, would not be a cognitive functional test. 

And then defining common objects as a key term. For example: 

> A common object is an everyday item, appliance, term, and so on. For example:
>
> ...
> ...
>
>  A common object is not something that requires domain-specific knowledge to recognise. For example, differentiating a skyscraper from a tent is differentiating common objects. However, differentiating an art-deco and neo-classical buildings is not differentiating common objects. 
>
>  Some forms of object recognition may require an understanding of a particular culture. For example, taxis can appear differently in different locales. This is an issue for many people, including people with disabilities, but it is not considered an accessibility-specific issue.Some forms of object recognition may require an understanding of a particular culture.

rachaelbradley:
This PR creates an AA and AAA SC as a compromise to allow recognizing common objects as an exception.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/pull/2042.html" title="Last updated on Sep 21, 2021, 10:06 PM UTC (9191b45)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/2042/79b3471...9191b45.html" title="Last updated on Sep 21, 2021, 10:06 PM UTC (9191b45)">Diff</a>


alastc:
I think we need to update the index prior to merging, to include the AAA version.

PR #2044, opened 2021-09-20
Outlining non-scoped things.

### LINKED ISSUE #1875: 3.2.7: Controls that require scrolling to reach
We gather that controls that the user can scroll to would pass this SC. We gather the following scenarios would pass:

* T&C modals that require a user to scroll to get to the controls that allow them to accept the terms.
* Cases where more options/fields become visible because of a selection elsewhere on the form. Think of filling out an address and the City field is not visible until the State is selected.
* Similarly, an e-reader that allows the user to select some text and then pops up some buttons to allow the user to interact in different ways with that text would pass.

If that's the case, it may be worth clarifying that scrolling is acceptable.

However, what would happen in the following two different cases?

* With the examples above _if there are no visible scrollbars_ to alert the user to the fact that there's more content to scroll through before they get to the controls required to proceed?
* If table column sort controls only appear on hover/focus, but the user needs to use them in order to find the information they're looking for in the table?

This issue has been raised on behalf of [TPGi](https://www.tpgi.com), based on an internal consultation process. [An article containing all of our comments](https://www.tpgi.com/wcag-22-working-draft-feedback/) can be found on our blog, along with an introduction and acknowledgments.

alastc:
Closes #1875


understanding/22/visible-controls.html
```diff
@@ -63,6 +63,14 @@ <h3>When Needed</h3>
...
+      
+      <p>Many scenarios are out of scope because they do not rely on hover or focus, for example:</p>
+      <ul>
+         <li>Options/fields become visible because of a <b>selection</b> elsewhere on the form. For example, filling out an address and the City field is not visible until the State is selected.</li>
```
patrickhlauke, line 69: "filling out ... is not visible" sounds awkward/wrong. perhaps `For example, more specific address and City fields only appearing once the user has selected a State."

perhaps even a clearer example (as this seems a bit artificial): `For example, address/city/zip code fields appearing only once a user has checked the option to receive information by post` or similar?

understanding/22/visible-controls.html
```diff
@@ -63,6 +63,14 @@ <h3>When Needed</h3>
...
+         <li>An e-reader that allows the user to <b>select</b> some text and then pops up some buttons to allow the user to interact in different ways with that texts.</li>
```
patrickhlauke, line 70: the various `some` make this sound a bit handwavy. perhaps `allows the user to <b>select</b> a text passage and then pops up contextual toolbar with buttons that allow the user to interact...`

understanding/22/visible-controls.html
```diff
@@ -63,6 +63,14 @@ <h3>When Needed</h3>
...
+         <li>Controls that become visible upon scrolling, such as a button at the bottom of Terms and Conditions modal.</li>
```
patrickhlauke, line 71: perhaps clearer (assuming that's what it meant) `such as a confirmation button that only appears once the user has scrolled through the entirety of a Terms and Conditions modal dialog`

understanding/22/visible-controls.html
```diff
@@ -66,9 +66,9 @@ <h3>Embedded Controls</h3>
       
       <p>Many scenarios are out of scope because they do not rely on hover or focus, for example:</p>
       <ul>
-         <li>Options/fields become visible because of a <b>selection</b> elsewhere on the form. For example, filling out an address and the City field is not visible until the State is selected.</li>
-         <li>An e-reader that allows the user to <b>select</b> some text and then pops up some buttons to allow the user to interact in different ways with that texts.</li>
-         <li>Controls that become visible upon scrolling, such as a button at the bottom of Terms and Conditions modal.</li>
+         <li>Options/fields become visible because of a <b>selection</b> elsewhere on the form. For example, address and City fields only appearing once the user has selected a State.</li>
+         <li>An e-reader that allows the user to <b>select</b> a text passage and then pops up contextual toolbar with buttons that allows the user to interact with that text.</li>
```
patrickhlauke, line 70: `pop up contextual toolbar with` > `pop up a contextual toolbar with`

understanding/22/visible-controls.html
```diff
@@ -63,6 +63,14 @@ <h3>When Needed</h3>
...
+      
+      <p>Many scenarios are out of scope because they do not rely on hover or focus, for example:</p>
+      <ul>
+         <li>Options/fields become visible because of a <b>selection</b> elsewhere on the form. For example, address and City fields only appearing once the user has selected a State.</li>
+         <li>An e-reader that allows the user to <b>select</b> a text passage and then pops up contextual toolbar with buttons that allows the user to interact with that text.</li>
+         <li>Controls that become visible upon scrolling, such as a confirmation button that only appears once the user has scrolled through all the content in a Terms and Conditions dialog.</li>
```
patrickhlauke, line 71: re-reading this, perhaps swap out `once` with `when` (to avoid making it sound like it's anything to do with "only show this 1 time")

Issue #2050, opened 2021-09-22
What is a complex shape?


awkawk:
The definition for perimeter is:

> Perimeter
> continuous line forming the boundary of a shape not including shared pixels, or the minimum bounding box of a complex shape, whichever is shortest. For example, the perimeter calculation for a rectangle is 2h+2w -4, where h is the height and w is the width and the corners are not counted twice. The perimeter of a circle is 2𝜋r.

I suggest that we remove "complex". A circle isn't complex, but what about a triangle/pentagon/dodecahedron? We have no ability to define complex so should just remove this term. Alternatively, someone should come up with a definition for complex (good luck!).

alastc:
I think the idea was that, for HTML/CSS at least, anything not a rectangle is 'complex'. If something _is_ rectangular then the bounding box should be the same as the perimeter. 

I agree it's not needed here though.

PR #2051, opened 2021-09-22
Removing complexity

awkawk:
See #2050


patrickhlauke:
non-rectangular?

alastc:
Need to add an example of a circle focus indicator around a star.

PR #2057, opened 2021-09-25
2.5.3 - control with label but no accname (understanding note + new failure technique)

### LINKED ISSUE #2045: Does SC 2.5.3 fail when there is a visual text label but the control has no accessible name?
Would a control that has a visible text label but not an accessible name associated with the control fail SC 2.5.3?  I am aware that it would fail other criteria such SC 1.3.1 and 4.1.2.   I want to make sure that folks are consistently failing appropriate criteria.    

If it's determined that it does fail - should we then link some of the techniques for providing accessible names into the SC understanding document?

patrickhlauke:
Also cleans-up 2.5.3 understanding document markup/structure
- adds new F111 Control with visible label text but no accname (improperly numbered F109 initially)
- add failure technique in Understanding document
- added notes in Understanding about UI components with visible text and no name failing both 4.1.2 and 2.5.3
- removed direct reference to WCAG 2.1
- resolved discussion with ACT on where to report some of the failure examples

Closes #2045


mraccess77:
Thanks @patrickhlauke looks good to me.

detlevhfischer:
Looks good to me too.

jake-abma:
+1

patrickhlauke:
Any news on this @alastc ?

patrickhlauke:
Just resolved the merge conflict (caused by more recent failure techniques that were added in the time since this PR was first filed), so this is ready to merge again...

awkawk:
I don't think that checks 2 or 3 are needed for this.

patrickhlauke:
@awkawk you made me look a the test procedure again, and on reflection I actually modified the procedure to be more specific. have a look

techniques/failures/F111.html
```diff
@@ -0,0 +1,83 @@
...
+    
+		<section id="tests">
+			<h2>Tests</h2>
+			<section class="test-procedure">
+				<h3>Procedure</h3>
+				<p>For all controls with a visible label (e.g., link text, button text, programmatically linked label, images in links with text), check that:</p>
+				<ol>
+					<li>The control has a visible text label.</li>
+					<li>The control has an accessible name.</li>
+          <li>The accessible name contains a match for the string of the visible label.</li>
```
mbgower, line 57: Since "match" can mean a few different things, I made this more closely align to the normative text

```suggestion
          <li>The accessible name contains the text string that appears as the visible lablel.</li>
```

bruce-usab:
On deck for 1/26 call.

bruce-usab:
Preview link for [F111.html](https://raw.githack.com/w3c/wcag/patrickhlauke-issue2045/techniques/failures/F111.html)

understanding/21/label-in-name.html
```diff
@@ -29,7 +30,7 @@ <h3>Identifying label text for components</h3>
...
+<p>For the purposes of this SC, text should not be considered a visible label if it is used in a symbolic manner, rather than directly <q>expressing something in human language</q> as per the definition of <a>text</a> in WCAG. For example, <a href="images-of-text">1.4.5 Images of Text</a> describes considerations for "symbolic text characters." In the images of text example "B", "I", and "ABC" appear on icons in a text editor, where they are meant to symbolize the functions for Bold, Italics, and Spelling, respectively. In such a case, the accessible name should be the function the button serves (e.g., "Spell check" or "Check spelling"), not the visible symbolic characters. A similar text editor is shown in the figure below.</p>
```
dav-idc, line 33: Should the hash be in the anchor link still?

```suggestion
<p>For the purposes of this SC, text should not be considered a visible label if it is used in a symbolic manner, rather than directly <q>expressing something in human language</q> as per the definition of <a>text</a> in WCAG. For example, <a href="#images-of-text">1.4.5 Images of Text</a> describes considerations for "symbolic text characters." In the images of text example "B", "I", and "ABC" appear on icons in a text editor, where they are meant to symbolize the functions for Bold, Italics, and Spelling, respectively. In such a case, the accessible name should be the function the button serves (e.g., "Spell check" or "Check spelling"), not the visible symbolic characters. A similar text editor is shown in the figure below.</p>
```

bruce-usab: @alastc or @mbgower please note syntax question from David.

patrickhlauke: > Should the hash be in the anchor link still?

ah, good catch. i just checked other docs, and it seems the hash should not be there (separately, we should really try to just make a cheat-sheet of some of the more common XSLT magic in effect here)

techniques/failures/F111.html
```diff
@@ -0,0 +1,83 @@
...
+			<p>The objective of this Failure is to describe situations where speech input users cannot reliably speak the name of a control because although it has a visible label, it lacks an accessible name.</p>
+			<p>When speech input users interact with a web page, they usually speak a command followed by the reference to some visible label (like the text in a button or a link, or the text labelling some input). If the control lacks an accessible name, speech users won't be able to activate the control using its visible label.</p>	
+		</section>
+		
+		<section id="examples">
+			<h2>Examples</h2>
+      
+      <section class="example">
+				<h3>A text input with a visible label, but without an accessible name</h3>
+				<p>The text input is preceded by a visible text label "Enter name", but the text is not marked up as a <code>&lt;label&gt;</code> for the input, and there is no alternative way (e.g. <code>aria-label</code>) to provide the input with an accessible name at all.</p>
```
scottaohara, line 32:
```suggestion
				<p>The text input is preceded by a visible text label "Enter name", but the text is not marked up as a <code>&lt;label&gt;</code> for the input, and there is no alternative way (e.g., <code>aria-label</code>) to provide the input with an accessible name at all.</p>
```

techniques/failures/F111.html
```diff
@@ -0,0 +1,83 @@
...
+				<pre><code>
+					&lt;p&gt;Enter name&lt;/p&gt;
+					&lt;input type="text"&gt;
+        </code></pre>
+      </section>
+
+			<section class="example">
+				<h3>A text input with a visible label and aria-labelledby pointing to a non-existent id</h3>
+				<p>The text input is preceded by a visible text label "Enter name". The text's container has an <code>id</code> of <code>nameEntry</code>, but the input has an <code>aria-labelledby</code> referencing a non-existent <code>name-entry</code>. As a result, the input lacks an accessible name altogether.</p>
```
scottaohara, line 41:
```suggestion
				<p>The text input is preceded by a visible text label "Enter name". The text's container has an <code>id</code> of <code>nameEntry</code>, but the input has an <code>aria-labelledby</code> referencing a non-existent <code>name-entry</code> id. As a result, the input lacks an accessible name altogether.</p>
```

techniques/failures/F111.html
```diff
@@ -0,0 +1,83 @@
...
+					&lt;input type="text" aria-labelledby="name-entry" &gt;
+        </code></pre>
+      </section>
+    </section>
+    
+		<section id="tests">
+			<h2>Tests</h2>
+			<section class="test-procedure">
+				<h3>Procedure</h3>
+				<p>For all controls with a visible label (e.g., link text, button text, programmatically linked label, images in links with text), check that:</p>
```
scottaohara, line 53:
```suggestion
				<p>For all controls with a visible label (e.g., link text, button text, programmatically linked label, images in links or buttons with text, etc.), check that:</p>
```

scottaohara (approved):
marking as approve with my minor editorial updates

techniques/failures/F111.html
```diff
@@ -0,0 +1,83 @@
+<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml" lang="en">
+	<head>
+		<title>Failure of Success Criterion 2.5.3 due to a control with visible label text but no accessible name</title>
```
WilcoFiers, line 4: This creates an inconsistency with ACT. Our [Form field has non-empty accessible name](https://www.w3.org/WAI/standards-guidelines/act/rules/e086e5/) rule maps to only to 4.1.2. It does not allow failing form fields with empty accessible names under 2.5.3. I can appreciate the argument for putting it here, I'm not sure why it wouldn't primarily be a 4.1.2 issue.

@kengdoj You may want to have a look at this.

WilcoFiers (changes requested):

techniques/failures/F111.html
patrickhlauke: I don't think WCAG has a concept of something failing "primarily" one SC over another. Maybe this technique should be expanded to fail both 2.5.3 and 4.1.2?

WilcoFiers: Agreed, and I can definitely see an argument for doing both. I've asked Kathy to take it up in ACT.

alastc: This procedure is a sub-set of things that could fail 4.1.2. I.e. it requires that there is a visible label, and that the text is contained in the accname.
From the TF discussion: We don't think this mis-alignment should hold things up, it is ok to have slightly different failures for different SCs, but we'll give some time to think about this one.

scottaohara: would submit this is also an example of a 1.3.1 failure.  i'm not sure it matters if ACT doesn't log all the SCs that could be failed for a particular pattern, so long as it is consistent in what the act checks can find.

dbjorge (changes requested):
1. Missing an update to the `Failure Techniques` section of `label-in-name.html` to reference the new technique?
2. If we do retitle to note applicability to 1.3.1 and/or 4.1.2, consider also adding a ref to those understanding docs' `Failure Techniques` sections?

techniques/failures/F111.html
WilcoFiers: I'm currently on a hiatus from ACT. Please coordinate with them.

patrickhlauke:
@alastc question: should I also add the technique to 1.3.1 and 4.1.2? I held off it at the time, because I wasn't sure if doing so would cause issues if the 1.3.1/4.1.2 understanding docs are also still used for WCAG 2.0 and 2.1. if not, I'm happy to also add the technique to the respective understanding docs (and tidy them a tiny bit) in this PR

patrickhlauke:
@alastc so as you merged this, assume the answer to my question above is: no, don't add the failure technique to 1.3.1 and 4.1.2

PR #2066, opened 2021-10-01
Dragging rects and connector lines

### LINKED ISSUE #1560: Does the Definition of "Dragging Movements" Intentionally Limit the Scope to Moving Objects/Items?
The definition of "Dragging Movements" appears to exclude when a user defines a path of their own (not content-defined) by dragging, rather than moving an object by dragging. Is this intentional or an oversight? 

Path-based gestures, where the _content_ defines a path, are covered under WCAG 2.1 SC 1.5.1. 

And, until you read the definition of Dragging Movements, WCAG 2.2 SC 2.5.7 sounds like it covers any dragging movement, including:

- leaving breadcrumbs
- defining a path that items will follow later (as in a strategy game)
- white board drawings, where precise pixel-by-pixel line drawing is unnecessary
- press and hold and move through space to collect items in that space (as in harvesting a field)
- press and hold to draw a selection rectangle
- press and hold to apply color to a shape
- press and hold to connect representations of wires, where additional wire appears as you drag, rather than an animation that seems like the user is moving the wire

The definition does seem to cover cases where you press and hold to, for example, lead a character through a maze.

detlevhfischer:
Added an example for dragging movements showing selection rectangles or drawing connector lines to address #1560

Closes #1560


understanding/22/dragging-movements.html
```diff
@@ -45,6 +45,8 @@ <h2>Intent of Dragging Movements</h2>
 
       <p>Some tasks may typically involve dragging, such as sorting items in a list, but as there are other ways to achieve that task they are not considered essential.</p>
       
+      <p>A special case are dragging movements that invoke animations keyed to the pointer position, for example, selection rectangles that set the first x/y rectangle coordinate at the pointer position via a pointer down-event, and the second x/y coordinate, after a dragging movement, at the next up-event. A similar example is would be a connecting line drawn between two different items on the screen, as in an allocation test where users are required to draw a line between questions and corresponding answers. In these cases, the dragging movement would require an alternative way to accomplish the same action that does not rely on the dragging movement. For example, two separate single tap or click actions may define the rectangle coordinates or the start and end points of a connecting line.</p>
```
patrickhlauke, line 48: `A similar example is would be` > `A similar example would be`

detlevhfischer: Done.

Issue #2062, opened 2021-09-30
Understanding 1.4.11 - symbolic text characters


fstrr:
Unless I'm missing something in the Understanding document, I think that clarification is needed for the level of contrast for components that only contain what Label In Name describes as [symbolic text characters](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name.html#symbolic-text-characters). Taking the common example of a close button with only an X symbol in it, at an equivalent size of 16px:

1. `<button aria-label="close">X</button>` : does the letter X require a minimum contrast of 4.5:1 as it's text?
2. `<button aria-label="close"><svg aria-hidden="true" class="button-symbol">…</svg></button>`: does the SVG require a minimum contrast of 3:1 contrast as it's an image?

patrickhlauke:
the "X" in this case is not text per se ... it's not representing the "letter X", but is used as an icon, so yes, 1.4.11 applies and it only normatively needs to be 3:1 contrast minimum

JAWS-test:
I agree with @patrickhlauke, but think that an addition to Understanding in this regard would be useful

patrickhlauke:
Made a first naive stab at it ... #2067

PR #2067, opened 2021-10-01
Add section on symbolic text characters to 1.4.11 understanding

patrickhlauke:
Closes https://github.com/w3c/wcag/issues/2062


fstrr:
This is good—thanks. A couple of things:

1. I tried the image in VoiceOver and the alt text is announced as "…a button with a greater than / greater than character", which is a little confusing. I haven't tried other screen readers, so don't know if the result is the same.
2. Do we need the word "technically"? I think it would make it clearer without it.
3. I don't think we can say "non-text characters" as [Images Of Text](https://www.w3.org/WAI/WCAG21/Understanding/images-of-text.html#examples) says this for Symbolic Text Characters: "Some of the buttons use text characters that do not form a sequence that expresses something in human language."
4. If we call the X inappropriate then the > is as well, I think. What about this as a first sentence: "Even though the two buttons inappropriately use text characters — an uppercase X, for a "Close" button, and a > character, for a right-pointing arrow — they count as symbolic text characters that represent functionality."
5. This is the only content on the page that uses "SC" as an abbreviation for Success Criterion—it would be more consistent as Success Criterion (n.b, this page also can't decide if that should be Title Cased or not).

That was more than a couple of things. Sorry.

patrickhlauke:
Any news on this @alastc ?

bruce-usab (changes requested):
I prefer for @fstr edits to be incorporated.

alastc (approved):

understanding/21/non-text-contrast.html
```diff
@@ -388,6 +388,16 @@ <h4>Infographics</h4>
 				<p>An infographic can use text which meets the other criteria to minimise the number of graphical objects required for understanding. For example, using text with sufficient contrast to provide the values in a chart. A long description would also be sufficient because then the infograph is not relied upon for understanding.</p>
 
 				</section>
+				
+				<section>
+					<h4>Symbolic text characters</h4>
+					<p>For the purposes of this SC, when text is used in a symbolic manner, rather than directly <q>expressing something in human language</q> as per the definition of <a>text</a> in WCAG, it is considered <em>non-text content</em>.</p>
```
patrickhlauke, line 394:
```suggestion
					<p>For the purposes of this Success Criterion, when text is used in a symbolic manner, rather than directly <q>expressing something in human language</q> as per the definition of <a>text</a> in WCAG, it is considered <em>non-text content</em>.</p>
```

patrickhlauke:
@fstrr just changed "SC" to "Success Criterion" ... good catch

understanding/21/non-text-contrast.html
```diff
@@ -388,6 +388,16 @@ <h4>Infographics</h4>
 				<p>An infographic can use text which meets the other criteria to minimise the number of graphical objects required for understanding. For example, using text with sufficient contrast to provide the values in a chart. A long description would also be sufficient because then the infograph is not relied upon for understanding.</p>
 
 				</section>
+				
+				<section>
+					<h4>Symbolic text characters</h4>
+					<p>For the purposes of this Success Criterion, when text is used in a symbolic manner, rather than directly <q>expressing something in human language</q> as per the definition of <a>text</a> in WCAG, it is considered <em>non-text content</em>.</p>
+					<figure id="figure-text-symbols">
+						<img alt="A button using an uppercase 'X' and a button with a '>' / 'greater-than' character" src="img/buttons-text-symbols.png" />
+						<figcaption>Even though technically the two buttons use text characters — an uppercase <q>X</q>, often (inappropriately) used for "Close" buttons, and a <q>&gt;</q> character, to act as a right-pointing arrow — they count as non-text characters/symbols. Their contrast ratio of just above 3:1 passes this SC.</figcaption>
```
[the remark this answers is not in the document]
patrickhlauke: i might even remove "inappropriately" altogether. it's inappropriate if the control then doesn't have any additional accName, but that goes beyond the scope of what this SC wants to mention/talk about

understanding/21/non-text-contrast.html
```diff
@@ -388,6 +388,16 @@ <h4>Infographics</h4>
 				<p>An infographic can use text which meets the other criteria to minimise the number of graphical objects required for understanding. For example, using text with sufficient contrast to provide the values in a chart. A long description would also be sufficient because then the infograph is not relied upon for understanding.</p>
 
 				</section>
+				
+				<section>
+					<h4>Symbolic text characters</h4>
+					<p>For the purposes of this Success Criterion, when text is used in a symbolic manner, rather than directly <q>expressing something in human language</q> as per the definition of <a>text</a> in WCAG, it is considered <em>non-text content</em>.</p>
+					<figure id="figure-text-symbols">
+						<img alt="A button using an uppercase 'X' and a button with a greater-than character" src="img/buttons-text-symbols.png" />
+						<figcaption>Even though the two buttons inappropriately use text characters — an uppercase <q>X</q>, often used for "Close" buttons, and a <q>&gt;</q> character, to act as a right-pointing arrow — they count as non-text characters/symbols. Their contrast ratio of just above 3:1 passes this Success Criterion.</figcaption>
```
patrickhlauke, line 397:
```suggestion
						<figcaption>Even though the two buttons use text characters — an uppercase <q>X</q>, often used for "Close" buttons, and a <q>&gt;</q> character, to act as a right-pointing arrow — they count as non-text characters/symbols. Their contrast ratio of just above 3:1 passes this Success Criterion.</figcaption>
```

patrickhlauke:
@giacomo-petri suggested during our WCAG 2.x backlog call today to add a matching note under 1.4.3 (and presumably 1.4.6).

understanding/21/non-text-contrast.html
```diff
@@ -388,6 +388,16 @@ <h4>Infographics</h4>
 				<p>An infographic can use text which meets the other criteria to minimise the number of graphical objects required for understanding. For example, using text with sufficient contrast to provide the values in a chart. A long description would also be sufficient because then the infograph is not relied upon for understanding.</p>
 
 				</section>
+				
+				<section>
+					<h4>Symbolic text characters</h4>
+					<p>For the purposes of this Success Criterion, when text is used in a symbolic manner, rather than directly <q>expressing something in human language</q> as per the definition of <a>text</a> in WCAG, it is considered <em>non-text content</em>.</p>
```
patrickhlauke, line 394:
```suggestion
					<p>When text characters are used as symbols – used for their visual appearance, rather than  <q>expressing something in human language</q>– they fall under the definition of  <a>non-text content</a>.</p>
```

understanding/21/non-text-contrast.html
```diff
@@ -388,6 +388,16 @@ <h4>Infographics</h4>
 				<p>An infographic can use text which meets the other criteria to minimise the number of graphical objects required for understanding. For example, using text with sufficient contrast to provide the values in a chart. A long description would also be sufficient because then the infograph is not relied upon for understanding.</p>
 
 				</section>
+				
+				<section>
+					<h4>Symbolic text characters</h4>
+					<p>When text characters are used as symbols – used for their visual appearance, rather than <q>expressing something in human language</q> – they fall under the definition of  <a>non-text content</a>.</p>
```
patrickhlauke, line 394:
```suggestion
					<p>When text characters are used as symbols – used for their visual appearance, rather than <q>expressing something in human language</q> – they fall under the definition of <a>non-text content</a>.</p>
```

understanding/21/non-text-contrast.html
```diff
@@ -388,6 +388,16 @@ <h4>Infographics</h4>
 				<p>An infographic can use text which meets the other criteria to minimise the number of graphical objects required for understanding. For example, using text with sufficient contrast to provide the values in a chart. A long description would also be sufficient because then the infograph is not relied upon for understanding.</p>
 
 				</section>
+				
+				<section>
+					<h4>Symbolic text characters</h4>
+					<p>When text characters are used as symbols – used for their visual appearance, rather than <q>expressing something in human language</q> – they fall under the definition of <a>non-text content</a>.</p>
```
patrickhlauke, line 394: Turned this sentence around to make it more consistent. I'm assuming that since "non-text content" is a definition, the publishing script for WCAG will automagically turn this `<a>non-text content</a>` into the appropriate link to the `#dfn-non-text-content`

patrickhlauke:
> @giacomo-petri suggested during our WCAG 2.x backlog call today to add a matching note under 1.4.3 (and presumably 1.4.6).

started to look at this, but while it would be easy enough to add an extra paragraph to the intent of 1.4.3 and 1.4.6, it would then have only be shown for the 2.1 and 2.2 versions of the understanding pages, and this sort of conditional processing isn't possible with the way WCAG is currently published (though we should really look at this for the future, as this is quite limiting). Didn't want to go down the whole "From WCAG 2.1 onwards, there's also 1.4.11..." type cross-referencing.

So, while having an extra para/note in 1.4.3 and 1.4.6 would be nice, leaving this aspect for now.

bruce-usab (approved):
Thanks @patrickhlauke !
Agreed, your rephrase of the line is an improvement.

alastc (approved):

understanding/21/non-text-contrast.html
```diff
@@ -388,6 +388,16 @@ <h4>Infographics</h4>
...
+					<figure id="figure-text-symbols">
+						<img alt="A button using an uppercase 'X' and a button with a greater-than character" src="img/buttons-text-symbols.png" />
+						<figcaption>Even though the two buttons use text characters — an uppercase <q>X</q>, often used for "Close" buttons, and a <q>&gt;</q> character, to act as a right-pointing arrow — they count as non-text characters/symbols. Their contrast ratio of just above 3:1 passes this Success Criterion.</figcaption>
```
detlevhfischer, line 397: Don't want to complicate things - but is a greater-than character really something that "expresses something in human language"? Is it not a mathematical symbol? One possible way to edit this is to say: 

> "When text characters like "x" or "i", or mathematical symbols like ">", are used as symbolic images, that is, used for their visual appearance, rather than <q>expressing something in human language or in a mathematic formula</q> – they fall under the definition of <a>non-text content."

I can live with it as it is as well, though.

PR #2075, opened 2021-10-06
Shuffling target size files

### LINKED ISSUE #2074: 2.5.5 understanding doc has an empty example
It's not clear in the 2.2 understanding document if an example is missing or if some stray example markup needs to be deleted: https://www.w3.org/WAI/WCAG22/Understanding/target-size-enhanced.html#examples

alastc:
Closes #2074


michael-n-cooper:
@alastc I updated this branch from main so it could merge, which could have caused side effects. Right now this PR updates just understanding/21/target-size-enhanced.html, if that looks right this can merge.

PR #2081, opened 2021-10-12
Adding sentence on large text areas

### LINKED ISSUE #2017: Focus indicators for large editing areas
In #1930 @mbgower wrote:

> In the 4th bullet of the Typical focus indicators list ( ⅓ through Understanding), it addresses insertion points. It got me wondering about something - whether we should have language about text editors (think of code editors) where the insertion point is the ONLY form of focus indicator. Maybe such an interface isn’t a UIC? Should that be noted?

I wanted to separate this off into it's own issue.

I wonder if we need a similar concept to target-size where an interactive area (whether rich-text editor or colour selector) is not in itself a discrete control?

If I'm in a code editor like MS Code, an outline around the whole editing area is not particularly useful or desirable. However, for most text inputs, it is necessary.

Is there a line we can draw there? E.g. a rich-text or code editor isn't itself a UIC, but a combination of UICs.

alastc:
Closes #2017


understanding/22/focus-appearance-minimum.html
```diff
@@ -232,7 +232,7 @@ <h4>Unusual shapes and gradients</h4>
...
                 <p>If the contrast of background colors that change are close enough to need to be tested for each combination then they would likely not meet the user need of people with low vision in certain scroll combinations and would likely fail in certain combinations as well. In these cases it would be an easy solution to use a double ring focus indicator or some other mechanism to indicate focus such as a solid box with a border to guarantee there is sufficient contrast across variations of background images or background gradients.</p>
 
-                <p>Some components may be very large, such as editing areas in a Content Management System or code-editor that fill the screen. There maybe some circumstances where a large editing area may not be as useful, however, it may still be beneficial to some. There is no exception for this scenario.</p>
+                <p>Some components may be very large, such as editing areas in a Content Management System or code-editor that fill the screen. There maybe some circumstances where a large focus indicator may not be as useful, however, it may still be beneficial to some. There is no exception for this scenario.</p>
```
patrickhlauke, line 235: that `, however,` looks a bit awkward. perhaps `... may not be as useful. However, it may ...` or `... may not be as useful – however, it may ...`

mraccess77:
It sounds like we keep going back and forth on this whether it is or is not required.  The wording says it won't be beneficial to some but may be to others so it needs to meet the requirements.  That's fine with me - but it still feels somewhat squishy.

PR #2107, opened 2021-10-24
Add explicit mention of SPAs and dynamic changes of title to 2.4.2 understanding

### LINKED ISSUE #2102: Clarify in understanding SC 2.4.2 Page Title applicability to single page applications
The definition of web page as it's applied to a url that doesn't change in an single page application seems to imply that when the main content of an SPA changes the page title doesn't need to change, however, some people read it differently and would say that since the context changes the page title needs to be change because the title no longer describes the purpose of the url.

As we have discussed for 2.2 SPAs - some SPAs do have urls the change with hash tags. changes and then it's more clearly a requirement.

Some discussion was on WebAIM a few years ago on this topic.
https://webaim.org/discussion/mail_thread?thread=8564

So, I'm hoping that we can make it more clear if the context changes but the URL remains unchanged does a page title of a single page application need to change or not.      I think we probably agree it's ideal - but is it required?

patrickhlauke:
* this marries up nicely with the existing example already present talking about a web application for banking
* minor additional tweak: change `his` to `their` in the wording for the banking example

Closes #2102

### DIFF
--- a/understanding/20/page-titled.html
+++ b/understanding/20/page-titled.html
@@ -113,7 +113,7 @@
          <dt>A Web application</dt>
          <dd>A banking application lets users inspect their bank accounts, view past statements,
                and perform transactions. The Web application dynamically generates titles for each
-               Web page, e.g., "Bank XYZ, accounts for John Smith" "Bank XYZ, December 2005 statement
+               Web page, e.g., "Bank XYZ, accounts for Alex Smith" "Bank XYZ, December 2005 statement
                for Account 1234-5678".</dd>
       </dl>
 


patrickhlauke:
Any news on this @alastc ?

LaurenceRLewis:
I don't see the changes in relation to SPA in the Understanding document as described in the pull request ['Add explicit mention of SPAs and dynamic changes of title to 2.4.2 understanding #2107'](https://github.com/w3c/wcag/pull/2107/commits/f78ab83c1413706ad6ebbc45f13e71f4aa1753e2) 

"_In cases such as Single Page Applications (SPAs), where various distinct pages/views are all nominally served from the same URI and the content of the page is changed dynamically, the title of the page should also be changed dynamically to reflect the content/topic of the current view._"

Was this every implemented?

patrickhlauke:
> I don't see the changes [...] Was this every implemented?

@LaurenceRLewis as this pull request is still open, and not merged, i think there's your answer

hidde:
Just stumbled upon this PR and would very much support it to be merged in, very useful addition.

marie450:
This PR would resolved a lot of discussion between our redaction and accessibility teams as all of our traditional HTML documents are replace by SPA.

understanding/20/page-titled.html
```diff
@@ -36,6 +36,12 @@ <h2>Intent of Page Titled</h2>
          it is not required to use the name of the document or web application; other things
          may also describe the purpose or the topic of the page. 
       </p>
+
+      <p>In cases such as Single Page Applications (SPAs), where various distinct pages/views are
+         all nominally served from the same URI and the content of the page is changed dynamically,
+         the title of the page should also be changed dynamically to reflect the content/topic of
```
maryjom, line 42: Suggestion for readability: "content or topic" rather than "content/topic".

patrickhlauke, line 42:
```suggestion
         the title of the page should also be changed dynamically to reflect the content or topic of
```

understanding/20/page-titled.html
```diff
@@ -109,29 +109,14 @@ <h2>Examples of Page Titled</h2>
...
-               								       
-               <strong>A Web application.</strong>
-               							     
-            </p>
-            
-            <p>
-               A banking application lets a user inspect their bank accounts, view past statements,
+         </dd>
+         <dt>A Web application</dt>
+         <dd>A banking application lets a user inspect his bank accounts, view past statements,
```
mbgower:
```suggestion
         <dd>A banking application lets users inspect their bank accounts, view past statements,
```

Issue #1965, opened 2021-07-18
3.3.7 Further guidance on 2-factor independent methods 


abijames:
For some organisation, particularly in financial services and payments, authentication must be provided from 2 independent sources. For example, [Strong Customer Authentication apply](https://www.fca.org.uk/firms/strong-customer-authentication) in the UK and [Payment Services Regulations ](https://ec.europa.eu/info/business-economy-euro/banking-and-finance/consumer-finance-and-payments/payment-services/payment-services_en) in the EU apply to financial services and will soon apply to e-commerce sites. This requires authentication using 2 independent sources through a combination of two out the three categories - knowledge, possession and inherence (biometrics). Also, authentication may be required after login, for example to process a payment or edit sensitive data but the current understanding document is only referring to authentication at the login stage. Referring to these different types of authentication methods and scenarios within the understanding documentation will assist with wider adoption  as without clearer guidance payment services platforms are less likely to adopt the accessible authentication requirements as they must balance it against their local regulators requirements and their duty to protect customers from fraud.

I would like to propose the following edits to the understanding document

1. Change the first sentence to include scenarios where authentication takes place after log-in
".. and secure method to log in, access content and undertake some tasks."

2. Add the following paragraph to expand on different 2-factor authentication methods:

Many organizations are required to use 2-factor authentication that combines independent sources to confirm a user’s identity. These sources can consist of combining authentication through:      
- knowledge (e.g. password, letters in a passphrase or memorized swipe path);
- possession (e.g. through receipt of a one time password generated or received on a device, scanning of a QR code on an external device);</li>
- biometrics (e.g. fingerprint scanning, facial recognition or keystroke dynamics). 

Most knowledge based authentication methods rely on a cognitive function test and so mechanisms to assist users must be available. When authentication relies on performing an action on a separate device, it should be possible to complete the action on the same device to avoid the need to transcribe information. As it may not be possible to know what device-based authentication methods are available to a user, offering a choice of methods can allow them to choose the path that most suits them.

3. Add 2 further examples of accessible authentication when 2-factor authentication is required

<li>A website that requires 2-factor authentication displays a QR code which can be scanned by an app on a user’s device to confirm identity.</li>
         <li>A website that requires 2-factor authentication sends a notification to a user’s device. The user  must enter their device’s authentication mechanism (e.g. user defined PIN, fingerprint, facial recognition) to confirm identity.</li>

[...]

PR #2120, opened 2021-11-01
Adding suggestions from issue #1965

fstrr:


fstrr:
Adding suggestions from issue #1965

Issue #585, opened 2019-01-16
WCAG 2.1 Understanding 2.3.1 - missing/vague dimension definitions


patrickhlauke:
In https://www.w3.org/WAI/WCAG21/Understanding/three-flashes-or-below-threshold.html it mentions

> The 1024 x 768 screen is used as the reference screen resolution for the evaluation. The 341 x 256 pixel block represents a 10 degree viewport at a typical viewing distance.

This omits / handwaves one fairly important aspect: saying that a monitor is "1024 x 768" is mostly meaningless without any physical dimensions of the monitor itself. I can set a giant cinema display to run at "1024 x 768", or a super-small LCD screen. Unless a monitor size itself is defined, or the "typical viewing distance" is explained (would this "typical distance" be the one that results in the pixels matching up with the definition for the "reference pixel"?), this is very vague.

I seem to remember discussion ages ago that this SC/Understanding's wordings came from a time when 1024x768 was the norm/top-end, and "most" monitors were generally of a certain size (14" or so?) ... but now, this feels strangely anachronistic without any anchoring/rationale behind it.

alastc:
Hi Patrick,

So is the question about the original rational, or primarily about converting it to a more modern size?

The first might take some digging, but taking the rational of the time I assume we could convert that to a more modern definition.

E.g. the <a href="https://www.w3.org/TR/CSS21/syndata.html#x40">reference pixel</a> is about 0.0213 degrees at the intended viewing distance, so the "10 degree viewport at a typical viewing distance" is (10 / 0.0213 = 469) 469px. Hmm, might need to round that off/up.

The technique [G176](https://www.w3.org/WAI/WCAG21/Techniques/general/G176) goes into a lot more detail and should be considered at the same time, care to do a PR?

awkawk:
The resolution defined was because of the typical monitor resolution at the time. How about we change:

The 1024 x 768 screen is used as the reference screen resolution for the evaluation. The 341 x 256 pixel block represents a 10 degree viewport at a typical viewing distance.

to

If a 1024 x 768 screen on a desktop monitor is used as the reference screen resolution for an evaluation then a 341 x 256 pixel block represents a 10 degree viewport at a typical viewing distance.

Would this help?

alastc:
If it's updated to match CSS pixels, it appears we need to update the value as well, 341x256 is not 10 degrees according to the spec. The commenter in issue #553 came to the same conclusion.

patrickhlauke:
[...]

Issue #1132, opened 2020-05-24
2.3.2 Three Flashes (in the wild!) - or not


detlevhfischer:
Hi, I just put this here because I rarely come across pages with flashing content - today I did, and I re-read the SC to check if this page
https://www.interviewmagazine.com/culture/chris-kraus
would fail. 
The first question is the frequency of flashing: more than three flashes in any 1 second period probably refers to the appearance of the changed state, not any change between states? If the former is correct, then the flashing here is too slow to be a FAIL; if each state change is counted, it may be just within the second. 
As to the second condition (10 degrees visual field on screen) I have no clue how to test that reliably except for taking cover in the first note of the "general flash and red flash thresholds" glossary entry that says the approximation of a rect of 341 x 256 pixels is an approximation when content is "viewed at 1024x768 pixels" (what difference would viewport size make here, I wonder - is it not the zoom factor that would need mentioning).
The flashing image has 600 x 300px so I guess the size would be large enough for a FAIL, but the frequency arguably not.
Please take this as a Sunday pastime - I know there are more important things to discuss...

patrickhlauke:
based on https://www.w3.org/TR/WCAG21/#dfn-flashes a flash is s *pair* of changes, i.e. the change from b/w to w/b and back to b/w counts as *one* flash.

as for the second part...i think i've mentioned before how the whole "degree of visual field" wording there is incomprehensible, and outdated, since it doesn't take into account different monitor sizes, viewing distances, etc (or at least, it doesn't explain them properly). zoom will of course be a factor in the real world, but i believe zoom is out of scope for SCs (as they're down to user preference / a user's change). I'll note as well that PEAT https://trace.umd.edu/peat (to my knowledge still the only "proper" tool that's supposedly able to make determinations about whether or not flashes are acceptable or not) only works (at least last time i checked it) by resizing a browser window to 1024x768 and then doing its analysis...which of course completely misses testing on any other resolutions, and any other changes in breakpoints/layout. and yes, doesn't take viewing distance etc into account. in terms of viewing distance, the only sensible approach is to anchor everything on CSS ideal pixels, of course. but the definition/explanation would need to be freshened up to account for stuff...and we need tools that are better than PEAT (where you can at least test at different breakpoints)

patrickhlauke:
x-ref https://github.com/w3c/wcag/issues/553 and https://github.com/w3c/wcag/issues/585

patrickhlauke:
[...]

Issue #553, opened 2018-12-09
Problems with WCAG 2.0 Flash Definition


dd8:
I think there are a number of assumptions in the flash definition:
https://www.w3.org/TR/WCAG21/#dfn-general-flash-and-red-flash-thresholds

Some of these are no longer true due to the introduction of high pixel density displays and the updated definition of the `px` unit in CSS 3:

1) If the definition is meant in terms of device pixels then the note about 341 x 256 pixels doesn't work for high DPI retina displays.

2) If the definition is meant in terms of CSS pixels then the size calculations are wrong since the CSS spec defines the visual angle of a CSS pixel: https://www.w3.org/TR/css-values-3/#reference-pixel (10 visual degrees is 10 degrees / 0.0213 degrees per pixel = 469 CSS pixels)

3) Universally available browser zoom means that content falling under the sizes in ` 2.3.1 Three Flashes or Below Threshold` threshold can become large enough to cause seizure problems when zoomed. Does this mean that anything failing `2.3.2 Three Flashes` should fail 2.3.1 due to browser zoom?

patrickhlauke:
My personal interpretation here would be:

1. back in the days of WCAG 2.0 where this was defined, a physical/device pixel was the same as a CSS pixel...but nowadays, the definition should be interpreted as CSS pixel.

2. i'll trust your maths skills here - for my own part, the whole ".006 steradians within any 10 degree visual field" is incomprehensible and impenetrable (and the same holds true for any developer i've spoken to about this). i'd also note that the definition for the reference pixel is idealised, and real-world devices and user agents, though they often come close, don't really take this into account.

3. i'd say that browser zoom is a user choice/setting, and not something that you'd take into account when testing (similar to the fact that if a user goes closer to a screen, even assuming the browser/OS followed the actual idealised reference pixel dimension, that would also increase the size of any flashing area within their field of view).

dd8:
I think CSS pixel is probably the best/only option - and is backwards compatible with the 96-dpi pixel common when WCAG 2.0 was written. It's also defined as a visual angle, which ties in well with the rest of the definitions (or at least the parts I understand)

It's also not defined what the 'maximum relative luminance' is in this definition. It's probably the relative luminance of black vs white (i.e maximum contrast) but difficult to be certain given how opaque the rest of this definition is.

[...]

PR #2127, opened 2021-11-08
Update three-flashes-or-below-threshold.html

mbgower:
A first attempt at rewriting flashing assessment criteria

x-ref https://github.com/w3c/wcag/issues/1132

Closes https://github.com/w3c/wcag/issues/585


understanding/20/three-flashes-or-below-threshold.html
```diff
@@ -38,12 +38,13 @@ <h2>Intent of Three Flashes or Below Threshold</h2>
...
          viewing distance. (The 10 degree field is taken from the original specifications and
          represents the central vision portion of the eye, where people are most susceptible
          to photo stimuli.)
       </p>
+      <p>With the proliferation of mobile and devices of many sizes, as well as the adoption of <a href="https://www.w3.org/TR/WCAG21/#dfn-css-pixels">CSS pixels</a> as a density-independent unit of measurement, the prior assessment criteria seem outdated. However, an image of a consistent size uses up relatively the same percentage of a user's visual field on any device. On a large screen, the image takes up less size, but the large screen takes up a larger part of the visual field. On a mobile screen, the image may take up most or all of the screen; however, the mobile screen itself takes up a smaller portion of the user's visual field. So the same dimension of the flashing content, represented in CSS pixels can still provide a consistent means of assessment. Substituting CSS pixels for the original pixel block means that the assessment becomes 341 x 256 CSS pixels, or a flashing area of 87,296 CSS pixels.</p>
```
patrickhlauke, line 47: Nothing we can do about it, but perhaps worth mentioning that regardless of the ideal pixel font, what the specification itself (and content developers) can't account for is the actual viewing distance that a user chooses. Users that are closer to their screens than the idealised viewing distance will be affected even by flashing areas that normatively pass. The same problem also applies to users who rely on zoom/screen magnification. Conversely, users that are further away from the screen than the idealised viewing distance will tolerate flashing areas that are larger than the threshold.

patrickhlauke: Not sure "mobile and devices of many sizes" sounds right. Perhaps just "With the proliferation of devices with varying screen sizes (from small-screen smartphones to very large desktop and living room displays" or similar?

patrickhlauke: i do wonder if we could shave off that awkward looking single pixel and round it to 340 x 256 ... but i guess that would be a hard normative change too far

patrickhlauke:
while you're in that general area, might also be worth updating the links/resources that have relocated.

```
<a href="http://trace.wisc.edu/peat/">Trace Center Photosensitive Epilepsy Analysis Tool (PEAT)</a>
```

change to

```
<a href="https://trace.umd.edu/peat/">Trace Center Photosensitive Epilepsy Analysis Tool (PEAT)</a>
```

and likewise

```
<a href="http://trace.wisc.edu/peat/photosensitive.php">Information about Photosensitive Seizure Disorders</a>
```

change to

```
<a href="https://trace.umd.edu/information-about-photosensitive-seizure-disorders/">Information about Photosensitive Seizure Disorders</a>
```

I'd note that PEAT as a tool is also fairly useless nowadays in a world of responsive web design, as it effectively can only analyse content at a resolution of 1024x768 and does so by resizing your browser (where it can) to that size. might be worth noting somehow?

patrickhlauke:
ah, i see i'm flogging my own dead horse with my comment about PEAT ... i *may* have mentioned it once or twice before, sorry.

also x-ref https://github.com/w3c/wcag/issues/553 and https://github.com/w3c/wcag/issues/585

understanding/20/three-flashes-or-below-threshold.html
mbgower: latest commit addresses @patrickhlauke's comments

patrickhlauke:
i've done a separate PR for updating the links (as mentioned in https://github.com/w3c/wcag/pull/2127#issuecomment-963640893) https://github.com/w3c/wcag/pull/2132 (as that's a more straightforward change that is, likely, non-controversial)

alastc:
> i've done a separate PR for updating the links (as mentioned in [#2127 (comment)](https://github.com/w3c/wcag/pull/2127#issuecomment-963640893)) #2132 (as that's a more straightforward change that is, likely, non-controversial)

Your PR for that was merged in, so no need to replicate here.

understanding/20/three-flashes-or-below-threshold.html
```diff
@@ -38,12 +38,15 @@ <h2>Intent of Three Flashes or Below Threshold</h2>
...
          represents the central vision portion of the eye, where people are most susceptible
          to photo stimuli.)
       </p>
+      <p>With the proliferation of devices of varying screen sizes (from small hand-helds to large living room displays), as well as the adoption of <a href="https://www.w3.org/TR/WCAG21/#dfn-css-pixels">CSS pixels</a> as a density-independent unit of measurement, the prior assessment criteria seem outdated. However, an image of a consistent size uses up relatively the same percentage of a user's visual field on any device. On a large screen, the image takes up less size, but the large screen takes up a larger part of the visual field. On a mobile screen, the image may take up most or all of the screen; however, the mobile screen itself takes up a smaller portion of the user's visual field. So the same dimension of the flashing content, represented in CSS pixels can still provide a consistent means of assessment. Substituting CSS pixels for the original pixel block means that the assessment becomes 341 x 256 CSS pixels, or a flashing area of 87,296 CSS pixels.</p>
```
patrickhlauke, line 47: `the prior assessment criteria seem outdated` reads a bit weird, especially since the next sentence clarifies why they're not. maybe `the prior assessment criteria *may* seem outdated`

patrickhlauke: `the assessment becomes 341 x 256 CSS pixels` reads slightly off. maybe more accurately `the threshold for the combined area of flashing becomes ...`

patrickhlauke:
note that this still leaves questions open like https://github.com/w3c/wcag/issues/553#issuecomment-471900362 (about that relative luminance, and that `0.80`.

however, i do think this closes https://github.com/w3c/wcag/issues/585 for me, so i'll add the github incantation to the top comment here

understanding/20/three-flashes-or-below-threshold.html
```diff
@@ -38,12 +38,15 @@ <h2>Intent of Three Flashes or Below Threshold</h2>
          did not allow any flashing (even of a single pixel) within a broad frequency range
          (3 to 50 Hz). This Success Criterion is based on existing specifications in use in
          the UK and by others for television broadcast and has been adapted for computer display
-         viewing. The 1024 x 768 screen is used as the reference screen resolution for the
+         viewing. In WCAG 2.0, the 1024 x 768 screen was used as the reference screen resolution for the
```
awkawk, line 41: a 1024 x 768 screen

PR #2130, opened 2021-11-08
Update H99.html

mbgower:
Renamed technique, modified example styling and test procedure, and added a new example for a go-to-page input mechanism.

### DIFF
--- a/techniques/html/H99.html
+++ b/techniques/html/H99.html
@@ -1,11 +1,11 @@
 <!DOCTYPE html>
 <html xmlns="http://www.w3.org/1999/xhtml" lang="en">
 	<head>
-		<title>Provide page list navigation</title>
+		<title>Provide a page-selection mechanism</title>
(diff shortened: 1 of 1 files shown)


bruce-usab (commented):
Thanks for this technique!  I would like to add a couple more examples.

Also is <q>print equivalent of the edition</q> the best term?  Alternative paper version?

Current:
> These locations typically come from a print equivalent of the edition and allow synchronization in, for example, a mix of print and digital publications in environments such as schools and universities.

Proposed:

These page break locations typically come from a print edition and allow synchronization between versions.  Classrooms at schools and universities frequently have a mix of print and digital publications.  Third party materials, for example study guides and teacher aids, usually reference print page numbers.  Government and legal publications, for example the [U.S. Federal Register](https://www.federalregister.gov/documents/current), traditionally give precedence to numbering from a (facsimile but canonical) print edition,.

PR #2132, opened 2021-11-09
Update links for old trace.wisc.edu resources

patrickhlauke:
* point to new sources for PEAT and "Information about Photosensitive Seizure Disorders" on trace.umd.edu
* point to archive.org stored versions of the contrast ratio examples
* remove links to the "About Decibels" article

Note that this still leaves lots of links too trace.wisc.edu/wcag_wiki which I can't seem to locate a new location for


patrickhlauke:
copying in @GreggVan who might know if resources (like the "About Decibels" which I removed, and the `wcag_wiki` ones, which I've left but are 404ing) are on some new/better location

Issue #2136, opened 2021-11-15
H39 Using caption elements to associate data table captions with data tables needs a rewritten test


mbgower:
H39 is considered sufficient to meet an aspect of 1.3.1; however, the context needs to be better expressed, and the test rewritten to reflect the fact that a caption for a table is not needed. 

[1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships) states:
> Information, structure, and relationships conveyed through presentation can be programmatically determined or are available in text.

If text exists that seems to provide a name or title for a table, then 1.3.1 requires that there be programmatic indication and proper association for the caption.. That's where H39 comes from. 

But if there is no visible text name/title for the table, then there is nothing conveyed through presentation. Therefore 1.3.1 doesn't apply.

Rather than a 3-step test procedure, the test should be written as a conditional:

Existing
>For each data table:
Check that the table has content that is presented as a table caption.
Check that the table includes a caption element.
Check that the content of the caption element identifies the table.

Proposed
For each data table that has content that is presented as a table title or caption:
1. Check that the table includes a caption element.
2. Check that the content of the caption element identifies the table.

mbgower:
I created a pull request that makes these changes. I also adjusted the wording of the original third bullet in the procedure, because H39 does not cover the quality of the text used to title a table; only that the text be coded as a caption

awkawk:
@mbgower I don't think that there is anything wrong with H39 as written. IF you want to apply the technique, then all three checks must pass. If you don't want to apply the technique (possibly because the table is a layout table, possibly because the table doesn't need a caption) that is ok too, because those won't meet the technique (the same is true for the original or modified wording).

awkawk:
Also, how did you change the final bullet? It looks the same?

patrickhlauke:
to me #2137 makes sense, as it doesn't substantially change anything, but just adds a reminder of what the scope of the technique itself is (i.e. `The objective of this technique is to programmatically associate captions for data tables where captions are provided in the presentation.`) and gives it as a prerequisite to running the tests, rather than seemingly implying that all tables need that (a reader just skimming this technique may indeed get that impression)

mbgower:
I agree it is possible to read it that way. But I also have someone arguing that this procedure means you need to have a caption, period.

I don't want that to be inferred, and I also think the wording of the third bullet over-reached what this technique is for.

mbgower:
[...]

PR #2137, opened 2021-11-15
Update H39.html

mbgower:
Update to address https://github.com/w3c/wcag/issues/2136
Tighten the language of [H39 Using caption elements to associate data table captions with data tables](https://www.w3.org/WAI/WCAG22/Techniques/html/H39) to isolate it to only passing 1.3.1 Info and Relationships:
1. if there is a visible text title/caption for the table, then
2. make sure it is referenced so that the title/caption is programmaticalyy associated.


techniques/html/H39.html
```diff
@@ -27,16 +27,15 @@ <h3>An appointment calendar with a <code class="el">caption</code>
       </section>
    </section><section id="tests"><h2>Tests</h2>
       <section class="procedure"><h3>Procedure</h3>
-         <p>For each data table: </p>
+         <p>For each data table that has content that is presented as a title or identifier:</p>
```
patrickhlauke, line 30: maybe instead of `indentifier`, more like `description`? or perhaps having that in addition to `identifier`? just because i've seen folks use `<caption>` to provide something more like a summary/explanation for a table.

techniques/html/H39.html
```diff
@@ -27,16 +27,15 @@ <h3>An appointment calendar with a <code class="el">caption</code>
...
          <ol>
-           <li>Check that the table has content that is presented as a table caption.</li>
             <li>Check that the table includes a <code class="el">caption</code> element.</li>
-            <li>Check that the content of the <code class="el">caption</code> element identifies the table.</li>
+            <li>Check that the text that titles or identifies the table is included in the <code class="el">caption</code> element.</li>
```
patrickhlauke, line 33: depending on the above comment, this will then obviously also be affected

techniques/html/H39.html
patrickhlauke: ah, expanding this PR further (to remind myself of the actual untouched content) I see that the "title or identifier" wording is actually in the main prose. while it goes beyond the scope of this PR, it *might* be worth tweaking it there too

mbgower: pushed commit to address. I was a little leery of using it since I didn't want to get into confusion over the old table summary attribute, but  I agree 'description' is better.

patrickhlauke:
i'd reopen and make ve this forward

fstrr:
> i'd reopen and make ve this forward

I understand the first two words, but the rest have lost me :)

patrickhlauke:
> > i'd reopen and make ve this forward
> 
> 
> 
> I understand the first two words, but the rest have lost me :)

sorry...pool+phone+autocorrect don't make it easy ;)

techniques/html/H39.html
```diff
@@ -42,13 +42,13 @@ <h3>Procedure</h3>
         <ol>
 					<li>Check that the table has content that is presented as a table caption.</li>
```
patrickhlauke, line 43:
```suggestion
```

techniques/html/H39.html
```diff
@@ -40,15 +40,14 @@ <h2>Tests</h2>
...
-          <li>Check that the content of the <code class="language-html">caption</code> element identifies the table.</li>
-         </ol>
+          <li>Check that the text that titles or describes the table is included in the <code class="el">caption</code> element.</li>
+        </ol>
       </section>
       <section class="results">
          <h3>Expected Results</h3>
          <ul>
-					<li>#1, #2, and #3 are true.</li>
+           <li>#1, #2, and #3 are true.</li>
```
patrickhlauke, line 50:
```suggestion
           <li>#1 and #2 are true.</li>
```

patrickhlauke:
fixed it up to resolve the merge conflict

bruce-usab (approved):

bruce-usab:
closes #2136

techniques/html/H39.html
```diff
@@ -40,15 +40,14 @@ <h2>Tests</h2>
 				<h3>Procedure</h3>
         <p>For each data table:</p>
         <ol>
-					<li>Check that the table has content that is presented as a table caption.</li>
           <li>Check that the table includes a <code class="language-html">caption</code> element.</li>
-          <li>Check that the content of the <code class="language-html">caption</code> element identifies the table.</li>
-         </ol>
+          <li>Check that the text that titles or describes the table is included in the <code class="el">caption</code> element.</li>
```
fstrr, line 44:
```suggestion
          <li>Check that the text that titles or describes the table is included in the <code class="language-html">caption</code> element.</li>
```

techniques/html/H39.html
```diff
@@ -40,15 +40,14 @@ <h2>Tests</h2>
 				<h3>Procedure</h3>
         <p>For each data table:</p>
         <ol>
-					<li>Check that the table has content that is presented as a table caption.</li>
           <li>Check that the table includes a <code class="language-html">caption</code> element.</li>
```
detlevhfischer, line 43: Check if, rather than check that? Because the rest is condtional on there being a caption at all?

detlevhfischer (commented):
See inline comment.

techniques/html/H39.html
mbgower: Since this PR is not proposing to alter that phrasing in this PR, I don't think it's appropriate to tackle here. My suggestion is to add this as another element in https://github.com/w3c/wcag/pull/3915

alastc:
I'm not sure that this solves the mis-conception about whether captions are required, but I thought the top of the description (old and new) already did so. No harm from the change.

Could folks give this a thumbs up for it to get merged...

PR #2138, opened 2021-11-19
Moved some contents out of SC intent and added as additional benefit

### LINKED ISSUE #1881: WCAG 2.2 3.2.6 Consistent Help understanding - remove explanation of help mechanisms
The whole understanding document https://www.w3.org/WAI/WCAG22/Understanding/consistent-help.html includes a whole lengthy section explaining what the various help mechanisms should provide (and because those aren't the focus of the SC, they're couched in weak "should" statements - see https://github.com/w3c/wcag/issues/1874)

That's completely out of scope for the SC itself, so the understanding document - by focusing on those help systems, rather than the actual "meat" of the SC which is the consistent location of the link/access TO help - becomes very confusing and ambiguous for auditors who're just after information relating to "what do i need to do to pass/fail this SC?".

While interesting, these various explanations/descriptions of what different forms help can take etc are orthogonal to the SC - most of this should be removed, in my view.

> Self help methods beyond the site, such as using internet search to find the contact information for an organization, can be too difficult. Further, the user’s disability may make it more difficult to find the help available (such as a “contact us” link, phone number, or support page) if the information is not present within a few interactions (e.g., displayed in the header, or via a menu). In addition, for some users with disabilities, struggling to complete a task on a site may cause additional cognitive challenges when searching for help within the site.
>
>When a user is quickly able to find help, they are able to complete the task even if they encounter challenges.
>
>At least one of the following mechanisms to get help should be included on all websites:
>
>* Human contact details such as a phone number, email address, hours of operation.
>* Human contact mechanism such as a messaging system, chat client, contact form, social media channel.
>* Self-help option such as an up-to-date Frequently Asked Questions, How Do I page, Support page.
>* A fully automated contact mechanism such as a chatbot.
>
>While it would be best for all sites to offer at least one of these types of help, this SC does not require help be provided. The SC requires that when help is available that it be in a consistent location.
>
>The human contact details should enable the user to connect with the organization or the part of the organization that can assist with the content. For example, an online jobs / recruitment portal should provide a contact method for the team that supports the recruitment portal and not a catch-all for the entire company. Each layer of contact added prolongs the time before the user will receive help.
>
>The human contact mechanism enables a person to express what they are looking for using their own words. For some with cognitive disabilities, this may be the best way for them to find an answer to their problem.
>
>For pages for which no human support is available, a self-help option should say that no human support is available. Self-help options should go beyond allowing the user to search within the site. Contextual help is still recommended (See SC 3.3.5 for more information) but a self-help option should provide a single location that makes it easier for people with cognitive disabilities to understand what help is available without having to hunt for it. While some people may easily be able to identify that no support would be available for a particular type of Web site, this may not be apparent to some users with disabilities.
>
>If a chatbot is provided, it should meet other WCAG Success Criteria for your conformance level. Chatbots which work more effectively for everyone, and particularly for people with cognitive disabilities should:
>
>* recognize misspelled words,
>* provide human contact details if the chatbot is unable to provide a satisfactory response after 3 attempts, and
be dismissed with a single interaction, and recalled using a link or button.
```

### PR DESCRIPTION
**Pull request by a11ydoer**
[Preview here](https://raw.githack.com/w3c/wcag/issue1881-consistent-help/understanding/22/consistent-help.html).

Closes #1881

### CODE REVIEW

File: understanding/22/consistent-help.html [hunk 1]
```diff
@@ -54,6 +54,27 @@ <h2>Intent of Consistent Help</h2>
...
+         <li>Users that experience cognitive fatigue or cognitive shut down will be able to reserve their energy for the task, instead of using it to find support.</li>
+         <li>Enabling users (especially those with cognitive disabilities) to find solutions while expressing their question using their own words increases their chances of success for completing a task.</li>
+      </ul>
+
+      <h2>Additional Benefits of Consistent Help</h2>
         <p>Self help methods beyond the site, such as using internet search to find the contact information for an organization, can be too difficult. Further, the user’s disability may make it more difficult to find the help available (such as a “contact us” link, phone number, or support page) if the information is not present within a few interactions (e.g., displayed in the header, or via a menu). In addition, for some users with disabilities, struggling to complete a task on a site may cause additional cognitive challenges when searching for help within the site.</p>
 
         <p>When a user is quickly able to find help, they are able to complete the task even if they encounter challenges.</p>
```
patrickhlauke, line 80: Could I suggest that after this paragraph, an extra heading be added, along the lines of

```
<h2>Help Mechanisms</h2>
```

and, on the following paragraph, changing the first sentence from

```
At least one of the following mechanisms to get help should be included on all websites:
```

to

```
Typical help mechanisms include:
```

(noting that the para after the list includes mention that it would be best for sites to include at least one)

understanding/22/consistent-help.html
```diff
@@ -54,6 +54,27 @@ <h2>Intent of Consistent Help</h2>
...
+      <h2>Benefits of Consistent Help</h2>
+
+      <ul>
+         <li>This Success Criterion helps people who may have difficulty locating help that is not part of the page they are using. Ensuring they can connect with help allows them to complete the task.</li>
+         <li>When a user can easily find help for their specific questions, they are more successful completing tasks.</li>
+         <li>Users that experience cognitive fatigue or cognitive shut down will be able to reserve their energy for the task, instead of using it to find support.</li>
+         <li>Enabling users (especially those with cognitive disabilities) to find solutions while expressing their question using their own words increases their chances of success for completing a task.</li>
+      </ul>
+
+      <h2>Additional Benefits of Consistent Help</h2>
```
patrickhlauke, line 77: I would suggest removing this heading here, as the following still counts as "benefits" so can remain under the same heading. Then, adding the heading later on (per my other comment).

understanding/22/consistent-help.html
patrickhlauke: This would then also eliminate the contentious "should" word.

understanding/22/consistent-help.html
```diff
@@ -54,11 +54,34 @@ <h2>Intent of Consistent Help</h2>
 
         <p>Without help, some users may abandon the task. They may also fail to correctly complete a task, or they may require assistance from people who do not necessarily keep private information secure.</p>
 
+        <p>It is not the intent of this Success Criterion to require authors to provide help information on PDFs or other static documents that may be available for viewing/download from the Web pages. It is also not the intent to require contact information if:</p>
```
patrickhlauke, line 57: Since the `While it would be best for all sites to offer at least one of these types of help, this SC does not require help be provided...` is now quite far away and in a separate section, it may be good here to have, before this paragraph about PDF/static docs, a very clear similarly structure paragraph first that says this as part of the actual intent:

`<p>It is not the intent of this Success Criterion to require authors to provide help / access to help. The Criterion only requires that when one of the listed forms of help is available across multiple pages that it be in a consistent location.</p>`

and then slim down the "While it would be best..." paragraph further down - see next comment

understanding/22/consistent-help.html
```diff
@@ -67,40 +90,20 @@ <h2>Intent of Consistent Help</h2>
          </ul>
          <p>While it would be best for all sites to offer at least one of these types of help, this SC does not require help be provided. Also, the order of the types of help listed in the SC does not imply priority. The SC requires that when one of the forms of help listed is available that it be in a consistent location.</p>
```
patrickhlauke, line 91: following on from my previous comment about moving the sense of this directly into the intent section, this could then be slimmed down to just

`The order of the types of help listed in the SC does not imply priority.`

understanding/22/consistent-help.html
```diff
@@ -54,11 +54,11 @@ <h2>Intent of Consistent Help</h2>
 
         <p>Without help, some users may abandon the task. They may also fail to correctly complete a task, or they may require assistance from people who do not necessarily keep private information secure.</p>
 
-        <p>It is not the intent of this Success Criterion to require authors to provide help information on PDFs or other static documents that may be available for viewing/download from the Web pages. It is also not the intent to require contact information if:</p>
+      <p>It is not the intent of this Success Criterion to require authors to provide help / access to help. The Criterion only requires that when one of the listed forms of help is available across multiple pages that it be in a consistent location. It does not authors to provide help information on PDFs or other static documents that may be available for viewing/download from the Web pages. It is also not the intent to require contact information if:</p>
```
patrickhlauke, line 61: `It does not authors to provide` > `It does not require authors to provide`

patrickhlauke: ah, got in there too late with this last edit suggestion

patrickhlauke: filed separately https://github.com/w3c/wcag/pull/2158

PR #2143, opened 2021-11-29
update terminology: color blind

shawna-slh:
https://github.com/w3c/low-vision-a11y-tf/discussions/121#discussioncomment-1549295
https://github.com/w3c/low-vision-a11y-tf/discussions/121#discussioncomment-1667201


kfranqueiro:
Checking on this old PR which targets the `wcag20` folder which currently doesn't affect builds; if we want to make this change, we should probably also update the same wording in `understanding/20/use-of-color.html`.

bruce-usab:
Discussed and had consensus on backlog call 10/25.

bruce-usab (approved):

kfranqueiro:
RE my previous question: discussed with backlog TF on 2025-02-14, and there was agreement to remove the old change under `/wcag20/` to avoid confusion and noise. I have updated the PR accordingly so it only applies the equivalent update under `/understanding/`.

PR #2144, opened 2021-11-30
Adjust to proposed change.

### LINKED ISSUE #1693: Findable help (3.2.6) demands contact links within PDF's for instance?
Currently this SC demands that all pages have at least 1 option to access the information in the same relative order.

1. I think this needs to be "all pages within the set of web pages".
2. The current definition of "set of web pages" and "web page" causes for example PDF documents to be counted among web pages within that set of web pages (because they are most likely made by the same organization with the same purpose as the website).

That means that every PDF now for example needs a link to the contact page. I don't think that is what this SC is trying to accomplish.

alastc:
Closes #1693

### DIFF
--- a/guidelines/sc/22/consistent-help.html
+++ b/guidelines/sc/22/consistent-help.html
@@ -5,7 +5,7 @@
     <p class="conformance-level">A</p>
 	<p class="change">New</p>
     
-    <p>For each <a>web page</a> within a <a>set of web pages</a> that provides one or more of the following ways of finding help, access to at least one form of help is included in the same relative order on each page:</p>
+    <p>For each <a>web page</a> within a <a>set of web pages</a> that provides one or more of following ways of finding help somewhere within that set, access to at least one form of help is included in the same relative order on each page within the set of web pages if it is included on that page:</p>
 
     <ul>
         <li>Human contact details;</li>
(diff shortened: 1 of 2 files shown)


guidelines/sc/22/consistent-help.html
```diff
@@ -5,7 +5,7 @@ <h4>Consistent Help</h4>
     <p class="conformance-level">A</p>
 	<p class="change">New</p>
     
-    <p>For each <a>web page</a> within a <a>set of web pages</a> that provides one or more of the following ways of finding help, access to at least one form of help is included in the same relative order on each page:</p>
+    <p>For each <a>web page</a> within a <a>set of web pages</a> that provides one or more of following ways of finding help somewhere within that set, access to at least one form of help is included in the same relative order on each page within the set of web pages if it is included on that page:</p>
```
patrickhlauke, line 8: `one or more of following ways` sounds off. i'd keep the original `one or more of the following ways`

patrickhlauke:
I think overall this single sentence is still trying way too hard to convey at least 3 separate concepts:

* two or more pages within a set of web pages can provide access to help - not necessarily all of them
* pages may offer access to one or more types of help
* at least one method of accessing help that is common across those pages that do offer it must be in the same relative order (i.e. if there's two links two two separate types of help, only one needs to be in the same relative order across the two or more pages)

that's quite a lot of dense concepts (and a lot of conditionals) to try and put into one single SC, let alone in a single sentence...

PR #2145, opened 2021-11-30
Updated SC text

### LINKED ISSUE #2098: 3.2.6 Consistent Help wording is perhaps confusing, understanding uses "should" which may make it sound normative
The current normative wording for the SC

> For each web page within a set of web pages that provides one or more of the following ways of finding help, access to at least one form of help is included in the same relative order on each page:

may be slightly confusing, because on skim-reading it it appears to say that pages need to include at least one of these forms of help. Of course, careful re-reading shows that it's scoped only to pages that do have one or more of those forms of help, but perhaps the wording can be tweaked/turned around, if it's not too late?

though admittedly it's not completely non-trivial to come up with a better wording...maybe if the list can be - awkwardly - put in the middle like so?

> For each web page within a set of web pages that provides one or more of the following ways to find help
>
> * Human contact details;
> * Human contact mechanism;
> * Self-help option;
> * A fully automated contact mechanism
> 
> then access to at least one of these is included in the same relative order on each page

---

As originally noted in https://github.com/w3c/wcag/pull/1834 the understanding document for 3.2.6 also has a few instances of using the words `should` - while it's perhaps clear that it's not the normative `SHOULD`, and the understanding docs themselves are only informative, it may give the impression that it's nonetheless making normative claims/statements.

> At least one of the following ways to get help **should** be included on all websites:

it may be that the "While it would be best..." sentence could be moved before the bullet list, and combined with the spirit of this to basically say (heavily paraphrased): normatively this SC doesn't specify the types of help, but as a best practice, these are the sorts of things that would be good to have...

Further instances:

> The human contact details **should** enable the user to connect with the organization or the part of the organization that can assist with the content. For example, an online jobs / recruitment portal **should** provide a contact method for the team that supports the recruitment portal and not a catch-all for the entire company. Each layer of contact added prolongs the time before the user will receive help.

> Self-help options **should** go beyond allowing the user to search within the site. Contextual help is still recommended (See SC 3.3.5 for more information) but a self-help option **should** provide a single location that makes it easier for people with cognitive disabilities to understand what help is available without having to hunt for it.

> If a chatbot is provided, it **should** meet other WCAG Success Criteria for your conformance level. Chatbots which work more effectively for everyone, and particularly for people with cognitive disabilities **should**:

> If the human contact is not available during certain hours or certain days, information **should** be provided so the user can tell when it will be available.

alastc:
Closes #2098

### DIFF
--- a/guidelines/sc/22/consistent-help.html
+++ b/guidelines/sc/22/consistent-help.html
@@ -5,7 +5,7 @@
     <p class="conformance-level">A</p>
 	<p class="change">New</p>
     
-    <p>Any of the following help mechanisms that are repeated on multiple <a>web pages</a> within a <a>set of web pages</a> occur in the same relative order to other page content, unless a change is initiated by the user:</p>
+    <p>If a <a>web page</a> contains any of the following help mechanisms, and those mechanisms are repeated on multiple web pages within a <a>set of web pages</a>, they occur in the same relative order to other page content, unless a change is initiated by the user:</p>
 
     <ul>
         <li>Human contact details;</li>
--- a/guidelines/sc/22/consistent-help.html
+++ b/guidelines/sc/22/consistent-help.html
@@ -5,7 +5,7 @@
     <p class="conformance-level">A</p>
 	<p class="change">New</p>
     
-    <p>Any of the following help mechanisms that are repeated on multiple <a>web page</a> within a <a>set of web pages</a> occur in the same relative order to other page content, unless a change is initiated by the user:</p>
+    <p>Any of the following help mechanisms that are repeated on multiple <a>web pages</a> within a <a>set of web pages</a> occur in the same relative order to other page content, unless a change is initiated by the user:</p>
 
     <ul>
         <li>Human contact details;</li>
--- a/guidelines/sc/22/consistent-help.html
+++ b/guidelines/sc/22/consistent-help.html
@@ -5,7 +5,7 @@
     <p class="conformance-level">A</p>
 	<p class="change">New</p>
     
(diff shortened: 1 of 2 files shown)


guidelines/sc/22/consistent-help.html
```diff
@@ -5,7 +5,7 @@ <h4>Consistent Help</h4>
     <p class="conformance-level">A</p>
 	<p class="change">New</p>
     
-    <p>For each <a>web page</a> within a <a>set of web pages</a> that provides one or more of the following ways of finding help, access to at least one form of help is included in the same relative order on each page:</p>
+    <p>Any of the following help mechanisms that are repeated on multiple <a>web page</a> within a <a>set of web pages</a> occur in the same relative order to other page content, unless a change is initiated by the user:</p>
```
patrickhlauke, line 8: `multiple <a>web page</a>` should probably be `multiple <a>web pages</a>`

understanding/22/consistent-help.html
```diff
@@ -21,15 +21,17 @@ <h2>Consistent Help</h2>
       <blockquote>
          <p class="conformance-level">A</p>
 
-         <p>For each <a>web page</a> within a <a>set of web pages</a> that provides one or more of the following ways of finding help, access to at least one form of help is included in the same relative order on each page:</p>
-            <ul>
+         <p>Any of the following help mechanisms that are repeated on multiple <a>web page</a> within a <a>set of web pages</a> occur in the <a>same relative order</a> to other page content, unless a change is initiated by the user:</p>
```
patrickhlauke, line 24: same as above `multiple <a>web page</a>`

PR #2160, opened 2021-12-07
Michael's page-break def

### LINKED ISSUE #2155: Consider revision to the definition of Page break locators
Suggest changing the definition https://w3c.github.io/wcag/guidelines/22/#dfn-pagebreak-locators
to remove confusion on the use of the word "page", since it has a specific meaning in WCAG not intended here

programmatically determinable destination markers that are arranged in a meaningful sequence to represent a locator serving the same purpose as page breaks in a printed document

alastc:
Closes #2155

### DIFF
--- a/guidelines/terms/22/page-break-locators.html
+++ b/guidelines/terms/22/page-break-locators.html
@@ -2,7 +2,7 @@
 <dd class="new">
 	<p class="change">New</p>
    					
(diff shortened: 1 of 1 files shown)


bruce-usab (approved):
I think <q>printed</q> is aspirational.   Maybe <q>formatted for printing</q>?

PR #2182, opened 2022-01-12
Implementing Andrew's suggested update

alastc:
Note - didn't include the "focus and unfocused states", as we'd recently agreed not to use the term state.

### DIFF
--- a/guidelines/sc/22/focus-appearance-minimum.html
+++ b/guidelines/sc/22/focus-appearance-minimum.html
@@ -5,7 +5,7 @@
     <p class="conformance-level">AA</p>
 	<p class="change">New</p>
     
-    <p>When <a>user interface components</a> receive keyboard focus, the focus indicator meets the following:</p>
+    <p>When <a>user interface components</a> receive keyboard focus, the focus indicator meets all the following:</p>
     <ul>
         <li><strong>Contrasting area:</strong> The focus indicator has an area with a contrast ratio of at least 3:1 between the colors when the component is focused and when it is not focused, and that area is at least as large as either:
         <ul>
--- a/guidelines/sc/22/focus-appearance-minimum.html
+++ b/guidelines/sc/22/focus-appearance-minimum.html
@@ -5,15 +5,14 @@
     <p class="conformance-level">AA</p>
 	<p class="change">New</p>
     
-    <p>When <a>user interface components</a> receive keyboard focus, an area of the focus indicator meets the following:</p>
+    <p>When <a>user interface components</a> receive keyboard focus, the focus indicator meets the following:</p>
     <ul>
-        <li><strong>Minimum area:</strong> The area is either:
+        <li><strong>Contrasting area:</strong> The focus indicator has an area with a contrast ratio of at least 3:1 between the colors when the component is focused and it is not focused, and that area is at least as large as either:
         <ul>
-            <li>at least as large as the area of a 1 <a>CSS pixel</a> thick <a>perimeter</a> of the unfocused component, or</li>
-            <li>at least as large as the area of a 4 CSS pixel thick line along the shortest side of a <a>minimum bounding box</a> of the unfocused component, and no part of the area is thinner than 2 CSS pixels.</li>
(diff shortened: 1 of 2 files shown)


guidelines/sc/22/focus-appearance-minimum.html
```diff
@@ -5,15 +5,14 @@ <h4>Focus Appearance (Minimum)</h4>
     <p class="conformance-level">AA</p>
 	<p class="change">New</p>
     
-    <p>When <a>user interface components</a> receive keyboard focus, an area of the focus indicator meets the following:</p>
+    <p>When <a>user interface components</a> receive keyboard focus, the focus indicator meets the following:</p>
     <ul>
-        <li><strong>Minimum area:</strong> The area is either:
+        <li><strong>Contrasting area:</strong> The focus indicator has an area with a contrast ratio of at least 3:1 between the colors when the component is focused and it is not focused, and that area is at least as large as either:
```
patrickhlauke, line 10: > when the component is focused and it is not focused

maybe me, but i'd add an extra "when" there

> when the component is focused and when it is not focused

patrickhlauke: > and that area

perhaps better

> and its area

understanding/22/focus-appearance-minimum.html
```diff
@@ -20,15 +20,14 @@ <h2>Focus Appearance (Minimum) Success Criterion text</h2>
         <blockquote>
             <p class="conformance-level">AA</p>
         
-            <p>When <a>user interface components</a> receive keyboard focus, an area of the focus indicator meets the following:</p>
+            <p>When <a>user interface components</a> receive keyboard focus, the focus indicator meets the following:</p>
             <ul>
-                <li><strong>Minimum area:</strong> The area is either:
+                <li><strong>Contrasting area:</strong> The focus indicator has an area with a contrast ratio of at least 3:1 between the colors when the component is focused and it is not focused, and that area is at least as large as either:
```
patrickhlauke, line 25: as above

guidelines/sc/22/focus-appearance-minimum.html
awkawk: +1/-1 means I can go either way, but I prefer "that" here

patrickhlauke: "that area" to me implies that you're defining the "focus indicator" as an area.

"its area" treats "focus indicator" as more than just an area, and the area as one of its properties. if that makes sense...

alastc: "the area"?

patrickhlauke: "the area of the indicator"

alastc: That wouldn't work, the area in scope can be smaller than the entire focus indicator (e.g. with a gradient fade-out). I think maybe it's best as 'that area', because it scopes that at the start of the sentence.

bruce-usab (approved):
Oops, missed an edit.

bruce-usab (changes requested):
I still think it would be better for:
>     the item with focus is not entirely hidden by author-created content

To be at the top, before the list requirements pertaining to focus indicator.

patrickhlauke (approved):

bruce-usab:
Thanks for the ping @alastc but the extra condition is still following the bullet list of other requirements.  This would be new/different/bad.  Three alternatives for this bit <q>when an item receives focus, it is not entirely hidden by author-created content</q>
1. move up and into opening condition
2. make it a bullet
3. somehow phrase as an exception (and leave it where it is)

alastc:
> Thanks for the ping @alastc but the extra condition is still following the bullet list of other requirements. 

Hi Bruce,

I don't think those options work because the scope is different. The bullets are the requirements for the **focus-indicator**. The scope of that additional item is that the **element** itself is not hidden, not the indicator.

mbgower:
I am closing this since it has been addressed with language in subsequent merged PRs. The current wording is at https://w3c.github.io/wcag/guidelines/22/#focus-appearance-minimum. if you disagree, please re-open.

PR #2184, opened 2022-01-13
Fix typos in Understanding WCAG 2.2

damiansiana11y:
Line 98: delete extra “the”
Techniques specific to WCAG 2.2 are added to the the techniques for WCAG 2.0,

Should be: 
Techniques specific to WCAG 2.2 are added to the  techniques for WCAG 2.0

Line 54: Delete space
Understanding WCAG 2.2 is a  guide to 

Should be: 
Understanding WCAG 2.2 is a guide to

### DIFF
--- a/understanding/index.html
+++ b/understanding/index.html
@@ -51,7 +51,7 @@
 		</section>
 		<section class="informative introductory" id="intro">
 			<h2>Introduction</h2>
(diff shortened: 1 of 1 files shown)


alastc:
Hi @damiansiana11y, thanks for this. 
I've merged the others, but in this case I think the section you edited has moved somewhere else (toc.html?) 

I'll close this one.

Also, we generally don't worry about extra spaces, they aren't rendered in HTML.

PR #2194, opened 2022-01-21
Update understanding on 1.4.4

rachaelbradley:
Update the understanding document to clarify that zoom alone can meet this SC.

### DIFF
--- a/understanding/20/resize-text.html
+++ b/understanding/20/resize-text.html
@@ -59,7 +59,7 @@
       </p>
       
       <p>Content satisfies the Success Criterion if it can be scaled up to 200%, that is, up
-         to twice the width and height. Authors may support scaling beyond that limit, however,
+         to twice the width and height. Content that can be scaled up to 200% using user agent zoom meets this Success Criteria, even if the text alone cannot be scaled up to 200%.  Authors may support scaling beyond that limit, however,
          as scaling becomes more extreme, adaptive layouts may introduce usability problems.
          For example, words may be too wide to fit into the horizontal space available to them,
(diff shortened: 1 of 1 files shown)


understanding/20/resize-text.html
```diff
@@ -59,7 +59,7 @@ <h2>Intent of Resize text</h2>
       </p>
       
       <p>Content satisfies the Success Criterion if it can be scaled up to 200%, that is, up
-         to twice the width and height. Authors may support scaling beyond that limit, however,
+         to twice the width and height. Content that can be scaled up to 200% using user agent zoom meets this Success Criteria, even if the text alone cannot be scaled up to 200%.  Authors may support scaling beyond that limit, however,
```
patrickhlauke, line 62: "Success Criteria" > "Success Criterion"

detlevhfischer, line 62: I am not sure is "even if the text alone cannot be scaled up to 200%." will be understood correctly. What is meant is "even if the text only-scaling function of a user agent alone cannot scale up text to 200%".
The way it is written may leave people to think that if they use browser zoom and the text is simulataneously scaled down via CSS at narrower breakpoints as they zoom, this would meet 1.4.4

detlevhfischer (changes requested):

alastc:
@detlevhfischer - Apologies for any "dismissed review" notification you received, I couldn't work out how else to allow merging. I have implemented your suggestion.

patrickhlauke:
I believe this PR has actually been superseded by the changes that were made in https://github.com/w3c/wcag/pull/2270 - closing, unless you feel this is still needed (in which case, feel free to reopen)

PR #2207, opened 2022-02-03
clarifying name, role, value

### LINKED ISSUE #2142: does a control that exposes an incorrect/inappropriate role fail 4.1.2 or not
A recent discussion bubbled this one up again...take the classic example of a link that acts as a button (triggers in-page functionality), or something like a select that for some reason is exposed as a button, or similar - basically, controls that expose a role, but one that does not actually reflect reality of what the controls do.

```
<a href="#" onclick="...">Trigger</a>
```

```
<select ... role="button">
  <option ...> ... </option>
  <option ...> ... </option>
  <option ...> ... </option>
</select>
```

Are these failures of 4.1.2? My take so far has been yes. However, I've now heard arguments that 4.1.2 only says that controls need to expose "a" role, and not necessarily the correct one... (ditto for attributes/properties)

The understanding for 4.1.2 is very generic here and doesn't quite commit to anything here https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html

(and yes, ironically ages ago in a different discussion, I have been making the point that for the `name` part of this SC, it does indeed not specify that the name needs to be correct or not, because there are more specific other SCs that deal with appropriateness of labels https://github.com/w3c/wcag/issues/929 ... but in this case, for the `role`, it makes sense to me that there's an implicit "correct role" interpretation here - same as for `value`, `state`, etc ... as otherwise what's the point if something that has exposes things completely wrongly passes?)

scottaohara:
closes #2142

In an attempt to add clarity that it's not just that a name, role and value are exposed but the **correct** name role and value are exposed, I have simply added the word "accurate" to two sentences.  This will help imply that it's not just "information" that needs to be exposed to AT, but "accurate information" about what the control represents.

If more modifications are necessary here, please let me know and I can make those changes, or the editors can make changes as they see fit.

### DIFF
--- a/understanding/20/name-role-value.html
+++ b/understanding/20/name-role-value.html
@@ -14,18 +14,18 @@
       
       
       <p>The intent of this Success Criterion is to ensure that Assistive Technologies (AT)
-         can gather information about, activate(or set) and keep up to date on the status of
+         can gather accurate information about, activate (or set) and keep up to date on the status of
          user interface controls in the content.
       </p>
       
       <p>When standard controls from accessible technologies are used, this process is straightforward.
          If the user interface elements are used according to specification the conditions
-         of this provision will be met. (See examples  of Success Criterion 4.1.2 below)
+         of this provision will be met. (See examples of Success Criterion 4.1.2 below)
       </p>
       
       <p>If custom controls are created, however, or interface elements are programmed (in
          code or script) to have a different role and/or function than usual, then additional
-         measures need to be taken to ensure that the controls provide important information
+         measures need to be taken to ensure that the controls provide important and accurate information
          to assistive technologies and allow themselves to be controlled by assistive technologies.
       </p>
       
@@ -35,7 +35,7 @@
          of user interface control state are whether or not a checkbox or radio button has
          been selected, or whether or not a collapsible tree or list node is expanded or collapsed.
       </p>
-      
+     
       <div class="note">
          
          <p>Success Criterion 4.1.2 requires a programmatically determinable name for all user


understanding/20/name-role-value.html
```diff
@@ -14,18 +14,18 @@ <h2>Intent of Name, Role, Value</h2>
       
       
       <p>The intent of this Success Criterion is to ensure that Assistive Technologies (AT)
-         can gather information about, activate(or set) and keep up to date on the status of
+         can gather accurate information about, activate (or set) and keep up to date on the status of
```
patrickhlauke, line 17: suggest changing `accurate` to `appropriate`, as even "accurate" implies there's always an unambiguously correct/proper/accurate one

understanding/20/name-role-value.html
```diff
@@ -14,18 +14,18 @@ <h2>Intent of Name, Role, Value</h2>
...
       <p>When standard controls from accessible technologies are used, this process is straightforward.
          If the user interface elements are used according to specification the conditions
-         of this provision will be met. (See examples  of Success Criterion 4.1.2 below)
+         of this provision will be met. (See examples of Success Criterion 4.1.2 below)
       </p>
       
       <p>If custom controls are created, however, or interface elements are programmed (in
          code or script) to have a different role and/or function than usual, then additional
-         measures need to be taken to ensure that the controls provide important information
+         measures need to be taken to ensure that the controls provide important and accurate information
```
patrickhlauke, line 28: ditto

bruce-usab (approved):

bruce-usab (commented):
Would <q>correct</q> be better than <q>appropriate</q>?

patrickhlauke:
"correct" implies that there's a binary "right/wrong". "appropriate" allows for there to be a certain level of ambiguity, a sliding scale of appropriateness. the latter better matches reality

bruce-usab:
> "correct" implies that there's a binary "right/wrong". "appropriate" allows for there to be a certain level of ambiguity

Agreed.  I am okay with appropriate.  I would prefer <q>not incorrect</q> but that is almost a double negative.

scottaohara:
@bruce-usab @alastc could this be added to the next survey, or for the 2.x sub-group to determine if it's good to merge?

patrickhlauke (approved):

understanding/20/name-role-value.html
```diff
@@ -14,33 +14,40 @@ <h2>Intent of Name, Role, Value</h2>
...
       </p>
       
       <p>If custom controls are created, however, or interface elements are programmed (in
          code or script) to have a different role and/or function than usual, then additional
-         measures need to be taken to ensure that the controls provide important information
+         measures need to be taken to ensure that the controls provide important and appropriate information
          to assistive technologies and allow themselves to be controlled by assistive technologies.
       </p>
+     
+      <p>What roles and states are appropriate to convey to assistive technology will be determined
```
patrickhlauke, line 32:
```suggestion
      <p>What roles and states are appropriate to convey to assistive technology will depend
```

understanding/20/name-role-value.html
```diff
@@ -14,33 +14,40 @@ <h2>Intent of Name, Role, Value</h2>
...
+         by what the control represents. Specifics about such information are defined by other
```
patrickhlauke, line 33:
```suggestion
         on what the control represents. Specifics about such information are defined by other
```

(goes hand in hand with the previous suggestion)

understanding/20/name-role-value.html
```diff
@@ -14,33 +14,40 @@ <h2>Intent of Name, Role, Value</h2>
...
          been selected, or whether or not a collapsible tree or list node is expanded or collapsed.
       </p>
-      
+     
       <div class="note">
          
          <p>Success Criterion 4.1.2 requires a programmatically determinable name for all user
-            interface components. Names may be visible or invisible. Occasionally, the name must
-            be visible, in which case it is identified as a label. Refer to the definition of
+            interface components. Names may be visible or invisible. Occasionally, the name needs
```
patrickhlauke, line 49: I'd love to see some about the fact that, normatively, the name doesn't necessarily need to even human-understandable. As for visible/invisible, maybe something a bit more generic about how there is a difference between label and name - that often they may be the same, but not necessarily, and then reference both the glossary definitions AND maybe hint at other SCs like label in name.

patrickhlauke (changes requested):
A few tiny suggested tweaks/additions - see code comments

understanding/20/name-role-value.html
awkawk: What do you mean that names don't need to be human-understandable? The name in 4.1.2 is "text by which software can identify a component within Web content to the user"

understanding/20/name-role-value.html
```diff
@@ -14,33 +14,40 @@ <h2>Intent of Name, Role, Value</h2>
...
+            to be visible, in which case it is identified as a label. Refer to the definition of
```
awkawk, line 50: perhaps "in which case it is identified as a label as well as a name." ?

understanding/20/name-role-value.html
patrickhlauke: I mean there's no actual requirement in the normative text that that name needs to reflect what the control does, that it's appropriate, understandable. the name could be "xx2" or some nonsense...that is then what the sofware uses to identify the component to the user. there's no qualitative statement about the name in the SC.

awkawk: Sure, but that part is in 1.1.1, right?

patrickhlauke: not necessarily, no - it may fail other criteria of course (as you imply, if it's an image, but then has an accName that is nonsensical like "xx2", it would fail 1.1.1 ... or if it has visible sensible text like "Open the pod bay doors" but has an overriding `aria-label` accName of "xx2" then it would fail 2.5.3 Label in Name). but strictly for 4.1.2, the accName can be anything, doesn't have to be human-understandable (it can pass 4.1.2, but then fail other more specific SCs)

scottaohara:
@patrickhlauke per our discussion today, can you either:
1 update the PR with the changes you think are needed
2 approve this PR and then we'll make a follow-on

understanding/20/name-role-value.html
patrickhlauke: Resolving this here for this PR, will open a separate issue/PR for this so as not to hold up this specific one here

understanding/20/name-role-value.html
patrickhlauke: resolving this for now, will capture this in a separate issue/PR

patrickhlauke (approved):

bruce-usab (approved):
Also looks good to me!  Thank you @scottaohara and @patrickhlauke

understanding/20/name-role-value.html
```diff
@@ -14,33 +14,40 @@ <h2>Intent of Name, Role, Value</h2>
...
          code or script) to have a different role and/or function than usual, then additional
-         measures need to be taken to ensure that the controls provide important information
+         measures need to be taken to ensure that the controls provide important and appropriate information
          to assistive technologies and allow themselves to be controlled by assistive technologies.
       </p>
+     
+      <p>What roles and states are appropriate to convey to assistive technology will depend
+         on what the control represents. Specifics about such information are defined by other
+         specifications, such as <a href="https://www.w3.org/TR/wai-aria/">WAI-ARIA</a>, or the
+         appropriate platform standards. Another factor to consider is whether there is sufficient
```
alastc, line 35:
```suggestion
         relevent platform standards. Another factor to consider is whether there is sufficient
```

mbgower, line 35:
```suggestion
         relevant platform standards. Another factor to consider is whether there is sufficient
```

scottaohara:
last thing to do here is to add a link to the definition of a11y supported

Issue #2226, opened 2022-02-12
Focus Appearance and what counts as the "real" size of an element/component


patrickhlauke:
Splitting this off from https://github.com/w3c/wcag/issues/2222#issuecomment-1035099989, based on @alastc's answer to my question https://github.com/w3c/wcag/issues/2222#issuecomment-1036230167

I would strongly urge to reconsider just focusing on the element's raw CSS dimensions, but rather what their apparent dimensions are.

See the expanded example here https://codepen.io/patrickhlauke/pen/xxPdQOd - it makes no sense that one would fail, while the other one would pass, and all that simply because of how they're coded when visually they're exactly the same. there's a risk here that we create another SC that we then have to preface to clients with "yes, it's not logical, and i really can't explain its actual real-world practical benefit...but it is what it is and you have to follow the wording of the SC even if it makes no sense"

patrickhlauke:
In my mind, the SC should focus on what the apparent/perceived perimeter/size of the control is, and not rummage around in the underlying markup/CSS to determine what the actual element's sizing is. you're opening up a pandora's box if you start to then also take into account things like clipping paths, and SVG shapes, etc, which will simply lead to instances of "it normatively fails, but only because of the way you built this control behind the scenes, and even though the way you built it is perfectly valid"

JAWS-test:
I agree with you, but I think we should clarify what is the perimeter of an object that has no frame. For example, a link that consists only of text. Is the perimeter 
* then the irregular shape of the letters
* the smallest rectangle in which all the letters fit in
* the size of the click area
* the size of the focus frame
* something else

JAWS-test:
I think the simplest thing would be to always take the visual size of the element as the scale, regardless of click area or other invisible sizes of the element. And then it would be best that, for irregular objects, the smallest rectangle that can be placed around the visible object (and is not rotated) is used to determine the size. Everything else can become very complex: Of course, with a circle I can take the circumference of the circle, but what about a link that consists of many letters, or just one letter, or a font icon: do I then have to take the circumference as well? Calculating this quickly becomes too much complex (Latin letters are already difficult, but what about Chinese calligraphy?). To be able to test the criterion sensibly, I would suggest drawing a rectangle around the visible object.
At most, the question would be where a visible object begins if the border of the object itself consists of a colour gradient, such as a shadow. I would suggest that all visible pixels, regardless of contrast, should then be taken into account.

patrickhlauke:
[...]

PR #2230, opened 2022-02-14
Focus-appearance - Idea for bounding the content

alastc:
[Preview](https://raw.githack.com/w3c/wcag/focus-appearance-content-term/understanding/22/focus-appearance-minimum.html) (if rawgit is working).

### DIFF
--- a/understanding/22/focus-appearance-minimum.html
+++ b/understanding/22/focus-appearance-minimum.html
@@ -20,20 +20,20 @@
         <blockquote>
             <p class="conformance-level">AA</p>
         
-            <p>When a keyboard operable element has <a>focus</a>, the focus indicator meets all the following:</p>
+            <p>When a <a>user interface component</a> has keyboard focus, the focus indicator meets all the following:</p>
             <ul>
-                <li><strong>Contrasting area:</strong> The focus indicator has an area with a contrast ratio of at least 3:1 between the colors when the element is focused and when it is not focused, and that area is at least as large as either:
+                <li><strong>Contrasting area:</strong> The focus indicator has an area with a contrast ratio of at least 3:1 between the colors when the component is focused and when it is not focused, and that area is at least as large as either:
                 <ul>
-                    <li>the area of a 1 <a>CSS pixel</a> thick <a>minimum bounding box</a> around the content of the element, or</li>
-                    <li>the area of a 4 CSS pixel thick line along the shortest side of a <a>minimum bounding box</a> of the content of the element, and no part of the area is thinner than 2 CSS pixels.</li>
+                    <li>the area of a 1 <a>CSS pixel</a> thick <a>minimum bounding box</a> around the content of the component, or</li>
+                    <li>the area of a 4 CSS pixel thick line along the shortest side of a <a>minimum bounding box</a> of the content of the component, and no part of the area is thinner than 2 CSS pixels.</li>
                 </ul></li>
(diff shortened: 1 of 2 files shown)


patrickhlauke:
On the face of it, looks good, but based on our recent discussions (element CSS border-box vs visual vs "what a user may think is a reasonable representation of the control") the minimum bounding box definition may need further tweaking/expanding/nuance

alastc:
Yep, this is basically a backup idea if the ones in the survey / comments don't work.

The [min-bounding box definition](https://www.w3.org/TR/WCAG22/#dfn-minimum-bounding-box) probably needs some clarification that the 'content' could be text or non-text, and doesn't include things like backgrounds / borders / box-shadows.

patrickhlauke:
> doesn't include things like backgrounds / borders / box-shadows

even box-shadows are a funny one. while yes, some sites use CSS box shadows to effectively make an outline, other sites might use a very subtle, smudged, drop-shadow-y shadow. it's arguably if a real-world user when asked would say that the extent of the drop shadow is also part of the component/element, or if it's "just a shadow behind it".

detlevhfischer:
This could work. I guess with all word-smithing we will not get out of real-world situations where measurement will be difficult (think of icons as content of buttons that are somewhat larger). I wonder whether there good ways to calculate bounding boxes around content, especially if you effectively need a 3:1 cut-off rule for visual components that include gradients / differently colored areas.

patrickhlauke:
> I wonder whether there are established ways of automatically calculating bounding boxes around content

i fear that any automated way of doing it would make certain assumptions that then can lead to false positives/negatives. and it will again depend on what exactly the bounding box is calculated automatically on (the actual thing that receives focus, or some deeper nested container in that ... as thrashed out in far too much back-and-forth in #2226 )

alastc:
> think of icons as content of buttons that are somewhat larger

I think what we should stress in the understanding document is that, despite the exact measure, the best way of meeting it is an outline. It makes it easier to calculate that an outline passes because there shouldn't be any niche cases there. 

It is only if you shrink beneath the content of the control or use an odd shape you'd be calculating it.

> I wonder whether there are established ways of automatically calculating bounding boxes around content

Possibly, but you'd probably have to take a warning-only approach rather than definite fail. Conversely, it would be easy to find certain types of pass, e.g. 2px outline that contrasts with the unfocused state.

understanding/22/focus-appearance-minimum.html
```diff
@@ -20,20 +20,20 @@ <h2>Focus Appearance (Minimum) Success Criterion text</h2>
...
         
-            <p>When <a>user interface components</a> receive keyboard focus, an area of the focus indicator meets the following:</p>
+            <p>When a <a>user interface component</a> has keyboard focus, the focus indicator meets all the following:</p>
             <ul>
-                <li><strong>Minimum area:</strong> The area is either:
+                <li><strong>Contrasting area:</strong> The focus indicator has an area with a contrast ratio of at least 3:1 between the colors when the component is focused and when it is not focused, and that area is at least as large as either:
                 <ul>
-                    <li>at least as large as the area of a 1 <a>CSS pixel</a> thick <a>perimeter</a> of the unfocused component, or</li>
-                    <li>at least as large as the area of a 4 CSS pixel thick line along the shortest side of a <a>minimum bounding box</a> of the unfocused component, and no part of the area is thinner than 2 CSS pixels.</li>
+                    <li>the area of a 1 <a>CSS pixel</a> thick <a>minimum bounding box</a> around the content of the component, or</li>
```
36degrees, line 27: Minor thing – perimeter is still a defined term but no longer appears in the success criteria.

PR #2234, opened 2022-02-17
Add technique and understanding content for text indicator

### LINKED ISSUE #1888: Adobe Comment on 3.2.7 Visible Controls
We have significant concerns about this proposed SC. As written, this will be difficult or impossible to implement for many authoring tools where content in a design canvas has dozens (or thousands) of editable objects. Adding a visible indicator to each object on a design canvas would likely make the content far less understandable to users, and may make the user interface overwhelming to some users.  To address this, we recommend adding an exception for User Interface components within the design area for a web page that is being used to author content. 

We also feel that it is very important to highlight the note more by making it normative text to indicate that UI Components that are made available via progressive disclosure are not in scope. 

These issues must be addressed in order for Adobe to support inclusion of this SC.

rachaelbradley:
This PR, if approved, closes #1888

### DIFF
--- a/techniques/general/hidden-control-3.html
+++ b/techniques/general/hidden-control-3.html
@@ -1,0 +1,63 @@
+<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml" lang="en">
+	<head>
+		<title>Add text to a control stating all content is editable.</title>
+		<link rel="stylesheet" type="text/css" href="../../css/editors.css"/>
+	</head>
+	<body>
+		<h1>Add text to a control stating all content is editable.</h1>
+		<section id="meta">
+			<h2>Metadata</h2>
+			<p id="sc">3.2.6 Hidden Controls</p>
+			<p id="type">Sufficient</p>
+		</section>
+		<section id="applicability">
+			<h2>Applicability</h2>
+			<p class="instructions">Any technology</p>
+		</section>
+		<section id="description">
+			<h2>Description</h2>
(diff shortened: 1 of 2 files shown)


techniques/general/hidden-control-3.html
```diff
@@ -0,0 +1,63 @@
...
+		<section id="applicability">
+			<h2>Applicability</h2>
+			<p class="instructions">Any technology</p>
+		</section>
+		<section id="description">
+			<h2>Description</h2>
+			<p>The objective of this technique is to provide the information needed to identify a control when all contents of a larger component are editable. This primarily applies to a large design canvans with multiple editable components with it.</p>
+			<p>To meet this technique, add text that:
+			<ul><li> is programatically and visually adjacent to the component, and</li>
+			<li> states that all content is editable could serve as the indicator. </li></p>
```
gradualclearing, line 23: Delete "could see as the indicator"

techniques/general/hidden-control-3.html
```diff
@@ -17,10 +17,10 @@ <h2>Applicability</h2>
 		</section>
 		<section id="description">
 			<h2>Description</h2>
-			<p>The objective of this technique is to provide the information needed to identify a control when all contents of a larger component are editable. This primarily applies to a large design canvans with multiple editable components with it.</p>
+			<p>The objective of this technique is to provide the information needed to identify a control when all contents of a larger component are editable. This primarily applies to a large design canvas with multiple editable components with it.</p>
 			<p>To meet this technique, add text that:
 			<ul><li> is programatically and visually adjacent to the component, and</li>
```
patrickhlauke, line 22: while not directly related to this change, while you're here, might be worth removing the redundant extra space at the start/end of the `<li>`s

Issue #2243, opened 2022-03-02
G183 Example 2 doesn't pass the G183 test procedure


dd8:
Example 2 describes text that has a 3.9:1 contrast with the surrounding text. It does not mention focus or hover style differences.

> The hypertext links in a document are medium-light blue (#3366CC) and the regular text is black (#000000). Because the blue text is light enough, it has a contrast of 3.9:1 with the surrounding text and can be identified as being different than the surrounding text by people with all types of color blindness, including those individuals who cannot see color at all.

https://www.w3.org/WAI/WCAG21/Techniques/general/G183#example-2

This appears to fail steps 3 and 4 of the test procedure:

1) Locate all instances where color alone is used to convey information about text.
2) Check that the [relative luminance](https://www.w3.org/TR/WCAG21/#) of the color of the text differs from the relative luminance of the surrounding text by a contrast ratio of at least 3:1.
3) Check that pointing (mouseover) to the link causes a visual enhancement (such as an underline, font change, etc.)
4) Check that moving keyboard focus to the link causes a visual enhancement (such as an underline, font change, etc.)

Expected Results
1) Checks 2, 3, and 4 are all true.

Related: https://github.com/w3c/wcag/issues/201

patrickhlauke:
https://github.com/w3c/wcag/pull/2244

PR #2244, opened 2022-03-02
Technique G183: expand example 2 to actually illustrate the technique's point

patrickhlauke:
Closes https://github.com/w3c/wcag/issues/2243

Note that, in addition, the links to the glossary definition `<a>relative luminance</a>` are broken / not correctly converted https://www.w3.org/WAI/WCAG21/Techniques/general/G183 - they just end up pointing to https://www.w3.org/TR/WCAG21/#
However, I suspect this is an issue for @michael-n-cooper to look into (rather than trying to hardcode a url to the correct term definition https://www.w3.org/TR/WCAG21/#dfn-relative-luminance)

### DIFF
--- a/techniques/general/G183.html
+++ b/techniques/general/G183.html
@@ -24,7 +24,7 @@
       </section>
       <section class="example">
          
-            <p>The hypertext links in a document are medium-light blue (#3366CC) and the regular text is black (#000000). Because the blue text is light enough, it has a contrast of 3.9:1 with the surrounding text and can be identified as being different than the surrounding text by people with all types of color blindness, including those individuals who cannot see color at all.</p>
+            <p>The hypertext links in a document are medium-light blue (<code>#3366CC</code>) and the regular text is black (<code>#000000</code>). Beyond the difference in color, the links have no other styles differences compared with the regular text. Because the blue text is light enough, it has a contrast of 3.9:1 with the surrounding text and can be identified as being different from the surrounding text by people with all types of color blindness, including those individuals who cannot see color at all. In addition, the links have <code>:focus</code> and <code>:hover</code> styles that reintroduce the underline when the links receive keyboard focus or when a mouse pointer hovers over them.</p>
          
       </section>
    </section><section id="tests"><h2>Tests</h2>
--- a/techniques/general/G183.html
+++ b/techniques/general/G183.html
@@ -24,7 +24,7 @@
       </section>
       <section class="example">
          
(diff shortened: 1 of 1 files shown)


dd8 (commented):
That fixes issue 2243 (https://github.com/w3c/wcag/issues/2243). Thanks!

dd8 (approved):
The change fixes https://github.com/w3c/wcag/issues/2243

michael-n-cooper:
> However, I suspect this is an issue for @michael-n-cooper to look into

#2240 will address the location of the relative luminance file

patrickhlauke:
Confirming that the problem with the relative link I mentioned seems to be fixed in https://www.w3.org/WAI/WCAG22/Techniques/general/G183

techniques/general/G183.html
```diff
@@ -24,7 +24,7 @@ <h3>Colors that would provide 3:1 contrast with black words and 4.5:1 contrast w
       </section>
       <section class="example">
          
-            <p>The hypertext links in a document are medium-light blue (#3366CC) and the regular text is black (#000000). Because the blue text is light enough, it has a contrast of 3.9:1 with the surrounding text and can be identified as being different than the surrounding text by people with all types of color blindness, including those individuals who cannot see color at all.</p>
+            <p>The hypertext links in a document are medium-light blue (<code>#3366CC</code>) and the regular text is black (<code>#000000</code>). Beyond the difference in color, the links have no other styles differences compared with the regular text. Because the blue text is light enough, it has a contrast of 3.9:1 with the surrounding text and can be identified as being different from the surrounding text by people with all types of color blindness, including those individuals who cannot see color at all. In addition, the links have <code>:focus</code> and <code>:hover</code> styles that reintroduce the underline when the links receive keyboard focus or when a mouse pointer hovers over them.</p>
```
alastc, line 27:
```suggestion
            <p>The hypertext links in a document are medium-light blue (<code>#3366CC</code>) and the regular text is black (<code>#000000</code>). Beyond the difference in color, the links have no other styles differences compared with the regular text. Because the blue text is light enough, it has a contrast of 3.9:1 with the surrounding text and can be identified as being different from the surrounding text by people with all types of color blindness, including those individuals who cannot see color at all. In addition to the contrast difference, the links have <code>:focus</code> and <code>:hover</code> styles that reintroduce the underline when the links receive keyboard focus or when a mouse pointer hovers over them.</p>
```

bruce-usab (approved):

PR #2260, opened 2022-03-14
Improve technique references

### LINKED ISSUE #1585: What does “Sufficient as a way to meet an unwritten technique” mean ?
What does _“Sufficient as a way to meet an unwritten technique”_ mean when it appears beside a listed Success Criterion under the **Applicability** section of a WCAG Technique page?  For example, in [Technique H69 Providing heading elements at the beginning of each section of content](https://www.w3.org/WAI/WCAG21/Techniques/html/H69.html), under **Applicability** it says, _“Success Criterion 2.4.1: Bypass Blocks (Sufficient as a way to meet an unwritten technique).”_  Does this mean that simply providing headings for sections of content that are repeated across multiple web pages is sufficient for meeting [2.4.1 Bypass Blocks](https://www.w3.org/WAI/WCAG21/Understanding/bypass-blocks)?  This seems problematic, since, although this may work for screen reader users, keyboard users would not be able to skip these sections unless they had installed a browser extension.  I have the same issue with [Technique ARIA11 Using landmarks to identify regions of a page](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA11).

michael-n-cooper:
Fixes #1585

### DIFF
--- a/xslt/generate-techniques.xslt
+++ b/xslt/generate-techniques.xslt
@@ -64,9 +64,10 @@
 	</xsl:template>
 	
 	<xsl:template name="technique-link">
-		<xsl:param name="technique" select="."/>
+		<xsl:param name="technique"/>
 		<xsl:choose>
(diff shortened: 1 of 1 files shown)


alastc (approved):
I don't have a local build setup to test this, but if you have then please go ahead.

Issue #2256, opened 2022-03-10
1.4.8: Visual Presentation - line spacing (leading) clarification


patrickhlauke:
One of the points of 1.4.8 Visual Presentation states:

> * Line spacing (leading) is at least space-and-a-half within paragraphs ...
In the understanding document https://w3c.github.io/wcag/understanding/visual-presentation.html it expands this further to

> By space and a half within paragraphs we mean that top of one line is 150% further from the top of the line below it than would be true when the text is 'single spaced' (the default spacing for the font). ...

Now, in CSS, the closest property to influence the spacing of lines is `line-height`. Line height is the height of the font itself, plus half-leading at the top and bottom.

The *default spacing for the font*, when you just don't touch `line-height`, or have it set to `line-height: normal`, will vary for each font - it will be whatever the user agent deems to be the ideal default line-height (top and bottom half-leading) based on the font metrics (supposedly anyway - see https://www.w3.org/TR/css-inline-3/#line-height-property which is still in draft).

Anecdotally, from some testing, for the standard Times New Roman in Windows (at least on my machine), Chrome gives makes `line-height: normal` about equivalent to `line-height: 1.14`. So, when 1.4.8 asks for 150% further space, does this mean that for that particular font, an author must define a minimum `line-height: 1.71` (`1.14 * 1.5 = 1.71`)? And for each font used, must authors first try and guess what the actual `line-height: normal` equates to, to be able to then do the 150% of it?

This would further be complicated by the fact that authors cannot guarantee which exact font will be used on a user's machine in the end - with font substitutions, possibility that a web font won't load, etc (and even differences in metrics across different versions/cuts of the same font with nominally the same name).

Or, for simplicity and consistency, are we treating `line-height: 1` to be the *default spacing for the font*, meaning that to pass this an author must define a minimum of `line-height: 1.5`? (I've seen this interpretation being used on a few AAA audits, anyway)

patrickhlauke:
Depending on the answer to this, there might need to be a small note/clarification in the understanding (particularly around what *default spacing for the font*, as the anchor for the other 1.5x calculation, actually means in practice), and possibly a technique with the live example code.

awkawk:
[...]

PR #2262, opened 2022-03-15
Tweak 1.4.8 understanding to clarify that 'space-and-a-half' is 150% line height

patrickhlauke:
Closes https://github.com/w3c/wcag/issues/2256

### DIFF
--- a/understanding/20/visual-presentation.html
+++ b/understanding/20/visual-presentation.html
@@ -45,9 +45,9 @@
          of a paragraph. It is best if there are several different options, for instance, space-and-a-half
          and double spacing for line spacing. By space and a half within paragraphs we mean
          that top of one line is 150% further from the top of the line below it than would
-         be true when the text is 'single spaced' (the default spacing for the font). By Paragraph
-         spacing that is 1.5 times larger than the line spacing we mean that the spacing from
-         the top of the last line of 1 paragraph is 250% farther from the Top of the first
+         be true when the text is 'single spaced' (i.e., the line height is 150% the size of the
+         font used). By Paragraph spacing that is 1.5 times larger than the line spacing we mean that
+         the spacing from the top of the last line of 1 paragraph is 250% farther from the Top of the first
          line of the next paragraph (i.e., that there is a blank line between the two paragraphs
          that is 150% of the single space blank line). 
       </p>


JAWS-test:
Unsure if the number for the paragraph spacing (250%) is correct, because 1.5 times 1.5 = 2.25 and not 2.5.
This is already in the old Understanding, but I feel like the 250% doesn't really fit the wording in the SC ("and paragraph spacing is at least 1.5 times larger than the line spacing"). Perhaps I also misunderstand the wording in the SC...?

patrickhlauke:
made a further attempt to also clarify the paragraph spacing part a bit (the way I understand it, at least).

note that, fundamentally, I also think the **normative** text of the SC is *wrong* in how it uses the term "Line spacing (leading)", because it's not saying the leading per se (i.e. the empty space between lines) needs to be a space-and-a-half (as that would equate to `line-height: 250%` (100% for the text itself, 150% for the leading), but that the "Line height" needs to be 1.5 times the regular font size... but not going to touch the normative side here. but that is certainly the root cause of the weird confusion ... and only the fact that this is AAA, so not revisitied all that often by authors/auditors, has masked this problem for so long.

patrickhlauke:
otherwise, the normative text should really be changed from

`Line spacing (leading) is at least space-and-a-half within paragraphs, and paragraph spacing is at least 1.5 times larger than the line spacing.`

to

`Line height (including leading) is at least space-and-a-half within paragraphs, and paragraph spacing is also at least space-and-a-half.`

patrickhlauke:
following up: if the SC indeed intended to mean that *leading* needs to be 150% (leading to a line height of 250% altogether), then the images/examples provided in the understanding are wrong, as is technique C21 https://www.w3.org/WAI/WCAG21/Techniques/css/C21 (as then that should set `line-height: 2.5` rather than `line-height: 1.5`)

patrickhlauke:
changing this to a draft PR, as the above points (ported back to the issue https://github.com/w3c/wcag/issues/2256) are rather fundamental to determine whether or not this is correct (or if indeed it's the technique C21 that's wrong/misleading)

patrickhlauke:
@alastc @awkawk any thoughts/discussion on this would be appreciated

bruce-usab:
Current:
> Line spacing (leading) is at least space-and-a-half within paragraphs

Proposed:
> Line height (including leading) is at least space-and-a-half within paragraphs

That is a nice editorial patch.

Current:
> and paragraph spacing is at least 1.5 times larger than the line spacing.

Proposed:
> and paragraph spacing is also at least space-and-a-half.

No!

One (important) benefit of 1.4.8 is to have noticeably more visible white space between paragraphs as compared to lines of text.  I don't know that it has to be 2.25.  Could be (proposed):

> and paragraph spacing is at least twice the line height.

patrickhlauke:
> and paragraph spacing is at least twice the line height (that is, double spacing between paragraphs).

again, this gets confusing because of the loose use of "spacing" and what that means exactly.

In essence, is it asking here to - in layperson's terms - have enough space between paragraphs that a line of text (with its ample line height) would fit in there? like there's an empty line between the paragraphs?

(and yes, we still have the problem of what the whole house of cards is actually anchored on)

awkawk:
I think that the understanding changes are ok. I don't think that we are using the term leading incorrectly as in typography it refers to the measurement from baseline to baseline (see https://helpx.adobe.com/indesign/using/leading.html, https://www.indesignskills.com/tutorials/leading-typography/, https://fonts.google.com/knowledge/glossary/line_height_leading).

patrickhlauke:
confusingly, there are conflicting definitions. some refer back to the original meaning (strips of lead being added between the movable type characters), so it's essentially just the inter-line space (between the descenders of one line and the ascenders of the next line). others indeed use leading to mean overall line height (baseline to baseline).

it would probably help if we added an extra diagram as well to illustrate what we/the spec mean when it talks about leading etc

GreggVan:
I think leading refers to both

It all goes back to typesetting days.    The characters were each separate little blocks.   Lead was placed both between letters (for spacing) and between lines of type (for line spacing)

So it doesn’t surprise me that we see leading as being defined both ways.   

I think the term Line Spacing that someone suggested is better than using leading (which is an ancient and jargony term that is based on history - and tells you nothing if you don’t know the jargon.

Line spacing however suggests what it means.   And if we mean only the space between lines - we can say  "added space between lines of type" 


Just a suggestion 



Gregg Vanderheiden
***@***.***

bruce-usab:
I am pleased there is no compelling reason to change the phrasing of the SC.

> In essence, is it asking here to - in layperson's terms - have enough space between paragraphs that a line of text (with its ample line height) would fit in there? like there's an empty line between the paragraphs? (generally, that's what's understood when talking about double-spaced lines of text)

Some people need a little more space between lines of text to facilitate more fluid reading.  Some people need extra space between paragraphs to help keep their place when reading long tracks of text.  Users of screen magnification is one example.  If every line is double spaced, paragraph breaks are harder to find (for everyone, but for some people it is really difficult).  This SC's requirement to open up paragraph spacing &ndash; independent and regardless of the line spacing &ndash; is important.



 
> (and yes, we still have the problem of what the whole house of cards is actually anchored on)

patrickhlauke:
Thanks for 'splaining. I think most of us here are aware of what the SC is trying to achieve...it's more about trying to work out how to more unambiguously convey the concepts here (vis-a-vis confusion on which definition of "leading" is meant, etc).

Anyway, will try to update this some more and then make it ready for review.

PR #2266, opened 2022-03-17
fix broken link, fix duplicate id attribute

fstrr:
1. A link to Use Of Color was 404ing.
2. There was a duplicate id attribute that I remediated


alastc:
Thank you :-)

Issue #2169, opened 2022-01-04
Seeking guidance on 1.4.4 browser zoom, text size zoom etc


jamesjacobs:
After reading through 1.4.4 Resize text many times and also issues raised here relating to it I'm still not 100% convinced I've understood the requirements correctly. Hoping someone can illuminate me 😄 

Put simply, if a site displays correctly and doesn't cut off, clip, truncate or obscure text when the **page is zoomed in browser** (for example, by hitting CMD/CTRL + in Chrome to zoom the whole page) and doesn't prevent scaling or pinch and zoom on mobile - is that a pass? 

**Or** should a user be able to **zoom/increase only text** (for example, via methods such as FireFox text-only zoom or iOS Safari text zoom using the AA text sizing button)? If that results in text being cut off etc when a full page zoom displays correctly - is that a fail of 1.4.4?

Thanks in advance

patrickhlauke:
the SC is silent about how users go about resizing text. in essence it demands that there be at least *one* way for users who need larger text to achieve this. no specifics about which methods a site must support. so yes, if the site works well with full-page zoom (even if it leads to bidirectional scrolling), then it's a pass - regardless of whether or not doing an explicit text-only resize on that same site would lead to problems. as long as one method works, it's a pass.

fstrr:
You might find the note in [Failure of Success Criterion 1.4.4 when resizing visually rendered text up to 200 percent causes the text, image or controls to be clipped, truncated or obscured](https://www.w3.org/WAI/WCAG21/Techniques/failures/F69.html) useful:

> "The Working Group has discovered many misunderstandings about how to test this failure. We are planning to revise this failure in a future update. Until then, if the content passes the success criterion using any of the listed sufficient techniques, then it does not meet this failure."

bruce-usab:
@jamesjacobs &mdash; please do reply back to this issue thread if you are satisfied with @patrickhlauke response (or explain what more you might be hoping for).  For better or worse, it is a pretty lightweight requirement.

jamesjacobs:
Thanks for the confirmation all, that's really helpful. There seems to be so many conflicting opinions on this on the web, which resulted in our confusion.

Thanks again. Will close the ticket down.

selfthinker:
> the SC [...] demands that there be at least _one_ way for users who need larger text to achieve this

The SC doesn't say that anywhere.

The Understanding page is, as usual, clear as mud. ;-) 
[...]

PR #2270, opened 2022-03-19
Expand/clarify understanding for 1.4.4 Resize Text, harmonise 1.4.8 Visual Presentation

patrickhlauke:
Relates to https://github.com/w3c/wcag/issues/2169 and questions about whether a site that break using "text-only" resizing fails, even if zooming works


detlevhfischer:
Rendered view via [githack](https://raw.githack.com/w3c/wcag/9c002e5066ea1c8e79d9a7ebd0ac9a1364140bef/understanding/20/resize-text.html)

understanding/20/resize-text.html
```diff
@@ -15,31 +15,35 @@ <h2>Intent of Resize text</h2>
       
       <p>The intent of this Success Criterion is to ensure that visually rendered text, including
          text-based controls (text characters that have been displayed so that they can be
-         seen [vs. text characters that are still in data form such as ASCII]) can be scaled
+         seen, rather than text characters that are still in data form such as ASCII) can be scaled
```
detlevhfischer, line 18: I fail to get "text characters that can be displayed so they can be seen, rather than text characters that are still in data form such as ASCII", and I am sure I won't be the only one.
Is the aim to exclude from scaling requirements things like an (i) icon which may be text or svg or rendered via font icon, since it would be considered a graphic? All of these can be "seen", i.e. this is about the visual display pf text anyway, and i a line be drawn it is probably one between text proper and text-based icons? Or am I missing the point entirely?

patrickhlauke: admittedly, i had no clue what that bit is about, so left it untouched, apart from removing the horrible-looking `[vs. ...]` stuff

patrickhlauke: i'd say this should also either be clarified or, if we can't work out anymore what it actually means, removed...

patrickhlauke:
@bruce-usab i like your rewrite there. so is that exception there (about images of text and captions) what the cryptic "in data form such as ASCII" actually means to mere mortals?

patrickhlauke:
took @bruce-usab's suggestion (with a slight bit of language tweaking), which now nicely also mentions the exception (where previously the document is mostly silent about it, except for a bit about images of text).

this also removes the cryptic mention of "text characters that are still in data form such as ASCII" which i still fail to understand. again, if anybody wiser than me has any idea what that meant, and why it was there/if it's essential to bring that back, happy to reinstate something, but it'll need a bit of clarification as well, as currently it seems to baffle people.

understanding/20/resize-text.html
patrickhlauke: removed the ASCII bit - see comment below https://github.com/w3c/wcag/pull/2270#issuecomment-1075082743

JAWS-test:
@patrickhlauke maybe you also want to remove the ASCII part of 1.4.8? - it is the same there

patrickhlauke:
@JAWS-test good point, might as well

bruce-usab:
I do not know who wrote <q>text characters that are still in data form such as ASCII</q> but I am guessing it is a reference to situations where ASCII text is being used (say, a timed-text file for captioning) but there is no feasible way (due to constraints of the video player) to provide the end-user a means to control the font size.  This is less of a problem nowadays.  I agree that it is not helpful to mention ASCII per se.

patrickhlauke:
@alastc any chance this PR and https://github.com/w3c/wcag/pull/2194 could be reviewed/discussed/merged soon?

yatil:
Could we somehow form a community group that reviews and merges those changes largely independently from AG WG? It feels like AG WG has not enough time to prioritize this, and a dedicated group for these editorial changes would probably be more efficient, especially if the red tape is removed.

patrickhlauke:
@alastc @mbgower ping?

understanding/20/resize-text.html
```diff
@@ -131,7 +134,7 @@ <h2>Examples of Resize text</h2>
          </li>
          
          <li>
-            A user uses a zoom function in his user agent to change the scale of the content.
+            A user uses a full-page zoom function in their user agent to change the scale of the content.
```
mbgower, line 137: This is personal bias. I get jarred every time I see a user/their combination. I intentionally rewrite these as users/their. 
I realize that with 3rd-person pronoun preferences, "they" is becoming viewed as singular, but always found it awkward notwithstanding, and believe making it agree 'traditionally' by usage is never wrong, and arguably desirable for clarity in some situations.

Totally understand my views on subject/pronoun agreement will be viewed as quaint/old-fashioned/pitiable by some...

mbgower (approved):

understanding/20/resize-text.html
patrickhlauke: while yes, this view does seem a bit old to me... we can move to plural.

though looking at this closer, the "Users use" pattern always grates me. How would we feel about "People use..." ?

patrickhlauke: decided to turn this list item around (also to avoid the "user uses ... user agent"

understanding/20/resize-text.html
```diff
@@ -20,30 +20,38 @@ <h2>In brief</h2>
...
+         but text is most critical. 
       </p>
       
       <p>The scaling of content is primarily a user agent responsibility. User agents that
          satisfy 
-         <a href="https://www.w3.org/TR/WAI-USERAGENT/guidelines.html#tech-configure-text-scale">UAAG 1.0 Checkpoint 4.1</a> allow users to configure text scale. The author's responsibility is to create Web
-         content that does not prevent the user agent from scaling the content effectively.
+         <a href="https://www.w3.org/TR/WAI-USERAGENT/guidelines.html#tech-configure-text-scale">UAAG 1.0 Checkpoint 4.1</a>
+         allow users to configure text scale through a number of mechanisms - including zoom (of the entire page's content),
+         magnification, text-only resizing, and allowing the user to configure a reference size for rendered text.
```
mbgower, line 35:
```suggestion
         magnification, text-only resizing, and allowing the user to configure a size for rendered text.
```

understanding/20/resize-text.html
```diff
@@ -20,30 +20,38 @@ <h2>In brief</h2>
...
          with user agent support for resizing text, including text-based controls, or by providing
          direct support for resizing text or changing the layout. An example of direct support
          might be via server-side script that can be used to assign different style sheets.
       </p>
+
+      <p>Content satisfies the Success Criterion if it can be scaled up to 200% using at least one text scaling
+         mechanism supported by user agents.</p>
+      
+      <p>The author cannot rely on the user agent to satisfy this Success Criterion for HTML
+         through a specific text scaling mechanism if users do not have access to a user agent with that particular mechanism.
```
alastc, line 47:
```suggestion
         through a specific text scaling mechanism if people do not have access to a user agent with that particular mechanism.
```

understanding/20/resize-text.html
```diff
@@ -20,30 +20,38 @@ <h2>In brief</h2>
...
+      </p>
       
-      <p>If the author is using a technology whose user agents do not provide zoom support,
-         the author is responsible for providing this type of functionality directly or for
-         providing content that works with the type of functionality available in the user agent.
-         If the user agent doesn't provide zoom functionality but does let the user change the
+      <p>If the author is using a technology whose user agents do not provide support for specific text scaling mechanisms,
+         the author is responsible for providing this type of functionality directly, or to provide
```
mbgower, line 51:
```suggestion
         the author is responsible for providing this type of functionality directly, or providing
```

understanding/20/resize-text.html
```diff
@@ -20,30 +20,38 @@ <h2>In brief</h2>
...
+      
+      <p>The author cannot rely on the user agent to satisfy this Success Criterion for HTML
+         through a specific text scaling mechanism if users do not have access to a user agent with that particular mechanism.
+      </p>
       
-      <p>If the author is using a technology whose user agents do not provide zoom support,
-         the author is responsible for providing this type of functionality directly or for
-         providing content that works with the type of functionality available in the user agent.
-         If the user agent doesn't provide zoom functionality but does let the user change the
+      <p>If the author is using a technology whose user agents do not provide support for specific text scaling mechanisms,
```
alastc, line 50:
```suggestion
      <p>If the author is using a technology whose user agents do not provide support for specific text scaling mechanisms (for example, a closed system such as kiosks),
```

understanding/20/resize-text.html
```diff
@@ -20,30 +20,38 @@ <h2>In brief</h2>
...
          direct support for resizing text or changing the layout. An example of direct support
          might be via server-side script that can be used to assign different style sheets.
       </p>
+
+      <p>Content satisfies the Success Criterion if it can be scaled up to 200% using at least one text scaling
+         mechanism supported by user agents.</p>
+      
+      <p>The author cannot rely on the user agent to satisfy this Success Criterion for HTML
+         through a specific text scaling mechanism if people do not have access to a user agent with that particular mechanism.
+      </p>
```
mbgower, line 48:
```suggestion
```

Issue #2219, opened 2022-02-08
Clarify what is meant by "When user interface components receive keyboard focus," in SC 2.4.11


mraccess77:
I had understood this to meant that when the user scrolled the component out of view that it didn't need to have a focus indicator - but that while it is visible and focus hasn't been moved the focus needed to persist.   Others bring up the point that it only means at the exact moment it receives focus does it need to have an indicator and the indicator can disappear.    Can we please define what is required.   In addition, if is in fact that it only has to briefly appear on focus and can disappear a second later - are there any other success criteria in WCAG that do require focus persist while the control is visible in the viewport and the user has not moved focus?

This difference seems to be related to discussion in #2201

patrickhlauke:
I'm sure we discussed some of this previously in some other issue, but can't find it now. I think it was in the context of whether something like `:focus-visible` would also be acceptable, and whether or not controls need to have persistent visible focus indication even if they were focused programmatically or from a mouse click/touch tap...

stevefaulkner:
I find the idea that a transitory indication of focus is adequate to conform both incorrect and objectionable. Focus indicators represent a state of being focused, change in the display indication when the control is in a focused state implies a change in state, which has not occurred.

awkawk:
I've had questions about this as well. @mbgower had some examples I think.

I've grappled with this SC a lot - just as we are asking what the examples are where the focus needs to disappear, I'm also wondering how much of a problem it is that the focus is not persistent. I can see how it is a problem if the focus isn't shown - even with a mouse people have this location problem and Windows has included a feature to help people find the mouse (https://support.microsoft.com/en-us/windows/find-your-mouse-pointer-fast-dbc1d222-778c-da15-5218-cb8336074554). That said, I know that when I use the keyboard to navigate that I regularly lose track of the focus even when it is quite prominent - I'm thinking about other things / the UI has a lot going on and even though the focus is prominent it can still get lost / etc - and the way that I reorient is to tab a couple of times and shift-tab if needed.

Has focus disappearance been a major problem that 2.4.7 Focus Visible hasn't been addressing?

mraccess77:
I would consider a fading indicator a failure of SC 2.4.7.

bruce-usab:
I must confess, it had not occurred to me to consider a fading indicator under SC 2.4.7.  Sorry to say, I do not think a fail is unambiguous.  Ping to my colleague @kengdoj to say if this detail is covered under Trusted Tester.

[...]

Issue #2222, opened 2022-02-10
Focus Appearance: Don't fail 1px outlines and rounded shapes


WilcoFiers:
In the [January 18th](https://www.w3.org/2022/01/18-ag-minutes) meeting, discussed using the border-box as the way to determine the size of components in HTML. The following note was to be added (PR pending):

> Note: In HTML the size of a component is measured up to and including the CSS border.

I didn't notice this at the time, but going back through the understanding document today I realised that this now introduces new problems. In particular, the round button example from the understanding document may not meet the SC anymore:

> **Unusual shapes and gradients**
> If you have an unusual shape (not rectangular), some mathematics may be required if the focus indicator is on the edge of the size requirement. For example, if the focusable control is a circle with a diameter of 100px, the circumference would be 314 pixels. A 1px (or greater) outline would meet the size criterion.
> ![Two orange circular buttons with an icon in the middle. The right hand circle has a dark outline around the whole circle to indicate focus](https://www.w3.org/WAI/WCAG22/Understanding/img/focus-indicator-circle.png)

If we're saying that in HTML, border-box needs to be used, that means the perimeter of that button is 100x100, and so the minimum focus area would be 400px, not 314px. This wouldn't just be true for circular buttons. This would be true for any control with rounded corners.

I'm not sure what to suggest as a solution here. One option could be to say it's border-box, excluding border-radius. The note as it was proposed sort of suggests it might be. That further complicates things though, and it then begs the question; if we're excluding border-radius, why not clip or overflow?

Even if we excluded border-radius, there are other ways to achieve that rounded / cut corner effect there the exception wouldn't be allowed. SVGs for example, or background images. In particular the [CSS `paint` API](https://css-tricks.com/the-css-paint-api/) is relevant there.

bruce-usab:
Instead of making border-box might the way _the_ way to determine the size of components in HTML, might it just be noted as _one_ way?  I agree that rounded corners and circles should not be excluded.  Another approach is to work the match metrics from the other direction &mdash; for the example above, this could min height or width of 100px and min focus/target area of 314px.

patrickhlauke:
as in, imagine something roughly like this...

```html
<button>
  <span>I'm a button</span>
</button>
```

```css
button { width: 100vw; height: 100vh; display: flex; justify-content: center; align-items: center; background: transparent; border: none; }
button span { background: #eee; padding: 1em; border: 1px #000 solid; }
```

[...]

PR #2281, opened 2022-03-22
Focus-Appearance rework

alastc:
Based on the recent meetings.

[Preview the SC text and revised understanding doc](https://raw.githack.com/w3c/wcag/focus-appearance-rework/understanding/22/focus-appearance-minimum.html).

To do: The updates to the AAA version.

Closes: #2219, #2222, #2226,

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Apr 27, 2022, 9:01 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fwcag%2F7e26284b38d50723d7ba5b84fe4db043634e860d%2Fguidelines%2Findex.html%3FisPreview%3Dtrue)

```

😭  Sorry, there was an error generating the HTML. Please report this issue!
Specification: http://labs.w3.org/spec-generator/uploads/bnr0Mh?isPreview=true&publishDate=2022-04-27
ReSpec version: 32.1.4
File a bug: https://github.com/w3c/respec/
Error: Error: Evaluation failed: Timeout: document.respec.ready didn't resolve in 29140ms.
    at ExecutionContext._evaluateInternal (/u/spec-generator/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:221:19)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async ExecutionContext.evaluate (/u/spec-generator/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:110:16)
    at async generateHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:221:12)
    at async toHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:92:18)
    at async Object.generate [as respec] (file:///u/spec-generator/generators/respec.js:15:44)
    at async file:///u/spec-generator/server.js:244:48

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/wcag%232281.)._
</details>


bruce-usab:
@alastc &mdash; Please see [my comment in the Google doc](https://docs.google.com/document/d/1TI_DsJjfg9RW_A9C1XfqgtfYqNMfnxeCz6lxjpOQ9rk/edit?disco=AAAAXBDHDx8).  I believe a rigid reading of <q>continuous</q> for the encompasses definition to be problematic.  I don't have a good suggestion, except to drop that adjective from the definition (which then leaves the possibility of a dashed border with too much empty space).

alastc:
>  I believe a rigid reading of continuous for the encompasses definition to be problematic. 

Hi @bruce-usab, remember that the exception is still based on area, so any non-continuous indicators would fall through to the exception.

bruce-usab:
@alastc &mdash; I wrote something similar in the Google doc.
> any non-continuous indicators would fall through to the exception

I do not agree that this is good enough, because the exception is part of the SC, not the definition.

On the survey, I asked:  Does a UI like [marching ants](https://en.wikipedia.org/wiki/Marching_ants) meets the definition for encompasses (and I am of the strong opinion that it should)?

I was scribe, so I did not track your reply on the call.  You said <q>**Yes**, because of the exception.</q>  That was enough for me at the time.

With the benefit of hindsight, I wish your answer had been more along the line:  <q>**No**, but it is okay, because of the exception.</q>  That might have clued me in more (at the time of the call) to note that I think it is a significant problem to use <q>encompasses</q> with such a particular caveat, which puts our use of the term outside the mainstream understanding of the word.

My suggestion is to _move_ the <q>continuous</q> qualifier from the definition to the text of the SC.

bruce-usab:
Thanks @alastc.&nbsp; I am more comfortable with a specialize particular meaning for <q>surrounds<.q> than <q>encompasses</q> &mdash; not that I can give a good defense as to why.  I think it might be better _not_ to have a defined term in this case, but I very much prefer what you have here:  <q>surrounds</q> in the SC, and <q>encompasses</q> as part of the definition of the defined term.

mbgower:
@bruce-usab My concern with using “surrounds” is that it fails to address some common outliers, such as a button that changes its color. By making the definition for “encompass” include “surrounds” as well as some related but distinct terms, we manage to tackle all those edge cases while maintaining a word that has a broader meaning than “surrounds” while including it. Hope that helps

alastc:
@mbgower That changing colour technique wouldn't pass the primary part of the SC anyway, that would fall through to the exception (and pass). 

Bruce's point is just about which term matches the intent best, and I'm fairly ambivalent about that (I thought they both did, but others have said that 'encompasses' wasn't quite right).

mbgower:
Why wouldn't it pass the primary?
If you have a grey button that contrasts 3:1 with a white background, and you make it black, the pixels of the focus indicator contrast against themselves, and the black contrasts against the white background. Since the change "includes the whole of the" button, it meets the primary.

mbgower:
I assume using "embiggen" is a joke? Otherwise, I'm not sure why you would propose replacing what you call a "fancy word" with a term i've never heard of which, when I looked it up, seems to have been first used on a Simpson's episode.

mbgower:
If the problem here is that you're trying to make it clear a dashed line fails, might I suggest:
"solidly bounds, surrounds or includes the whole of"

mbgower:
"Continuously encompass" just seems awkward to me. 
There may not be a way of wording this elegantly. Do we have to resort to "surrounds the visual presentation of the user interface component with an unbroken line"? I find that overly prescriptive, personally.

 For my money "solidly bounds, surrounds or includes the whole of" as a definition of encompasses is pretty good.

bruce-usab:
@mbgower &mdash; "solidly bounds, surrounds or includes the whole of" is fine as a concept.  It is _not_ fine as definition for "encompasses" because a minority of people using that word would assume "solidly".  The fact that we dare not leave encompasses undefined illustrates why there is a problem.

This is where I believe we are with the top half of the SC:

When a user interface components has keyboard focus, the focus indicator:
- Encompasses the visual presentation of the user interface component;
- Has a contrast ratio of at least 3:1 between its pixels in the focused and unfocused states;
- Has a contrast ratio of at least 3:1 against adjacent colors.

So how about instead:

When a user interface components has keyboard focus, the focus indicator:
- Solidly bounds, surrounds or includes the whole of the visual presentation of the user interface component;
- Has a contrast ratio of at least 3:1 between its pixels in the focused and unfocused states;
- Has a contrast ratio of at least 3:1 against adjacent colors.

mbgower:
See both https://github.com/w3c/wcag/pull/2283 and https://github.com/w3c/wcag/pull/2284 for variations on this

bruce-usab (changes requested):
With this PR, I am still seeing <q>encompasses</q> as a defined term with <q>solidly</q> as a required characteristics.  That detail goes above and beyond usual meaning of the wordr.

But I am also seeing definitions for focus and perimeter which I am comfortable with.  So I am not confident that I looking at the most recent version.

mbgower:
@bruce-usab most up to date of my proposed changes are in 2284 (linked in my last comment). Note that they are cumulative, so it helps to look at 2283 first to see the progression from 2281

Issue #1698, opened 2021-03-25
Visible Required label and 3.3.2: Labels or Instructions


LaurenceRLewis:
Understanding Success Criterion 3.3.2: Labels or Instructions
https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html

Benefits - bullet point 2
_Providing clear and unambiguous labels and instructions (including clear identification of required fields) can prevent users from making incomplete or incorrect form submissions, which prevents users from having to navigate once more through a page/form in order to fix submission errors._

Is labelling only the optional fields rather than the required fields a fail of 3.3.2: Labels or Instructions? If so, does anyone have any user data to back this up.

I believe it is non-compliant, however, need to get information on why.

JAWS-test:
Understanding Examples: 

> In a form which contains both required and optional fields, the required fields and/**or the optional fields** are clearly labeled as such. 

Thus, it is not a violation if only the optional fields are marked. However, it is more common to mark the required fields

LaurenceRLewis:
Thanks @JAWS-test I missed that in the understanding doc.

LaurenceRLewis:
@JAWS-test I can’t find that paragraph in the understanding document. Can you provide the link to it please.

JAWS-test:
https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html#examples

goetsu:
@detlevhfischer @patrickhlauke to be clear can you confirm working group position on these specific cases : 

- If all fields are required (like a login / password field) 
-- without any visually information that all fields are mandatory
-- without any required or aria-required attributs on the fields 
- If all fields are required (like a login / password field) 
-- without any visually information that all fields are mandatory 
-- with a required or aria-required attributs on the fields

goetsu:
@patrickhlauke thanks for your personal pragmatic opinion that I share ;) 
I think that the understanding text of the SC or at least the text of the [H90 technique](https://www.w3.org/WAI/WCAG21/Techniques/html/H90) will gain to be made more clear on such case with a note in the description for example like "When all fields inside a form a required event if it's a good pratice there is no formal requirement for them to have a visual  indication that they are required before the user submit the form"

bruce-usab:
@goetsu &mdash; I think it is fair to characterize similar WG discussions to focusing on the exact phrasing of the SC.  So I agree with @patrickhlauke analysis.

[...]

PR #2282, opened 2022-03-23
Change for case where all fields are mandatory

goetsu:
- add information to clarfy that when all fields mandatory this technique is just a best pratice not a requirement
- making test procedure applicable only when form contain both required and optional controls

### DIFF
--- a/techniques/html/H90.html
+++ b/techniques/html/H90.html
@@ -2,7 +2,8 @@
       <p>HTML and XHTML controls that use external labels.</p>
    </section><section id="description"><h2>Description</h2>
       <p>The objective of this technique is to provide a clear indication that a specific form control in a Web application or form is required for successful data submission. A symbol or text indicating that the control is required is programmatically associated with the field by using the <code class="el">label</code> element, or the <code class="el">legend</code> for groups of controls associated via <code class="el">fieldset</code>. If a symbol is used, the user is advised of its meaning before the first use. </p>
-   </section><section id="examples"><h2>Examples</h2>
+      <p>Even when all fields inside a form are required, it is a good practice to have a visual indication that they are required before the user submits the form.
+  </section><section id="examples"><h2>Examples</h2>
       <section class="example">
          <h3>Using text to indicate required state</h3>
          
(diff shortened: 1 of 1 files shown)


bruce-usab:
If merged, closes https://github.com/w3c/wcag/issues/1698

patrickhlauke:
@bruce-usab if you add that to the PR's description/first comment, github will do this automagically (I just edited it to that effect)

bruce-usab:
> if you add that to the PR's description/first comment, github will do this automagically

Thanks!  That was indeed the effect I was trying to invoke.

bruce-usab (approved):
Looks good, thanks!

techniques/html/H90.html
```diff
@@ -2,7 +2,8 @@
       <p>HTML and XHTML controls that use external labels.</p>
    </section><section id="description"><h2>Description</h2>
       <p>The objective of this technique is to provide a clear indication that a specific form control in a Web application or form is required for successful data submission. A symbol or text indicating that the control is required is programmatically associated with the field by using the <code class="el">label</code> element, or the <code class="el">legend</code> for groups of controls associated via <code class="el">fieldset</code>. If a symbol is used, the user is advised of its meaning before the first use. </p>
-   </section><section id="examples"><h2>Examples</h2>
+      <p>Even when all fields inside a form are required, it is a good practice to have a visual indication that they are required before the user submits the form.
```
kfranqueiro, line 5:
```suggestion
      <p>Even when all fields inside a form are required, it is a good practice to have a visual indication that they are required before the user submits the form.</p>
```

bruce-usab (approved):
Added the closing `</p>` Ken spotted

mbgower:
I'm a little leery of this change on its own. We've already spent quite a bit of time clarifying that the word "required" as used in Labels or Instructions means "needed".
So I'd like to pay a bit more attention to this technique and bring it in line with that, rather than this very simple change.

I'm also not sure I agree it is only a best practice.

bruce-usab:
Discussed on backlog call 2/7 and @mbgower offered to make another pass.  On call, had several examples where the context means there is no needed to mark fields as being required.  Common example is login fields.

mbgower:
I think the note added in the first paragraph is running counter to this sufficient technique. The name of this technique is "Indicating required form controls using label or legend". If one did not mark each field that is required, this technique wouldn't be met.

There _are_ ways to handle required fields without indicating each via the label or legend (e.g., "All fields are required", "All fields are required unless marked as optional", etc).  I think the contents of this note belongs in another sufficient technique which covers those, whether new or additions. A possible candidate where info could be added is https://www.w3.org/WAI/WCAG21/Techniques/general/G184

I also have some concerns that changing the first check has altered the test and scope of the technique
> <li>Check if some of the form controls are optional</li>

Remember that this is a sufficient technique. It's not the only way to meet the requirement. Currently, one meets the technique by marking required fields via the label or legend. With the addition of this first check qualification, if all my fields are required and I mark them all, I do not meet this test procedure anymore, since it can only be met if some forms are optional.

Again, it might be worth seeing if 184 can be leveraged for what is not a technology-specific approach.

PR #2288, opened 2022-03-30
Adding design canvas exception

### LINKED ISSUE #1888: Adobe Comment on 3.2.7 Visible Controls

alastc:
Closes #1888

### DIFF
--- a/guidelines/sc/22/visible-controls.html
+++ b/guidelines/sc/22/visible-controls.html
@@ -9,6 +9,7 @@
     <ul>
         <li>The same functionality is available through a component on the same page, or on a different step in a multi-step process, without requiring pointer hover or keyboard focus to make it visible;</li>
         <li>The component provides keyboard-only functionality;</li>
+        <li>The trigger for hover of focus is an editable item within a region of the page dedicated to editable items;</li>
         <li>A mechanism is available to make the components persistently visible;</li>
         <li>Hiding the visual indicator is <a>essential</a>.</li>
     </ul>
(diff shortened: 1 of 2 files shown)


bruce-usab (approved):
Am I correct that this is Typo?
> The trigger for hover **of** focus

Should be:  The trigger for hover **or** focus

guidelines/sc/22/visible-controls.html
```diff
@@ -9,6 +9,7 @@ <h4>Visible Controls</h4>
     <ul>
         <li>The same functionality is available through a component on the same page, or on a different step in a multi-step process, without requiring pointer hover or keyboard focus to make it visible;</li>
         <li>The component provides keyboard-only functionality;</li>
+        <li>The trigger for hover of focus is an editable item within a region of the page dedicated to editable items;</li>
```
awkawk, line 12: Maybe "dedicated to editable or selectable items"?

PR #2293, opened 2022-03-31
Alternative definition of page-break-locator

### LINKED ISSUE #2259: Are CSS break-before/after indicators page break locators?
There are a few ways in CSS to indicate where to insert / where not to insert a page break when printing. Looking at the definition of page break locators, it seems like they might apply. However the understanding document / technique doesn't mention them. Here's an example of one:

```html
<h1>Hello world</h1>
<p>Lorem ipsum dolor sit amet consectetur?</p>
<h2>More stuff</h2>
<p>Lorem ipsum dolor sit amet consectetur?</p>
<h2>More stuff</h2>
<p>Lorem ipsum dolor sit amet consectetur?</p>
<style>
  h2 { break-before: page }
</style>
```

It's not clear to me why this SC should apply to CSS breaks, however looking at the normative language I think it does. Could this be clarified?

alastc:
Closes #2259

### DIFF
--- a/guidelines/terms/22/page-break-locators.html
+++ b/guidelines/terms/22/page-break-locators.html
@@ -2,7 +2,7 @@
 <dd class="new">
 	<p class="change">New</p>
    					
-  <p><a>programmatically determinable</a> destination markers that are arranged in a meaningful sequence to represent a locator serving the same purpose as page breaks in a printed document.</p>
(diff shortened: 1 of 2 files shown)


alastc:
Not agreed, will update the understanding doc instead.

patrickhlauke:
should the branch be deleted as well?

Issue #2241, opened 2022-03-01
[typo] focus should not necessarily be set to the **_first_** interactive element in the dialog


philljenkins:
Regarding example in Understanding Success Criterion 2.4.3: Focus Order
https://w3c.github.io/wcag21/understanding/focus-order.html

Some auditors and developers are interpreting example no. 3 as a "requirement", that  literally the "**_fist_**" interactive element should always receive focus first.  That is not the intent of the SC and confuses the reader, since many times it is more desirable for the focus to be set on the primary interactive element.  For example, on a log-in screen it would not be desirable for the focus to be set on a "cancel X button" that might appear at the top right of a dialog or page that redundantly includes a "Cancel" button at the bottom of the dialog or page when the primary purpose of the dialog is to enter some input such as a name, ID, or some other primary input.
  
> 3. A Web page implements modal dialogs via scripting. When the trigger button is activated, a dialog opens and focus is set to the **_first_** interactive element in the dialog. As long as the dialog is open, focus is limited ...

Please change "first" to "primary".

JAWS-test:
I think it depends on the context whether first or primary is better. That's why I wouldn't be in favor of replacing first with primary. I would like to see an addition: "first or primary". Also, it could be explained when first and when primary makes sense. 

Justification: For example, the login pop up may have a drop down list for the application language or other relevant but not primary settings before the username input field. If these intial do not have focus, blind users will have difficulty recognizing that there are other form fields before the username input field.

patrickhlauke:
it might not even need to go to an interactive element at all, but to something like the first non-interactive heading in the dialog (for content-heavy dialogs, rather than dialogs that just contain mainly form controls) - see https://github.com/w3c/aria-practices/issues/442

patrickhlauke:
though it bears reminding auditors/developers that understanding document is non-normative, and the examples are just that...examples. non-exhaustive, and not necessarily binding.

alastc:
Indeed, these are examples. We could add a duplicate of that example which said "primary" instead of "first". It isn't prescribing what you have to do, it is an example of something that would pass. (NB: I don't think we should add a duplicate, that would just add text for very little benefit.)

patrickhlauke:
or change the example from suggesting a particular place for focus to land (first interactive? primary? somewhere else?) and just say that focus goes to the modal dialog and leave it as an exercise for the reader to work out what that means (the main thing is that it doesn't remain in the underlying page that triggered the modal, which is the usual problem we find)

[...]

PR #2306, opened 2022-04-08
update language on setting focus in a modal dialog

fstrr:
It seems like there is agreement in the discussion on issue #2241 on changing the language for setting focus in a modal dialog. This PR updates the Understanding document with that change.

### DIFF
--- a/understanding/20/focus-order.html
+++ b/understanding/20/focus-order.html
@@ -127,7 +127,7 @@
          </li>
          
          <li>A Web page implements modal dialogs via scripting. When the trigger button is activated,
-            a dialog opens and focus is set to the first interactive element in the dialog. As
+            a dialog opens and focus is set within the dialog. As
             long as the dialog is open, focus is limited to the elements of the dialog. When the
             dialog is dismissed, focus returns to the button or the element following the button.
          </li>


patrickhlauke (approved):

fstrr:
closes #2241

mbgower (commented):
I wonder if "on or within" would be better? I'm pretty sure the html5 dialog now allows for focus to be placed in a variety of places, including on the dialog element itself. So "within" seems a bit too specific.

scottaohara:
to be fair though, this example is saying it has implemented a modal dialog via scripting, so is this even an example of using the dialog element?

the default for a dialog in html is still the first focusable element within.  it's only specific author overrides / fallback scenarios that the dialog element itself can receive focus.  

tldr; yes, dialogs can have focus set on or within.  but in the context of the current text / high level examples presented in this file, does mentioning that even matter, as how exactly this dialog is implemented is still rather nebulous.

understanding/20/focus-order.html
```diff
@@ -127,7 +127,7 @@ <h2>Examples of Focus Order</h2>
          </li>
          
          <li>A Web page implements modal dialogs via scripting. When the trigger button is activated,
-            a dialog opens and focus is set to the first interactive element in the dialog. As
+            a dialog opens and focus is set within the dialog. As
```
mbgower, line 130:
```suggestion
            a dialog opens and focus is set on or within the dialog. As
```

PR #2327, opened 2022-04-27
Gundula's suggestions

### LINKED ISSUE #2323: Bounding box for separated links
From Gundula: 
<hr>

I do not agree to the following part of the definition for bounding box:
"Where a component consists of disconnected parts, such as a link that wraps onto multiple lines, each part is considered to have its own bounding box."

With the existing definition, a focus indicator is suggested to fully enclose (with four solid lines) each of the disconnected parts, resulting in the appearance of several focus indicators. Next, the link could be named something like "a more detailed view", the part on the first line being "a". As a result the length of the shortest side is the length of the "a", which does not follow the intention of the SC.

Suggestion:
add "In such a case, the base for further measurement is the concatenation of all boxes for disconnect parts. In the example of a link that wraps onto multiple lines the resulting box would enclose the same link if it would not wrap."

The definition of minimum bounding box implies there is only one minimum bounding box for a given UI element. Consequently, the normative text should say 'the minimum bounding box', not 'a minimum bounding box'.

alastc:
Closes #2323

### DIFF
--- a/understanding/22/focus-appearance-minimum.html
+++ b/understanding/22/focus-appearance-minimum.html
@@ -33,7 +33,7 @@
                 <li>The focus indicator and the indicator's background color are not modified by the author, or</li>
                 <li>An area of the focus indicator meets all the following:
                     <ul>
-                        <li>Is at least as large as the area of a 1 CSS pixel thick <a>perimeter</a> of the unfocused component, or an area at least as large as a 4 CSS pixel thick line along the shortest side of a <a>minimum bounding box</a> of the unfocused component;</li>
+                        <li>Is at least as large as the area of a 1 CSS pixel thick <a>perimeter</a> of the unfocused component, or an area at least as large as a 4 CSS pixel thick line along the shortest side of the <a>minimum bounding box</a> of the unfocused component;</li>
                         <li>Has a contrast ratio of at least 3:1 against the same pixels in the unfocused state;</li>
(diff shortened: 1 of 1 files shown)


understanding/22/focus-appearance-minimum.html
```diff
@@ -49,7 +49,7 @@ <h2>Focus Appearance (Minimum) Success Criterion text</h2>
             </dd>
             <dt class="new"><dfn>minimum bounding box</dfn></dt>
             <dd class="new">
-                the smallest enclosing rectangle aligned to the horizontal axis within which all the points of a shape lie. Where a component consists of disconnected parts, such as a link that wraps onto multiple lines, each part is considered to have its own bounding box.
+                the smallest enclosing rectangle aligned to the horizontal axis within which all the points of a shape lie. Where a component consists of disconnected parts, such as a link that wraps onto multiple lines, each part is considered to have its own bounding box. In such a case, the base for further measurement is the concatenation of all boxes for disconnect parts. In the example of a link that wraps onto multiple lines the resulting box would enclose the same link if it would not wrap.
```
patrickhlauke, line 52: `In the example of a link that wraps onto multiple lines the resulting box would enclose the same link if it would not wrap.`

perhaps adding a comma, and changing `would` to `did`?

`In the example of a link that wraps onto multiple lines, the resulting box would enclose the same link if it did not wrap.`

PR #2330, opened 2022-04-28
Updating offset definition.

### LINKED ISSUE #2159: Target spacing (minimum): multiple overlapping elements interact oddly with offset
In working on automated tests, I ran into something I hadn't realised before. When an element is overlapped by multiple other elements, it is possible to pass the SC with a very small target on the outer element. In particular scenario H4 from my codepen:

https://codepen.io/wilcofiers/pen/abZxPow?editors=1100
Screenshot 2021-12-07 at 14 21 57

H4 shows a rectangle with two squares on top of it. The squares touching edges horizontally. The squares have a 24x24 diameter and so pass. The rectangle is only visible as a 6 pixel wide perimeter around the two squares.

The reason this passes is because to the offset from the rectangle to either square is a diagonal from their top-right corners for the left square, or their top-left corners for the right square. For both squares the offset is slightly more than 30 pixels, so the rectangle meets the offset exception. The odd thing here is if those two squares were treated as a single target, the SC would fail. But because there are two of them, the SC passes.

I have no suggestion on how to fix this, without causing a good deal of other problems that might get overlooked. Given how unusual this case is, I think it's acceptable to ignore it. But I did want to point it out so we're conscious of it.

alastc:
Closes #2159


guidelines/terms/22/target-offset.html
```diff
@@ -1,5 +1,5 @@
 <dt class="new"><dfn data-lt="target offsets">target offset</dfn></dt>
 <dd class="new">
 	<p class="change">New</p>	
-   <p>the distance measured from the farthest point of a target to the nearest point of the second target. Offset includes the target and spacing around the target. The target offset from A to B may be different then the offset from B to A, if the size of these targets differ.</p>
+   <p>the distance measured from the farthest point of a target (A), to the nearest point of the second target (B). The offset includes the target and spacing around the target. Targets other than A and B to not contribute to the spacing. The offset from A to B may be different then the offset from B to A, if the size of these targets differs.</p>
```
patrickhlauke, line 4: typo `to not contribute` > `do not contribute` ?

patrickhlauke: typo `different then the offset` > `different than the offset`

understanding/22/target-size-minimum.html
```diff
@@ -37,7 +37,7 @@ <h4>Target Size (Minimum)</h4>
             <dl><dt class="new"><dfn data-lt="target offsets">target offset</dfn></dt>
             <dd class="new">
                 <p class="change">New</p>
-            <p>the distance measured from the farthest point of a target, to the nearest point of the second target. Offset includes the target and spacing around the target. The offset from A to B may be different then the offset from B to A, if the size of these targets differs.</p>
+            <p>the distance measured from the farthest point of a target (A), to the nearest point of the second target (B). The offset includes the target and spacing around the target. Targets other than A and B to not contribute to the spacing. The offset from A to B may be different then the offset from B to A, if the size of these targets differs.</p>
```
patrickhlauke, line 40: same as above. `do not contribute`, `different than the offset`

guidelines/terms/22/target-offset.html
mbgower: addressed

understanding/22/target-size-minimum.html
mbgower: implemented

Issue #1270, opened 2020-08-09
Add H97 to related techniques for H69


carmacleod:
Please add [H97: Grouping related links using the nav element](https://www.w3.org/WAI/WCAG21/Techniques/html/H97.html)
to the list of Related Techniques for [H69: Providing heading elements at the beginning of each section of content](https://www.w3.org/WAI/WCAG21/Techniques/html/H69).

PR #2338, opened 2022-04-29
Update H69 to address issue #1270

fstrr:
- add reference to H97
- make links more descriptive inline with other Technique documents

Closes https://github.com/w3c/wcag/issues/1270


patrickhlauke:
@fstrr a heads-up: if you include `Closes #XXXX` in the first comment on the pull request, github will automagically tie the issue and PR together so when this is merged, it auto-closes the issue as well

mbgower (commented):
@michael-n-cooper i thought these complementary references were added automatically. Or is this a case where just because the list techniques cites the heading, it doesn't mean we need to have the heading citing the list technique? I do think that it's a weaker association going in this direction...

alastc:
@fstrr could you check the "Resolve conflicts" on this PR please?

I assume the conflicts come from your technique updates, so you can probably spot which bits are new more easily than I can...

fstrr:
@alastc Tis done.

Issue #1308, opened 2020-08-14
ARIA11: Add clause about form role needing an accessible label


carmacleod:
Elements with role="form" need to have an accessible label before they are considered to be a landmark.

Please add a new sentence to the [ARIA11 Technique](https://w3c.github.io/wcag/techniques/aria/ARIA11) to that effect. Something similar to what will be added for region in #1269:

> form: A region of the document that represents a collection of form-associated elements, some of which can represent editable values that can be submitted to a server for processing. It maps to <code>form</code> in HTML, and needs an accessible label to be included as a landmark for assistive technology.

NOTE: The label requirement for form role was added in ARIA 1.2, which has not reached Rec yet, so this can wait until it does.
**UPDATE:** ARIA 1.2 reached W3C Candidate Recommendation status on March 2, 2021.

PR #2339, opened 2022-04-29
Update ARIA11 to address issue #1308

fstrr:
Adds text for ARIA 1.2 form role.

Closes https://github.com/w3c/wcag/issues/1308


techniques/aria/ARIA11.html
```diff
@@ -26,7 +26,7 @@ <h1>Using ARIA landmarks to identify regions of a page</h1>
...
          <li><code class="language-html">main</code>: A region that contains a page's main content.</li>
          <li><code class="language-html">region</code>: A region that contains a perceivable section of the page containing content that is sufficiently important for users to be able to navigate to the section. A <code class="language-html">region</code> landmark isn't exposed as a landmark region unless it has an accessible name.</li>
-         <li><code class="language-html">form</code>: A region of the document that represents a collection of form-associated elements, some of which can represent editable values that can be submitted to a server for processing.</li>
+         <li><code class="language-html">form</code>: A region of the document that represents a collection of form-associated elements, some of which can represent editable values that can be submitted to a server for processing. It maps to <code class="language-html">form</code> in HTML, and needs an accessible name to be exposed as a landmark.</li>
```
scottaohara, line 29: not against the idea of mentioning what element this maps to in html - but now this is the only one that does.

there are a bunch of other nits i have with this doc (outside of the scope of this PR) - would it be ok to rectify those with this PR, or do we need to merge this and then i make another PR to make those corrections?

fstrr: Depends! I'm increasingly not a fan of PRs that wander off topic, but if it's just a couple of minor things then we could add them. I could also back out the mapping content to put in another PR if your bunch of nits is large.

scottaohara (approved):
changing my review to an approval - since any issues i have with this doc are out of scope for this PR

Issue #788, opened 2019-06-17
[ Techniques ] 2 wrong URL pointing to "Understanding" in header


code-elegant:
"Editors' Draft:" and "Official version:" links are broken at the very beginning of:
https://www.w3.org/WAI/WCAG21/Techniques/

**Techniques for WCAG 2.1**
Updated 7 June 2019

    Authors: Accessibility Guidelines Working Group
    Editors: Alastair Campbell, Michael Cooper, Andrew Kirkpatrick
    Editors' Draft: https://w3c.github.io/wcag/techniques/ ← here
    Official Version: https://www.w3.org/WAI/WCAG21/techniques/ ← here
    File Issues: https://github.com/w3c/wcag/issues/

https://github.com/w3c/wcag/blob/master/techniques/index.html#L22
https://github.com/w3c/wcag/blob/master/techniques/index.html#L23

`<li><strong>Editors' Draft:</strong> <a href="https://w3c.github.io/wcag/understanding/">https://w3c.github.io/wcag/techniques/</a></li>`

`<li><strong>Official Version:</strong> <a href="https://www.w3.org/WAI/WCAG21/Understanding/">https://www.w3.org/WAI/WCAG21/techniques/</a></li>`

fstrr:
Looks like the links were fixed but then were pointing to the not-correct URLs. Created a pull request for this: https://github.com/w3c/wcag/pull/2340

PR #2340, opened 2022-04-29
update to make links correct (issue #788)

fstrr:
- Editors Draft and Official Version URLs didn't match the visible link text


fstrr:
closes #788

alastc:
That's... odd.

PR #2343, opened 2022-04-29
combining the 2 user agent exceptions into 1

mbgower:
I cannot see anything we lose combining the 2 user agent exceptions.
I have written this as a branch from my more discussed changes, but I humbly offer it as what looks like a useful reduction.

### DIFF
--- a/guidelines/sc/22/focus-appearance-minimum.html
+++ b/guidelines/sc/22/focus-appearance-minimum.html
@@ -14,8 +14,7 @@
     <p>Exceptions:</p>
 
     <ul>
-        <li>The focus indicator is determined by the user-agent and cannot be adjusted by the author, or</li>
-        <li>The focus indicator and the indicator's background color are not modified by the author, or</li>
+        <li>The focus indicator and the indicator's background color are determined by the user agent and not modified by the author, or</li>
         <li>An area of the focus indicator meets all the following:
             <ul>
                 <li>At least as large as the area of a 1 CSS pixel thick perimeter of the unfocused component, or is at least as large as a 4 CSS pixel thick line along the shortest side of the minimum bounding box of the unfocused component, and</li>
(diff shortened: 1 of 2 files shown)


patrickhlauke (approved):

mbgower:
@patrickhlauke I had two more thoughts:
1.  should this be an "or" not an "and"? Is there a situation where someone other than the author has modified it -- say a CMS? Or is that still considered the author?
2. is there any reason why this can't be even further reduced down to "The focus indicator and the indicator's background color are determined by the user agent"

patrickhlauke:
> 1. should this be an "or" not an "and"? Is there a situation where someone other than the author has modified it -- say a CMS? Or is that still considered the author?

I think that distinction (actual human author, or the CMS) is unnecessarily pedantic. I think we've always considered "by the author" to be a shorthand for "whatever arrives at the UA/browser".

> 2. is there any reason why this can't be even further reduced down to "The focus indicator and the indicator's background color are determined only by the user agent"

I think the "modified by the author" bit is helpful to give some context on what is meant by "determined by the user agent". Otherwise, people could misinterpret "determined" as meaning "calculated" or something (based on the CSS that was sent?)

mbgower:
@rachaelbradley The whole unmodified user agent concept is probably a concept that could be better spelled out in 3.0

alastc:
I wonder if there are instances where you can't change the indicator, but you can change the background.

I.e. changing the background of the page would mean any native select drop-down control would fail.

The advantage of them being separate is that the "can't change" is focus-indicator only, but "haven't changed" is both indicator and background to the indicator.

mbgower:
> I wonder if there are instances where you can't change the indicator, but you can change the background.

Ah, nice one! Yes, I guess that is the edge case we have to protect against here. Okay, I will incorporate that into the Understanding document.

Issue #2237, opened 2022-02-23
Does 2.4.11 apply to active descendants of a combobox?


WilcoFiers:
When you expand a combobox or select, users can choose an option using arrow keys. In ARIA this is called the active descendant. So in addition to having a focused combobox, there is an active option. Question is, does 2.4.11 apply to active descendants in addition to the focused control? If so, does the focused control need to have the focus indicator while there is an active descendant?

More generally, I think an intuitive understanding of what is focused can occasionally contradict the programmatically determined focus state. It seems ambiguous to me which one of those to use when the two contradicts, such as in the case of active descendants.

https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.1pattern/listbox-combo.html

![Focused combobox, with an expanded dropdown with the second item selected](https://user-images.githubusercontent.com/530687/155290166-f87ee118-be2f-4bac-9698-2f7e555eb7ec.png)

patrickhlauke:
> More generally, I think an intuitive understanding of what is focused can occasionally contradict the programmatically determined focus state. It seems ambiguous to me which one of those to use when the two contradicts, such as in the case of active descendants.

this situation (that "focused" doesn't necessarily mean "the element that represents `document.activeElement`) has been touched on quite a few times before yes. most recently in that lengthy back-and-forth issue #2226, some mention of this as it relates to 2.4.7 here in #1001, but I'm fairly sure there were even earlier discussions)

i think long story short is that SCs mean "focus" in the more general "logical" sense, whatever the user perceives/understands as having focus. not necessarily the programmatic element that the browser understands as having focus (which yes, will make automated testing "interesting", as there may even be fully custom implementations of things - think `<canvas>` where the entire idea of a focus is "virtual" and handled completely by the author - that simply won't be auto-testable)

patrickhlauke:
but more specifically to your point about `activedescendant` ... this is a very interesting case where focus, even logically, is "in two places at once" effectively (even more so for comboboxes that aren't just like selects, but allow freeform typing in the text edit as well as providing a list of options/suggestions). agree that at the very least the understanding doc should devote some explanation to these "special" situations.

patrickhlauke:
> I think we're sort of trying to have it both ways here Patrick

are we? where? (i admit i've lost track of where the very latest thinking/wording on 2.4.11 is)

WilcoFiers:
Well, the way I understand this:

[...]

PR #2345, opened 2022-05-02

mbgower:
An attempt to address the concerns expressed for the later comments in https://github.com/w3c/wcag/issues/2237.
The closing paragraph has become a final exception bullet. A note has been added providing an example as well as information on selection versus focused.
UPDATE: After discussion, returned final point to a paragraph again.


bruce-usab (approved):
I am fine with this, but please note that this version takes away the page author/owner choice..

> these requirements _**are**_ applied to the sub-component instead.

I had expected to find <q>are allowed</q> or similar.

mbgower:
Thanks, Bruce. We can certainly use the 'may' language as a fallback, but my sense is that with the revision and the State note, we may be able to get away with something a bit emphatic.

mbgower:
Note some other tweaking options include:

option 1, remove the if and make it an "and"
- Where a user interface component has active sub-components and a sub-component receives its own focus indicator, these requirements are applied to the sub-component instead.


option 2, same as option 1, but add in "may"

- Where a user interface component has active sub-components and a sub-component receives its own focus indicator, these requirements may be applied to the sub-component instead.

alastc:
> Where a user interface component has active sub-components and a sub-component receives its own focus indicator, these requirements are applied to the sub-component instead.

I think with the associated note (outlining 'selection') we can be emphatic about that, so it is an either/both scenario.

I just wonder about the "and a sub-component" in a complex widget that might have different types of sub-component (e.g. calendar widget with days selected + focused + month buttons).

How about: 
"Where a user interface component has active sub-components which receives their own focus indicator, these requirements are applied to the sub-component instead."

mbgower:
Having gone through a bunch more examples, I think we may have to stick with the "may" language. Here is the challenge: I don't think we want to fail more discrete but clear selection indicators by judging them as focus indicators, and it is possible that someone can do that in many of these cases. For example, think of many tablists where the tab selection is just a thickish underline of the selected tab, or some other form of 'okay' selection indication... 
image
image
image
image
I'm not a huge fan of some of those examples (borderline text contrast issues, etc), but things similar that can pass all other wcag criteria are very common in the wild and there is one thing making them arguably 'not that bad' for interaction, which is **proximity**.
 If you imagine putting a focus rectangle around the whole of each of these interactions, and having them employed as a synchronized tab, I think you can make a case that the group focus indicator is sufficient to orient the user within what is a contained space. It's a lot easier to locate the subtle selection state within a focused component with subcomponents than it is to find a focus indicator _anywhere_ on the page.
I don't want to overplay this proximity consideration, but to me it is a factor in this discussion. Selection just doesn't need to be as obvious as focus.
But it's entirely possible for someone to interpret the selection indicator as a focus indicator (rightly or wrongly). If we have "may" language, then it reduces something that is usable from failing a misinterpretation of whether a subcomponent focus indicator exists or if it is simply selection is being shown.

On the other hand, I want to make sure that where there _is_ a crappy focus indicator on a subcomponent that needs it (an unsynchronized tablist, a date picker, a multi-select dropdown), that should fail.

So I'm a little torn. Ultimately, the takeaway for me is that the "may" language will get us to a better place while still providing a fail-safe against people starting to confuse selection with focus -- and thus beginning to apply overly vigorous requirements to the selection state.

alastc:
Ok, well, it was you suggesting it so if you withdraw the suggestion, we can carry one. Presumably with "these requirements **_may_** be applied to the sub-component instead"

mbgower:
@bruce-usab in discussions on the Friday call, it was felt that the sub-component information is not really an exception, and fits better in its former location. Can you abide it there with the updated language?
Please have a look at the latest diff to see how it's ended up

bruce-usab (approved):
Thanks for the ping.   I think this is fine.

I agree that the concluding paragraph is not really an exception, and having it in parallel with the actual exceptions is awkward.

mbgower:
> I think this is fine.

Thanks, Bruce. 

@alastc I have confirmed that the normative language in this PR now matches what was approved and merged in the SC itself. So this PR should be considered non-normative.
It basically covers the modifications to the note and some changes to the Understanding document.

guidelines/sc/22/focus-appearance-minimum.html
```diff
@@ -24,8 +24,7 @@ <h4>Focus Appearance (Minimum)</h4>
             </ul>
         </li>
     </ul>
+                
+    <p>Where a user interface component has active sub-components, if a sub-component receives a focus indicator, these requirements can be applied to the sub-component instead.</p>
```
alastc, line 28: Is the note supposed to appear on the index page as well? If so it needs to be here.

alastc:
Saw the email, so I've merged this.

PR #2391, opened 2022-05-11
Dealing with #2355

### LINKED ISSUE #2355: Accessible Auth understanding update
Steve Faulkner pointed out that:

<hr>

I am unclear about the exception:
> Exception: When the cognitive function test is to recognize common objects or content the user provided to the website.

This appears to except forms of Captcha that ask users to choose pictures of boats (for example)

The understanding doc states:

> If a [CAPTCHA](https://www.w3.org/TR/turingtest/) is used as part of an authentication process, there must be a method that does not include a cognitive function test.
 
but then appears to contradict itself
 
> If the test is based on something the website has set such as remembering or transcribing a word, or recognizing a picture the website provided, that would be a cognitive functional test. Recognizing common objects, or a picture the user has provided is a cognitive function test, however, it is excepted at the AA level.

<hr>

I (Alastair) think the "must" needs to be adjusted as it is left over from a previous version.

alastc:
Closes #2355


mbgower (approved):

alastc:
Approved: https://www.w3.org/2022/11/15-ag-minutes#item08

Issue #2292, opened 2022-03-31
Understanding SC 2.4.11: add additional inline-text link examples


jenoe:
Currently there are quite a few examples of button-style components listed in the Understanding WCAG document and just one example showing an issue with an outline focus indicator in a line-break situation.

Inline-text links are arguably a very common usage of this SC and currently a big share of those would not meet the new requirements by default. Therefore I would like to see a few good expamples helping developers and designers to meet the new SC.

mbgower:
@alastc I was just looking at this the other day. The wording of the bounding box is quite prescriptive:
> Where a component consists of disconnected parts, such as a link that wraps onto multiple lines, each part is considered to have its own bounding box.

You'll recall I identified that there are a few different implementations of this, some of which break the bound across lines, and others that enclose the whole link. I wonder if a slight wording change could address this:

> Where a component consists of disconnected parts, such as a link that wraps onto multiple lines, each part can also be considered to have its own bounding box.

JessePinkman48:
Cool

alastc:
Hi @mbgower, we've got discussion on the normative text in #2323, this one is about adding examples in the understanding doc.

@JessePinkman48 a 1px outline would generally pass (with contrast), it is only if you use a border that you get oddities with wrapping links.

We can add examples, it is just lower priority than getting the normative text ready for the next stage.

mbgower:
@jenoe This issue was discussed on working group calls. The general feeling on the working group was that both implementations of link focus were adequate, but existing wording in bounding box was resulting in the use of <code>border</border> failing. I have made a new PR to address this, with an update in the definition of Perimeter and changes in the wording in the link examples. Please review and let me know if this addresses your concern.
https://github.com/w3c/wcag/pull/2420

mbgower:
@jenoe Since the changes were merged, and there has been no feedback from you since, I'm going to assume this is resolved. If you do not think your concern has been addressed, please feel free to re-open.

mbgower:
Just reviewed the Understanding document and realized that the material on links needs some specific text about the 'link on single line' added to the minimum bounding box definition.

jenoe:
@mbgower I think the text and the second example now is much clearer than it was in March. From my side the issue is resolved.

I agree that an (inital) sentence covering the single line situation would be helpful.

mbgower:
@alastc I've made a small PR to address this. Once it is incorporated, this can be closed, since the issue opener has indicated it is resolved.

[...]

PR #2420, opened 2022-05-13
Focus-appearance perimiter and wrapping links

mbgower:
Adjusted the definition of perimeter to accommodate the very visible use of CSS focus for links
@alastc I had a pre-existing branch where I had already begun this work, so just carried out it there.

Addresses https://github.com/w3c/wcag/issues/2292
https://github.com/w3c/wcag/issues/2323


mbgower:
Updated language in Understanding document to match reswizzle of the normative text.

bruce-usab (approved):
Per my 6/28 survey comment, I think this is phrased correct.

bruce-usab:
> Relooking at this, I don't think "component" is necessary in the addition. i think it can just be "link" to clarify anyone reading it cold. As well, I don't think we need the last 'if smaller' wording, so it could read:

@mbgower -- on the call, the concern I heard from @WilcoFiers is that most SC use <q>component</q> so that should be the focus of this SC.  I am more than a little sympathetic with that concern, so my suggestion was to give <q>link</q> explicitly as an example a component where wrapping is an issue.  Otherwise, there is an implication that link and component are synonyms.

The awkwardness is that (text) links are the only kind of component where (text) wrapping is possible -- so <q>link</q> is not merely an example but the only.

PR #2435, opened 2022-05-17
Issue 291

### LINKED ISSUE #291: ARIA11 should reference role="region", and not role="application"
[ARIA11](https://www.w3.org/TR/WCAG20-TECHS/ARIA11.html) ([latest](https://w3c.github.io/wcag/techniques/aria/ARIA11)) lists "application" as a landmark role, but it is [no longer considered to be a landmark role](http://w3c.github.io/aria/aria/aria.html#landmark_roles).
This change was made [12-Nov-2015](http://w3c.github.io/aria/aria/aria.html#other-substantive-changes-since-the-wai-aria-1.0-recommendation):
> 12-Nov-2015: Changed the superclass of application from landmark to structure

Also, ARIA11 should mention the [generic role "region"](http://w3c.github.io/aria/aria/aria.html#region), which _is_ one of the landmark roles.
Note that the HTML `<section>` element maps to region role by default.

Note also that the spec for region mentions that:
> Authors **must** give each element with role region a brief label that describes the purpose of the content in the region.

AT do not expose region role or section element as a landmark _unless it has a label_, so the technique should require a label for using region role or section element as a landmark.

Please also add a link to this [excellent browser extension for landmark navigation](https://www.paciellogroup.com/blog/2017/05/improving-access-to-landmark-navigation/) in the resources section.

fstrr:
closes #291


fstrr:
hey, @mbgower

Thanks for the feedback.

Would "Using semantic HTML elements to identify regions of a page" be a better document title?

The "For instance, if a navigation role is used multiple times on a page, each instance may have a unique label specified using `aria-labelledby` content is in the [current ARIA11 Technique](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA11), so has been there for a good while. That said, the new fourth test procedure does move that away from "may". I think it's probably fair that adding a label is best practice rather than a failure, so taking out that new Test and related Expected Result is probably the way to go.

From memory, Internet Explorer needs roles added to semantic HTML for the landmarks to appear (for example: `<main role="main">`). Every other browser and screen reader combination supports landmarks. I can add some content saying that ARIA doesn't need to be added where HTML can be used.

fstrr:
documents updated per your feedback (and some other things)

mbgower:
@fstrr I have a concern that although you've changed the name of the HMTL5 document, you are using "landmark" primarily in the content. However, I'm not sure anyone would think of the HTML5 equivalents as being "landmarks". I went to change the terms in the article and ran into something of a Pandora's box, in that there does not seem to be consistency of terminology about these elements, even within the HTML5 specification.
Some seem to be called "[sectioning](https://html.spec.whatwg.org/multipage/dom.html#sectioning-content)" elements, but that term does not cover all of them. The [main](https://html.spec.whatwg.org/multipage/grouping-content.html#hierarchically-correct-main-element) element for example is categorized as flow content (as are all these things), which is the parent container for sectioning; however it is not listed as a sectioning element. Using "flow" doesn't seem accurate because there are many things that are not equvalents of landmarks which are also in this category.
Another term I've heard is "semantic html", but that is treated much more broadly in the specification (anything with meaning).
I've also heard "structural semantic content" used, but the word "structural" doesn't appear in the specification.
I'm not sure how to proceed or what to recommend!
Here's how I'd written the first paragraph before deciding to just post this comment:
> <p>Technologies that support <a href="https://html.spec.whatwg.org/multipage/">HTML5</a>. </p>
    </section>
    <section id="description">
      <h2>Description</h2>      <p>The purpose of this technique is to provide programmatic access to sections of a web page. Some HTML semantic elements (or "sections") programmatically identify sections of a page. They serve the same function as ARIA landmarks. Browsers don't expose such tags directly to users, but they they are made available by assistive technology (AT) to users orient themselves to a page and help them navigate easily to various sections of a page. They also can help sighted keyboard-only users navigate to sections of a page using a <a href="https://www.tpgi.com/improving-access-to-landmark-navigation/">browser plugin</a>.</p>

fstrr:
Moved this PR to a draft. Will continue to work on this to clarify it. I'm trying to avoid the word "section" where possible to hopefully avoid confusion with the `section` element and also "sectioning content" which, as you point out, doesn't help us here.

Also, as HTML is a living standard and version-less, I'm just referring to it as HTML instead of HTML5.

(As a side note, I'd like to clear up all the outdated references to HTML and XHTML in the files at some point.)

fstrr:
Does this [ARIA Practices 1.1 ARIA Landmarks Example help](https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/examples/landmarks/HTML5.html)?

> "It is important to understand that many HTML5 sectioning (e.g. main,nav, aside ...) elements by default define ARIA landmarks."

mbgower:
> "It is important to understand that many HTML5 sectioning (e.g. main,nav, aside ...) elements by default define ARIA landmarks."

It depends how technical we want to get. I am not the best person to nail this down, but my undersatnding is as follows... What aria landmarks do is better described as creating an explicit landmark role for the element, whereas the html5 sectioning element creates an implicit role. _Maybe_ by that token we can say that the html sectioning elements create implicit landmark roles? @scottaohara may be able to give us the best suggested wording.

fstrr:
There's also [H97](https://www.w3.org/WAI/WCAG21/Techniques/html/H97.html), which relates to this. 

It also needs a re-write to fix the horizontal scrolling, multiple typos, and an example that seems redundant (slightly surprised that page got approved). I'm working on that.

alastc:
Approved: https://www.w3.org/2023/05/16-ag-minutes#item12

@michael-n-cooper - I updated the links in info & relationships to be relative (they were hard-coded to wcag21). I'll check when it is published, but hopefully that works...

PR #2452, opened 2022-05-20
multiple typos

fstrr:
ref: #273


fstrr:
closes #273

fstrr:
Fixing typos reported in 2017. Also removed one redundant link.

ARIA4 raises a question as to whether we should have our own code examples or whether we should just link to the ARIA spec. Now that ARIA Authoring Practices 1.2 is out, having our own code means more maintenance work for us to update our content.

PR #2453, opened 2022-05-20
typos, outdated content

fstrr:
Closes #275

- typos were originally reported in 2017.
- the specific one for ARIA7 related to a note specifically about a bug fix for IE9 and IE10, so instead of fixing the reported typo, I removed the content and updated the related code example to remove the workaround.
- also fixed horizontal scrolling of an example and two product names


fstrr:
closes #275

Note: 275 references 5 files, but this PR only includes 3. This is because the other two files have been corrected in the interim.

Issue #2314, opened 2022-04-19
Understanding for 2.1.4 Character Key Shortcuts may require some disambiguation


patrickhlauke:
From a recent [discussion on Slack about 2.1.4 Character Key Shortcuts](https://web-a11y.slack.com/archives/C7KTV5CBA/p1650395025786859), it seems there's some confusion about `Shift` as a modifier key. Dumping some of the discussion here verbatim:

> Bruce B  3 hours ago
> Another question about SC 2.1.4. Just want to clarify, do punctuation and symbol characters include those that need the Shift key to be accessed? For example, the "?" key takes two key simultaneous key presses Shift + /. Would this still be considered a single character key shortcut and thus one of the three options would be required (turn off, remap, active on focus only)? (edited) 
>
> Patrick H. Lauke  3 hours ago
yes, it's about the single character/punctuation mark...as on some keyboard layouts, ? may be accessible directly without Shift.
> only non-printable characters/key combinations are excluded
>
> Bruce B  2 hours ago
Thank you @Patrick H. Lauke. That is what I thought but I guess I was second guessing myself because the SC makes a note to point out that both lower case and upper case letters are included, and you need Shift for those, so I wasn't 100% sure if they were making an exception for Shift when it concerns letters. I see now that it is the character itself, not the actual keys that are important.
>
[...]
> Bruce B  36 minutes ago
> @Patrick H. Lauke As you suggested, if on some keyboards ? might be directly accessible without Shift, then could Shift + ? be considered two keys, one of them being a non-printable key, and thus meet 2.1.4 as is? Well, I guess on that keyboard the Shift would produce a different character than ? so it would be impossible to have the key combination Shift + ?? I guess my ultimate question is whether Shift can ever count as a non-printable keyboard key. (edited) 
>
> andrewmacpherson  17 minutes ago
> I'm unsure if it's safe to assume that a question mark will always require shift.
> The USB Human Interface Device scan codes don't differentiate between slash and  question-mark. Assuming a keyboard follows USB-HID, question mark is always shift + slash. That would also be true for a USB keyboard with a dedicated question-mark key.
> For instance, I have a bunch of keyboards with programmable firmware. I could make a dedicated question mark key, but under the hood the keyboard will send USB shift and slash codes to the computer.
> But... not all keyboards follow USB HID :-(
[...]

Issue #1950, opened 2021-07-08
Is a video player considered a single component for purposes of single character shortcuts SC 2.1.4?


mraccess77:
For the purposes of SC 2.1.4 should a video player be treated as one component where when a control in the video player is focused single characters are allowed or must it be when individual buttons are focused only where it is allowed?

Allowing the video player as a whole to be treated like a component would seem to help keyboard users rather than to have to turn off single character shortcuts and would mean that only when the video container is focused would speech users have to navigate out to dictate.

Refer to the discussion on Vimeo's github page.  Today Vimeo's shortcuts work across the entire page not just within the video - but I want to be sure that the right recommendation is being given in the request for them to make a change.

https://github.com/vimeo/player.js/issues/732#issuecomment-876203730

bskibinski:
https://github.com/vimeo/player.js/issues/732#issuecomment-876203730
Julez and other accessibility experts already sum it up well :)

So the video as a whole cannot be treated like a user interface element.
so the fullscreen shortcut is only allowed when the fullscreen button itself is in focus.

This is also my experience with accessibility audits, I thought it was permitted with youtube (keyboard shortcuts only work when the video has focus) but it failed the audit, and we had to disable it.

patrickhlauke:
cross-referencing a lengthy discussion on wai-ig mailing list from 2 years ago ... i think the conclusion at the time was that no, it doesn't https://lists.w3.org/Archives/Public/w3c-wai-ig/2019JulSep/0021.html

> so the fullscreen shortcut is only allowed when the fullscreen button itself is in focus

which then makes the shortcut pointless since the user isn't "shortcutting" anything, they had to reach that button ... and might as well just hit space/enter.

scottaohara:
Agreed.  If the actual control needs to be focused to use the keyboard shortcut, then it might as well not be implemented at all which seems detrimental to helping people.

patrickhlauke:
it's a very weird decision not to see the whole of the video player as a standalone component of sorts. once focus is anywhere in the player, shortcuts should be allowed. just for myself, i can't count the number of times i've taken advantage of these sorts of shortcuts for youtube etc as a user ... including some hidden power-user features like the "rewind 10 seconds" / "skip ahead 10 seconds" which youtube has, but just doesn't provide buttons for (`j` and `l`, where `k` is also a shortcut for play/pause...)

bskibinski:
[...]

PR #2455, opened 2022-05-22
Add note to 2.1.4 Character Key Shortcuts understanding

patrickhlauke:
Make the point that while the SC itself refers to "character keys", it's nothing to do with how many keys are pressed.
Also corrects a typo, gives the document a proper `<title>` (though irrelevant, as this is changed on publication, it's still nice for good housekeeping), and uses `<kbd>` for the characters mentioned.

Closes https://github.com/w3c/wcag/issues/2314


bruce-usab (commented):
Looks good!

patrickhlauke:
@alastc any chance this could be discussed/considered at some point soon?

detlevhfischer (approved):
Should the revised understanding have something on videos processing single printable key shortcuts once they receive focus? The question has been popping up (can't remember where) whether the whole of the video as it receives focus can count as "user interface component" to allow it to be "active only on focus". I can see a good case for allowing that, and I'am not aware of drawbacks. Not sure whether this question should be lumped together with this clarifying rewrite though...

patrickhlauke:
@detlevhfischer i'd keep that thing about video players separate, as a resolution to https://github.com/w3c/wcag/issues/1950

understanding/21/character-key-shortcuts.html
```diff
@@ -15,58 +15,57 @@ <h2>In brief</h2>
...
-<p>A user interacting with a webpage or web app that doesn't use single-character shortcuts doesn't have this problem. Inadvertent strings of characters from the speech application are not interpreted as shortcuts if a modifier key is required. A speech user filling in a text input form may find that a phrase that is accidentally picked up by the speech microphone results in stray text being entered into the field, but that is easily seen and undone. The Resources section of this page contains links to videos demonstrating these types of issues.</p>
+      <p>The intent of this Success Criterion is to reduce accidental activation of keyboard shortcuts. Character key shortcuts work well for many keyboard users, but are inappropriate and frustrating for speech input users – whose means of input is strings of letters – and for keyboard users who are prone to accidentally hit keys. To rectify this issue, authors need to allow users to turn off or reconfigure shortcuts that are made up of only character keys.</p>
```
mbgower, line 22:
```suggestion
      <p>The intent of this Success Criterion is to reduce accidental activation of keyboard shortcuts. Character key shortcuts work well for many keyboard users, but are inappropriate and frustrating for speech input users, whose dictation is interpreted as strings of letters, and for keyboard users, who are prone to accidentally hit keys. To rectify this issue, authors need to allow users to turn off or reconfigure shortcuts that are made up of only character keys.</p>
```

understanding/21/character-key-shortcuts.html
```diff
@@ -15,58 +15,57 @@ <h2>In brief</h2>
...
+			<div class="note">
+				<p>Even though this Success Criterion refers to <q>character keys</q>, note that it's not relevant whether a shortcut can be activated using a single physical key on a keyboard, or if it requires a combination of keys to be pressed. For instance, on most full-size US and UK keyboard, the <kbd>?</kbd> (question mark) symbol is accessed using <kbd>Shift</kbd>+<kbd>/</kbd> (forward slash key next to the right-hand <kbd>Shift</kbd> key). The <kbd>é</kbd> (lowercase e with an acute accent) requires the use of <kbd>AltGr</kbd>+<kbd>e</kbd>. The specific key combination required for certain characters will also vary depending on the user's keyboard layout. However, shortcuts that use these characters still fall under the requirements of this Success Criterion. What matters is that a shortcut relies on a printable character (letters, punctuation, numbers, symbol characters), and not the number of physical keyboard keys that users need to press to trigger it.</p>
```
fstrr, line 24: I think you said last week that you'd note that the AltGr example was specific to Windows. Also, if you're interested, you can nest `kbd` elements when you're referencing multiple keys: https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-kbd-element

patrickhlauke: Done, and done

bruce-usab (approved):
Still looks good to me!

understanding/21/character-key-shortcuts.html
```diff
@@ -15,62 +15,61 @@ <h2>In brief</h2>
...
-<p>A user interacting with a webpage or web app that doesn't use single-character shortcuts doesn't have this problem. Inadvertent strings of characters from the speech application are not interpreted as shortcuts if a modifier key is required. A speech user filling in a text input form may find that a phrase that is accidentally picked up by the speech microphone results in stray text being entered into the field, but that is easily seen and undone. The Resources section of this page contains links to videos demonstrating these types of issues.</p>
+      <p>The intent of this Success Criterion is to reduce accidental activation of keyboard shortcuts. Character key shortcuts work well for many keyboard users, but are inappropriate and frustrating for speech input users, whose dictation is interpreted as strings of letters, and for keyboard users, who are prone to accidentally hit keys. To rectify this issue, authors need to allow users to turn off or reconfigure shortcuts that are made up of only character keys.</p>
```
alastc, line 22:
```suggestion
      <p>The intent of this Success Criterion is to reduce accidental activation of keyboard shortcuts. Character key shortcuts work well for many keyboard users. However, they can be inappropriate and frustrating for speech input users whose dictation is interpreted as strings of letters, and for keyboard users who are prone to accidentally hit keys. To rectify this issue, authors need to allow users to turn off or reconfigure shortcuts that are made up of only character keys.</p>
```

alastc: Trying to reduce sentence complexity.

patrickhlauke: sure. i also like how this makes it less of an absolute statement and tempers it a touch.

patrickhlauke:
@detlevhfischer sorry, a super-late rejoined to this: in a11y slack https://web-a11y.slack.com/archives/C7KTV5CBA/p1736926950552289 somebody pointed out that http://3needs.org/en/testing/code/kb-shortcuts.html seems to omit/not test all printable characters, such as `é` - any chance it could be modified/amended

detlevhfischer:
@patrickhlauke Thanks, I'll pass this on to Telekom MMS, this is not our Bookmarklet I just used it to show its operation on a test page. This thing is that on a German (and presumably also English) keyboard, producing é actually involves hitting two keys so I am not sure if these special characters would really need to be covered - where do you stop? Other languages have other diacritical characters...

patrickhlauke:
> This thing is that on a German (and presumably also English) keyboard, producing é actually involves hitting two keys

well that's the point of this PR here. it's not about "how many physical keys do you need to press", but rather "is it a printable character". and yes, in theory, you need to test the *entire* unicode ranges of printable characters...

besenwagen:
> where do you stop?

In the context of printable [Latin-1 Supplement](https://en.wikipedia.org/wiki/Latin-1_Supplement) characters, 255 seems a much better choice than 191. 🤷‍♀️

PR #2477, opened 2022-05-24
Charter 2022 alternates joint updates

rachaelbradley:
These are the changes from our meeting.


bruce-usab (commented):
Okay, but I have suggested a possible work item for WCAG 2 Final which @alastc did not include with his verbal tally last week.  Something like:
1.3.1.1 simple tables
1.3.1.2 complex tables
1.3.1.3 forms something something something
1.3.1.4 etc.

PR #2485, opened 2022-05-25
Focus-appearance re-structure aligning with recent comments.

alastc:


guidelines/sc/22/focus-appearance.html
```diff
@@ -9,15 +9,15 @@ <h4>Focus Appearance</h4>
...
+                <li><a>Encloses</a> the visual presentation of the user interface component, and</li>
+                <li>has a contrast ratio of at least 3:1 between its pixels in the focused and unfocused states, and</li>
+                <li>has a contrast ratio of at least 3:1 against adjacent colors.</li>
             </ul>
         </li>
         <li>An area of the focus indicator meets all the following:
             <ul>
                 <li>At least as large as the area of a 1 <a>CSS pixel</a> thick <a>perimeter</a> of the unfocused component, or is at least as large as a 4 CSS pixel thick line along the shortest side of the <a>minimum bounding box</a> of the unfocused component, and</li>
-                <li>has a contrast ratio of at least 3:1 against the same pixels in the unfocused state, and </li>
+                <li>has a contrast ratio of at least 3:1 between its pixels in the focused and unfocused states, and</li>
```
WilcoFiers, line 20: I think I'd prefer to keep "the same pixels". If a component moves up 2 pixels during focus, we don't want people arguing there is no change in contrast because the colors didn't change.

PR #2500, opened 2022-06-02
Update G219.html - note to swipe gesture alternative to clarify that swiping alone would not suffice as alternative

### LINKED ISSUE #2406: "directional horizontal swipe" to pass 2.5.7: Dragging Movements?
Fourth example in G219 talks about using a "directional horizontal swipe" to pass 2.5.7: Dragging Movements

This doesn't seem Sufficient as a technique for the User and Functional Needs this SC is trying to tackle...

https://w3c.github.io/wcag/techniques/general/G219

> In a Kanban implementation for process management, tasks can be dragged horizontally from one ‘swimming lane’ to another in order to change the status of tasks. **A directional horizontal swipe over an item moves it to the adjacent lane**. This process can be repeated until the item is in the desired lane.

detlevhfischer:
Adding note to swipe gesture alternative to clarify that swiping alone would not suffice as alternative.

We could also delete swipe als alternative, but I think it useful to cover it since might be a good quick alternative (as long as some other alternative involving more steps, such as a popup menu selection, would also available).

Closes #2406


techniques/general/G219.html
```diff
@@ -20,8 +20,7 @@ <h2>Applicability</h2>
...
-      
+      <p>Such dragging movements are difficult or impossible to carry out for some users with motor disabilities. The alternative to dragging movements operates the underlying function by one or several <a>single pointer</a> activations. A single tap or click may reveal controls (arrows) to move a target in a stepwise fashion; open a drop-down menu where the drop position can be selected; or allow moving it to an ajacent postion by a swipe gesture. (Regarding the swipe gesture alternative, note that the swipe gesture alone would not be sufficient since Success Criterion 2.5.1 Pointer Gestures requires that functionality can operated with a single pointer <strong>without</strong> a path-based gesture, unless a multipoint or path-based gesture is essential.)</p>
```
patrickhlauke, line 23:
```suggestion
      <p>Such dragging movements are difficult or impossible to carry out for some users with motor disabilities. The alternative to dragging movements operates the underlying function by one or several <a>single pointer</a> activations. A single tap or click may reveal controls (arrows) to move a target in a stepwise fashion; open a drop-down menu where the drop position can be selected; or allow moving it to an ajacent postion by a swipe gesture. (Regarding the swipe gesture alternative, note that the swipe gesture alone would not be sufficient since Success Criterion 2.5.1 Pointer Gestures requires that functionality can be operated with a single pointer <strong>without</strong> a path-based gesture, unless a multipoint or path-based gesture is essential.)</p>
```

bruce-usab:
Discussed on backlog call 10/25.  Kernal of question raised worth keeping.  Needs closer review.

patrickhlauke:
Superseded by https://github.com/w3c/wcag/pull/5060

PR #2506, opened 2022-06-03
Issue 2505: re-title H54 to use "term" instead of "word"

### LINKED ISSUE #2505: H54 (Using the dfn element to identify the defining instance of a word) should be "term" not "word"
[H54](https://www.w3.org/WAI/WCAG21/Techniques/html/H54) should be titled "Using the dfn element to identify the defining instance of a term". The code example in the document defines a term rather than one word and the [HTML spec for the `dfn` element](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-dfn-element) says:

> "The dfn element [represents](https://html.spec.whatwg.org/multipage/dom.html#represents) the defining instance of a term."

fstrr:
Closes #2505 

Also:
1. Update TOC page.
2. Update Unusual Words.
3. Update Resources link from HTML 4.01 to HTML Living Standard.
4. Update code example to define more terms that were in it.

### DIFF
--- a/techniques/html/H54.html
+++ b/techniques/html/H54.html
@@ -1,47 +1,76 @@
-<!DOCTYPE html><html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml"><head><title>Using the dfn element to identify the defining instance of a word</title><link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove"></link></head><body><h1>Using the <code class="el">dfn</code> element to identify the defining instance of a word</h1><section class="meta"><p class="id">ID: H54</p><p class="technology">Technology: html</p><p class="type">Type: Technique</p></section><section id="applicability"><h2>When to Use</h2>
-      <p>HTML and XHTML</p>
-   </section><section id="description"><h2>Description</h2>
-      <p>The objective of this technique is to use the <code class="el">dfn</code> to mark the use of a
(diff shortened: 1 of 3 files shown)


michael-n-cooper:
@alastc the TOC is generated, but for the moment it does require the static files (I could maybe change that). It doesn't harm things to have changes to the file since the generator will update, so I'm ok to merge this.

kfranqueiro:
This PR might warrant being closed and re-created:

- `toc.html` files are no longer included in the repo (because they are indeed generated)
- there seem to be far more changes to H54 than just its title; many of them seem to be whitespace but for instance I also see "dfn element" incorrectly changed to "def element"

I would also note that a [counterpoint](https://github.com/w3c/wcag/issues/2505#issuecomment-1192308794) was added on the original issue, regarding whether this change should land at all.

patrickhlauke:
I've made a counter-counter point https://github.com/w3c/wcag/issues/2505#issuecomment-3202152792 as well, just to make it trickier :)

patrickhlauke:
Following on from @kfranqueiro's suggestion about redoing this as a fresh PR, would it make sense to close this one here, and to make a new one to address the original #2505 ? (I'm moving the latter to WCAG 2.x todo)

bruce-usab:
Ping to @fstrr to please close this PR (or create new one).

PR #2510, opened 2022-06-05
Fix link and link text related WAI-ARIA

momdo:
- Currently, URL https://www.w3.org/TR/wai-aria/ is not WAI-ARIA 1.0.
- URL https://www.w3.org/TR/aria-in-html/ is now Using ARIA. 
- [HTML to Platform Accessibility APIs Implementation Guide](https://www.w3.org/TR/html-aapi/) is no longer in active maintenance. See [HTML Accessibility API Mappings 1.0](https://www.w3.org/TR/html-aam-1.0/) instead.


patrickhlauke (approved):

momdo:
Most of the intended changes in this pull request have already been merged, so I will close it.

Issue #2479, opened 2022-05-24
Clarification 1.2.5: Audio Description versus 1.2.7 Extended Audio Description 


PICARDBRB:
For a video with a text media alternative under 1.2.3 Level A, and the author must meet Level AA criterion, and where the audio track of a prerecorded video does not have sufficient pauses for audio description does the content meet 1.2.5 because the author cannot technically embed the audio descriptions as described in the video?   **OR** must the author  use a technique to create a second audio track with descriptions,  the author would need to apply a technique that meets 1.2.7? 

1.2.5 Audio description, describes the requirement to provide audio description for all prerecorded video content in synchronized media.  The intent is to include audio in the existing pauses in the dialogue although this is not stated explicitly in the Success Criterion statement.  

1.2.7 Extended Audio Description: Where pauses in foreground audio are insufficient to allow audio descriptions to convey the sense of the video, extended audio description is provided for all prerecorded video content in synchronized media.

It appears to me that if the author must create a second audio track (and there is no choice) to meet 1.2.5, then what is the purpose of 1.2.7?

Thanks for your help

bruce-usab:
> For a video with a text media alternative under 1.2.3 Level A

Providing a text alternative for time-based media (or not) is not a factor with the situation you describe, because the AA requirement is to provide description.  It improves accessibility, but text the equivalent is above and beyond what 1.2.5  requires.

> ...and the author must meet Level AA criterion, and where the audio track of a prerecorded video does not have sufficient pauses for audio description

The AA SC 1.2.5 requirement is for audio description.  Per the WCAG2 definition for audio description, emphasis added, <q>narration is added during existing **pauses in dialogue**.</q>.

You mention SC 1.2.7, but that might not be the only option.  Does the video require?
> narration added to the soundtrack to describe important visual details that cannot be understood from the main soundtrack alone

If the video is using, for example, orchestral music to set the mood &mdash; then to meet 1.2.5 then the default background audio level needs to be lowered, and voice narration dubbed over it.

On the other hand, if there is constant talking/narration/dialogue in the video, it is probably okay (from a 1.2.5 perspective) because the literal textual requirement of the SC is being met.

> does the content meet 1.2.5 because the author cannot technically embed the audio descriptions as described in the video? 

The content meets 1.2.5 if a version with audio description is available.  That can be the main/default version.  That can be a version with additional narration (dubbed over pauses in dialog).  That can be a version using extended audio description.

[...]

PR #1790, opened 2021-05-10
Tweak understanding for 1.2.3 and 1.2.5

patrickhlauke:
* turn the "During existing pauses..." sentence around, so as not to suggest that an exception exists
* add a note for 1.2.5 that explicitly says that a lack of gaps is not an excuse/exemption
* expanded the note to also touch on cases where AD is insufficient/partial
* in both 1.2.3 and 1.2.5, wrap up example transcript in a `<blockquote>` for clarity
* general code cleanup (removing unnecessary empty lines, mixed spaced+tabs indentation)

Closes https://github.com/w3c/wcag/issues/1768, Closes https://github.com/w3c/wcag/issues/4460, Closes https://github.com/w3c/wcag/issues/782, Closes https://github.com/w3c/wcag/issues/2479, Closes https://github.com/w3c/wcag/issues/3254

If accepted, this PR supersedes https://github.com/w3c/wcag/pull/2517

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/pull/1790.html" title="Last updated on Apr 29, 2024, 11:23 PM UTC (f91c843)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/1790/c9e2cfc...f91c843.html" title="Last updated on Apr 29, 2024, 11:23 PM UTC (f91c843)">Diff</a>

PR #2517, opened 2022-06-07
Addressing issue #2479: add note that if not enough pauses for AD, then authors MUST fall back on alternative or FAIL

DavidMacDonald:
Added language to clarify that The language of the success criterion is an OR statement between the requirement for an alternative for time-based media OR an audio description. Therefore, if an audio description cannot be provided because there are insufficient spaces in the dialog to insert the Audio Description, then an alternative for time-based media would be required.


bruce-usab (changes requested):
Please replace "required" with "necessary for good accessibility" or similar.  I don't agree that "required" is correct.

erfansafari510:
like

patrickhlauke:
in the context of the "other" debate we've been having about videos that don't have sufficient pauses for AD ... this may still be relevant. retracting my thumbs-down, but this may be ground already covered by other more recent PRs that try to more fundamentally address the problem (but have been mired in lengthy discussions) https://github.com/w3c/wcag/issues/4460

this PR *may* be superseded by #1790 if that gets accepted/merged

Issue #2509, opened 2022-06-04
3.3.2: provide failure examples/techniques for `placeholder` and `<select>` relying on default `<option>` only


patrickhlauke:
tangential to https://github.com/w3c/wcag/issues/2484 it seems nowhere in the understanding, nor the techniques, do we have anything about what I believe is now the commonly accepted idea that `placeholder` alone fails 3.3.2 as it's not a permanent label/instruction, and along similar lines, having a `<select>` that is only labelled by virtue of its initial default `<option>`.

https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html

suggest that we could do with a few more examples or failure techniques along these lines.

x-ref https://github.com/w3c/wcag/issues/864 and https://github.com/w3c/wcag/issues/1136

detlevhfischer:
Here is a PR #2532 to adress this, and this is a [rendered version of the changed 3.3.2 understanding text](https://raw.githack.com/w3c/wcag/648209001cae1500e06206d1132edf4e109cdb50/understanding/20/labels-or-instructions.html) with changes icluded (2nd and 3rd para in section Intent)

patrickhlauke:
as this topic keeps bubbling up...any chance we could get this looked at at some point?

patrickhlauke:
another good example of a potential failure of 3.3.2 would be form fields that rely solely on the use of `title` as visible label...

A11yTea:
Hi everyone, I am currently writing an issue for this failure and was surprised that this has not been merged yet. I looked at it and it looks good. I would just change "form control" for "input".

patrickhlauke:
> surprised that this has not been merged yet

the AGWG/WCAG 2.x backlog TF moves in mysterious (often surprisingly slow) ways ;)

A11yTea:
Hi @patrickhlauke !
Do you know if this got lost in the mysterious "AGWG/WCAG 2.x backlog TF" void?

bruce-usab:
@A11yTea would you be willing to draft a proposed failure technique?

I just did a quick search to see what I could find as the bones of failure:

From understanding [contrast minimum](https://www.w3.org/WAI/WCAG22/Understanding/contrast-minimum) and  [contrast enhanced](https://www.w3.org/WAI/WCAG22/Understanding/contrast-enhanced.html):
[...]

Issue #2277, opened 2022-03-22
Ambiguous labels and SC 2.5.3


JAWS-test:
The page https://codepen.io/jaws-test/pen/JjMRwVe contains some labels that I think are difficult to evaluate in terms of 2.5.3, but are very common:
* Label for a radio button to activate an option and the same label for an input field to configure the option.
* One label for 2 fields 
* One label for 3 fields that have a placeholder 
* No label, but a button that serves as a label 

In the following, I am only concerned with the correct answer to satisfy SC 2.5.3. I do not want to discuss the correct answers for SCs 1.3.1, 2.4.6, 3.3.2 or 4.1.2 here.

## Label for a radio button to activate an option and the same label for an input field to configure the option.

What would be the correct name for the first radio button:
* Every (not meaningful but at first glance the visual label)
* Every . day (visible label without content of the input)
* Every 3. day (name according AccName specification)

And for the last radio button:
* Every month
* Every month on
* Every month on Monday

What would be the correct name for the first input field:
* Every
* Every . day
* Every 3. day

Can the name of the input be the same as the name of the radio button?

Or would 2.5.3 not be applicable because too complex?
Or should such forms be avoided because they cannot fulfill 2.5.3 and 1.3.1, 2.4.6, 3.3.2, 4.1.2 at the same time?

## One label for 2 fields 

What is correct name for the first input field
* First name (does not match with visible label)
* First and last name (is not meaningful)
* First and last name, here: First name (too long)

## One label for 3 fields that have a placeholder 

What is correct name for the last input field
* Date (not meaningful)
* dd (not meaningful, according to Understanding placeholder can serve as name)
* Date: dd (probably acceptable, contains both forms of the visible label)
* Date: day (maybe the best solution, but ignores the placeholder as name)

## No label, but a button that serves as a label 

What is correct name for the search field
* Search (the group name)
* e.g. CA 95383 (the placeholder)
* Search for zip code (the button)
* zip code (because "Search" is already in the group name)

JAWS-test:
@mbgower My point was not to learn how to do it better (I provided examples for that myself), but to learn what would be a violation and what would not be a violation of SC 2.5.3 when evaluating existing pages with this design according to WCAG

mbgower:
Okay, before proceeding with that exercise, can you clarify that you think these all pass Labels or Instructions?

JAWS-test:
> can you clarify that you think these all pass Labels or Instructions?

[...]

PR #2532, opened 2022-06-24
Update labels-or-instructions.html

detlevhfischer:
In response to #2509 I have tried to revise the current [Understanding text of 3.3.2 Labels and instructions](https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html), section Intent, adding the reqirement that the label should be permanently available. 

- `placeholder` (not sufficient as label since it disappears)
- default `option` of `select` (generally not sufficient as visible label unless the context of the form makes its meaning obvious)

I can think of cases where mandating an extra visible label for `select` might be wrong, as in cases where selects are embedded in inline texts. This may happen in language tests or probably other contexts (say, "It was `<select><option>wonderful</option><option>so, so</option><option>horrific</option></select>` to meet so many new people.") Maybe not a good example, but you get the idea.  I would shy away from mandating an explicit visible label for `select` in all possible cases. 

I would add examples to underpin these additions once there is a bit of feedback on the proposed changes, and it appears the changes makes sense to other WG members.

### DIFF
--- a/understanding/20/labels-or-instructions.html
+++ b/understanding/20/labels-or-instructions.html
@@ -23,6 +23,8 @@
          only when the individual control has focus especially when instructions are long and
          verbose.
       </p>
+     <p>Visible labels must remain present when users interact with a form control. Text provided on fields via the placeholder attribute disappears once users enter text. The placeholder text may provide helpful examples of the type of content expected, but is not sufficent on its own as a visible label.</p>
+     <p>A select must generally have a separate label to identify the category of options included. Thus, the label "City" may describe the category under which all possible options (names of cities) fall. Depending on the context, users may not be able to infer the category (and the meaning of the select)  if only the option is presented. The default option is generally not sufficient as visible label unless the context of the form makes its meaning visually obvious for all options that the user may select.</p>
       
       <p>The intent of this Success Criterion is not to clutter the page with unnecessary information
          but to provide important cues and instructions that will benefit people with disabilities.
@@ -78,7 +80,6 @@
       
       
       <ul>
-         
          <li>A field which requires the user to enter the two character abbreviation for a US state
             has a link next to it which will popup an alphabetized list of state names and the
             correct abbreviation.


JAWS-test:
@detlevhfischer 

"It was ... to meet so many new people" is a good label for the select, so I see no problem for 3.3.2. 

Problems for 2.5.3. with such labels I have already [addressed](https://github.com/w3c/wcag/issues/2277)

detlevhfischer:
> "It was ... to meet so many new people" is a good label for the select, so I see no problem for 3.3.2.
But that was exactly my point: the surrounding text makes an **extra** label describing the select category superfluous.

patrickhlauke (commented):
While I like the additions, I'm wondering if they get a bit too specific for the high-level "intent" section. It might be better to have just the first sentence "Visible labels must remain present when users interact with a form control." in the intent, and then use the placeholder and select as failure examples in the bullet list later on?

understanding/20/labels-or-instructions.html
```diff
@@ -32,6 +32,8 @@ <h2>Intent of Labels or Instructions</h2>
          only when the individual control has focus especially when instructions are long and
          verbose.
       </p>
+     <p>Visible labels must remain present when users interact with a form control. Text provided on fields via the placeholder attribute disappears once users enter text. The placeholder text may provide helpful examples of the type of content expected, but is not sufficent on its own as a visible label.</p>
```
```suggestion
     <p>Visible labels must remain present when users interact with a form control. Text provided on fields via the placeholder attribute disappears once users enter text. The placeholder text may provide helpful examples of the type of content expected, but is not sufficient on its own as a visible label.</p>
```

understanding/20/labels-or-instructions.html
```diff
@@ -32,6 +32,8 @@ <h2>Intent of Labels or Instructions</h2>
          only when the individual control has focus especially when instructions are long and
          verbose.
       </p>
+     <p>Visible labels must remain present when users interact with a form control. Text provided on fields via the placeholder attribute disappears once users enter text. The placeholder text may provide helpful examples of the type of content expected, but is not sufficient on its own as a visible label.</p>
```
fstrr: I think these remove some possible wiggle room.

```suggestion
     <p>Visible labels must remain visible when users interact with the label's related form control. Text provided on fields via the <code>placeholder</code> attribute disappears once users enter text. The <code>placeholder</code> text may provide helpful examples of the type of content expected, but is not sufficient on its own as a visible label.</p>
```

PR #2539, opened 2022-07-06
xslt stylesheet fix for json output

np-at:
I added a fallthrough case in the handling of the `version` property to output an empty array instead of being a no-op (which resulted in invalid json).

Prior observed output from `json` job:
```		{
			"id": "WCAG2:text-alternatives",
			"alt_id": ["text-equiv"],
			"num": "1.1",
                        "versions": ,       <==================
			"handle": "Text Alternatives",
			"title": "Provide text alternatives for any non-text content so that it can be changed into other forms people need, such as large print, braille, speech, symbols or simpler language.",
			"successcriteria": [				
. . .
```

After fix: 

```
		{
			"id": "WCAG2:text-alternatives",
			"alt_id": ["text-equiv"],
			"num": "1.1",
			"versions": [],
			"handle": "Text Alternatives",
			"title": "Provide text alternatives for any non-text content so that it can be changed into other forms people need, such as large print, braille, speech, symbols or simpler language.",
			"successcriteria": [
. . .
```

I couldn't any guidance on how to do a technical pull request, so let me know if I need to resubmit in a different format.


michael-n-cooper:
It appears this overlaps with a fix I had made: https://github.com/w3c/wcag/commit/50cae78dc19d33b10f0ffa5bdef6f6c2319e1e26. My fix is more basic but sufficient, closing this draft PR.

PR #2549, opened 2022-07-15
Adding H99 to multiple ways

alastc:
An update to put H99 (Provide a page-selection mechanism) under the Multiple Ways SC.

Effectively this technique wasn't referenced by any understanding document, so this is being added as an advisory technique for multiple-ways. Patrick updated the linking to use the JS file, and Matt Garish added a paragraph on usage for epub.

The linking is done, just need to check the content.


patrickhlauke:
@alastc just checking if this is still "active", and if so this will probably need to be tweaked now that technique association is done separately/differently

EDIT: done just now

understanding/20/multiple-ways.html
```diff
@@ -55,6 +55,13 @@ <h2>Benefits of Multiple Ways</h2>
             to use a hierarchical navigation scheme that may be difficult to understand.
          </li>
          
+         <li>For readers of digital publications like EPUBs, being able to scan and navigate to the
+         	sections of a work using a table of contents makes it easier both to locate content and
+         	to find their position again. Providing an index allows readers to locate specific
+         	information they are interested in by topic. And in mixed print and digital
```
ljoakley:
```suggestion
         	information they are interested in by topic. In mixed print and digital
```

bruce-usab (approved):

PR #2567, opened 2022-08-04
Update general-flash-and-red-flash-thresholds.html

rachaelbradley:
Potential updates to the notes under general flash and red flash thresholds.


michael-n-cooper (changes requested):
There is a MathML version of the luminance contrast algorithm in guidelines/relative-luminance.xml, which should probably also be updated.

alastc:
@michael-n-cooper Do we need to to update the relative luminance MathML?  We haven't updated that, have we? 

This update adds a some commentary about non-sRGB spaces, but doesn't update the luminance calc.

PR #2568, opened 2022-08-05
Tweaks to improve inter-rate reliability

mbgower:
Although the intention was that the first set of bullets in Focus Appearance applied to the whole focus indicator, during testing I found it was possible to mis-interpret that ( for instance, find SOME of the pixels of the focus indicator passing all three bullets).
Adding "entire" resolves this potential misinterpretation. It makes it clear that where the entire focus appearance does not pass, the second scenario comes into play: an area must be calculated and the criteria apply to that area.

- Made scope of first set of bullets more emphatic.
- Made 3rd bullet wording match second set of bullets.
- Made tweak to definition to make it read a bit less awkwardly on definition substitution


mbgower:
An example which may help illustrate why I think this emphatic wording is necessary...
This example uses the defaults from Firefox, but ignore that and imagine these are custom.
Screenshot 2022-07-18 at 7 23 47 AM
^ Expanded select list of four car manufacturers with no focus, and Saab selected using contrasting grey to indicate selection. List has a 1 px dark grey border
Screenshot 2022-07-18 at 7 24 14 AM
^ Same select but with focus added using a light blue surrounding indicator plus a dark blue indicator which replaces the pixels of the dark grey border AND expands 2px beyond the pixels into the whitespace around them. Grey indicator for selected item is replaced with dark blue for all pixels and the letters of the selected item are inverting to white to retain text contrast

---

The light blue indicator clearly surrounds the entire component, so one can say that part of the indicator surrounds, but then those light blue pixels do not contrast with the background sufficiently for sub-bullet 2. We can’t pick and choose some pixels that pass the first sub-bullet and then choose other pixels that pass the second, right? Are we allowed to say that as long as some of the pixels meet all three of the sub-bullets, it passes? I don’t think that’s our intent, but it’s a legitimate question, given the current wording.
 
As well, the dark blue overlaps with the grey component border. It does include MORE pixels, such that part of it also surrounds the component…
 
My read on this is that all the pixels have to meet all three sub-bullets. I believe that was the intent, and my changes make it emphatic.
 
I have similar qualms when applying the first bullet’s criteria to the subcomponent. CLEARLY, the light grey to dark blue selection is sufficient contrast wise, but since the pixels are exactly the same, it doesn’t “solidly bound or surround”. I originally used the word encompasses, which I DO think would have allowed the change of every pixel, but we dropped that.
 
So my read (for both component and subcomponent) is that you have to go to area, both of which pass that.

guidelines/sc/22/focus-appearance.html
```diff
@@ -7,11 +7,11 @@ <h4>Focus Appearance</h4>
     
     <p>When a <a>user interface component</a> has keyboard focus, one or both of the following is true:</p>
```
WilcoFiers, line 8: This, as AGWG had voted and agreed on in [February 22nd](w3.org/2022/02/22-ag-minutes#item06)

```suggestion
    <p>When the keyboard focus indicator of a <a>user interface component</a> is visible, one or both of the following is true:</p>
```

guidelines/sc/22/focus-appearance.html
```diff
@@ -7,11 +7,11 @@ <h4>Focus Appearance</h4>
...
     <ol>
-        <li>The <a>focus indicator</a>:
+        <li>The entire <a>focus indicator</a>:
```
WilcoFiers, line 10:
```suggestion
        <li>The entire <a>focus indicator</a> meets all the following:
```

guidelines/sc/22/focus-appearance.html
```diff
@@ -7,11 +7,11 @@ <h4>Focus Appearance</h4>
...
             <ul>
                 <li><a>encloses</a> the visual presentation of the user interface component, and</li>
                 <li>has a contrast ratio of at least 3:1 between the same pixels in the focused and unfocused states, and</li>
-                <li>has a contrast ratio of at least 3:1 against adjacent colors.</li>
+                <li>has a contrast ratio of at least 3:1 against adjacent non-focus-indicator colors.</li>
```
WilcoFiers, line 14: I think this is confusing, more than it is helpful.

WilcoFiers (changes requested):

guidelines/sc/22/focus-appearance.html
alastc: I'm fairly sure that was superseded, I'll look it up.

mbgower: Yeah, I think you're right. it's just a result of this branch not being updated when other changes were made, I think.

guidelines/sc/22/focus-appearance.html
mbgower: I think "meets all the following" is redundant given "the entire" in front of it.Not something I particularly care about (in other words, if you feel strongly, I can live with it), but I just don't like adding more words that add little value.

guidelines/sc/22/focus-appearance.html
mbgower: It matches the wording in the second set of bullets and it's pretty important because you can have a situation where you could arguably fail a two-tone focus indicator (or just a non-solid colour) because someone could argues there isn't 3:1 contrast between adjacent colours. Saying "non-focus-indicator" makes it emphatic that one is ONLY assessing this bullet against something adjacent to the focus indicator, not pixels within the indicator itself.

GreggVan:
I hope there will be a diagram in the    Understanding WCAG 2.x  document because I don’t know what each of  those line items means.  

Can you help me here for now…
If we have a button of color A
And a background of color B
And a focus indicator of color  C
Q1:  What has to contrast with what - when? 
Is it     3:1 between A and B    
AND   3:1 between  C with A 
AND   3:1 between  C with B    ? 

Or is it something else?

Q2:  And if the button has graded colors AND the background is an image behind the button —  I presume the rules are applied locally at each point around the edge of the button over the background - yes? 


Thanks 


gregg

———————————

guidelines/sc/22/focus-appearance.html
alastc: Tracing forward from Feb, the start was changed when we accepted the "[reboot](https://docs.google.com/document/d/1TI_DsJjfg9RW_A9C1XfqgtfYqNMfnxeCz6lxjpOQ9rk/edit#heading=h.tcutgbbczlhr)" version.

guidelines/sc/22/focus-appearance.html
alastc: @mbgower isn't that undone by starting off the bullets with "entire"? 
If you are looking at the entire focus indicator, you have to take the edges of that for adjacent, even if it creates contrast within itself by having light & dark areas?
Either way it's going to fall through to the measures bullets.

alastc:
Hi Gregg,

The changes here are to make it clearer that the 1st set of bullets applies to the entire indicator (not just parts of it).

Q1: This hasn't changed for a long time, the indicator "C" needs to contrast with the pixels it replaces (i.e. the background of the indicator, wherever that is), and have adjacent contrast with A&B unless it is at least 2px thick.

In the 1st set of bullets this is stricter and it is just asking for contrast with A&B, no size measure. But in the second set of bullets the requirement hasn't changed and is extensively covered in the understanding doc.

Q2: Yes.  Coloured backgrounds add a lot of complexity, a useful technique in that scenario is to pull the indicator within the background of the component.

GreggVan:
OK - understood

Two things
1) we still should put a diagram in Understanding to explain what all the terms used in the SC mean.  


2) check my understanding — because I am misunderstanding or I think there might be a problem…  


So if the focus indicator is 2 px wide — it can be the same color as the button?   
(Or am I understanding it wrong?)


If I am understanding it correctly 
Isnt this a problem for a person who has low vision and use a magnifier that does not track the focus cursor (because that makes them dizzy and is disorienting to them) 

They are using a form with a submit button. — When the focus could shifts to it - there would be no indication that it has focus except that it is bigger  than it used to be (the indicator is the same color as the submit button so it just make it look like a bigger button)   ( So the low vision person would need to remember how big the submit button was last time they looked at it with their magnifier) ? 



Am I understanding it correctly?
Is there a problem for this person — or people who don’t use a magnifier but just use "close vision"  (leaning into the screen to look at it with tunnel-like close view vision) 

If there is I’m not sure we can do something about it  at this point — but trying to understand if I understand it or not. 

Thx


gregg

———————————

alastc:
Hi Gregg,

I think you're understanding correctly, that kind of scenario can be an issue for people with low vision. 

However, you have to also consider what kind of indicators you'd be banning if you require adjacent contrast without allowing for thickness.

Although when the colours are identical it isn't great:
Three black buttons, the middle one is visibly thicker.

If you required adjacent contrast it wouldn't allow for very thick but non contrasting outlines, these would fail:
Three green buttons, the centre one has a thick blue outline.

three grey buttons with a thin black border, the centre one has a thick black border.

GreggVan:
Thanks  - the examples are very helpful in this discussion 

Comments below marked      GV:    

> On Aug 11, 2022, at 2:59 PM, Alastair Campbell ***@***.***> wrote:
> 
> 
> Hi Gregg,
> 
> I think you're understanding correctly, that kind of scenario can be an issue for people with low vision.
> 
> However, you have to also consider what kind of indicators you'd be banning if you require adjacent contrast without allowing for thickness.
> 
> Although when the colours are identical it isn't great:
>  <https://user-images.githubusercontent.com/216630/184247855-d02e947d-ab3c-47f8-a178-84db4c4c61d7.png>    GV: yes - not great when they are right next to each other — and completely unusable if there was only one button there (like a submit button) with nothing else to compare it to. 
   

> If you required adjacent contrast it wouldn't allow for very thick but non contrasting outlines, these would fail:
>  <https://user-images.githubusercontent.com/216630/184249601-0d13e3f3-2847-4ced-b545-e1d47256bd3e.png>   GV: Yes — for those with color vision — this works.  But we elsewhere say that we should not have any content that requires color vision.   So look at that in black and white.  It begins to look like example 1 above.      If you put just one pixel of white in between the button and the highlight though it would be so much easier to see — and would pass fine. 


>  <https://user-images.githubusercontent.com/216630/184248878-7907f7ef-f98e-4007-9147-f458972cb2e2.png>   GV: why would this one fail?   It has contrast with both the button and the background.   (And it would also be clear if the button was not next to the others.

Or am I missing something?

 
(

alastc:
> But we elsewhere say that we should not have any content that requires color vision.   So look at that in black and white.  It begins to look like example 1 above. 

Not that much:
image

Contrast is a pretty blunt tool, AFAIK there isn't a problem with blue-green differentiation unless you're a very rare mono-chromate (is that the right term?). The contrast is on the edge (2.8:1) but given the size difference it's a difficult argument with a designer.

> why would this one fail?

You might need to zoom in, but those buttons have a 1px dark border, and then the focus style is an adjacent thick dark outline. because the border expands by 3px it looks different, but it still doesn't have adjacent contrast.

The primary bullets we are conveying what is desirable: adjacent contrast regardless of thickness.
Then if people are skirting closer to the minimum we have a reasonable compromise if they want to get into the complication of sizing/thickness etc.

GreggVan:
I agree that what is below would / should pass

And as to the argument - you could say that the 1 px border has been expanded into a 4 px border — and therefore should pass. 

And I don’t think that anyone would challenge that as passing.   So I’m not sure what the problem is there.


On the other hand — the suggested language would cause the other two examples to Pass  — when they clearly should fail.  

If you put the button with focus and without focus about 8 inches apart and asked which one had focus — (for the other two examples - not this one) 
You would have to look back and forth a number of times to figure out which had focus — if you could at all with confidence. 
It is only if they are right next to each other that you can tell and then only for one of them I think. 


Since there is a fix for this— what is the downside of the fix?   That is what I am missing.  

Also it would be a good idea to mention the one or two pixel gap between button and highlight.  That also is both very good and makes lots of things work that wouldn’t otherwise.  NOT a requirement to do so - but a mention of it as a powerful idea.

Best

G

guidelines/sc/22/focus-appearance.html
mbgower: Oh, for the two-tone, you're right. Some of that is going to fail (by design) so you would be assessing the area criteria.

mbgower:
@GreggVan 
> Also it would be a good idea to mention the one or two pixel gap between button and highlight.  That also is both very good and makes lots of things work that wouldn’t otherwise.  NOT a requirement to do so - but a mention of it as a powerful idea.

You should take a few minutes to look at the current state of the [Understanding document](https://w3c.github.io/wcag/understanding/focus-appearance.html). I feel like most of your concerns are addressed by it, but if you feel that's not the case, I'd love to find out where so it can be updated.

PR #2574, opened 2022-08-08
Replace misleading image  in 1.4.11 understanding with clearer example, tweak all other button example images with bigger text

### LINKED ISSUE #2494: Understanding Success Criterion 1.4.11: Non-text Contrast - Figure 15
- URL: https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html
- Issue description: Figure 15 is marked as non passing use of color, but it's actually passing 1.4.1.
- Screenshot:
Screenshot of Figure 15 with description: Figure 15 The change of background within the component is not in scope of non-text contrast. However, this would not pass Use of color

- Expected result: This is actually not failing 1.4.1 Use of Color, because text color is changing from white to black, which is not only a change in hue, but also a change in lightness (> 3:1).
As per 1.4.1
> If content is conveyed through the use of colors that differ not only in their hue, but that also have a significant difference in lightness, then this counts as an additional visual distinction, as long as the difference in relative luminance between the colors leads to a contrast ratio of 3:1 or greater.

Keeping text to be white, we will have a contrast ratio issue, but at least it will fail the Use of color criterion and will be consistent with description below.

Giacomo

patrickhlauke:
the current image to illustrate how a simple change of background colour is not in scope of 1.4.11 non-text contrast actually muddies things, because the focused button also has a different text colour and a white halo/shadow around the text.

the current confusing image

replacing the image with a cleaner example image that *only* changes the background colour - and, to keep it more consistent with the other examples, it uses the light blue for the non-focused buttons and the dark blue for the focused one.

the amended example image

Closes #2494

EDIT: in addition, this also makes all "Example" text in the button images bigger (21px) and bold, to avoid any doubt that they'd pass 1.4.3 as "large text"


fstrr (approved):

patrickhlauke:
expanded this PR to also tweak all "Example" blue buttons, making their text bigger/bold

fstrr (approved):

patrickhlauke:
@alastc any chance this could be discussed and hopefully merged at some point soon?

patrickhlauke:
as this example has just caused some confusion in a WAI-IG discussion...can this please be looked at soon? @alastc @mbgower

detlevhfischer:
> EDIT: in addition, this also makes all "Example" text in the button images bigger (21px) and bold, to avoid any doubt that they'd pass 1.4.3 as "large text"

Since size of text in images is always difficult to judge across platforms, I wonder why we can't have an example that meets 4,5:1? Apart from the muddying aspects noted, this was the case in the old example.

patrickhlauke:
@detlevhfischer if somebody has the time/inclination to redo all the example images *again* ... sure.

mbgower:
@detlevhfischer I suggest we could add a note to the Understanding stating that all images examples of buttons are intended to represent large text, and so only require a 3:1 contrast ratio against the button background, or similar

patrickhlauke:
@alastc @iadawn I've just noticed that the version on https://w3c.github.io/wcag/understanding/non-text-contrast.html is still showing the old (and, in the case of figure 15, incorrect) version (even though https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast.html is correct). any idea why the github.io version lags behind the actual repo content?

PR #2578, opened 2022-08-09
Focus appearance updates from Wilco

alastc:
Some parts of: https://docs.google.com/document/d/1kUjK65oIWBYXN1x_4HchCMg3ujcfLwD7WzfO7G2uM9s/edit#

### DIFF
--- a/guidelines/sc/22/focus-appearance.html
+++ b/guidelines/sc/22/focus-appearance.html
@@ -7,9 +7,9 @@
     
     <p>When a <a>user interface component</a> has keyboard focus, one or both of the following is true:</p>
     <ol>
-        <li>The <a>focus indicator</a>:
+        <li>The <a>focus indicator</a> meets all the following:
             <ul>
-                <li><a>encloses</a> the visual presentation of the user interface component, and</li>
+                <li><a>encloses</a> the user interface component, and</li>
                 <li>has a contrast ratio of at least 3:1 between the same pixels in the focused and unfocused states, and</li>
                 <li>has a contrast ratio of at least 3:1 against adjacent colors.</li>
             </ul>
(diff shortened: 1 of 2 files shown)


WilcoFiers (changes requested):
Can you fix the typo in the minimum bounding box definition too? "(such **has** hypertext links)" should be "(such **as** hypertext links)"

guidelines/sc/22/focus-appearance.html
```diff
@@ -30,8 +30,10 @@ <h4>Focus Appearance</h4>
         <li>The focus indicator is determined by the user agent and cannot be adjusted by the author, or</li>
         <li>The focus indicator and the indicator's background color are not modified by the author.</li>
     </ul>
+  
+    <p class="note">The position, size, and shape of user interface components and sub-components is determined by its visual presentation. The visual presentation includes, but is not limited to the component’s visible content, border, component-specific background, and effects emanating from the visible component such as shadows or glow effects.</p>
```
patrickhlauke, line 34: this sets out to initially define the position/size/shape of the UIC. but this seems to come out of the blue, as nothing up to this point talked about any of those things. perhaps it needs a slightly different intro ... something more like "What is perceived as the user interface component and any sub-components (to determine whether or not a focus indicator actually encloses it) depends on its visual presentation." or similar? Basically, need to somehow glue to together so it's clear this tries to explain what needs to be taken into consideration when determining what the focus indicator actually needs to *enclose*?

mbgower:
Thanks for the tweaks that removed "is not limited to" and "size".
You may want to think about saying "area" instead of "size" in the other usage, resulting in it being "...(to determine enclosure or area..." . Otherwise, looks good

alastc:
Closing as we merged a different branch that covered the same things.

PR #1040, opened 2020-02-07
Errata/clarification of 2.2.1 Timing Adjustable extension

patrickhlauke:
Based on the discussion in https://lists.w3.org/Archives/Public/w3c-wai-gl/2020JanMar/0161.html this is a first attempt at making it more explicit what the confusing "at least ten times" means. It appears this has been misinterpreted by some as "ten times" meaning that a site needs to provide ten separate warnings/options to extend, when in fact it appears this, like the previous point about adjusting timing, is meant as "ten times the length of time".

PR #2581, opened 2022-08-09
Errata/clarification of 2.2.1 Timing Adjustable extension (counterproposal)

patrickhlauke:
A counterproposal to https://github.com/w3c/wcag/pull/1040 that aims to keep backwards-compatibility with the 2.0/2.1 version which some are arguing did in fact intend the "ten times" in the extend bullet to refer just to the number of opportunities to extend).

In the original, there is no clarification of how long each extension should be, which opens this particular bullet up for being completely gamed by authors (and lead to paradoxical situations where a site that passes this by virtue of offering 10 really short extensions still doesn't actually help real-world users that need more time, not in the same way that the other bullets do).
For backwards-compatibility, this keeps the idea of 10 opportunities (if that was indeed the original intent), but adds a further clause about the length of the extensions (and yes, technically this would only need 9 opportunities to be on par with the "adjust" bullet, but for backwards-compatibility this keeps the "ten").

This would then make the bullet more useful (it actually does enforce providing users with more time, rather than being gameable), and be backwards-compatible (while being stricter).

### DIFF
--- a/guidelines/sc/20/timing-adjustable.html
+++ b/guidelines/sc/20/timing-adjustable.html
@@ -32,7 +32,8 @@
          
          <p>The user is warned before time expires and given at least 20 seconds to extend the
             time limit with a simple action (for example, "press the space bar"), and the user
-            is allowed to extend the time limit at least ten times; or
+            is given ten opportunities to extend the time limit (with each extension having the
+            same duration as the original time limit); or
          </p>
          
       </dd>


patrickhlauke:
I would suggest re-surveying this with four options:

- https://github.com/w3c/wcag/pull/1040
- https://github.com/w3c/wcag/pull/2581
- no PR for this, but: just change `and the user is allowed to extend the time limit at least ten times;` to `and the user is given ten opportunities to extend the time limit;` (and also give a rationale in the understanding document why the WG thinks this will help users in any way)
- leave the SC as is (but explain what is meant in the understanding, *and* give a rationale why the WG thinks this will help users in any way)

bruce-usab:
Sorry, I am not in favor of this one either.

> with each extension having the same duration as the original time limit

If the original time limit was reasonably generous, this qualifier is needlessly excessive.

patrickhlauke:
> If the original time limit was reasonably generous

big "if" though, since the SC doesn't say anything about the original time limit either

guidelines/sc/20/timing-adjustable.html
```diff
@@ -32,7 +32,8 @@ <h4>Timing Adjustable</h4>
          
          <p>The user is warned before time expires and given at least 20 seconds to extend the
             time limit with a simple action (for example, "press the space bar"), and the user
-            is allowed to extend the time limit at least ten times; or
+            is given ten opportunities to extend the time limit (with each extension having the
```
backwardok, line 35: Probably want to keep the "at least" language in there
```suggestion
            is given at least ten opportunities to extend the time limit (with each extension having the
```

GreggVan:
The intent of the SC was to give users 10 time more than the time limit typically provided.

If 90 minutes is excessive for the typical user - then that could be reduced to what is needed.

If 90 minutes is how long a typical user needs to complete whatever activity they need the resource for - then the user with disabilities should be given the option of 10 times that time to complete the activity.  And that was not seen as excessive for people with some disabilities.


NOTE 

 if they are talking about a resource like a movie — the time limit would only apply to the time needed to operate the movie - not watch it. 

If they are talking about a book or some thing that you can typically read in 90 minutes - then indeed the SC was intended to allow for people with disabilities to be able to have up to 10 times that length to read the read the thing.


It is ok to clarify the SC - or to disagree.  But we should not now be changing it or reinterpreting it to be other than what was intended when the working group reached consensus and added it.  As co-chair at the time -  I can attest to what was intended, and it can also be seen in the previous bullet what was intended and also in the Understanding document.      

BUT in applying it — 
it was intended to give 10 times the time for the person to complete some activity.    We can clarify that.  
It was not intended to give them 10 time the time it takes a roast to cook (but yes 10 times the time to prepare it) — if you will.      

 So if there was a form — 10 times the time to complete it
If there was a wait time for a response from the organization - no - not 10 times.    
And not 10 times the warranty period 
but yes 10 times the time to fill out a warranty claim.

I hope this helps

Gregg Vanderheiden
***@***.***

cstrobbe:
Alternative rewording:

> The user is warned before time expires and given at least 20 seconds to extend the time limit **with the default duration by means of** a simple action (for example, "press the space bar"), and the user is allowed to extend the time limit at least ten times;

This adds "with the default duration" in the middle of the sentence instead of adding a longer parenthesis at then end, and replaces "with a simple action" with "by means of ..." to avoid repetition of "with".

GreggVan:
That works 

Or switch the "by" and "with"  to read 

> The user is warned before time expires and given at least 20 seconds to extend the time limit by the default duration with a simple action (for example, "press the space bar"), and the user is allowed to extend the time limit at least ten times;
> 



gregg

———————————

GreggVan:
Or even a bit clearer — add ''another"   to either version 

> The user is warned before time expires and given at least 20 seconds to extend the time limit by another default duration with a simple action (for example, "press the space bar"), and the user is allowed to extend the time limit at least ten times;
> 
> The user is warned before time expires and given at least 20 seconds to extend the time limit with another default duration by means of a simple action (for example, "press the space bar"), and the user is allowed to extend the time limit at least ten times;



gregg

———————————
Co-Founder Raising the Floor. http://raisingthefloor.org <http://raisingthefloor.org/>
The Global Public Inclusive Infrastructure (GPII) http://GPII.net <http://gpii.net/>
The Morphic project  https://morphic.org <https://morphic.org/>

bruce-usab:
My suggestion for making this sort of normative change would be to replace the current Extend bullet with two options:

Current:
> The user is warned before time expires and given at least 20 seconds to extend the time limit with a simple action (for example, "press the space bar"), and the user is allowed to extend the time limit at least ten times; or

Proposed:
> &bull; Extend1of2:&nbsp; The user is warned before time expires and given at least 20 seconds to extend the time limit with a simple action (for example, "press the space bar"), and the user is allowed at least ten opportunities to extend the time limit; or
> &bull;
> Extend2of2:&nbsp;  The user is warned before time expires and given at least 20 seconds to extend the time limit with a simple action (for example, "press the space bar"), and the user is allowed enough of these opportunities to extend the time limit to at at least ten times the length of the original default setting; or

FWIW, I don't think we have time for a normative change to 2.2.1.








Extend

GreggVan:
I agree on normative change at this point.  The chairs can chime in to confirm that or not.  



But to clarify your suggestion….  are you suggesting changing one bullet into two?

If two - then they both say the same thing.   The first is just not very clear - and that is what the discussion is all about. 
This provision was for people with cognitive, language, and learning disabilities  -  

Changing the current text to your second version ...
>  The user is warned before time expires and given at least 20 seconds to extend the time limit with a simple action (for example, "press the space bar"), and the user is allowed enough of these opportunities to extend the time limit to at at least ten times the length of the original default setting; or
> 

..  could be done though — since this would not be a normative change since that was the original intent of the SC and would provide the same coverage for people with cognitive, language, and learning disabilities  (and other disabilities) 



gregg

———————————

bruce-usab:
> But to clarify your suggestion….  are you suggesting changing one bullet into two?

Yes.

> If two - then they both say the same thing.   

No, that is not correct.  The current Extend -- as written -- requires ten opportunities.  And my opinion is that the SC should have an the editorial change to make this explicit.

Then we could _add_ a bullet which gives the choice of the 20 second warning and opportunity for extended time of ten times default.

Current Adjust bullet (emphasis added):

> &bull; Adjust:&nbsp; The user is allowed to adjust the time limit **before** encountering it over a wide range that is at least ten times the length of the default setting; or

 So my proposed Extend2of2 bullet is similar to that requirement, but without the **before** caveat.

cstrobbe:
@bruce-usab 

> The current Extend — as written — requires ten opportunities. And my opinion is that the SC should have an the editorial change to make this even more explicit.

How many opportunities need to be provided is not what we are debating here, unless I have misunderstood the issue. The question is by how much time each opportunity extends the timeout. What Gregg and I are saying is that each opportunity extends the timeout by the original duration of the timeout. So if the original timeout was 20 minutes, each opportunity gives the user an additional 20 minutes. That's what we regard as the original intent.

> Then we could _add_ a bullet which gives the choice of the 20 second warning and opportunity for extended time of ten times the length of the default time limit.

Based on what I wrote above, this comes down to moving the original intent to a separate bullet instead of editing the original bullet. I don't understand the reasoning for this. If you reject adding something like "by the default duration" to the original bullet because you regard that as non-editorial, how does adding another bullet get around this? Or am I completely misunderstanding your objection?

cstrobbe:
The issue that has been identified is that the current "Extend" condition does not state how much time should be added, so ten times 1 minute would meet the letter of the SC (i.e. if you ignore the original intent).

Bruce Bailey proposed to replace the current "Extend" condition with the following:

> * Extend1of2:  The user is warned before time expires and given at least 20 seconds to extend the time limit with a simple action (for example, "press the space bar"), and the user is allowed at least ten opportunities to extend the time limit; or
> * Extend2of2:  The user is warned before time expires and given at least 20 seconds to extend the time limit with a simple action (for example, "press the space bar"), and the user is allowed enough of these opportunities to extend the time limit to at least ten times the length of the original default setting; or

Extend1of2 does not state how much time should be added, i.e. there is no requirement to add the duration of the original timeout. As a consequence the loophole is not closed.

Extend2of2 uses the wording "at least ten times the length of the original default setting" from the "Adjust" condition (with the addition of "original") but does not require that the user be given ten opportunities.

As far as I can see, this is a proposal to change the normative text without closing the original loophole. Extend2of2 comes down to a general technique for meeting the "Adjust" condition of the SC. It is essentially a variation of [G180](https://www.w3.org/WAI/WCAG21/Techniques/general/G180.html) but with the addition of a providing a warning and giving the user at least 20 seconds to do a simple action. If we don't close the loophole, I would write that up as a technique rather than going through the process for changing normative content.

bruce-usab:
> The issue that has been identified is that the current "Extend" condition does not state how much time should be added

I do not agree that it is at all a problem that the current Extend bullet does not state how much time should be added.  There is no loop hole.

> so ten times 1 minute would meet the letter of the SC

Yes, and that is fine.  Presumably the site owner (in this hypothetical case) believes that to be an acceptable customer experience _for everyone_.  The PWD still has 20 seconds and simple action to get that minute.

> i.e. if you ignore the original intent.

I am not ignoring the original intent:  providing end-users enough time to read and use content by mitigating time limits.  2.2.1 requires timing adjustable but provides flexibility to the site owner providing six distinct options.

PR #2584, opened 2022-08-10
Update timing-adjustable.html

bruce-usab:
Editorial, clarifying that <q>ten times</q> in Extend bullet does not imply ten times as much time.

Closes PR #1040.
Closes PR #2581.

### DIFF
--- a/guidelines/sc/20/timing-adjustable.html
+++ b/guidelines/sc/20/timing-adjustable.html
@@ -32,7 +32,7 @@
          
          <p>The user is warned before time expires and given at least 20 seconds to extend the
             time limit with a simple action (for example, "press the space bar"), and the user
-            is allowed to extend the time limit at least ten times; or
+            is allowed at least ten opportunities to extend the time limit; or
          </p>
          
       </dd>


awkawk:
@bruce-usab My review on this is that the text is clear and I agree with what it says, but since I also believe that this is what it says already and since the Understanding document doesn't really clarify it sufficiently we should fix it in Understanding rather than changing the normative text.

awkawk (commented):
(see comment above - I didn't follow the expected Github review process)

GreggVan:
Disagree strongly with this

This would be a major normative change -  rewriting the SC to mean something other than what it was originally meant to be.

(I’m not sure what we will want it to be in a future version — but 10 times  the typical time is indeed what was intended in the original SC.  Good or bad) 

Gregg Vanderheiden
***@***.***

bruce-usab:
> My review on this is that the text is clear and I agree with what it says

I agree.

> but since I also believe that this is what it says already and since the Understanding document doesn't really clarify it sufficiently we should fix it in Understanding rather than changing the normative text.

My preference is this edit, but I am also okay with leaving this SC alone.  FWIW, I do not believe this particular pernicious misunderstanding can be corrected in Understanding.

bruce-usab:
@alastc &mash; please close as you see fit.  This PR is probably a good candidate for wontfix

bruce-usab:
> This would be a major normative change -  rewriting the SC to mean something other than what it was originally meant to be.

@GreggVan &mdash; it may be the case that we did not end up writing what you (and others) originally meant to have written.  As written, the bullets are independent from each other.  My proposed change (in this PR) is editorial on its face.

Current 2.2.1 Extend bullet:
> The user is warned before time expires and given at least 20 seconds to extend the time limit with a simple action (for example, "press the space bar"), and the user is allowed to extend the time limit at least ten times;

Could have been:
> The user is warned before time expires and given at least 20 seconds to extend the time limit with a simple action (for example, "press the space bar") and the user is allowed enough opportunities to extend the time limit to at least ten times the duration of the original time limit;

bruce-usab:
Closing as we did not get AGWG concurrence on the underlying tension between RAR and how SC is understood.

PR #2589, opened 2022-08-10
Adding red-flash note for ISO 9241-391

alastc:


patrickhlauke (commented):
instead of unordered lists of plain text, could/should the algorithms be wrapped up as `<pre><code>...</code></pre>` blocks?

alastc:
Closing, we accepted #2619 for 2.2, but not previous versions.

PR #2600, opened 2022-08-11
Removing redundant note text

awkawk:
The H5 heading is added by the class so shouldn't be in the note text.


alastc:
We have this in another PR (or two) as well, but I think I can merge this without causing issues.

PR #2604, opened 2022-08-12
Removed "the" from focus indicator

### LINKED ISSUE #2598: Focus indicator definition
The glossary item for focus indicator reads:
the pixels that are changed to visually indicate when a user interface component is in a focused state.

The context from 2.4.11 is:
When a [user interface component] has keyboard focus, one or both of the following is true:
The [focus indicator]:

So if this definition replaces the SC content the word "the" is repeated.

I suggest either:
pixels that are changed to visually indicate when a user interface component is in a focused state
or 
collection of pixels that are changed to visually indicate when a user interface component is in a focused state

AND, since this isn't a sentence I also removed the period at the end of the definition.

alastc:
Closes #2598

### DIFF
--- a/guidelines/terms/22/focus-indicator.html
+++ b/guidelines/terms/22/focus-indicator.html
@@ -1,7 +1,7 @@
 <dt class="new"><dfn>focus indicator</dfn></dt>
 <dd class="new">
 	<p class="change">New</p>	
-   <p>the pixels that are changed to visually indicate when a user interface component is in a focused state.
+   <p>pixels that are changed to visually indicate when a user interface component is in a focused state.
     </p>
    				
 </dd>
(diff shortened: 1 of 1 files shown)


guidelines/terms/22/focus-indicator.html
```diff
@@ -1,7 +1,7 @@
 <dt class="new"><dfn>focus indicator</dfn></dt>
 <dd class="new">
 	<p class="change">New</p>	
-   <p>the pixels that are changed to visually indicate when a user interface component is in a focused state.
+   <p>pixels that are changed to visually indicate when a user interface component is in a focused state.
```
awkawk, line 4: And remove the period at the end

PR #2608, opened 2022-08-12
editorial update to accessible-auth exception

### LINKED ISSUE #2592: Clarify exception in Accessible Authentication
The exception in SC 3.3.7 Accessible Authentication has a small structural problem. First off, it isn't written as a grammatically correct sentence, which is unlike other exceptions that aren't bulleted lists. Second, the comma sets the two halves apart from each other and suggests that they are not related.

Current: 
Exception: When the cognitive function test is to recognize objects, or content the user provided to the website.

So this could be understood as:
* When the cognitive function test is to recognize objects
* When the cognitive function test is to recognize content the user provided to the website.

The first bullet would be an issue if we are trying to avoid "click all of the images containing stop signs" type challenges. The understanding document indicates that this would be a cognitive function test, so the grammar of the exception potentially allows all of those again.

The second bullet is now ok (still not a sentence) in that it is exempting content that the user provided which will be familiar.

Trying to reconcile these, I am asking myself, "what is an object that a user provides to a web site?" I can't come up with anything - do we need that part? Can someone clarify why?

Here's my suggested rewording:
Exception: The cognitive function tests ask the user to recognize content that the user provided to the website.

Thoughts? Am I missing something?

alastc:
Closes #2592


patrickhlauke (approved):

guidelines/sc/22/accessible-authentication.html
```diff
@@ -6,7 +6,9 @@ <h4>Accessible Authentication</h4>
 	<p class="change">New</p>
     
     <p>For each step in an authentication process that relies on a <a>cognitive function test</a>, at least one other authentication method is available that does not rely on a cognitive function test, or a mechanism is available to assist the user in completing the cognitive function test.</p>
-    <p>Exception: When the cognitive function test is to recognize objects, or content the user provided to the website.</p>
+    
+    <p>Exception: The cognitive function tests ask the user to recognize content that the user provided to the website or to recognize objects.</p>
```
patrickhlauke, line 10: i'd add a comma after `website`. and perhaps add some kind of qualifier about what kind of objects? "common" objects? or is that too loaded/avoided on purpose?

alastc: I don't think you need a comma with an 'or'.
And yes, we're avoiding common objects as then it then needs defining what is common. After a long discussion, everyone agreed that we should just go with 'objects'.

patrickhlauke: reads a bit odd without the comma, but maybe it's just me. possibly due to the second clause being very short and seemingly just dangling at the end of that sentence

bruce-usab (approved):

PR #2609, opened 2022-08-12
update to add example of a cognitive function test

### LINKED ISSUE #2577: Clarify Accessible Authentication by including "remembering user names and passwords" in the SC text
Success criterion 3.3.7 Accessible Authentication currently reads:

> For each step in an authentication process that relies on a [cognitive function test](https://w3c.github.io/wcag/guidelines/22/#dfn-cognitive-function-test)...

The difficulty with that is that "cognitive function test" is easily misunderstood to mean "CAPTCHA". It would be helpful to explicitly mention the most common case; remembering user names and passwords. The SC would read as follows with that addition:

> For each step in an authentication process that relies on a [cognitive function test](https://w3c.github.io/wcag/guidelines/22/#dfn-cognitive-function-test) **such as remembering a username and password**, at least one other authentication method is available that does not rely on a cognitive function test, or a mechanism is available to assist the user in completing the cognitive function test.

Alternatively:

> For each step in an authentication process that relies on **remembering a username and password, or some other** [cognitive function test](https://w3c.github.io/wcag/guidelines/22/#dfn-cognitive-function-test), at least one other authentication method is available that does not rely on a cognitive function test, or a mechanism is available to assist the user in completing the cognitive function test.

fstrr:
closes #2577


guidelines/sc/22/accessible-authentication.html
```diff
@@ -5,7 +5,7 @@ <h4>Accessible Authentication</h4>
     <p class="conformance-level">AA</p>
 	<p class="change">New</p>
     
-    <p>For each step in an authentication process that relies on a <a>cognitive function test</a>, at least one other authentication method is available that does not rely on a cognitive function test, or a mechanism is available to assist the user in completing the cognitive function test.</p>
+    <p>For each step in an authentication process that relies on a <a>cognitive function test</a> such as remembering a password or solving a puzzle, at least one other authentication method is available that does not rely on a cognitive function test, or a mechanism is available to assist the user in completing the cognitive function test.</p>
```
patrickhlauke, line 8: I would put the additional text in parentheses, or delimited by dashes, or similar.

```
...cognitive function test</a> (such as remembering a password or solving a puzzle), at least
```

otherwise, it almost feels like this is providing an explicit subset of cognitive function tests that this applies to, and only to these (i.e. could be misread as "if it relies on either remembering a password or solving a puzzle, then this SC applies", which is not the intention)

patrickhlauke: also, suggest this same change should also be made to https://github.com/w3c/wcag/blob/main/guidelines/sc/22/accessible-authentication-no-exception.html so they don't go out of sync

bruce-usab (approved):

PR #2612, opened 2022-08-12
Update G220 for bug fixes, additional example

### LINKED ISSUE #2303: Does 3.2.6 Consistent Help allow for content that is under disclosure widget on one page in a set but not another?
Does 3.2.6 Consistent Help allow for content that is under disclosure widget on one page in a set but not another in the same set?

For example, the relative order is the same but on one page you must expand a section for contacting us and on another page it's in the same order but there is no disclosure control - it's just visible.

fstrr:
1. Edited punctuation that is possibly the cause of the <a href="https://w3c.github.io/wcag/techniques/general/G220">current rendered version of this page</a> ending after Example 2.
2. Removed reference to single-page apps.
3. Added an example that hopefully addresses comments in #2303

Closes #2303

### DIFF
--- a/techniques/general/G220.html
+++ b/techniques/general/G220.html
@@ -23,12 +23,19 @@
 		<section id="examples">
 			<h2>Examples</h2>
 			<section class="example">
-				<h3> A link at the top of the page</h3>
+				<h3>A link at the top of the page</h3>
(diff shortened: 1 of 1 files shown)


alastc:
Approved: https://www.w3.org/2022/11/15-ag-minutes.html#t05

PR #2617, opened 2022-08-13
Update focus-appearance.html

mbgower:
Some lingering language from before "Area" wording was brought in.


alastc:
editorial and non-normative, so merging.

Issue #2613, opened 2022-08-12
Add one-time-code to list of autocomplete attributes


fstrr:
[WCAG 2.1's list of input control purposes](https://www.w3.org/TR/WCAG21/#input-purposes) lists the relevant `autocomplete` attributes that relate to the user of the content and pertain only to information related to that individual. Included in that list is `current-password` and `new-password`, but [`one-time-code`](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-autocomplete-one-time-code) isn't, but feels like it should be.

patrickhlauke:
> The following input control purposes are intended to relate to the user of the content and pertain only to information related to that individual.

a one-time code isn't really (personal) information related to the user

yatil:
I agree with Patrick here. Maybe one could extend accessible authentication to include a requirement for using `one-time-code`.

patrickhlauke:
i seem to remember that as far as Accessible Authentication is concerned, as long as you can copy/paste into the field for the one-time passcode (OTP), you satisfy the requirements of that SC

patrickhlauke:
ah, i thought my edits to the understanding https://github.com/w3c/wcag/pull/1898 touched on OTPs as well, but they don't. might be something to add there...i'll have a think

yatil:
> i seem to remember that as far as Accessible Authentication is concerned, as long as you can copy/paste into the field for the one-time passcode (OTP), you satisfy the requirements of that SC

Correct, hence I asked about extending the SC to require it 😉

alastc:
Ah, that wasn't there when we did the WCAG 2.1 list, perhaps that something we can add. 

There's a task to review other new entries as well.

patrickhlauke:
> perhaps that's something we can add

but again, not really a piece of info related to the individual...

patrickhlauke:
> Correct, hence I asked about extending the SC to require it

note that 3.3.7 doesn't directly *require* the use of 1.3.5-specific input purpose identification. at its most basic, it requires that pasting and autofilling (if it works) isn't blocked, nothing more. i.e. you can *fail* 1.3.5, but still pass 3.3.7. at least that's my reading.

alastc:
> a one-time code isn't really (personal) information related to the user

It doesn't have to be personal info, in this case it is something the user has, and does relate to the user. (No one else should have the same code at that time.)

fstrr:
> There's a task to review other new entries as well.

[...]

PR #2618, opened 2022-08-14
Add section about temporary codes/OTPs to 3.3.7 understanding

patrickhlauke:
related to https://github.com/w3c/wcag/issues/2613

extends the understanding for 3.3.7 Accessible Authentication to directly mention temporary codes/one-time passwords, and noting that for these as well it's at least required that they can be autofilled or pasted (otherwise, it would fail due to requiring user to *transcribe* manually)

this PR also softens the language around 1.3.5, to make it clearer that if autofill from the browser/3rd party pwd manager fails for whatever reason (other than it being blocked actively by the site), then it's not necessarily a failure of 3.3.7 as long as at least copy/paste still works.

/cc @yatil @alastc @fstrr


patrickhlauke:
and for avoidance of doubt, I'd still maintain that an OTP does *not* fall under the definition of fields that "relate to the user of the content and pertain only to information related to that individual" ... since they keep varying, they're not a fixed piece of information.

edwardhorsford:
It would be helpful to clarify if a code going to a second device would fail or pass the criterion so long the input field allows pasting / autofill.

Eg: is it implicitly a fail if you know the code was sent to a second device (originating request on laptop, code to sms) and so isn't local? Or is it sufficient that pasting is allowed.

A reading of the current text could imply that if a service required SMS based codes and had no other means to sign in, then that would fail as it requires transcribing.

patrickhlauke:
@edwardhorsford the current understanding skirts around this issue a bit

> Most knowledge-based authentication methods rely on a cognitive function test and so mechanisms to assist users must be available. When authentication relies on performing an action on a separate device, it should be possible to complete the action without the need to transcribe information. It may not be possible to know what device-based authentication methods are available to a user; offering a choice of methods can allow them to choose the path that most suits them.

I'd say the difficulty here is determining whether or not, for sure, the user has no means other than transcribing. If it's SMS, and they're on a mobile/tablet device, then they can likely copy/paste, so it would pass. But the exact same page would fail if they're accessing it from a computer that doesn't receive SMS. so which is it, pass or fail?

Similarly for things like RSA keyfobs, which clearly require transcribing, but usually offer a software-only solution that a user can install on their machine.

Lastly, an auditor may not know if there's an alternative to whatever OTP code system the site they're auditing is used.

Any ideas of wording (either here in this PR, or as further additions to the understanding document) welcome

edwardhorsford:
@patrickhlauke For reference, I'm a designer who designs services for the UK Government - though am coming to this from the security angle. From a security perspective I think it's often *intentional* that codes get transcribed. Certain security techniques rely on this - they'll explicitly want the second factor going to a second device.

I was reading [this introduction to 2.2](https://www.tpgi.com/new-success-criteria-in-wcag22/) which has this to say about 2FA and this success criterion:
> Note however that the 2FA itself must also meet the same requirements. For example, if the 2FA sends a code to the user, that then requires them to type that code into their browser, this would fail 3.3.7 unless they can directly copy-paste from one to the other (because otherwise it’s a requirement to transcribe information). So if the 2FA required a separate device, like a code sent to the user’s cellphone which must be entered into a desktop browser, then that would fail this SC. Requiring the user to input only certain characters of an authentication string (such as, entering the 1st, 3rd and 6th digit of a passcode) would also fail, even if copy-pasting is possible, because the input is required in a format that is different from the original passcode.

If the intention of the criterion is that a service that requires transcribing a 2FA code is not compliant - it would be good to say this explicitly. It would also be good for more clarity on alternatives - your mention of software only solution makes me wonder if supporting security keys as an alternative might then pass?

edwardhorsford:
@patrickhlauke great, thank you for clarifying that. It would be helpful if this was explicitly mentioned.

Note this seems like it could be problematic - such prompts 'yes that's indeed me' can be open to phishing attacks (of course certain code flows can be too).

GreggVan:
+1 

How about adding at the front   "   A service that requires transcribing a 2FA code is not compliant."  
Maybe follow that with   "for example "  and add the detailed example that is there now - to make it clear.

 
Gregg Vanderheiden
***@***.***

bruce-usab:
My only concern is that there is something at AA that the Security Theater folks think is okay.

In the U.S. Federal government, the preferred 2FA standard (for employees) is PIV cards. For citizens, it is registering with [Login.gov](https://www.login.gov/) so that sites behavior (for the public and PWD) is the one I judge this SC against.

Yes, I know, cart before the horse.  But also, there seems to be growing awareness that PIV cards for 2FA is not a technology which industry has much adopted.

patrickhlauke:
Added a small upfront clarification per @GreggVan's suggestion.

Also added clarification about different types of two-factor authentication that do NOT count as cognitive function tests (mentioning Windows Hello, YubiKey, apps that ask the user to confirm that it's them on the other device, etc)

PR #2621, opened 2022-08-16
Audio control

alastc:
Michael Gower's suggestion for #1533

### DIFF
--- a/guidelines/sc/20/audio-control.html
+++ b/guidelines/sc/20/audio-control.html
@@ -4,8 +4,8 @@
    
    <p class="conformance-level">A</p>
    
-   <p>If any audio on a Web page plays automatically for more than 3 seconds, either a <a>mechanism</a> is available to pause or stop the audio, or a mechanism is available to control audio
-      volume independently from the overall system volume level.
+   <p>If any audio on a Web page plays automatically for more than 3 seconds, either a <a>mechanism</a> is available to pause or stop the audio or a mechanism is available to control audio
+      volume, independently from the overall system volume level.
    </p>
    
    <p class="note">Since any content that does not meet this success criterion can interfere with a user's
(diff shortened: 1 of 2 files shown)


guidelines/sc/20/audio-control.html
```diff
@@ -4,8 +4,8 @@ <h4>Audio Control</h4>
    
    <p class="conformance-level">A</p>
    
-   <p>If any audio on a Web page plays automatically for more than 3 seconds, either a <a>mechanism</a> is available to pause or stop the audio, or a mechanism is available to control audio
-      volume independently from the overall system volume level.
+   <p>If any audio on a Web page plays automatically for more than 3 seconds, either a <a>mechanism</a> is available to pause or stop the audio or a mechanism is available to control audio
```
patrickhlauke, line 7: personally, think it reads a bit awkard and long without that comma after "stop the audio", but i agree with that added comma in the next line

patrickhlauke:
any chance the second note from https://github.com/w3c/wcag/pull/1825/files could also be grafted onto this? then i'd be happy to close #1825 in preference to this PR

alastc:
On the paragraph, we had a comment that I agree with: "As not all user agents have this capability, the author may not rely on it. So this paragraph does not help the author, as they need to care themselves anyway."
(I think "care themselves" means they have to take care of it anyway.)

guidelines/sc/20/audio-control.html
alastc: Yea, it has a lot of "or"s, but I think that's the way to make clear it applies to both.

patrickhlauke:
> On the paragraph, we had a comment that I agree with: "As not all user agents have this capability, the author may not rely on it. So this paragraph does not help the author, as they need to care themselves anyway." (I think "care themselves" means they have to take care of it anyway.)

They can rely on it if they're only supporting a platform/user agent that does have it (though not applicable to open web, more closed systems).

But I do think it's worth having something that explains that this can't be relied on/does not pass. It helps auditors in determining whether it's a pass or fail...

PR #2624, opened 2022-08-17
Accesible auth: removing 'text' from the content-recognition

alastc:


patrickhlauke:
I'd note that *recognising* text would be a case of showing the user different bits of text, including text they provided, and having them pick out the correct one. I don't think any site would implement something like "here's three passwords as radio buttons ... pick the correct password out of the line-up", so I don't personally think this would contradict the fact that remembering and entering a password (or other info like mother's maiden name etc) is not excluded.

However, can live with this, as I currently can't think of a site/system that relies on the "recognise text you provided" either.

alastc:
Yea, anything text-based tends to be things like the "pick the 3rd, 5th and 9th characters from your password" approach, but that is very much the type of thing we're trying to catch.

patrickhlauke:
Yeah but that's not quite "recognise", still more a "remember and then enter". But as said, can live with this change.

bruce-usab:
I think we might want to be cautious with any use of "recognize text" as the default quick reading of that might be OCR.  It is okay in context, but I think we should avoid that particular phrasing if we can easily.

PR #2625, opened 2022-08-17
AWK's suggested version

alastc:
Around sub-components.


bruce-usab (approved):
Looks good!

Nitpick:  bullets should end with semicolon.  That is, <q>; and</q> instead of <q>, and</q>.

New feeling of dread question:

Is there any disconnect between 2.4.7 and its (1) keyboard operable UIC where the keyboard focus indicator is visible, (2) keyboard focus indicators generally, and (3) focus indicators (even more general).

alastc:
> When Chrome (temporarily) lights up a Skip Nav target (usually a heading ID), that is a focus indicator. 

Yes, but it isn't a UIC, so not in scope. (UIC is still heavily used in the scoping.)

awkawk:
@bruce-usab What do you mean by "When Chrome (temporarily) lights up a Skip Nav target (usually a heading ID), that is a focus indicator. "? Are you talking about the skip link itself or where it goes?

bruce-usab:
@awkawk &ndash; try Chrome Setting > Accessibility > Show a quick highlight on the focused object.

My understanding is that this feature is driving much of the persistent/fading conversation.

My recent comment in this thread (and what I brought up on Tuesday call) is about where it [triggering the skip link] goes.

This Chrome setting _also_ applies a "quick highlight" to keyboard operable UIC.

awkawk:
@bruce-usab so I think you are referring to the indicator that shows you where the skip link "lands" after you click on it - if so then I agree that it isn't a UIC so wouldn't apply here.

bruce-usab:
> Yes, but it isn't a UIC, so not in scope.

Great!

> UIC is still heavily used in the scoping.

It is not a keyboard operable UIC certainly.  Keyboard operable is explicitly part of 2.4.7.  I am happy for this not to be an issue for 2.4.11, but I felt compelled to ask.

guidelines/sc/22/focus-appearance.html
```diff
@@ -23,15 +23,16 @@ <h4>Focus Appearance</h4>
...
-    <p class="note">Note: Examples of sub-components that may receive a focus indicator are menu items in an opened drop-down menu. However, it may also be possible to indicate user interaction for such sub-components by relying strictly on a visual indication of which item is <em>selected</em>. Where selectable sub-components have no differentiated focus indicator, the visual indicator for sub-component selection is measured against <a href="https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast.html">1.4.11 Non-text Contrast</a> requirements, not against this Criterion.</p>
-    <p class="note">Note: Contrast calculations can be based on colors defined within the technology (such as HTML, CSS and SVG). Pixels modified by user agent resolution enhancements and anti-aliasing can be ignored.</p>
+    
+    <p class="note">What is perceived as the user interface component and any sub-components (to determine enclosure or size) depends on its visual presentation. The visual presentation includes, but is not limited to the component's visible content, border, component-specific background, and effects emanating from the visible component such as shadows or glow effects.</p>
```
awkawk, line 32: I am suggesting this modification because selectable sub-components are focused. They may be focused and selected, but focused doesn't stay on the top-level component. See https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#x6-3-focus-vs-selection-and-the-perception-of-dual-focus

guidelines/sc/22/focus-appearance.html
```diff
@@ -23,15 +23,16 @@ <h4>Focus Appearance</h4>
...
+
+    <p class="note">Examples of sub-components that may receive a focus indicator are menu items in an opened drop-down menu.</p>
```
awkawk, line 34: Might need to add an example or change to "An example of a sub-component.... is a menu item.."

guidelines/sc/22/focus-appearance.html
```diff
@@ -5,11 +5,11 @@ <h4>Focus Appearance</h4>
...
 	<p class="change">New</p>
     
-    <p>When the keyboard <a>focus indicator</a> is visible, one or both of the following is true:</p>
+    <p>When the keyboard <a>focus indicator</a> is visible, one or both of the following are true:</p>
     <ol>
-        <li>The focus indicator:
+        <li>The focus indicator meets all the following:
             <ul>
-                <li><a>encloses</a> the visual presentation of the user interface component, and</li>
+                <li><a>encloses</a> the <a>user interface component</a>  or sub-component that is focused, and</li>
```
awkawk, line 12: My concern here is due to things like cards where there are focusable sub-items and while the whole card is mouse-clickable the focus moves to the first item and hitting enter or clicking on that goes to the same place as clicking on the whole card. Sometimes it makes more sense to focus in on a smaller piece of a larger control.

guidelines/sc/22/focus-appearance.html
alastc: That's a useful resource, but as it is says "in some widgets, moving focus may also perform the select operation."

I.e. from a user point of view they are the same. 

Yet, when moving away the selected things stays selected. So if we require a focus and select indicators at the same time (and the selection has to meet non-text contrast), we setup a 3-color contrast problem.

guidelines/sc/22/focus-appearance.html
alastc: I emailed about this recently, but the card example isn't a sub-component.

The UIC in that case is the link, the thing that gets focus, and we can use the indicator to establish that. The clickable area isn't part of that analysis. There are several examples in the doc: https://docs.google.com/document/d/1P88HywmDChMklvKR_bwa2H4jW-CyJFj231HyPjjxY78/edit

mbgower:
I find it makes it a bit more convoluted.

Issue #1839, opened 2021-06-02
Understanding for 1.4.10 redefines normative requirement for 1.4.4


patrickhlauke:
https://www.w3.org/WAI/WCAG21/Understanding/reflow.html#the-relation-of-reflow-to-the-success-criterion-1.4.4-resize-text states that

> It is not required to achieve 200% text enlargement at every breakpoint, but it should be possible to get 200% text enlargement in some way.

This redefines/limits the scope of 1.4.4 (a non-normative document redefining the normative requirement). And it further goes against the full pages conformance requirement https://www.w3.org/TR/WCAG21/#cc2

> A full page includes each variation of the page that is automatically presented by the page for various screen sizes (e.g. variations in a responsive Web page). Each of these variations needs to conform (or needs to have a conforming alternate version) in order for the entire page to conform.

I know this has been a tricky subject in general (with previous discussions like https://github.com/w3c/wcag/issues/1671 and https://github.com/w3c/wcag/issues/704), but this seems a fairly egregious case of redefining things by the backdoor again (and not even from the understanding in 1.4.4, but from a completely separate understanding document) - at least on the face of it/how it can be interpreted.

I think I remember the context for this being "users may not be able to zoom to 200% before hitting another breakpoint, so technically at that original breakpoint they can't get to 200% since they now fell into the next breakpoint", but this nuance isn't there in the current wording. I would suggest this needs to be expanded - and also, more crucially, this needs to be presented in the understanding document for 1.4.4 as well, rather than expecting authors/auditors to find this information about 1.4.4 in a completely separate understanding document for 1.4.10

detlevhfischer:
Yes, I guess best delete this paragraph and amend 1.4.4 the normative text instead: #1671

patrickhlauke:
i suggest we first do the former, and then get consensus on how to change the normative wording of 1.4.4 (if we even can at this stage) / soften it somehow in the understanding at least (adding some "reasonableness" clause in the understanding along the lines of what we said above). probably two steps i'd say.

patrickhlauke:
Coming in late, but made an initial stab at this here https://github.com/w3c/wcag/pull/2630

Issue #2101, opened 2021-10-20
Is there an intended minimum viewport size for 1.4.4?


daniel-montalvo:
Ref:
https://github.com/act-rules/act-rules.github.io/issues/1318

For rule:
[Zoomed text node is not clipped with CSS overflow (59br37)](https://act-rules.github.io/rules/59br37)

The rule currently makes the following assumption:

> While success criterion 1.4.4 Resize text does not explicitly mention which viewport size has to be resized up to 200%, it is assumed that a viewport size of 1280 by 1024 is applicable. A 1280 by 1024 viewport size is explicitly mentioned under success criterion 1.4.10 Reflow.

Is this assumption reasonable even though 1.4.4 does not mention any viewport size?

mraccess77:
While 1.4.4 doesn't specifically provide a value - I'd treat it like an accessibility supported issue and what you are making the claim against. 1280 is a width to start from and is what I use for testing that.

Issue #1050, opened 2020-02-26
Change 1.4.10 Reflow to level A


alastc:
Creating this issue based on a thread started by @WayneEDick, who said:

> There is one thing you that would make a big difference for people with low vision. Change Reflow 1.4.10 from AA to A. This can be done with WCAG 2.2, we don't have to wait for Silver.
> 
> There are two good reasons to do this. It is vitally important for almost everyone with low vision. It is as important to low as text-to-speech is to blindness. The second reason is that the overwhelming majority of sightes nearly meet 1.4.10. So, no cutting edge technology is required. It is dead simple to make a case that works at 320 by 768.
> 
> The main reason AG got this wrong in the first place is that the Low Vision support communnity grossly underestimated the print size needed to read with low vision. The simple fact is that the print size needed to read is almost a linear function 1/(visual acuity). If your vision is 20/80 you need print that is 4 times as big. We already worked out how to make 400% into 1000% by varying viewing distance.
> 
> The second reason everyone got this wrong is reflow itself. Almost nobody, including me, realized
>  that horizontal scrolling takes as much work as it does. I was probably the most startled by my own discovery.
> 
> I suppose we didn’t make it Level A because we didn’t think it could be met so easily. It can. 
> Moving it to level A would be to give it the status it needs. 

Charles H replied:

> I am not sure where that linear function math and print target reside, but +1 to moving to A.
>  
> Bi-directional scroll effort is only part of the concern. I believe it also requires additional cognitive processing and reduces readability to track progress through a passage of text that requires scrolling to see. There also seems to be a case where scale does correctly reflow without a need for bi-directional scroll, but the result is still unreadable due to a large volume of word breaks.

Regarding the feasibility, @patrickhlauke said:

> I would say this is heavily dependent on the type of site. Functionality- and information-dense sites, particularly complex web applications, will have more trouble making a working/workable version at 320 CSS px width. It is not just the technology side that matters - as yes, technologies like responsive/adaptive web design are now fairly mature ... but it can be a challenge actually having to rethink how complex content and interactions can be made available in small-sized viewport.
> 
> And as most companies/sites (at least in my auditing experience) target a base of WCAG AA, changing the SC to A won't make that much of a difference in practice (and sites/applications that actively choose to ignore it for AA will likewise ignore it if it was A).

@WayneEDick replied:
> I think that the difficulties you cite concerning reflow for current websites are not harder than implementing any Level A criterion in 1.3. That is, if the developer hasn't designed for it.
[...]

PR #2630, opened 2022-08-22
Expand 1.4.10 Reflow note about 200% and breakpoints, add matching note to 1.4.4 Resize Text

patrickhlauke:
- Updates the wording in the Reflow Understanding document which discusses the connection between Reflow and Resize Text in attempt to make it clearer
- Adds similar reciprocal language into the Resize Text Understanding document

Closes https://github.com/w3c/wcag/issues/1839

Related: https://github.com/w3c/wcag/issues/2101 https://github.com/w3c/wcag/issues/704


patrickhlauke:
Putting this down as an initial strawman change. Happy for any improvements/suggestions to possibly take this further.

bruce-usab:
Can we change the 2nd 200% in first example sentence to use px?

> For example, if at the default browser setting of 100% zoom, text is set at 20px when the window is 1280px wide, the same 20px at 200% zoom would mean a text size of **200%**.

(I am holding my breath to see if that 200% becomes 40px or 28.3px.)

bruce-usab (changes requested):
Second 200% -> 28.3 px

bruce-usab (approved):

patrickhlauke:
not sure i follow where you got the 28.3px from.

that whole sentence, which I left mostly untouched, is a bit dense and confusing though. will see if i can make it clearer what's actually going on there.

patrickhlauke:
Done. reworded to possibly make a bit more explicit what the 20px at 100% / 10px at 400% was getting at.

JAWS-test:
I am unsure if this is the intention of the SC. I think the reason for 1.4.10 is, 
* to allow 400% zoom,
* to simplify reading at 400% zoom by not requiring scrolling.

Low vision users may need a high zoom factor to read content.

Therefore the first reason (400% zoom) is much more important than the second (no scrolling). Unfortunately, 1.4.10 is worded to focus only on the second reason, but the Understanding explains that the intent is actually primarily the first reason.

If we don't require 400% zoom now, we give up the point of 1.4.10. Then SC 1.4.4 and combination with SC 1.4.8 ("Text can be resized without assistive technology up to 200 percent in a way that does not require the user to scroll horizontally to read a line of text", at level AA instead of AAA) would be sufficient.

I am aware that Understanding already says that 200% magnification at 400% zoom is sufficient. In response to my previous inquiries about the meaning of this sentence, I was told here in the forum that this exception only applies to large text (such as headings), which would become much too large at 400%.

bruce-usab:
> not sure i follow where you got the 28.3px from.

That would be `20*sqrt(2)` (approximately).

I think at one point there was debate if 200% meant twice the ~number of pixels or twice the font size.  (A square 40px on at side has 4x the area as a 20px square.)  I was not entirely clear where AGWG landed on that detail.

patrickhlauke:
> > not sure i follow where you got the 28.3px from.
> 
> That would be `20*sqrt(2)` (approximately).
> 
> I think at one point there was discussion if 200% meant twice the ~number of pixels or twice the font size. (Since a square 40px on at side has 4x the area as a 20px sided square.) I was not entirely clear where AGWG landed on that detail.

First I heard of it in the context of "text size" (which usually refers to the height/x-height/etc of a font). It may have come about from other discussions about target size and/or size of enhanced focus area (where we do want to talk about area)

> > ...the same 20px at 200% zoom would mean a text size of 200%.
> 
> Should this not be?

It's a truism, but one that was there to begin with. And it then sets up the contrast to the next sentence where that isn't the case due to the new media query.

> > ...the same 20px at 200% zoom would mean a text size of 40px.

This whole thing would then veer into disambiguating between CSS pixels and display pixels, and at which display density that would be true. maybe something like "would mean text is displayed at an effective size of 40px (because it's not the actual measure that changes, but the way it's displayed/the CSS pixels are converted to display pixels)

> 
> Examples based on 16 css px would be a little more typical.

but make slightly uglier numbers...

GreggVan:
The second is (almost) as important as the first
(Both are nearly equally important)

When you require horizontal scrolling - it makes it very difficult to read anything - including finding the beginning of the next line.      Tracking back to find the next line is very hard when you need to horizontally scroll and cannot refer to the end of one line while also seeing the candidates for next line. 


Gregg Vanderheiden
***@***.***

JAWS-test:
@GreggVan

If I need 400% zoom, then
* I can't perceive the contents if at 400% zoom contents are not visible or are only enlarged to 200%.
* I find it very difficult to perceive the content if I have to scroll horizontally after each line.
* I can perceive the content only slightly harder if I have to scroll horizontally after a block of content.

The difference between NOT PERCEIVING and DIFFICULTLY PERCEIVING seems important to me.

JAWS-test:
See https://github.com/w3c/wcag/issues/658#issuecomment-474127528 and https://github.com/w3c/wcag/issues/1050#issuecomment-595918723

The exception that content does not need to be zoomed four times at 400% zoom was intended for large text like headlings. Unfortunately, this information never made it into Understanding, although it is completely incomprehensible to me why not, because it's the crucial information for 1.4.10 - otherwise 1.4.10 is worthless in terms of 400% zoom.

detlevhfischer:
@JAWS-test The 1.4.10 Reflow SC has nothing on differentiating between body text and headings and nothing on requiring 400% body text. (That line would be difficult to draw anyway.) Yes, there is a user need for zooming text to 400%, but it is not a normative requirement (yet). I think it is important to distinguish between "what this SC ideally means" and what is put down in the normative text.
I think @patrickhlauke s explanatory notes are spot on and help clarifying the application of both 1.4..4 and 1.4.10.

JAWS-test:
@detlevhfischer Yes, I am aware of that. But I think this is a mistake, because then according to 1.4.4 I can support 200% zoom and at 400% zoom I can theoretically even shrink the font four times to comply with 1.4.10 Reflow. However, this contradicts the intention formulated in the first sentence of the understanding:

> The intent of this Success Criterion is to support people with low vision who need to enlarge text and read it in a single column. When the browser zoom is used to scale content to 400% ...

mbgower:
I think we should include some illustrations, especially in the resize text Understanding document.

mbgower:
@JAWS-test regarding you're comments, I just wanted to ensure you were aware of what is currently stated under "The relation of Reflow to the Success Criterion 1.4.4 Resize Text" in the [existing Reflow Understanding](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html).

bruce-usab:
Briefly mentioned on TF call 4/12.  This PR should be part of Reflow discussion.

Leeze555:
ok thank

patrickhlauke:
@bruce-usab when zooming, the CSS px measures don't change (it's only the browser's internal mapping of "how many actual display pixels do i use to represent a single CSS pixel" changes). so i really don't think it'd be helpful to start talking about `px` here.

also, refreshing my mind about your "28.3 pixels" thing early on, note that when zooming to 200%, you're not zooming with some square root calculation ... you're doubling the horizontal and vertical dimensions, i don't believe any implementation of zooming considers the actual *area* of things, but just changes the horizontal and vertical dimensions of how a CSS pixel is represented by that particular factor.

Leeze555:
thanks

Leeze555:
ok

patrickhlauke:
@Leeze555 can you just ... shush and go away?

patrickhlauke:
Will see if I can add an illustration/diagram somehow (and clarify how the "at 400% zoom" relates back to the original size, not "320 CSS px and THEN 400%")

understanding/21/reflow.html
```diff
@@ -61,7 +61,7 @@ <h3>Avoiding scrolling in horizontally and vertically written languages</h3>
...
+        <p>For example, if at the default browser setting of 100% zoom, text is set at 20px when the window is 1280 CSS pixels wide, at 200% zoom it will visually appear at twice the size. Once the page reaches a width of 320 CSS pixels, at 400% zoom, the author may decide to set the text size to 10px: the text is now half the size in CSS pixels, but enlarged to 400%, meaning it is still displayed at twice the size compared to the default browser setting of 100% zoom. It is not required to achieve 200% text enlargement while remaining inside a specific breakpoint (as zooming may result in the variation for a new breakpoint becoming active), but it should still be possible to get 200% text enlargement in some way compared to the default 100% zoom.</p>
```
scottaohara, line 64:
```suggestion
        <p>For example, if at the default browser setting of 100% zoom, text is set at 20px when the window is 1280 CSS pixels wide, at 200% zoom it will visually appear at twice the size. After zooming by 400% the page reflows to fit within the 320 CSS pixel viewport, the author may decide to set the page's text size to 10px. The text is now half the original size in CSS pixels, but as it has been enlarged to 400%, so it is still displayed at twice the size compared to the default browser setting of 100% zoom. It is not required to achieve 200% text enlargement while remaining inside a specific breakpoint (as zooming may result in the variation for a new breakpoint becoming active), but it should still be possible to get 200% text enlargement in some way compared to the default 100% zoom.</p>
```

understanding/21/reflow.html
```diff
@@ -61,7 +61,7 @@ <h3>Avoiding scrolling in horizontally and vertically written languages</h3>
...
+        <p>For example, if at the default browser setting of 100% zoom, text is set at 20px when the window is 1280 CSS pixels wide, at 200% zoom it will visually appear at twice the size. After zooming by 400% the page reflows to fit within the 320 CSS pixel viewport, the author may decide to set the page's text size to 10px. The text is now half the original size in CSS pixels, but as it has been enlarged to 400%, so it is still displayed at twice the size compared to the default browser setting of 100% zoom. It is not required to achieve 200% text enlargement while remaining inside a specific breakpoint (as zooming may result in the variation for a new breakpoint becoming active), but it should still be possible to get 200% text enlargement in some way compared to the default 100% zoom.</p>
```
patrickhlauke, line 64:
```suggestion
        <p>For example, if at the default browser setting of 100% zoom, text is set at 20px when the window is 1280 CSS pixels wide, at 200% zoom it will visually appear at twice the size. After zooming by 400% the page reflows to fit within the 320 CSS pixel viewport, the author may decide to set the page's text size to 10px. The text is now half the original size in CSS pixels, but as it has been enlarged to 400%, it is still displayed at twice the size compared to the default browser setting at 100% zoom. It is not required to achieve 200% text enlargement while remaining inside a specific breakpoint (as zooming may result in the variation for a new breakpoint becoming active), but it should still be possible to get 200% text enlargement in some way compared to the default 100% zoom.</p>
```

understanding/20/resize-text.html
```diff
@@ -91,6 +91,11 @@ <h2>Intent of Resize Text</h2>
...
+         <p>For example, if at the default browser setting of 100% zoom, text is set at 20px when the window is 1280 CSS pixels wide, at 200% zoom it will visually appear at twice the size. Once the page reaches a width of 320 CSS pixels, at 400% zoom, the author may decide to set the text size to 10px: the text is now half the size in CSS pixels, but enlarged to 400%, meaning it is still displayed at twice the size compared to the default browser setting of 100% zoom. It is not required to achieve 200% text enlargement while remaining inside a specific breakpoint (as zooming may result in the variation for a new breakpoint becoming active), but it should still be possible to get 200% text enlargement in some way compared to the default 100% zoom.</p>
```
patrickhlauke, line 97:
```suggestion
         <p>For example, if at the default browser setting of 100% zoom, text is set at 20px when the window is 1280 CSS pixels wide, at 200% zoom it will visually appear at twice the size. After zooming by 400% the page reflows to fit within the 320 CSS pixel viewport, the author may decide to set the page's text size to 10px. The text is now half the original size in CSS pixels, but as it has been enlarged to 400%, it is still displayed at twice the size compared to the default browser setting at 100% zoom. It is not required to achieve 200% text enlargement while remaining inside a specific breakpoint (as zooming may result in the variation for a new breakpoint becoming active), but it should still be possible to get 200% text enlargement in some way compared to the default 100% zoom.</p>
```

patrickhlauke:
@scottaohara tweaked and applied your suggestion, and then made the same change to the other file as well to match it.

detlevhfischer:
Mulling this note, which is a good clarification in substance, I still have some questions as to where it belongs and what it means for UA without windows and UA-based zoom function. The note states:

> As with most other Success Criteria, this criterion applies to each variation of the page that is automatically presented for various screen sizes (e.g. media query variations in a responsive site).

I wonder about the "automatically presented"  - because it sounds tailored to user agents that have no windows (where the size and/or zoom factor may be changed any time) but use the entire screen as viewport (i.e., phones, tablets). And here we are in a situation where reflow may not work (or not be testable) when the UA does not offer a zoom function triggering reflow the same way as in desktop browsers. So the note does not really clarify if and how this SC would then apply in such windowless environments. Should content reflow when the viewport changes due to other applications loaded alongside the browser split screen mode, or when device orientation changes change the viewport width? Maybe this is another issue that must be addressed elsewhere.

The second thing is that the note talks most about the effect of reflow on text size and what that means related to 1.4.4. That's fine - but is this then not a note that would really belong to 1.4.4, with a note here (in 1.4.10) merely stating that the reflow requirement (as stated normatively)  is independent from the text resizing requirement, and refer to the discussion of text size under different viewpoint width /media quer conditions in 1.4.4?

mbgower:
@patrickhlauke any response to @detlevhfischer ?

mbgower:
@detlevhfischer since the changes received wide support, I am going to merge these changes. I invite you to open your last comments as issues or discussions.

PR #2632, opened 2022-08-23
Adding decorative effects definition

alastc:
Update based on the meeting today.

### DIFF
--- a/guidelines/sc/22/focus-appearance.html
+++ b/guidelines/sc/22/focus-appearance.html
@@ -30,7 +30,7 @@
         <li>The focus indicator and the indicator's background color are not modified by the author.</li>
     </ul>
     
-    <p class="note">What is perceived as the user interface component or sub-component (to determine enclosure or size) depends on its visual presentation. The visual presentation includes the component's visible content, border, component-specific background. It does not include <a>decorative effects</a> emanating from the visible component such as shadows or glow effects.</p>
+    <p class="note">What is perceived as the user interface component or sub-component (to determine enclosure or size) depends on its visual presentation. The visual presentation includes the component's visible content, border, and component-specific background. It does not include <a>decorative effects</a> emanating from the visible component such as shadows or glow effects.</p>
 
     <p class="note">Examples of sub-components that may receive a focus indicator are menu items in an opened drop-down menu, or focusable cells in a grid.</p>
 
--- a/guidelines/sc/22/focus-appearance.html
+++ b/guidelines/sc/22/focus-appearance.html
@@ -30,7 +30,7 @@
         <li>The focus indicator and the indicator's background color are not modified by the author.</li>
     </ul>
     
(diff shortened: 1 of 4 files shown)


patrickhlauke (approved):

awkawk:
We don't have any glossary items that are only used in notes. Is this in normative text anywhere?

bruce-usab:
> We don't have any glossary items that are only used in notes. 

Good catch.  We do have defined glossary terms which only appear the glossary --- to simplify other defined terms.  @patrickhlauke is that a semi-elegant option?

alastc:
If that's an issue, I guess there are two options:

1. Expand the note to incorporate the definition, e.g. 

> What is perceived as the user interface component or sub-component (to determine enclosure or size) depends on its visual presentation. The visual presentation includes the component's visible content, border, and component-specific background. It does not include decorative effects emanating from the visible component such as shadows or glow effects. Decorative effects are those added to an element that are outside of the element that, if removed, would not change the ability to identify the element.

2. Incorporate the definition into the SC text, e.g.

> An area of the focus indicator, excluding _decorative effects_, meets all the following:

I'd rather go with the current PR as it minimizes wordage and complication though...

alastc:
>     3. How about _adding_ separate note explaining use of the term?

Hmm, a second note to explain the term in the first note? I'm not sure how that is better than just extending the first note.

Also, I didn't put that SC addition in the right place, it would be:

> is at least as large as the area of a 1 [CSS pixel](https://w3c.github.io/wcag/guidelines/22/#dfn-css-pixels) thick [perimeter](https://w3c.github.io/wcag/guidelines/22/#dfn-perimeter) of the unfocused component or sub-component **(excluding decorative effects)**, or is at least as large as a 4 CSS pixel thick line along the shortest side of the [minimum bounding box](https://w3c.github.io/wcag/guidelines/22/#dfn-minimum-bounding-box) of the unfocused component or sub-component, and

Making the most complex bullet more complex...

bruce-usab:
IMHO extending the first note is better than adding a definition in the glossary.  It does not have be a separate note.  A stand-alone sentence at the end is fine.

Like @awkawk using glossary _only_ for a note does not feel right.  My larger objection is with a **single use** glossary terms.  [Note to self:  It might be worth a tally of those.]

It is, of course, counterproductive to have to explain a term while trying to make use of the term!  But it also needs to be easy for an inexperienced reader to verify that the term means what they think it does.   Hyperlinked text is pretty good for that!  But hypertext is not as good as having the explanation just as skippable &mdash; but also right there.  If not an extra note, how about one or two sentences <ins>_at the end_</ins> of one of the others?

Here is my attempt:

> ....The visual presentation includes the component's visible content, border, and component-specific background. It does not include decorative effects emanating from the visible component.   Decorative effects are outside of the visible component and if removed do not change the ability to visually identify the component.   Examples of decorative effects include shadows and glow effects around a user interface control.

awkawk:
Yeah, adding Normative text to explain non-normative text doesn't seem like a good idea.

alastc:
I've updated it, does that seem reasonable? CCing @WilcoFiers

bruce-usab:
+1 to this version.

PR #2641, opened 2022-08-25
several updates

### LINKED ISSUE #2173: MathML version of the relative luminance definition
Recently, we have been taking a closer looking at the [flash thresholds](https://w3c.github.io/wcag/guidelines/22/#dfn-general-flash-and-red-flash-thresholds).

Relatedly, can we delete the [last note](https://w3c.github.io/wcag/guidelines/22/#h-note-134) under [relative luminance](https://w3c.github.io/wcag/guidelines/22/#dfn-relative-luminance) which links to an [xml file](https://w3c.github.io/wcag/guidelines/22/relative-luminance.xml)?

If not, then probably we should update the file, at least a little bit.
- [x] delete (html style) comment
- [x] change <q>WCAG 2.0</q> (part of linked text, at top) to <q>WCAG 2.2</q>
- [x] change two URLs under references from hard-coded links (using 2.0 anchors) to relative
- [x] change one URL link 2.1 SC 1.4.3 to relative

It is all presentational MathML, but changing that to content MathML would be a pretty heavy lift.  Also lots of cruft like:  `<mn>8</mn><mi>b</mi><mi>i</mi><mi>t</mi>` which, as I understand it, should just be:  `<mi>8 bit</mi>`

I am willing to take a stab at the pull request.

michael-n-cooper:
fixes #2173

Make modern HTML
update links


michael-n-cooper:
This fixes the basic problems with the relative luminance definition. There are 2 steps further we could go:

* rename the file to relative-luminance.html, since it is valid HTML 5 content.
* replace the term file with the (fixed) MathML version, and do away with the separate file.

I think with modern browser support we can do the second, but I'd like input before extending the PR.

bruce-usab (approved):
Thank you @michael-n-cooper , I find nothing but improvements.  Page is better in every way.

I was hoping for more.  I will detail in a separate comment.

bruce-usab:
Current version of Edge and Chrome do not support MathML.  Safari is fine with it.  Sadly, I think we still need the stand-alone version.

+1 to rename the file to relative-luminance.html, since it is valid HTML 5 content 

My others asks, from easy to impractical:

[ ] Page 'title' and linked text (pointing to this file) is misleading and not correct.

Current:
>  MathML version of the relative luminance definition

Should be:
> page of the relative luminance definition using MathML to display the formulas.

Still just a bit too much presentational effects.  The only one left that makes me grate my teeth is:
> mathcolor="blue"

[ ] Can that formatting be moved be moved into a style sheet?

Is a version that entirely avoids Presentation Markup even possible?  I was hoping for nothing but Content Markup.  I know not the level of effort needed for what I am asking!

alastc:
Bruce is currently thinking that the compatibility isn't great, so it is best to keep it a separate file.

michael-n-cooper:
I think I've got this PR much better. I copied in the exact code of the revised definition from the guidelines, except the MathML, which I put in place and then updated that one value in it. The only thing missing is styling so it looks more like it looks in the guidelines, but that's more complicated than you'd think. I think it's readable and good to go. Oh - and I removed that annoying blue - I left it alone last edit pass but agree with your revulsion. :)

michael-n-cooper:
Oh, and BTW I'd like this to show up in the CR publication. I'm taking the liberty of merging this into my staging branch for that now, though it should go through review to merge into main. Please @alastc @bruce-usab please object as soon as possible if needed!

alastc (approved):

PR #2653, opened 2022-08-30
Focus Visible should also apply when there is only one control

WilcoFiers:
The current understanding document of 2.4.7 Focus Visible suggests there is an exception for when there is only one focusable element on the page. The actual SC text does not mention any such exception though, and so suggesting it's allowed in the understanding document seems out of place.

This case was discussed by the ACT Task Force, who is currently evaluating a rule which has this [one-component exception](https://www.w3.org/WAI/standards-guidelines/act/rules/oj04fd/proposed/#applicability). However none of the testers who looked at it considered this a valid exception, and would fail pages with a single control if it did not have a visible focus. We would propose removing this exception from the understanding document.

While rare, this type of issue can exist on for example redirect pages; A page that informs the visitor content is moved, and includes a single link to the new location.

### DIFF
--- a/understanding/20/focus-visible.html
+++ b/understanding/20/focus-visible.html
@@ -17,8 +17,6 @@
          keyboard focus.
       </p>
       
-      <p>It must be possible for a person to know which element among multiple elements has the keyboard focus. If there is only one keyboard actionable control on the screen, the success criterion would be met because the visual design presents only one keyboard actionable item.</p>
-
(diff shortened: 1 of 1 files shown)


alastc:
I wonder if there was an assumption that, if there's only one control, it wouldn't change if you tabbed past it? (Whereas you can tab through the browser chrome as well, so can tell when it has focus.)

patrickhlauke:
> I wonder if there was an assumption that, if there's only one control, it wouldn't change if you tabbed past it? (Whereas you can tab through the browser chrome as well, so can tell when it has focus.)

that seems the most likely explanation to me, yes. agree that the exception should ideally be removed as it seems nonsensical

patrickhlauke (approved):

bruce-usab (commented):
I am fine with removing this paragraph, as it might not be at all obvious to the end-user that there is only one UIC.

PR #2657, opened 2022-08-31
Update target-size from CFC thread

alastc:
We (mostly me) missed an update from the CFC thread:


bruce-usab (commented):
@WilcoFiers -- this fixes a typo, it is not a change to how anyone understood the meaning.

PR #2661, opened 2022-09-02
update ref to srgb

michael-n-cooper:


bruce-usab (approved):
I think it is sRGB rather than SRGB (lower case `s`)

svgeesus:
> I think it is sRGB rather than SRGB (lower case s)

Yes, in general usage.

The specification ref database is happy with either case though, and [emits all uppercase here](https://www.specref.org/?q=srgb)

Issue #2583, opened 2022-08-10
How does 2.4.12: Focus Not Obscured relate to opacity


WilcoFiers:
Simple question: When a component is focused, and the component is covered by an element with an opacity of 95% would that count as obscuring? What about if the opacity is 5%?

A very literal reading of the SC would suggest that even at 99.9% opacity, the component isn't "entirely hidden". A more reasonable reading of the SC might suggest that perhaps the component could be considered "fully obscured" when the opacity is such that the focus indicator no longer meets the 3:1 contrast requirement of 2.4.11 Focus Appearance?

This feels like a significant enough gap in the SC that I think it might be worth updating the normative text, or perhaps a note.

bruce-usab:
I don't really like that <q>fully obscured</q> is (1) subjective and (2) maybe too much of a loophole.

But what number could we pick for the SC?  I think Understanding could spell out some <q>safe harbor</q> examples.  I like your 3:1 suggestion for worse-case-acceptable opacity.

cstrobbe:
If the intent is to avoid other content overlapping the component that has focus, a rewording may be necessary, e.g.

> When a user interface component receives keyboard focus, other author-created content does not entirely overlap it.

(I don't know whether that wording brings back the issue that was addressed by replacing "by" with "due to" in an earlier revision of 2.4.12. )

And then similarly for SC 2.4.13 at level AAA:

> When a user interface component receives keyboard focus, no other author-created content overlaps it.

alastc:
I haven't come across an example of this, and we've been testing for it for a while now.

If there is support for a simple change, e.g. "the component is not entirely overlapped by author-created content." (nice idea @cstrobbe), that's possible.

Otherwise I think we're in "perfect is the enemy of the good" territory.

alastc:
Note from recent discussion on email and in meeting: 

- We should keep the current SC text, but…
- We should add something to the understanding document about semi-transparent things that could obscure, and link through to 2.4.11.

I.e. if the focus indicator goes behind authored content that is semi-opaque, that doesn’t fail focus-not-obscured, but should be tested against 2.4.11. (Which would be good motivation to use scroll-padding or similar so it doesn’t overlap!)

rachaelbradley:
[Email conversation ](https://lists.w3.org/Archives/Public/w3c-wai-gl/2022JulSep/0274.html)has some additional discussion.

mbgower:
[...]

PR #2671, opened 2022-09-09
Update focus-not-obscured-minimum.html

mbgower:
An attempt to rewrite Focus Not Obscured to both incorporate some wording on opacity ( to address https://github.com/w3c/wcag/issues/2583 )and to generally revise the content.


alastc:
Approved: https://www.w3.org/2022/11/22-ag-minutes.html#t10

PR #2681, opened 2022-09-21
1.4.11 Non-text Contrast understanding: expand explanation for Figure 10

patrickhlauke:
Provides some actual context about *why* the green border doesn't need to contrast with the button itself, and moves the easily missed extra prose that follows into the actual figure caption (as it's currently easy to miss when skimming over the document)


bruce-usab (approved):

mbgower:
Might be an idea to generate a preview copy so folks can easily review in context -- even a screen shot of the pertinent paragraphs in your local build would be useful to some for context.

detlevhfischer:
[Rendered view of 1.4.11 Understanding](https://raw.githack.com/w3c/wcag/ccd7e305e57dc819daba8c73f8a87f0700e6eeaa/understanding/21/non-text-contrast.html)

understanding/21/non-text-contrast.html
```diff
@@ -118,10 +118,9 @@ <h4 id="related-focus">Relationship with Focus Visible</h4>
 				<figure id="figure-focus-outer-green">
 					<img src="img/ntc-focus-outer-green.png" alt="Three blue buttons on a white background, the middle has a dark green outline outside of the botton's non-focused boundary." width="400" />
 					<figcaption>
-						The external green indicator (#008000) does contrast with the white background (#FFF) which the component is on, <strong>passing</strong> the criterion. It does not need to contrast with both the component background and the component.
+						The external green indicator (#008000) does contrast with the white background (#FFF) which the component is on, <strong>passing</strong> the criterion. It does not need to contrast with both the component background and the component, as visually the effect is that the button is noticeably larger, and it's not necessary for a user to be able to discern this extra border in isolation. Although this passes non-text contrast, it is not a good indicator unless it is very thick. <span class="wcag22">There is a criterion in WCAG 2.2 that addresses this aspect, <a href="../22/focus-appearance-minimum.html">Focus Appearance</a>
```
bruce-usab, line 121: Since [2.4.13](https://www.w3.org/TR/WCAG22/#focus-appearance) ended up a AAA, should that be reflecting in this paragraph?

> There is a <ins>criterion</ins> in WCAG 2.2 that addresses this aspect...

bruce-usab (approved):
Good with or without the "AAA" I ask about adding into line 121.

understanding/21/non-text-contrast.html
mbgower, line 121:
```suggestion
						The external green indicator (#008000) does contrast with the white background (#FFF) which the component is on, <strong>passing</strong> the criterion. It does not need to contrast with both the component background and the component, as visually the effect is that the button is noticeably larger, and it's not necessary for a user to be able to discern this extra border in isolation. Although this passes non-text contrast, it is not a good indicator unless it is very thick. <span class="wcag22">There is a AAA criterion in WCAG 2.2 that addresses this aspect, <a href="../22/focus-appearance-minimum.html">Focus Appearance</a>
```

understanding/21/non-text-contrast.html
```diff
@@ -118,10 +118,9 @@ <h4 id="related-focus">Relationship with Focus Visible</h4>
 				<figure id="figure-focus-outer-green">
 					<img src="img/ntc-focus-outer-green.png" alt="Three blue buttons on a white background, the middle has a dark green outline outside of the botton's non-focused boundary." width="400" />
 					<figcaption>
-						The external green indicator (#008000) does contrast with the white background (#FFF) which the component is on, <strong>passing</strong> the criterion. It does not need to contrast with both the component background and the component.
+						The external green indicator (#008000) does contrast with the white background (#FFF) which the component is on, <strong>passing</strong> the criterion. It does not need to contrast with both the component background and the component, as visually the effect is that the button is noticeably larger, and it's not necessary for a user to be able to discern this extra border in isolation. Although this passes non-text contrast, it is not a good indicator unless it is very thick. <span class="wcag22">There is a AAA criterion in WCAG 2.2 that addresses this aspect, <a href="../22/focus-appearance-minimum.html">Focus Appearance</a>.
```
gundulaniemann, line 121: In fact, the AAA requirement should not carry 'minimum' in its name or file name.

mbgower: @gundulaniemann Focus Appearance does not have "minimum" in its name, and changing the URL on another SC in this PR is out of scope.
Please feel free to open a separate issue to have this corrected.

scottaohara (approved):

understanding/21/non-text-contrast.html
alastc: The URL was a mistake, thanks for catching @gundulaniemann, Pat has updated it.

PR #2716, opened 2022-10-07
Update focus-appearance.html

mbgower:
Made the Understanding wording for inline links match the updated bounding box definition to address https://github.com/w3c/wcag/issues/2292


mbgower:
Added in some minor changes to the Understanding document, in regard to 'exception' and 'area' language, to respond to some other comments.

Issue #2713, opened 2022-10-06
2.5.8: Are all slider variants excluded or only sliders that look and are operated like sliders 


JAWS-test:
Please look at:

https://www.w3.org/WAI/ARIA/apg/example-index/slider/slider-rating.html

You can see 5 rating stars, which can be activated individually (in fact, even half stars can be activated, although this is not immediately obvious). Looking at it, I would assume that the stars have to be checked for the correct target size according to 2.5.8.

However, looking at the source code, I see that it is a slider and that this is considered as one control in 2.5.8. 

Now the question: Does the exception apply to all visual sliders or to all programmatic sliders or to both variants?

And what is meant by "granular values"? Are there sliders without granular values? Which ones? Or is that supposed to describe that every slider has granular values?

JAWS-test:
Thank you, @patrickhlauke 
If this opinion is shared by others, I would appreciate a addition in the Understanding to avoid misunderstandings

bruce-usab:
> If this opinion is shared by others, I would appreciate a addition in the Understanding to avoid misunderstandings

I think we can survey and then close this issue.  Or update Understanding and then close.

patrickhlauke:
there is a tentative PR change https://github.com/w3c/wcag/pull/2718

bruce-usab:
Ready for Survey tag might be premature, since it would be nice to knit up _all_ proposed changes for Understanding.

My recent thinking is that since 2.5.1 mentions "path-based gestures" that covers sliders.  2.5.7 Understanding can then be more about classic drag-and-drop style UI.

GreggVan:
Not sure what you mean by 
> My recent thinking is that since 2.5.1 mentions "path-based gestures" that covers sliders

So just for clarity

A  slider would not be a "path based gesture" since it only matters what the end point is.     You can jump from one point to another and it works fine. 

"Path-based" refers to things where the endpoints are not sufficient.  The entire path (and/or the speed along the path) is important in determining the outcome.
Examples
	- flight path of an airplane through obstacles or moving obstacles 
	- a signature
	- water color painting



gregg

------------------------------
Gregg Vanderheiden
***@***.***

alastc:
Some confusion on the conversation yesterday (and probably need Wilco to be involved to check is solves the problem raised). So will re-up it later.

bruce-usab:
[...]

PR #2718, opened 2022-10-09
Tweak explanation of sliders in 2.5.8 Target Size (Minimum) note

patrickhlauke:
the current "sliders with granular values" seems to be causing some confusion. the more common term for these is "continuous sliders"

Closes https://github.com/w3c/wcag/issues/2713

### DIFF
--- a/understanding/22/target-size-minimum.html
+++ b/understanding/22/target-size-minimum.html
@@ -31,7 +31,7 @@
                 <li><strong>User agent control:</strong> The size of the target is determined by the user agent and is not modified by the author;</li>
                 <li><strong>Equivalent:</strong> The function can be achieved through a different control on the same page that has an area of at least 24 by 24 CSS pixels.</li>
             </ul>
-            <p class="note">Targets that allow for values to be selected spatially based on position within the target are considered one target for the purpose of the Success Criterion. Examples include continuous sliders (which allow users to select very granular values along a range), color pickers displaying a gradient of colors, or editable areas where you position the cursor.</p>
+            <p class="note">Targets that allow for values to be selected spatially based on position within the target are considered one target for the purpose of the Success Criterion. Examples include continuous sliders (which allow users to select granular values along a range), color pickers displaying a gradient of colors, or editable areas where you position the cursor.</p>
             <p class="note">Are there issues with internationalization when describing inline links?</p>
             <p class="note">Are there issues with pop-over content overlapping targets triggering failures?</p>
             <dl><dt class="new"><dfn data-lt="target offsets">target offset</dfn></dt>
--- a/understanding/22/target-size-minimum.html
+++ b/understanding/22/target-size-minimum.html
@@ -31,7 +31,7 @@
                 <li><strong>User agent control:</strong> The size of the target is determined by the user agent and is not modified by the author;</li>
                 <li><strong>Equivalent:</strong> The function can be achieved through a different control on the same page that has an area of at least 24 by 24 CSS pixels.</li>
             </ul>
-            <p class="note">Targets that allow for values to be selected spatially based on position within the target are considered one target for the purpose of the Success Criterion. Examples include sliders with granular values, color pickers displaying a gradient of colors, or editable areas where you position the cursor.</p>
+            <p class="note">Targets that allow for values to be selected spatially based on position within the target are considered one target for the purpose of the Success Criterion. Examples include continuous sliders (which allow users to select very granular values along a range), color pickers displaying a gradient of colors, or editable areas where you position the cursor.</p>
             <p class="note">Are there issues with internationalization when describing inline links?</p>
             <p class="note">Are there issues with pop-over content overlapping targets triggering failures?</p>
             <dl><dt class="new"><dfn data-lt="target offsets">target offset</dfn></dt>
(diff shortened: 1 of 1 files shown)


detlevhfischer (approved):

patrickhlauke:
@alastc following your last change, this now removes any mention of continuous sliders. is that intentional?

melaniephilipp:
I’m wondering if some wires (i.e. SCs) got crossed in our discussion on issue #2713 today (or perhaps my brain is already on holiday). 

The original issue was about 2.5.8 Target Size and how to know when a slider's granularity is small enough to be part of the exemption “Examples include sliders with granular values, color pickers displaying a gradient of colors, or editable areas where you position the cursor”. 

But, if I remember correctly, somewhere along the line today it seems like talk turned toward sliders and 2.5.7 Dragging Movements – and the fact that you should be able single-point-click on sliders so they shouldn’t be excepted from 2.5.7 Dragging Movements at all. Also, comments on the issue itself turned from Target Size to Dragging Movements. 

The result was that sliders were removed entirely from the 2.5.8 Target Size exemption with this PR. Which is exactly the opposite of what @WilcoFiers proposed.

For reference, Wilco's survey comment is copied here: 

> I think what is "continuous" is fairly debatable. Is a 200 px wide slider with 5 options continuous? What about 50? what about 500? Where do you draw the line here. Even something like a color picker isn't truly continuous. It's a range between 0 and 255. It's all pretty arbitrary, and then starts to depend on things like how much pixel density does the screen have.
> 
> We should stay out of this and just exempt all sliders. WCAG has other requirements that ensure further accessibility of them, such as 2.1.1 keyboard.

patrickhlauke:
just picking up on @WilcoFiers's comment from the survey though

> WCAG has other requirements that ensure further accessibility of them, such as 2.1.1 keyboard

keyboard accessibility won't help a user that is reliant on a pointer device though, which is what 2.5.8 is about...

patrickhlauke:
Converting this to draft until clarification

patrickhlauke:
Re-added sliders, but just as "sliders" without any thin-slicing of how "granular" they may be. I believe this hits what @WilcoFiers intended.

If I got it wrong, please let me know. Leaving this as draft PR for now.

Incidentally ... am I missing a pointer to the meeting minutes? I'm not sure I saw minutes circulated on the mailing list

detlevhfischer (approved):
Looks good to me.

patrickhlauke:
I'll set this back to active PR / properly ready for review

JAWS-test:
I wish I hadn't said anything....
The original wording about the sliders was not perfect. But better than the last two suggestions:
* removing slider as exception is not useful, because of course there are sliders where each pixel represents a value and then 24 px click area is not possible
* adding slider as exception without specifying it is not useful either, because then everybody can mark a radiobutton group, a tablist, a tree, a selection list or other elements, which are used to select single values, as slider, in order not to have to keep a minimum size. A role=slider is sufficient, the visual appearance of the element does not matter, see: https://www.w3.org/WAI/ARIA/apg/example-index/slider/slider-rating.html

bruce-usab:
I think [sliders are essentially scrollbars](../../issues/2713#issuecomment-1327495817) so they do not warrant any special treatment.

But what about maps?  If every pixel is a target, do they all fail 2.5.8 or is there an exception?

patrickhlauke:
> But what about maps? If every pixel is a target, do they all fail 2.5.8 or is there an exception?

Excepted:

> **Essential**: A particular presentation of the target is [essential](https://w3c.github.io/wcag/guidelines/22/#dfn-essential) or is legally required for the information being conveyed;

but only if it's truly one of those maps where it's essential to be able to click on every pixel (e.g. to get precise lat/lon values)

bruce-usab:
I am very pleased to include maps into the _essential_ exception.  Anyone have concerns with that?

If maps are covered by this exception, then scrub-type-sliders would warrant the same exception from Target Size.

melaniephilipp:
@alastc @patrickhlauke Am I mistaken, or does this not appear in the current [WCAG 2.2 CR](https://www.w3.org/TR/WCAG22/#target-size-minimum) and [Understanding Doc](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html)?

patrickhlauke:
@melaniephilipp hmmm, you're right, this change doesn't seem to be in the `main` branch despite @alastc merging it at the time https://github.com/w3c/wcag/commit/42567419f9ad7b2163f239a4430cd6b67fa5ceaf

it appears that @alastc accidentally overwrote this change again a few days later. I mean, I assume it's accidental? https://github.com/w3c/wcag/commit/9a9ddba57393c2c77bfb889c26680ac46f54b92d

any chance you could rectify this and rerun this change, Alastair?

patrickhlauke:
(as an aside...i know these things happen, but please let's be absolutely careful/sure when merging something, particularly if it's been in flight for a while, to not accidentally overwrite other PRs/changes that happened in the meantime...I would have assumed this would have come up as a merge conflict at least. if not, always make sure to rebase/merge main back into your PRs before doing the final merge...otherwise we'll just keep overwriting things...)

melaniephilipp:
@alastc @patrickhlauke both this and the 2.4.11 Focus Not Obscured gap I mentioned earlier this week impact notes that appear on the face of the Spec. Does fixing them impact CR?

alastc:
Hi @patrickhlauke, from the [Jan 24th minutes](https://www.w3.org/2023/01/24-ag-minutes#item05) I can see we approved the version with 'sliders' but removing the granular phrasing, I'll re-instate that. The [later disucssions](https://www.w3.org/2023/02/14-ag-minutes#item12) were not on that topic, so it was a mistake to over-right that.

@melaniephilipp The focus-not-obscured on is in the survey for today.

patrickhlauke:
lovely, thank you @alastc

PR #2723, opened 2022-10-11
Correct typo in 2.1.4 understanding

patrickhlauke:
`Crition` > `Criterion`


patrickhlauke:
only just realised that this fix is already included in my old https://github.com/w3c/wcag/pull/2455 pull request, which has been languishing for a while. IF that gets merged, this will be superseded. but as i don't see any movement on it, leaving this open for now...

PR #2725, opened 2022-10-11
Update label-in-name.html

### LINKED ISSUE #2276: Notes in the label should not have to be part of the AccName (2.5.3)
I wonder if it would be useful to add in the Understanding to 2.5.3 that hints (that are in the visually visible label) do not have to be part of the Accessible Name to satisfy SC 2.5.3.

Example:
* Formatting hint (label: "Date (yyyy-mm-dd)" - AccName "Date" would be sufficient if input hint is transmitted elsewhere, e.g. as description).
* Accesskey (label "Save (Alt+S)" - AccName "Save" would be sufficient if accesskey is transmitted elsewhere, e.g. via aria-keyshortcuts)

Reason:
* Accessible Name should be concise.
* Accesskeys should be transmitted programmatically rather than in text form
* Input hints should be transmitted as description rather than as part of the name
* no one will try to call the element by its accesskey or input hint when using speech input

### LINKED ISSUE #2302: 2.5.3 Label in Name - aria-hidden on part of visual label
Hello :) 

I'd love to get a take from the working group on if folks feel this scenario is a failure of [2.5.3 Label in Name](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name.html). It's a real head-scratcher. Does it pass the test steps found in [F96](https://www.w3.org/WAI/WCAG21/Techniques/failures/F96#tests)? 

The intent within the code is to try to avoid the double announcement of 'required' with a screen reader by hiding that in the visual label with `aria-hidden` and instead relying on that from the `required` attribute. 

I've noticed that with some voice recognition software (iOS Voice Access & macOS Voice Control) if one says "Tap, First name required" it doesn't work as expected and fallbacks must be used. So the screen reader win may impact some voice recognition a bit.

``` html
<label for="first-name">
    First Name <span aria-hidden="true">(required)</span>
</label>
<input type="text" id="first-name" required>
``` 
First text name field with (required) in label)

Please excuse best practice / usability - strictly WCAG failure consideration here.

mbgower:
Adds examples and modifies wording to address two issues:
Closes #2302
Closes #2276


understanding/21/label-in-name.html
```diff
@@ -61,12 +65,16 @@ <h3>Accessible Name and Description Computation specification</h3>
...
-			</section>
+</section>
+<h4>Parenthetical text</h4>
+	<section>
+	<p>It is important to mention parenthetical text in labels in the context of accessible name versus description. In common usage, text in parentheses is considered secondary but relevant to meaning. Users of speech recognition would not typically announce text in parentheses as part of the input name. For that reason, parenthetical text may be optionally considered a description and left out of the accessible name.</p>
+	<p>However, where parenthetical information provides important context, as indication of a required field or guidance on input masks, this information must be provided programmatically in some other way to the user if that information is not included as part of the accessible name. For example, "Name (required)" and "Date (YYYY-MM-DD)" could accept "Name" and "Date" as the accessible names, but in order to pass 1.3.1 Info & Relationships authors would need to surface the required attribute programmatically through the html <code>required</code> attribute or <code>aria-required</code>, as well as the mask information (such as through <code>aria-describedby</code>).</p>
```
JAWS-test, line 77: HTML instead of html?

understanding/21/label-in-name.html
```diff
@@ -61,12 +65,16 @@ <h3>Accessible Name and Description Computation specification</h3>
...
  <p>It is important to note that the specification allows authors to override the name calculated through native semantics. Both <code>aria-label</code> and <code>aria-labelledby</code> take precedence in the name calculation, overriding the visible text as the accessible name even when the visible text label is programmatically associated with the control. For this reason, when a visible label already exists, <code>aria-label</code> should be avoided or used carefully, and <code>aria-labelledby</code> should be used as a supplement with care.</p>
 <p>Finally, <code>aria-describedby</code> is not included in the Accessible Name computation (instead it is part of the Accessible Description computation). By convention, text associated with a control through <code>aria-describedby</code> is announced immediately after the accessible name by screen readers. Therefore, the context of headings, instructions, and group labels can be provided through the accessible description to assist users of screen readers without affecting the experience of those who navigate using speech recognition software.</p>
-			</section>
+</section>
+<h4>Parenthetical text</h4>
```
detlevhfischer, line 74: Even if parenthetical content is the most common case, would it be better to indicate that there may be different ways to visually separate additional content that may then not need not be considered as part of the accname? For example, additional content may be visually separated by different formatting (label bold, additional text normal or italic); separation by a vertical bar; additional text placed  on a new line etc. etc. Should we really constrain this to parentheses only?

detlevhfischer (changes requested):

understanding/21/label-in-name.html
mbgower: @detlevhfischer Parentheses are the common use case. I'm not aware of text styling differences being treated differently by speech input users, and I have not seen the use of a separator. Do you have examples that you have in mind?
Ultimately, we're treading on soft ground here.  I'm inclined to keep the consideration contained and essential.

We already have guidance about text in **close** proximity being a primary driver, so I don't think material on a new line needs to be listed here.

understanding/21/label-in-name.html
mbgower: Done

bruce-usab (approved):
This is really good detailed information.

>  here parenthetical information provides important context, as indication of a required field or guidance on input masks ... as well as the mask information

I think we need a less technical term than <q>mask</q>.  But I do not have a suggestion at the moment.

bruce-usab:
Thanks Mike, that is better.  I agree that the prose needs to avoid <q>requirements</q>.   Below I use limits instead of restrictions:  I split the last sentence into two, so I could spell out the second limitation a little more. 

Where parenthetical information provides important context, such as indication of a required field or limitations on what is allowed for input, this information must be provided programmatically in some other way to the user, if that information is not included as part of the accessible name. For example, "Name (required)" and "Date (YYYY-MM-DD)" could accept "Name" and "Date" as the accessible names, but in order to pass 1.3.1 Info &amp; Relationships, authors would need to surface the required attribute programmatically as well as limits on the allowed data formatting (in this example, eight integers fitting the YYYY-MM-DD pattern).  <q>Required</q> should be surfaced through the HTML required attribute or aria-required.  The allowed  data formatting may be conveyed through aria-describedby.

bruce-usab (approved):

bruce-usab (approved):

understanding/21/label-in-name.html
```diff
@@ -61,12 +65,16 @@ <h3>Accessible Name and Description Computation specification</h3>
...
+	<p>However, where parenthetical information provides important context, such as indication of a required field or limitations on what is allowed for input, this information must be provided programmatically in some other way to the user if that information is not included as part of the accessible name. For example, "Name (required)" and "Date (YYYY-MM-DD)" could accept "Name" and "Date" as the accessible names, but in order to pass 1.3.1 Info & Relationships, authors would need to surface the required attribute programmatically, as well as any limits on the allowed data formatting (in this example, eight integers fitting the YYYY-MM-DD pattern). "Required" should be surfaced through the HTML <code>required</code> attribute or <code>aria-required</code>. The allowed data formatting may be through <code>aria-describedby</code>).</p>
```
scottaohara, line 77: Some minor wording adjustments.  Most notable one was to correct the reference to the "required attribute" when the sentence was talking about programmatically surfacing state.  

```suggestion
	<p>However, where parenthetical information provides important context, such as indication of a required field or limitations on what is allowed for input, this information must be provided programmatically in some other way to the user if that information is not included as part of the accessible name. For example, "Name (required)" and "Date (YYYY-MM-DD)" could accept "Name" and "Date" as the accessible names. However, in order to pass 1.3.1 Info & Relationships, authors would need to programmatically surface the required state, as well the noted limits on the allowed data formatting (in this example, eight integers fitting the YYYY-MM-DD pattern). The "required" state could be surfaced through the HTML <code>required</code> attribute or by using <code>aria-required="true"</code>. The allowed data formatting could be surfaced by being referenced using the <code>aria-describedby</code>) attribute.</p>
```

mbgower:
@scottaohara can you have a look at the example I added in the technique to illustrate this, and suggest any changes?

@mraccess77 I believe between the changes in the Understanding document and the additions to the technique, this has addressed your concerns. Please let me know.

techniques/general/G211.html
```diff
@@ -149,7 +149,30 @@ <h4>Range of inputs with few labels</h4>
...
+			<p>Technique G211 is not intended to complicate existing conventions for the construction of accessible form inputs. As described in the <em>Punctuation and capitalization</em> and <em>Parenthetical text</em> subsections of the Understanding document, information that would not normally be spoken when using speech recognition to navigate to controls does not need to be in the accessible name.  Where there are established ways of ensuring information and relationships conveyed visually are present programmatically, it is acceptable to leave parenthetical information and punctuation out of the accessible name.</p>
+			<p>The following snippet demonstrates possible techniques mentioned in the Understanding document. Since the required field is programmatically indicated, and the input restrictions on the date are surfaced through <code>aria-describedby</code>, the asterisk and parenthetical information is not required in the accessible name.</p>
+		
+
+<code>	
+&lt;label for="name"&gt;Name&lt;/label&gt; *<br />
+&lt;input type="text" name="name" id="name" required&gt;<br />
+&lt;label for="birth"&gt;Birthdate&lt;/label&gt; <span id="mask">(YYYY-MM-DD></span><br />
```
scottaohara: does the span for the mask also need to have its `<` and `>` changed to `&lt;` and `&gt;`??

patrickhlauke: yes, it should

mraccess77:
The main thought I have is that the check test is specific to excluding only parenthetical content that actually appears in parenthesis - in my experience most parenthetical text is likely not in parens but on a different line - on the other side of an input or somehow made to look different from the field name.  Could we update the check to account for this and match the discussion earlier in the document where we say "not typically used to focus the control by name?".

techniques/general/G211.html
mbgower: i thought I already correct that. Is it still showing up for you?

mbgower:
> in my experience most parenthetical text is likely not in parens but on a different line - on the other side of an input or somehow made to look different from the field name

So, the step instructions specifically say "information in parentheses". I think we should be very careful to not use the term "parenthetical" in a more abstract way here, because it will lead to confusion.

To address this potential misunderstanding, I have updated the heading in the Understanding document to read "Text in parentheses" and added a note directly below explaining that this literally means text appearing within a set of round brackets. I have done similar updates to the technique.

> in my experience most parenthetical text is likely not in parens but on a different line - on the other side of an input or somehow made to look different from the field name. Could we update the check to account for this

There is a _lot_ of guidance, both in this general technique and in the Understanding document, talking about how interpretation of the label should be done. I'm leery of trying to summarize that in the steps of a test.

I've added in "label" in steps 2 and 4 to make it clear we are talking about the text that has been identified as the label.

understanding/21/label-in-name.html
```diff
@@ -61,12 +65,17 @@ <h3>Accessible Name and Description Computation specification</h3>
...
 <p>Finally, <code>aria-describedby</code> is not included in the Accessible Name computation (instead it is part of the Accessible Description computation). By convention, text associated with a control through <code>aria-describedby</code> is announced immediately after the accessible name by screen readers. Therefore, the context of headings, instructions, and group labels can be provided through the accessible description to assist users of screen readers without affecting the experience of those who navigate using speech recognition software.</p>
-			</section>
+</section>
+<h4>Text in parentheses</h4>
+	<section>
+		<p>Note: The term "parenthical" in this section is describring text that literally appears within a set of round brackets (). It is not used in the more abstract sense of "information related to".</p>
```
patrickhlauke, line 76: `Note: The term "parenthical"` > `Note: The term "parenthetical"`

patrickhlauke: maybe overthinking it here, but would this first sentence somehow be interpreted as "if it's not in *round* brackets, but square or curly ones, then it doesn't count"?

mbgower: addressed

mraccess77:
I'm trying to understand what info is considered in or out of the label.  From what I understand you are saying that if it looks like it's in the label but it has parens then it doesn't need to be included.  It also sounds like other content that would not be considered part of the label would not be covered by the SC by nature of the definition of label and that this information may have other factors that set it apart but that we are not addressing those in this pull request?  Example, different font weight than label or whitespace between label and other text, etc.  If I understand correctly, that makes sense to me.

understanding/21/label-in-name.html
mbgower: From a common usage perspective, anything other than parentheses would be _highly_ unusual. Curly and square brackets are not by convention used in a form label. I wouldn't even want to dignify mentioning them here for fear someone sees that as some kind of advocation! And yeah, I think you're overthinking a bit :)

mbgower:
> From what I understand you are saying that if it looks like it's in the label but it has parens then it doesn't need to be included. 

Correct, as per https://github.com/w3c/wcag/issues/2276

> It also sounds like other content that would not be considered part of the label would not be covered by the SC by nature of the definition of label and that this information may have other factors that set it apart but that we are not addressing those in this pull request?

I'd say that is correct @mraccess77. I don't know how I keep finding myself inheriting maintenance of SCs that I was originally concerned/skeptical about, but Label in Name certainly continues the trend.

The concept of Label in Name seemed simple enough when first brought up, but as you note there are many things that may lead to someone thinking of a chunk of text as being (or not being) a label. The challenge is compounded where we can clearly say 'this is the accessible name' and yet there are many cases where even where folks can agree on what the visible label is, it's not clear that ALL of that visible label is really needed (or even helpful) as part of the Accessible name. 

Most of the written guidance in the Understanding doc and general technique have existed for several years now, so overall I think it seems to have worked well. These changes will hopefully add clarity in scenarios like the two issues that motivated these changes.

bruce-usab (approved):
Very nice, thanks!

understanding/21/label-in-name.html
```diff
@@ -61,12 +65,17 @@ <h3>Accessible Name and Description Computation specification</h3>
...
  <p>It is important to note that the specification allows authors to override the name calculated through native semantics. Both <code>aria-label</code> and <code>aria-labelledby</code> take precedence in the name calculation, overriding the visible text as the accessible name even when the visible text label is programmatically associated with the control. For this reason, when a visible label already exists, <code>aria-label</code> should be avoided or used carefully, and <code>aria-labelledby</code> should be used as a supplement with care.</p>
 <p>Finally, <code>aria-describedby</code> is not included in the Accessible Name computation (instead it is part of the Accessible Description computation). By convention, text associated with a control through <code>aria-describedby</code> is announced immediately after the accessible name by screen readers. Therefore, the context of headings, instructions, and group labels can be provided through the accessible description to assist users of screen readers without affecting the experience of those who navigate using speech recognition software.</p>
```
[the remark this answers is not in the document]
mbgower: The section on Accessible Name and Description already says quite a bit, including
> It is important to note that the specification allows authors to override the name calculated through native semantics. Both aria-label and aria-labelledby take precedence in the name calculation, overriding the visible text as the accessible name even when the visible text label is programmatically associated with the control. For this reason, when a visible label already exists, aria-label should be avoided or used carefully, and aria-labelledby should be used as a supplement with care.

I'm worried it's already into the technical weeds a bit for an understanding document. Arguably parts of the doc are already more appropriate for a technique, IMO.

understanding/21/label-in-name.html
```diff
@@ -61,12 +65,17 @@ <h3>Accessible Name and Description Computation specification</h3>
...
+	<p>However, where parenthetical information provides important context, such as indication of a required field or limitations on what is allowed for input, this information must be provided programmatically in some other way to the user if that information is not included as part of the accessible name. For example, "Name (required)" and "Date (YYYY-MM-DD)" could accept "Name" and "Date" as the accessible names. However, in order to pass 1.3.1 Info & Relationships, authors would need to programmatically surface both the required state and the limit on the allowed data formatting (in this example, eight integers fitting the YYYY-MM-DD pattern). The "required" state could be surfaced through the HTML <code>required</code> attribute or by using <code>aria-required="true"</code>. The allowed data formatting could be surfaced by being referenced using the <code>aria-describedby</code>) attribute.</p>
```
kengdoj, line 78: Is it possible to list outliers to this? In the US, area codes for telephone numbers are sometimes shown within parentheses. Ex: (202)555-5555. The area code should be part of the name.

nattarnoff: @kengdoj I think this use case falls under the "However, where parenthetical information provides important context, such as indication of a required field or limitations on what is allowed for input." I know how much easier it is top point developers to a line in a spec or documentation, but the possible number of cases that could be an outlier is potentially huge. 
Also, I question any label that includes the formatting. That really should be aria-describedby information.

chaals: I don't think this text increases understanding. Parenthetical information (relevant, but not necessary to understand), is probably better in the description than the label. Whether it is presented in parentheses (round brackets) or not seems more like a typographical accident.

I'm prone to putting a lot in parentheses. Some use commas and subsidiary clauses instead, others write more sentences.

working-examples/label-in-name-general/example3.html
chaals, line 1: This is an odd way to construct a phone number in the 21st century. In the US it is normal to have an area code and a number. There are still cases where there is an extension or line number, but while that used to be common decades ago it makes the example unusual and unlkely. (Like asking for gramophone playing speed and A/B sides for a piece of music recorded in 2014).

Far more usual is a country code and a number.

nattarnoff: Agreed. I know a lot of US citizens still consider the area code as separate from the number, but "local" calls are a misnomer anymore. Most cell phones require area code. Numbers are now portable across carriers and are attached to devices that aren't locked to a location. US phone numbers **are** 10 digits + country code. To make the phone number field anything but a single field for 10 characters makes little sense to me.

tombrunet: local calls being a "misnomer" somewhat depends on where you live, how dense the population is, and how many non-locals live there. Austin, TX didn't add a second area code until 2013. Prior to that, most people dialed with just 7 digits. I can still call a number with the same area code using 7 digits from my cell. Now that a lot of people have moved into the area with non-local numbers and newer numbers are different, I now am accustomed to always including an area code, but there's no requirement for me to do so.

I could certainly envision a website for a local company in a smaller city that doesn't service a lot of non-locals that doesn't have an expectation of an area code.

kengdoj (approved):
ACT (Accessible Conformance Testing) has reviewed, and we are fine with these updates.

understanding/21/label-in-name.html
mbgower: I'd agree with @nattarnoff that the example of the area code in parentheses doesn't seem to enter into the discussion on parentheses in a _label_. @kengdoj can you supply an example of an area code in a phone input label that would be problematic with the guidance listed in this subsection?

mbgower: > Parenthetical information (relevant, but not necessary to understand), is probably better in the description than the label.

@chaals I entirely agree, but there's an unfortunate gap between what people should do and what they DO.  As well as a gap between the intent of this SC and how it is being interpreted These changes were introduced to address issues opened by users, and have support. In fact, this has already been voted on and approved by the WG.

working-examples/label-in-name-general/example3.html
mbgower: If someone has a preferred example of a single text string labelling multiple inputs, happy to include. I stress that this PR was already approved by the WG. We're awaiting a review from ACT for the purpose of reviewing any rules considerations that result; otherwise this would have been merged almost a year ago.

mraccess77:
One example of a single label for multiple fields could be a query builder.  These tend to have multiple rows and columns that you can add criteria like operators and operands to and they tend to only have a column label at the top for clarity sake.

Issue #2251, opened 2022-03-07
SC 1.4.3 should say contrast ratio 4.50:1 instead of 4.5:1 


julezrulez:
I still have argues about contrast just under 4.5:1. For instance 4.49:1. I have had math several decades ago. I have learned there the following rounding rules: if numbers are written as 4.5 than this could be 4.45 through 4.54. So when measuring things, it should be very clear that contrast of 4.49 is almost right (which actually means that contrast fails).

So I propose to change the texts of this success criterion, like for instance:
- "The visual presentation of text and images of text has a contrast ratio of at least **4.50:1**, ...:"
- "Large-scale text and images of large-scale text have a contrast ratio of at least **3.00:1**;"

(changed text is marked by **bold**)

This might save a discussions with devs and avoid mistakes in advance.

JAWS-test:
I think 2 decimal places won't help, as even 4.4999 would be too little, although it can be rounded to 4.50. Therefore a note in the Understanding would be more meaningful

patrickhlauke:
related https://github.com/w3c/wcag/issues/200

patrickhlauke:
ah, forgot what the actual follow-up was on those. seems like it's also worth adding the same/similar note to non-text contrast and possibly use of color understanding as well? if yes, i can tackle that in a separate PR @alastc

julezrulez:
Thanks for your responses. It's almost clear to me. I understand that changing normative text is difficult. (The normative text however is where we measure accessibility with, so that text should be precise in that.). I can live with the text being put into the  understanding documents. The text in the pull requests are clear enough.

That raises the question where I can find the proposed texts. I expected to find it here: https://www.w3.org/WAI/WCAG21/Understanding/contrast-enhanced.html
Also in the WCAG 2.0 documentation I could not find the text.

patrickhlauke:
hmmm, indeed. I can't find those notes from the diffs @alastc referenced anywhere ...

* https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html
* https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast7.html
* https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum
*  https://www.w3.org/WAI/WCAG21/Understanding/contrast-enhanced
* https://www.w3.org/WAI/WCAG22/Understanding/contrast-minimum
[...]

PR #2254, opened 2022-03-09
Add note about rounding to non-text contrast understanding, fix `Success Criteria` to `Success Criterion`

patrickhlauke:
In discussion for https://github.com/w3c/wcag/issues/2251 it emerged that changes were pushed in 2016 to the understanding documents for 1.4.3 and 1.4.6, clarifying that values should not be rounded when evaluating pass/fail for contrast ratios - see https://github.com/w3c/wcag/issues/2251#issuecomment-1061917901

However, it now appears that this change at the time somehow fell through the cracks in the various reorganisations of the repositories here. And indeed, the two notes that were agreed by AGWG and merged at the time are nowhere to be found - not in any of these

* https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html
* https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast7.html
* https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum
*  https://www.w3.org/WAI/WCAG21/Understanding/contrast-enhanced
* https://www.w3.org/WAI/WCAG22/Understanding/contrast-minimum
* https://www.w3.org/WAI/WCAG22/Understanding/contrast-enhanced

This PR reintroduces those notes (since they were already agreed back in 2016, and are clearly still valuable) to 1.4.3 and 1.4.6, and also adds a matching note to 1.4.11.

Ideally, this note should just be added once to the contrast ratio definition https://w3c.github.io/wcag/21/guidelines/#dfn-contrast-ratio ... but that would be a change of normative wording which is doubtful will get through. so, while duplicative, this at least adds it to the non-normative understanding documents.

This note should not only go into 2.2, but also 2.1 (and ideally 2.0, but i doubt these get touched anymore)

EDIT: after merging main back into this PR, and some parallel changes that got pushed there earlier, this PR is now just correcting `Success Criteria` to `Success Criterion` (singular) in the contrast-minimum and contrast-enhanced understanding, and then adding the same note to non-text contrast understanding

Issue #200, opened 2016-07-08
Rounding and Color contrast


DavidMacDonald:
The color value #777777 is evaluated on some analyser tools as 4.5:1 (pass), and 4.48:1 (fail) on other tools on our resource list. 
https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html#visual-audio-contrast-contrast-resources-head
Tests here:
http://davidmacd.com/test/check-color.html

The actual value appears to be 4.48. 
https://twitter.com/jared_w_smith/status/751433009316794368

My personal opinion is that the SC wording of "at least 4.5:1" means it cannot be rounded up to the threshold from a lower number. 

But this started quite a discussion on twitter with some saying that science, math and case law points to 4.48 passing... it is argued that because we only had **one** digit in the decimal 4.5:1 for the SC then any **2** digit value can be rounded up or down according to rounding practices.( i.e., 4.44 would round to 4.4 failing and 4.45 would round to 4.5 which is a pass.)

The minimum contrast value up until the last draft of WCAG 2 was 5:1. We reluctantly accepted 4.5:1 on the last draft recommended by the Access Board ... if we had stayed at 5 then the under the argued rounding reasoning, 4.5:1 would still pass because it it rounds up to 5:1.

Let's make a decision and put it in the Understanding. I suggest something explicit such as.
>When this Success Criteria says "at least 4.5:1", it means this is the threshold, and that no number less than that can pass, even if it is rounded."

pauljadam:
Can the normative SC wording write it out to the decimal place it's expecting folks to round? i.e. 4.50:1.00 and 3.00:1.00? Or something that is very specific about the rounding if you can round or not?

DavidMacDonald:
Maybe in 2.1, but we even cringed at 4.5:1 instead of 5:2, adding more digits seems a little academic.

awkawk:
I think that if you were in a dispute over this somehow that:
1) the rest of the site must be fantastic as this is a very minor difference
2) you would be asked to provide a scientific basis for the treatment of the numbers.  4.5, as written in WCAG 2.0 has 2 significant digits.  To compare a number to it, in this case 4.48, you might subtract 4.5-4.48, which gives you 0.02, but since there can only be the same number of significant digits in the answer, the answer for the difference is actually 0.  This would mean that 4.46 and greater (or even 4.456 if you have a tool that gives that number of decimal places) could be viewed as meeting a 4.5:1

[...]

PR #2726, opened 2022-10-11
Rounding update

Myndex:
This clarifies the rounding note in the contrast SCs, essentially modifying PR #2254 as:

> _....the computed value must be rounded to a precision of at least one decimal place, i.e. 3.0:1. (For example, 4.45:1 would round up to the 4.5:1 threshold, but 4.44:1 would not)._

This pull request was suggested in issue #2251 

This is functionally equivalent to the meaning of these SCs per 2.0, and for 2.1 prior to PR #2254.

RATIONALE: When specifying if something should or should not be rounded, it is incomplete to say "don't round". What is needed is the specific desired precision to be rounded to.

In the present case, rounding to one decimal point is the difference of no more than one 8bit code value, as shown in the below example. This means that rounding to a higher precision brings no functional benefit. 

4 5 vs 4 45 examples shown, there is no visible difference as changing a single code value in this use case is below the threshold of human perception

This potentially closes issue #2251 

Thank you for reading.

Andy

### DIFF
--- a/understanding/20/contrast-enhanced.html
+++ b/understanding/20/contrast-enhanced.html
@@ -69,7 +69,8 @@
          </p>
          <p>The 7:1 and 4.5:1 contrast ratios referenced in this Success Criterion are intended to be
             treated as threshold values. When comparing the computed contrast ratio to the Success Criterion
-            ratio, the computed values should not be rounded (e.g. 4.499:1 would not meet the 4.5:1 threshold).</p>
+            ratio, the computed value must be rounded to a precision of at least one decimal place, i.e. 7.0:1.
+            (For example, 4.45:1 would round up to the 4.5:1 threshold, but 4.44:1 would not).</p>
          
       </div>
       
--- a/understanding/20/contrast-minimum.html
+++ b/understanding/20/contrast-minimum.html
@@ -63,8 +63,8 @@
          </p>
          <p>The 3:1 and 4.5:1 contrast ratios referenced in this Success Criterion are intended to be
             treated as threshold values. When comparing the computed contrast ratio to the Success Criterion
-            ratio, the computed values should not be rounded (e.g. 4.499:1 would not meet the 4.5:1 threshold).</p>
-         
+            ratio, the computed value must be rounded to a precision of at least one decimal place, i.e. 3.0:1.
+            (For example, 4.45:1 would round up to the 4.5:1 threshold, but 4.44:1 would not).</p>
       </div>
       
       <p>The previously-mentioned contrast requirements for text also apply to images of text
--- a/understanding/21/non-text-contrast.html
+++ b/understanding/21/non-text-contrast.html
@@ -16,7 +16,7 @@
 			<p>Low contrast controls are more difficult to perceive, and may be completely missed by people with a visual impairment. Similarly, if a graphic is needed to understand the content or functionality of the webpage then it should be perceivable by people with low vision or other impairments without the need for contrast-enhancing assistive technology.</p>
 
 			<div class="note">
-				<p>The 3:1 contrast ratios referenced in this Success Criterion is intended to be treated as threshold values. When comparing the computed contrast ratio to the Success Criterion ratio, the computed values should not be rounded (e.g. 2.999:1 would not meet the 3:1 threshold).</p>
+				<p>The 3:1 contrast ratios referenced in this Success Criterion is intended to be treated as threshold values. When comparing the computed contrast ratio to the Success Criterion ratio, the computed value must be rounded to a precision of at least one decimal place, i.e. 3.0:1. (For example, 2.95:1 would round up to the 3.0:1 threshold, but 2.94:1 would not).</p>
 			</div>
 
 			<section id="user-interface-components">


patrickhlauke:
Originally posted as a comment here https://github.com/w3c/wcag/issues/200, but probably more relevant here

To the point of "you can't just say 'no rounding' as there will always be rounding", the implicit meaning, I'd argue, of the common phrase "no rounding" is of course "up to the limit of the platform / variable type used". as this will likely be a float, and a float has - if i remember correctly - 15 decimals of precision, that would be my general understanding of what the implied precision is.

"No rounding" to me means something like this

> if (8.999999/2)>=4.5) { console.log('Pass'); } else { console.log('Fail'); }
Fail
whereas "with rounding" is something like this

> if (parseFloat((8.999999/2).toPrecision(2))>=4.5) { console.log('Pass'); } else { console.log('Fail'); }
Pass
Yes, even that first calculation effectively uses "rounding" in the sense that a float can't have more than 15 meaningful decimals...but that's hair splitting (and yes, ironic coming from me, the chief hairsplitter normally)

I'll also note that (and yes, we disagree on whether or not it's good or bad) the formula for the contrast calculation https://www.w3.org/TR/WCAG21/#dfn-contrast-ratio has those + 0.05 bits, which implies at least a precision of two decimals. so if we must specify a rounding/precision, then I'd say it should be at least 2 decimals

Myndex:
Hi @patrickhlauke 

> _....has those + 0.05 bits, which implies at least a precision of two decimals._ 

No, that does not imply any such thing. The "0.05 bits" are applied to both sides of the ratio divide, so they null out to 1.

### ${0.05 \over 0.05} = 1$

The useful precision is actually limited by the input luminance which for WCAG2 is the IEC sRGB piecewise calculated relative luminance. The effective/useful luminance precision is limited by both of **1)** the 8-bit-per-channel color values and **2)** human perception of an sRGB self-illuminated display. I've already demonstrated that one decimal point of precision is adequate for this given use case, particularly in view of how 8-bit-per-channel code values affect the luminance calculation.


> _as this will likely be a float, and a float has - if i remember correctly - 15 decimals of precision_

This depends on the language. A _FLOAT_ is typically 6 or 7 decimal digits of precision. A _DOUBLE_ is 15 digits. In Javascript all numbers are doubles (despite their indicated type, all stored as doubles). But for properly typed languages, other types may be used, and that leads us to mention briefly some of the problems working with floats.

For instance in javascript, `0.1 + 0.2 == 0.3 is FALSE`,
because `0.1 + 0.2 = 0.30000000000000004`

The point is, we are not also specifying that doubles must be used, nor should we.

What we _SHOULD_ be doing is specifying the minimum needed precision, which as I've shown is one decimal point in terms of the results from the WCAG 2 math.

And that is the purpose of this pull request.

patrickhlauke:
> The "0.05 bits" are applied to both sides of the ratio divide, so they null out to 1.

so you're saying that `(L1 + 0.05) / (L2 + 0.05)` can simply be taken as `L1 / L2`?

GreggVan:
That still leaves an ambiguity gap

Would suggest instead that we say (for the whole standard)  that 
the precision for each measurement in this recommendation should be calculated to the number of decimal places specified, with intermediate values rounded up or down with the midpoint (5) rounded up.  


Or some such




Gregg Vanderheiden
***@***.***

Myndex:
> _so you're saying that `(L1 + 0.05) / (L2 + 0.05)` can simply be taken as `L1 / L2`?_

No, not precisely — the effect is that instead of `#000` on `#fff` being infinite contrast (which it would with `L1 / L2`) max contrast is clamped at 21. For lower contrasts, there is a mild derating of the value, but it's simply not enough.

I've experimented with asymmetrical versions, as did Hwaung/Peli... still not fit for the use case tho. Dropping the 180 year old Weber variants and moving to math derived from the much more modern Stevens-type curves has proved to be more valuable in predicting contrast.

patrickhlauke:
> Would suggest instead that we say (for the whole standard) that the precision for each measurement in this recommendation should be calculated to the number of decimal places specified, with intermediate values rounded up or down with the midpoint (5) rounded up.

so for, say 1.4.12 Text Spacing you'd accept a line height of 1.45 times the font size, or 1.95 font height of spacing between paragraphs? for 1.4.7 Low or No Background Audio accepting 19.5 dB as good enough to pass the 20 dB threshold? i mean sure, let's put it to a vote. at least then we'd have some consistency/clarity of interpretation

GreggVan:
That still leaves an ambiguity gap

Would suggest instead that we say (for the whole standard)  that 
the precision for each measurement in this recommendation should be calculated to the number of decimal places specified, with intermediate values rounded up or down with the midpoint (5) rounded up.  


Or some such




Gregg Vanderheiden
***@***.*** ***@***.***>

Myndex:
Hi @patrickhlauke 

> _so for, say 1.4.12 Text Spacing you'd accept a line height of 1.45 times the font size, or 1.95 font height of spacing between paragraphs? for 1.4.7 Low or No Background Audio accepting 19.5 dB as good enough to pass the 20 dB threshold? i mean sure, let's put it to a vote. at least then we'd have some consistency/clarity of interpretation_

YES on all those counts, most definitely.

1) 1.4.12 text spacing is incorrect in the first place as it specifies a relationship to font size and that is not instructive. It **_should_** be related to x-height, NOT font-size.

Though even then, there are enough variations in glyph design that the 1.4.12 spec is an arbitrary value, and "not ideally instructive".

2) 1 dB is the minimum threshold for a just noticeable difference. There is no perceivable difference between 19.5dB and 20dB, and also, like other perceptions, aural perception is _EXTREMELY_ context sensitive.

Beyond that 1.4.7 has enough "things" to, eh, "assist" that I am not even going to discuss it today. Maybe in a few years when I've done with visual contrast. But as a side note these arbitrary single-dimensional SCs relating to perception are very triggering to me to say the least.


---

That's it for me today, you guys wore me out. I'll be back next week.

bruce-usab (changes requested):
@Myndex no harm asking, but the question of rounded is settled as a wontfix

bruce-usab:
Here is the 2016 issue thread:  https://github.com/w3c/wcag/issues/200

See [10/25/2016 AGWG minutes](https://www.w3.org/2016/10/25-wai-wcag-minutes.html#item04).

From that thread, the decision was recorded in the [decision wiki on 27 October 2016]( https://www.w3.org/WAI/GL/wiki/Decisions#2016) following [CFG via email](https://lists.w3.org/Archives/Public/w3c-wai-gl/2016OctDec/0310.html).

patrickhlauke:
based on the "won't fix", closing this PR

PR #2735, opened 2022-10-17
Add clarity for SC 3.2.6

### LINKED ISSUE #2688: Subtle grammar issue in 3.2.6
> If a [web page](https://www.w3.org/TR/WCAG22/#dfn-web-page-s) contains any of the following help mechanisms, and those mechanisms are repeated on multiple web pages within a [set of web pages](https://www.w3.org/TR/WCAG22/#dfn-set-of-web-pages), they occur in the same relative order to other page content, unless a change is initiated by the user

As I understand it, this bit:

`they occur in the same relative order to other page content`

should be written

`they occur in the same order relative to other page content`

In the first example, "relative" modifies the word "order" which suggests the order is the same to a certain degree when compared with other things that have an order.

In the second, the phrase "relative to" more clearly establishes the relationship with "other page content" which is what was intended.

Happy to submit a PR if y'all agree this change is correct.

dtex:
- Add link to defined term on SC 3.2.6
- Expand definition of "Same Relative Order"
- Closes #2688


bruce-usab:
@mbgower please do take a look if you can make the time on the [linked issue thread](https://github.com/w3c/wcag/issues/2688#issuecomment-1281091380).

TLDR:  [SC 3.2.6 Consistent Help](https://www.w3.org/TR/WCAG22/#consistent-help) uses "same relative order" in a way that is not wholly consistent with the ]glossary use of those three words](https://www.w3.org/TR/WCAG22/#dfn-same-relative-order) as referenced by [3.2.3 Consistent Navigation](https://www.w3.org/TR/WCAG22/#dfn-same-relative-order).

bruce-usab:
I am of the opinion that the least bad fix is to swap to words in the CR SC text, so that 3.2.6 is _not_ using the defined term.

> If a web page contains any of the following help mechanisms ... they occur in the **same <ins>order</ins> relative <del>order</del>**  to other  page content...

mbgower:
Yes, as noted in the issue, I think it makes more sense to NOT touch the definition and treat this as an editorial change by reversing two words.

As per @cstrobbe comment https://github.com/w3c/wcag/issues/2688#issuecomment-1262199960, I also think an updated to the Understanding document makes sense. 

> For this reason, the alternative wording "the same order relative to other page content" makes sense, but it wouldn't be bad if [Understanding SC 3.2.6](https://w3c.github.io/wcag/understanding/consistent-help.html) pointed out the difference in what "relative order" applies to in SC 3.2.6 and 3.2.3.

@dtex Are you comfortable both philosophically and technically with making the change to your PR?

dtex:
@mbgower I'm happy to change my PR. 

There were a couple of other spots where the "same relative order" defined term was being used, so I've also adjusted those. Apart from that, I think the Understanding document was already doing an excellent job of covering the intent, and SC 3.2.6 already covers it pretty well in the second note.

I also noticed 3.2.6 in the Understanding document did not have the latest edits, so I brought that up to date as well.

I have force pushed these updates to my clone so that no footprints are left on the "Same Relative Order" definition file.

mbgower:
Looks better, thanks.
I put one of your changes back (line 50) because it _was_ talking about help appearing in the same "relative order" on different pages. So that is a different meaning than talking about the same order relative to other content on a single page.

Hope that makes sense.

mbgower (approved):
I made minor editorial changes.

dtex:
> I put one of your changes back

This is back to the defined term. Should we add a link?

bruce-usab (commented):
Sorry I have lost track if we want "same relative order" (and linking to term in glossary) or not use those three words in that exact order?

bruce-usab (approved):
SC looks correct (i.e, not using the defined glossary term).  Please double-check that Understanding is consistent with that.

mbgower:
> This is back to the defined term. Should we add a link?

No, do not put the link in. You do not need to link to a defined term unless you're using it in that sense -- it's normally reserved for normative uses, as far as I know. Not for non-normative Understanding documents.

awkawk:
Whatever you decide on, please make sure to add the "erratumraised" label to the PR

bruce-usab:
> You do not need to link to a defined term unless you're using it in that sense...

We have had more than a few on-going discussions around using glossary terms and not linking when we should/could/might have.

IMHO it is better to avoid the ambiguity.

> Not for non-normative Understanding documents.

I agree that is much less of a problem in Understanding.  But there is still a risk in Understanding for someone later to "correct a typo" thinking the lack of a link to be a minor oversight.

alastc:
@awkawk Does it need an erratumraised on a not-yet-recommended doc?

Issue #2740, opened 2022-10-19
WCAG 2.2 Understanding Success Criterion 2.5.8: Target Size (Minimum)


maryjom:
Intent section:
In the paragraph right above the image of the examples, the last two sentences have "the the", should be "the".

Examples section:
The text that exists now:

> A row of buttons, each of which has a horizontal width of more than 24 CSS pixels, but a height of only 20 CSS pixels. Since there is sufficient spacing above and below the row of buttons, the Success Criterion passes.

Suggested text to clarify what the passing situation is before saying it passes.
>A row of buttons, each of which has a horizontal width of more than 24 CSS pixels, a height of only 20 CSS pixels, and vertical padding of 2 CSS pixels above and 2 CSS pixels below the row of buttons. Since there is sufficient spacing above and below the row of buttons, the Success Criterion passes.

alastc:
Merged #2743 so closing.

PR #2743, opened 2022-10-20
Update target-size-minimum.html

mbgower:
minor Understanding doc editorial changes for typos and improvement of image description as per https://github.com/w3c/wcag/issues/2740


understanding/22/target-size-minimum.html
```diff
@@ -146,7 +146,7 @@ <h2>Examples of Target Size (Minimum)</h2>
 
         <ul>
             <li>Three buttons are on-screen and the target area of each button is 24 by 24 CSS pixels. Since the target size itself is 24 by 24 CSS pixels, no additional spacing is required, the Success Criterion passes.</li>
-            <li>A row of buttons, each of which has a horizontal width of more than 24 CSS pixels, but a height of only 20 CSS pixels. Since there is sufficient spacing above and below the row of buttons, the Success Criterion passes.</li>
+            <li>A row of buttons, each of which has a horizontal width of more than 24 CSS pixels, a height of only 20 CSS pixels, and vertical padding of 2 CSS pixels both above and below the row of buttons. Since there is sufficient spacing above and below the row of buttons, the Success Criterion passes.</li>
```
awkawk, line 149: @mbgower now that I think about this - would this pass? If there was a button above this row of buttons there would be only 22 CSS pixels from the bottom edge of the in-row button to the bottom of the button above it. If this indicated 4 CSS pixels both above and below then it would unambiguously be ok.

detlevhfischer: @awkawk @mbgower This may still be from a time when offset was measured in another way. I guess the vertical height of the buttons would now have to be 22 so that 22 + 2 = 24.

detlevhfischer (changes requested):

understanding/22/target-size-minimum.html
awkawk: Right, so we can say that the buttons are 22 with 2 padding on top and bottom or that the buttons are 20 with 4 padding.

alastc:
Editorial, I'll merge...

PR #2746, opened 2022-10-20
Focus not obscured margin technique and working example

fstrr:
New technique and working example for Focus Not Obscured that will work for "minimum" and "enhanced" criteria.

[Preview](https://raw.githack.com/w3c/wcag/focus-not-obscured-margin-technique/techniques/css/C43.html)


detlevhfischer:
For convenience, here is a [GitHack preview of the proposed Technique "Using CSS margin and scroll-margin to un-obscure content" (C43)](https://raw.githack.com/w3c/wcag/d1cc05a6caef5bf1d64f19b32091db399facee76/techniques/css/C43.html)

alastc:
Notes from the friday WCAG 2.x meeting:

- Unfortunately making the cookie banner sticky at higher zoom levels makes it unusable, we need to revert that change.
- We should have a message above saying that it is a demonstration of scroll-padding when at larger sizes, and it passes by being not-sticky at smaller sizes.

patrickhlauke:
Btw, randomly, for an example in the wild ... I recently fixed this on the Bootstrap documentation site https://github.com/twbs/bootstrap/pull/37926

PR #2752, opened 2022-10-25
add "entirely" to example and future technique

### LINKED ISSUE #2423: Add "entirely" to Example 2.4.12: Focus Not obscured
https://w3c.github.io/wcag/understanding/focus-not-obscured.html#examples

"entirely" is missing from the text but important for the SC:

Change:

> Examples
> A page has a sticky footer (attached to the bottom of the vieport). When tabbing down the page the focused item is not hidden by the footer.

To:

> Examples
> A page has a sticky footer (attached to the bottom of the vieport). When tabbing down the page the focused item is not entirely hidden by the footer.

fstrr:
closes #2423

### DIFF
--- a/understanding/22/focus-not-obscured-minimum.html
+++ b/understanding/22/focus-not-obscured-minimum.html
@@ -46,7 +46,7 @@
       <h2>Examples of Focus  Not Obscured</h2>
 
       <ul>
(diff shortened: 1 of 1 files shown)


fstrr:
This has been implemented elsewhere, so isn't needed now.

PR #2753, opened 2022-10-26
2.5.7 Understanding ´Dragging Movements  - adding example

### LINKED ISSUE #2550: WCAG 2.2 SC 2.5.7 Understanding Document Example Could Be Replaced with an Example that Encourages Greater Equity
The Understand Document says: "An example is a color wheel where a color can be changed by dragging an indicator. In addition, text fields for the numerical input of color values allow the definition of a color without requiring dragging movements."

Numerical input of a color value requires more knowledge, more work, and most likely a separate reference document compared with dragging an indicator.  Meanwhile, a single pointer click equivalent of the functionality described is easy to imagine and implement, making the extra user effort unjustifiable. 

Perhaps instead:
"An example is a widget where you can drag a gift to one person in a photo of a group of people . In addition to the photo widget, there is a text field / menu control that allows users to give the gift to a person by choosing their photo and name from the menu or by typing their name, without requiring dragging movements."

detlevhfischer:
* Added another example (dragging gift to person provides menu alternative) suggested in #2550 
* The color wheel example was changed earlier to allow single click/tap selection of color

Closes #2550


alastc:
Approved: https://www.w3.org/2023/02/28-ag-minutes#t09

PR #2770, opened 2022-11-09
Adding 'until up event'

### LINKED ISSUE #2669: Dragging movement definition might be improved
The [definition for dragging movement](https://www.w3.org/TR/WCAG22/#dfn-dragging-movements) is defined as the following in the CR version of WCAG 2.2:

> an operation where the pointer engages with an element on the down event and the element (or a representation of its position) follows the pointer
> 
> NOTE
> Examples of draggable elements include list items, text elements, and images.

Now per that definition, clicking an element which follows the pointer is a “dragging movement”, even if an _up_ event occurs during the movement, if the element keeps following the pointer. I don’t think that this is what is meant with dragging movement and the Understanding is backing my understanding up with this explanation:

> When an interface implements functionality that uses dragging movements, users perform four discrete actions:
>
> 1. tap or click to establish a starting point, then
> 2. press and hold that contact while...
> 3. performing a repositioning of the pointer, before...
> 4. releasing the pointer at the end point.
> Not all users can accurately press and hold that contact while also repositioning the pointer. An alternative method must be provided so that users with mobility impairments who use a pointer (mouse, pen, or touch contact) can use the functionality.

I would therefore suggest the following rewrite of the definition for dragging movement:

> an operation where the pointer engages with an element on the down event and the element (or a representation of its position) follows the pointer, and the engagement ends on a subsequent up event

(I’m also not super sure what the note does here, everything that can be draggable is a draggable element, right? So I don’t find it helpful and would suggest to remove it.)

I’m not sure if that is the right language for this, but I’d like to make sure that this WCAG SC does not accidentally cover discrete two step processes.

alastc:
Closes #2669

### DIFF
--- a/guidelines/terms/22/dragging-movement.html
+++ b/guidelines/terms/22/dragging-movement.html
@@ -2,7 +2,7 @@
 <dd class="new">
 	<p class="change">New</p>
    					
-   <p>an operation where the pointer engages with an element on the down event and the element (or a representation of its position) follows the pointer</p>
+   <p>an operation where the pointer engages with an element on the down event and the element (or a representation of its position) follows the pointer until an up event</p>
    
    <p class="note">Examples of draggable elements include list items, text elements, and images.</p>
 </dd>


alastc:
Approved here: https://www.w3.org/2022/11/15-ag-minutes.html#t04

PR #2772, opened 2022-11-09
add mention of inline text links

### LINKED ISSUE #2679: 2.4.11 text-decoration as focus indication for links
The new requirement for focus indication fundamentally changes the focus style for links. For many years, links were styled with 
```
a {text-decoration:none}; 
a:hover, a:focus {text-decoration:underline}
```
This new requirement significantly improves the readability of focused links and I'm not arguing about it's purpose. However, this is going to have a long-tail impact on websites that have used this pattern. It's hard to quantify the impact, as this pattern has been so ingrained in web development styles. 

This change also affect web sites that are using an underline on links to make them more obvious and then removing the underline on hover and focus. Again, this change makes the link interaction more understandable. But it's going to impact an enormous number of web sites. 

```
a {text-decoration:underline}; 
a:hover, a:focus {text-decoration:none}
```

I have done a quick audit of our company web sites and this will require updates to footers and some inline links on most of our pages. I feel we are in better shape, as we already have defined an outline style for the majority of our links, but I appreciate how this will impact other creators significantly.

Ted Drake, Accessibility and Inclusive Design Leader (Intuit)

bruce-usab:
Updates Understanding for 2.4.11 Focus Appearance to mention that default inline text link underlines meets the exception:
> The focus indicator is determined by the user agent and cannot be adjusted by the author

Closes #2679

### DIFF
--- a/understanding/22/focus-appearance.html
+++ b/understanding/22/focus-appearance.html
@@ -310,7 +310,7 @@
                     <figcaption> Passes: Although the CSS <code>border</code> on a multi-line link does not enclose the link, the minimum bounding box definition allows it to be assessed as if it was on a single line. As well, the indicator's area is invariably more than four times the shortest-side calculation .</figcaption>
                 </figure>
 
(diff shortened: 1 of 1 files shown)


bruce-usab:
Closes #2679

bruce-usab (commented):
Looks good to me.

PR #2777, opened 2022-11-10
Update for open/closing interfaces

### LINKED ISSUE #2751: 2.4.12 Focus not Obscured (Minimum) and user opened / controlled content
While implementing WCAG 2.2’s 2.4.12 Focus not Obscured (Minimum), we encountered a scenario we’d like the Understanding document to provide clarification on: The applicability of 2.4.12 where content that can be opened / closed by the user impacts the visibility of other focusable content.

Regarding movable content, the Understanding document clarifies: “If the interface is configurable so that the user can move toolbars and non-modal dialogs around, then only the initial positions of user-movable content would be considered for testing and conformance of this success criterion.”

That doesn’t seem to directly address content where opening / closing of overlay content can be controlled by the user, that is: where there is no failure of 2.4.12 in the default state of the page, but when the user opens overlay content, focusable elements could end up entirely behind the content while the content is open.

More detailed examples follow.

One impactful and wide-spread example is a chat interface that is collapsed into a button in its default state. In that state, there isn’t a 2.4.12 failure. However, when the user opens the chat interface, it entirely covers some focusable content. For this scenario let’s also say that any content behind the chat can be accessed by collapsing the chat to access it again - without losing any of the chat activity.

Closely related are the messaging popup panels in the lower right corner of sites like Facebook and LinkedIn.

Another example is popup content that is actively opened / closed by the user on mouse click or Enter (as opposed to opening on hover or focus), sometimes referred to as a “toggle tip”. Content behind the “toggle tip” could contain focusable elements that are entirely covered. Toggle tips aren’t (usually) modal, so the user could tab to elements entirely behind the toggle tip overlay if they don’t choose to close it.

Those are just a few examples that come to mind.

Clarification along the lines of the one given for movable content would be very welcomed. 

Thank you.

alastc:
Closes #2751


bruce-usab:
I will take a run at integrating additional comments from 1/27 backlog call.

understanding/22/focus-not-obscured-minimum.html
```diff
@@ -36,13 +36,18 @@ <h2>Intent of Focus Not Obscured</h2>
...
 
+        <p>If the interface is configurable so that the user can manually open, close, and/or move components such as toolbars and non-modal dialogues, then only the initial default position and state of user-movable content is applicable for testing and conformance of this success criterion.  Similarly, if the interface contains content where user interation causes eleements to open/close or expand/collapse (i.e., a change is initiated by the user), only the initial default state of user-controlable components is in scope for this success criterion.</p>
+
+        <p>Converserely, if a component automatically opens and obscures focused elements, that behavior would fail this criterion.  For example, if a navigation drop-down opens upon focus, but does not close when focus moves to other element, the page would not pass if focusable elements were visually obscured.</p>
```
mbgower, line 43: IMO, a dropdown that opens on focus is verging on a failure of On Focus (not enough context to know) and is not a behaviour we should be advancing. So I don't think this serves as a good example.

understanding/22/focus-not-obscured-minimum.html
```diff
@@ -36,13 +36,18 @@ <h2>Intent of Focus Not Obscured</h2>
...
+
+        <p>Selecting or activating a component (for example, by mouse click or return key) is not in scope for this SC, since that woould be a change initiated by the user, and is distict from when the keyboard focus indicator is visible.</p>
```
mbgower, line 45: I don't understand what this is doing here, or what it's trying to 'cut off' as a misconception.

understanding/22/focus-not-obscured-minimum.html
```diff
@@ -91,7 +96,7 @@ <h3>Additional Techniques (Advisory)</h3>
          <h3>Failures</h3>
          <ol>
             <li>
-
+              It is essentially a bug (and a failure against this SC) if a component (for example, a hamburger menu) stays open on keyboard focus after keyboard focus is removed &ndash; when mouse-over behavior works as is typically expected (that is, the component automatically expands, but then collapses after the mouse cursor moves away from the component).
```
mbgower, line 99: I'm not sure why we're suddenly talking about mouse operation, and I am not agreeing that the 'expected' behaviour of mouse interaction is for menus to automatically expand on hover, if that is what is being inferred.

mbgower (changes requested):

understanding/22/focus-not-obscured-minimum.html
bruce-usab: I agree that we do not want to imply automatic expand on hover is expected behavior.

I also agree it would be better to avoid "mouse operation".

understanding/22/focus-not-obscured-minimum.html
```diff
@@ -36,13 +36,18 @@ <h2>Intent of Focus Not Obscured</h2>
...
 
-      <p class="note">If the interface is configurable so that the user can move toolbars and non-modal dialogs around, then only the initial positions of user-movable content would be considered for testing and conformance of this Success Criterion.</p>
+        <p>Where other content can overlap with a focused item, the focused element should not be hidden.  Typical types of content that can overlap focused items are sticky footers, sticky headers, or non-modal dialogues.  As a user tabs through the page (thus moving the keyboard focus indicator), these layers of content can obscure the focused item, including the focus indicator.  Such behavior is usually indiccative of a failure to satisfy this success criterion.</p>
```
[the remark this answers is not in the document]
bruce-usab: I will take another pass to remove redundancy.

mbgower (changes requested):
> It is essentially a bug (and a failure against this SC) if a component (for example, a hamburger menu) stays open on keyboard focus after keyboard focus is removed &ndash; when mouse-over behavior works as is typically expected (that is, the component automatically expands, but then collapses after the mouse cursor moves away from the component).

I'm not sure this entirely resolves it. I think the notion of equivalent experience/interaction between mouse and keyboard you're working towards can be useful. However, it still seems a little incomplete. The issue isn't only when it opens on keyboard focus; it's also when keyboard activation opens it, yet loss of keyboard focus does not collapse it, even though while a mouse user can click to open, when they click elsewhere, it collapses.

PR #2778, opened 2022-11-10
Separating off part to a note

### LINKED ISSUE #2427: Change definition and add parts to NOTE for "target offset"
Target offset now reads:
https://w3c.github.io/wcag/guidelines/22/#dfn-target-offsets

(Also change in Understanding! https://w3c.github.io/wcag/understanding/target-size-minimum.html#dfn-target-offset)

> the distance measured from the farthest point of a target to the nearest point of the second target. Offset includes the target and spacing around the target. The target offset from A to B may be different then the offset from B to A, if the size of these targets differ.

Now as the definition text can be replaced in a SC by the definition itself it should be changed to:

> distance measured from the farthest point of a target to the nearest point of the second target.
> 
> NOTE
> Offset includes the target and spacing around the target.
> 
> NOTE
> The target offset from A to B may be different than the offset from B to A, if the size of these targets differ.

If changed the SC reads as (changed definition to text of definition):

> Spacing: The distance measured from the farthest point of a target to the nearest point of the second target is at least 24 CSS pixels to every adjacent target;

alastc:
Closes #2427

### DIFF
--- a/guidelines/terms/22/target-offset.html
+++ b/guidelines/terms/22/target-offset.html
@@ -1,5 +1,6 @@
 <dt class="new"><dfn data-lt="target offsets">target offset</dfn></dt>
 <dd class="new">
 	<p class="change">New</p>	
-   <p>the distance measured from the farthest point of a target (A), to the nearest point of the second target (B). The offset includes the target and spacing around the target. The offset from A to B may be different than the offset from B to A, if the sizes of these targets differ.</p>
+   <p>the distance measured from the farthest point of a target (A), to the nearest point of the second target (B). The offset includes the target and spacing around the target.</p>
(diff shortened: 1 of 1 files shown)


patrickhlauke:
just checking: are you keeping merged branches around on purpose? if not, suggest deleting them as soon as something's merged, to avoid stale stuff piling up in this already crowded repository...

alastc:
Generally not, but I'm merging things during the meeting so quite often have to move on before I can delete it.

One exception in the WCAG22-typos branch, which in ongoing.

patrickhlauke:
no worries. picture me in a french maid outfit flitting around the repo (or, maybe, don't)

PR #2781, opened 2022-11-10
Updating XML to HTML source

alastc:
@michael-n-cooper I trying to fix the build error, the one at the moment is the lack of the relative luminance xml file, which is renamed to HTML.

So I've updated most instances of that filename... is that ok?


alastc:
Based on the conversation I reverted the xml > html change as @michael-n-cooper wanted to do that at the template level.

PR #2793, opened 2022-11-15
Issue2525 parsing syntactical

alastc:


JAWS-test:
@alastc  Was there a decision on this PR? (In connection with the discussion about https://github.com/w3c/wcag/pull/2797)

dd8:
I think this is problematic because the HTML Living Standard includes the content model in the HTML syntax. See
https://github.com/w3c/wcag/issues/2525#issuecomment-1330514991

apond1:
Bggh

Arkar202:
https://gva88.com/register?token=b25ef3c2ffa99c2

alastc:
The normative part is defunk now, just leaving open in case we want to incorporate this into understanding page?

patrickhlauke:
@alastc we need to decide if we want to pursue or drop this

patrickhlauke:
Discussed at the facilitator meeting today with @alastc @kfranqueiro @bruce-usab - agreed to close this as obsolete.

PR #2794, opened 2022-11-15
revise understanding doc to reference web components

### LINKED ISSUE #2495: 1.4.12 Text Spacing and scoped styles with shadow dom
I'm at a loss for how 1.4.12 Text Spacing is meant to align with instances of custom elements where their styles are encapsulated by the Shadow DOM, and other styles from the primary document are explicit **not supposed** to pierce the Shadow DOM to modify the styles of the internals of the custom element.  

For example: https://codepen.io/scottohara/pen/xxYWvGj

In that codepen I have a `<button>` defined in both the light and shadow dom.  The styles of the light DOM (where text spacing styles would be enabled by users) do not go into the Shadow DOM, and this is by design.

I can use the Stylus browser plugin and apply text spacing properties to that example, and the button in the light dom adjusts as expected.... and the button in the shadow dom doesn't adjust, which is as **I would expect**, but I'm not sure what the SC expects here.... and I can bet that users / auditors won't know what to expect here either.

Appreciate thoughts on what can be done here.

scottaohara:
closes #2495

this could potentially be expanded to further indicate that it's because content rendered in the Shadow DOM is purposefully encapsulated so that styles remain scoped to the web component and other external styles don't pass through the light tree into the shadow tree.


understanding/21/text-spacing.html
```diff
@@ -18,7 +18,7 @@ <h3 id="authorresp">Author Responsibility </h3>
...
+         <p>If the markup-based technologies being used are capable of overriding text to the Success Criterion's metrics, then this SC is applicable. For instance Cascading Style Sheet/HTML technologies generally allow for the specified spacing metrics. However, if discrete portions of a markup language do not support any of the listed text style properties, then that portion of the markup language is exempt. For example, web components are a part of HTML, but do not support modification of any of the listed text style properties if they have been implemented with a closed ShadowRoot. Similarly, such web components and other plugin technologies would need to have a built-in ability to modify styles to the specified metrics. Currently, this SC does not apply to PDF as it is not implemented using markup.</p>
```
danhendersonede, line 21: Super picky comment (I'm happy either way):

Shouldn't web components be a proper noun (Web Components), to differentiate itself from any component-like thing on the web?

mbgower: @danhendersonede I suspect we will need to capitalize the term to reduce confusion. I'll ensure it is discussed on the AGWG call.

alastc, line 21:
```suggestion
         <p>If the markup-based technologies being used are capable of overriding text to the Success Criterion's metrics, then this SC is applicable. For instance Cascading Style Sheet/HTML technologies generally allow for the specified spacing metrics. However, if discrete portions of a markup language do not support any of the listed text style properties, then it is not applicable to that portion of the content. For example, web components are a part of HTML, but do not support modification of any of the listed text style properties if they have been implemented with a closed ShadowRoot. Similarly, such web components and other plugin technologies would need to have a built-in ability to modify styles to the specified metrics. Currently, this SC does not apply to PDF as it is not implemented using markup.</p>
```

understanding/21/text-spacing.html
```diff
@@ -18,7 +18,7 @@ <h3 id="authorresp">Author Responsibility </h3>
...
+         <p>If the markup-based technologies being used are capable of overriding text to the Success Criterion's metrics, then this SC is applicable. For instance Cascading Style Sheet/HTML technologies generally allow for the specified spacing metrics. However, if discrete portions of a markup language do not support any of the listed text style properties, then it is not applicable to that portion of the content. For example, web components are a part of HTML, but do not support modification of any of the listed text style properties if they have been implemented with a closed ShadowRoot. Similarly, such web components and other plugin technologies would need to have a built-in ability to modify styles to the specified metrics. Currently, this SC does not apply to PDF as it is not implemented using markup.</p>
```
scottaohara, line 21:
```suggestion
         <p>If the markup-based technologies being used are capable of overriding text to the Success Criterion's metrics, then this SC is applicable. For instance Cascading Style Sheet/HTML technologies generally allow for the specified spacing metrics. However, if discrete portions of a markup language do not support any of the listed text style properties, then it is not applicable to that portion of the content. For example, Web Components are a part of HTML, but do not support modification of any of the listed text style properties if they have been implemented with a closed ShadowRoot. Such Web Components and other plugin technologies would need to have a built-in ability to modify styles to the specified metrics. Currently, this SC does not apply to PDF as it is not implemented using markup.</p>
```

understanding/21/text-spacing.html
```diff
@@ -18,7 +18,7 @@ <h3 id="authorresp">Author Responsibility </h3>
...
+         <p>If the markup-based technologies being used are capable of overriding text to the Success Criterion's metrics, then this SC is applicable. For instance, Cascading Style Sheet/HTML technologies generally allow for the specified spacing metrics. However, if discrete portions of a markup language do not support any of the listed text style properties, then it is not applicable to that portion of the content. For example, Web Components are a part of HTML, but do not support modification of any of the listed text style properties if they have been implemented with a closed ShadowRoot. Such Web Components and other plugin technologies would need to have a built-in ability to modify styles to the specified metrics. Currently, this SC does not apply to PDF as it is not implemented using markup.</p>
```
scottaohara, line 21: per the wg call today, this note already states that the closed shadowRoot web components would need to have a built-in ability to modify styles.  So I (incorrectly) thought that would have addressed the concerns that people were raising, since it's still staying that there would need to be a built-in way to adjust?

would appreciate feedback/suggestions on what could be clarified here - or is it that the note needs to more overtly state that if an author can avoid closed shadowroot, then they are recommended to go that route. And that maybe there needs to be more overt call outs to certain HTML features (provided by browsers) that do use this and cannot be changed by authors, and those should arguably be ok?  e.g., the date picker exapmle?

scottaohara:
didn't mean to close this.  going to try and come back to this since it's been lingering for awhile now.

PR #2796, opened 2022-11-16
Converting XML to HTML

alastc:
This is blocking the build: Updating file references to the relative luminance file. Hopefully straightforward and correct.

I left one reference in WCAG2ICT for http://www.w3.org/WAI/GL/WCAG20/relative-luminance.xml, assuming that's still correct for it's time.


alastc:
Damnit, I already did this last week!

PR #2797, opened 2022-11-16
Removing 4.1.1

alastc:
Cross-reference: https://github.com/w3c/wcag/issues/2525#issuecomment-1316949146

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/pull/2797.html" title="Last updated on Nov 16, 2022, 2:16 PM UTC (2b6caec)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/2797/b991776...2b6caec.html" title="Last updated on Nov 16, 2022, 2:16 PM UTC (2b6caec)">Diff</a>

### DIFF
--- a/guidelines/index.html
+++ b/guidelines/index.html
@@ -378,8 +378,6 @@
             <section class="guideline">
                 <h3>Compatible</h3>
                 <p>Maximize compatibility with current and future user agents, including assistive technologies.</p>
-                
-                <p class="ednote">Success Criterion 4.1.1 Parsing has been removed from the guidelines.</p>
 
                 <section data-include="sc/20/parsing.html" data-include-replace="true"></section>
 
--- a/guidelines/index.html
+++ b/guidelines/index.html
@@ -378,6 +378,8 @@
             <section class="guideline">
                 <h3>Compatible</h3>
                 <p>Maximize compatibility with current and future user agents, including assistive technologies.</p>
+                
+                <p class="ednote">Success Criterion 4.1.1 Parsing has been removed from the guidelines.</p>
 
                 <section data-include="sc/20/parsing.html" data-include-replace="true"></section>
 
--- a/guidelines/sc/20/parsing.html
+++ b/guidelines/sc/20/parsing.html
@@ -1,18 +1,8 @@
 <section class="sc">
    
    <h4>Parsing</h4>
+
    
-   <p class="conformance-level">A</p>
-   
-   <p>In content implemented using markup languages, elements have complete start and end
-      tags, elements are nested according to their specifications, elements do not contain
-      duplicate attributes, and any IDs are unique, except where the specifications allow
-      these features.
-   </p>
-   
-   <p class="note">Start and end tags that are missing a critical character in their formation, such
-      as a closing angle bracket or a mismatched attribute value quotation mark are not
-      complete.
-   </p>
+   <p>This criterion has been removed.</p>
    
 </section>
--- a/script/wcag.js
+++ b/script/wcag.js
@@ -81,24 +81,10 @@ function termTitles() {
 	});	
 }
 
-// clear removed sections after respec has run to preserve generated numbering
-function clearRemoved() {
-  const removeIds = ["parsing"];
-  removeIds.forEach(function(currentValue) {
-    var selector = ".tocline > a[href=\"#" + currentValue + "\"]";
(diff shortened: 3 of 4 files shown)


patrickhlauke:
just noting that this may break backwards-compatibility, or rather: it's possible for a site to be conformant to 2.2 but non-conformant to 2.1/2.0.

alastc:
Yep, and we did discuss that on the call. Partly these things should be caught by other SCs (primarily 1.3.1, 4.1.2), partly allowed by the exception "where allowed by the spec", and partly that it creates so much noise people are ignoring it anyway.

Also, a couple of people volunteered to do a mapping exercise of things 4.1.1 caught that should be caught by other SCs.

TestPartners:
People ignoring the SC doesn't sound like a valid reason for removing it. We have rarely had a client that wasn't able to meet this SC. It's actually one of the easier SCs to fix because all you need to do is write valid code. If people are not willing to do that, it's difficult to imagine them meeting all the other success criteria.

alastc:
> People ignoring the SC doesn't sound like a valid reason for removing it.

Unless they are ignoring it for good reason.

> We have rarely had a client that wasn't able to meet this SC. It's actually one of the easier SCs to fix because all you need to do is write valid code.

That's almost the opposite of my experience. With the exception of a publisher (who was also exporting to ePub), it is one of the most time consuming SCs to test and fix. Partly because you have to wade through so many results and filter down to what you actually need to fix. And then doing so rarely had a perceptible effect _unless_ it was also causing issues elsewhere (e.g. for 4.1.2).

I'd much rather spend time testing (or teaching people how to test) 1.3.1 and 4.1.2 efficiently. One of those methods can be validation, but the signal to noise ratio is not good in most cases I've come across.

alastc:
Sorry, I didn't realised you'd missed all the context.

Most of it is in #2525, but see also the [original reason for the SC](https://github.com/w3c/wcag/issues/770#issuecomment-1318749819), which no longer exists.

Also, the discussion in #2525 highlighted that most people (including AG members) have been interpreting it more widely than intended, failing more things than was intended. (I.e. the 'content model' was not intended to be covered.)

Added to the fact that the "original rationale" is no longer relevant, we've had a lot of people saying that, particularly at larger scales, it creates more "busy work" than work which actually improves accessibility for people.

My personal experience has been that the useful parts are caught under 1.3.1 and 4.1.2, and the effort in 'validating' webapps based in React/Angular etc vs testing the other SCs is not worth it.

So those have been the "good reasons" for ignoring it.

bruce-usab:
> it's possible for a site to be conformant to 2.2 but non-conformant to 2.1/2.0.

Yes, but only for a site that is non-conformant to 2.0/2.1 because of _only_ failures against SC 4.1.1 which don't map to other SC (already in 2.0/2.1) -- such as duplicated-but-unused IDs.

patrickhlauke:
> Yes, but only for a site that is non-conformant to 2.0/2.1 because of only failures against SC 4.1.1 which don't map to other SC (already in 2.0/2.1) -- such as duplicated-but-unused IDs.

sure, but that's still breaking backwards compatibility. so IF 4.1.1 gets removed (which, don't get me wrong, I'm all in favour of), this nuance should be called out in the intro for the spec/wherever it talks about being backwards-compatible

bruce-usab:
> this nuance should be called out in the intro for the spec/wherever it talks about being backwards-compatible

Agreed, and that was part of the conversation on the Tuesday call, but we tabled that part of the conversation.  There will be some details to work out!

alastc:
> sure, but that's still breaking backwards compatibility.

We also need to add a little nuance, the requirement it was addressing is backwards compatible, but there can be differences in the results of an audit.

joe-watkins:
@alastc Should this have been removed in `2.1` as well? 

https://www.w3.org/WAI/WCAG21/Understanding/parsing.html

Screenshot of 4.1.1 SC

alastc:
Arg. Yes. It's a template thing, I'll have to check with Michael.

joe-watkins:
@alastc Appears to be fixed! `4.1.1 Parsing` is back in WCAG 2.1 - shewey!

https://www.w3.org/WAI/WCAG21/Understanding/parsing.html

Screenshot of 4.1.1 Parsing Success Criterion page

PR #2798, opened 2022-11-16
Target offset definition - Wilco and Bruce's suggestions

### LINKED ISSUE #2695: Difficulties with inconsistency of Target Size (Minimum)
I've been working on an automated implementation of the Target Size (minimum) criterion in axe-core. (It's available via `npm install axe-core@latest` if anyone wants to try it.) It's working reasonably well, but I keep running into scenarios where something that looks like it should fail doesn't. Take the following list of links - things like this happen a lot in table of contents:

- [Link A](#)
    - [Nested link B with long text](#)
    - [Link C](#)

The links are 17 pixels tall, with 4 pixels whitespace between them. That would suggest they all fail. That's overlooking the fact that offset can be a diagonal. In this scenario, link C is actually the only link that fails the success criterion. The other two links have their offset measured on a diagonal, which results in them passing.

- The offset of Link A is the distance between its top-left corner, and the top-left corner of link B. That's 21 down, and 28 to the right. That's a diagonal distance of 31px, so it passes.
- The offset of link B to link A and C is much further still. It's right corners are over 100px away from link A and B, so pass
- Only the offset of link C is 21 pixels. Its furthest point is somewhere on the underline, and from their the closest point to link B is straight up.

I've seen this happen in a number of places. ToC, navigations, I've seen it in inline links lists too (Wikipedia has lots of these at the bottom of its pages). Thinking about it in terms of "is there a spot on the link I could try to hit without accidentally hitting the wrong thing", this approach does make sense. Link A sticks out on the left, link B sticks out on the right. That works, and maybe this is just some artefact of WCAG 2.2 that people are going to need to get used to.

On the other hand, if someone wants to click link B, should they just assume that link A and C inline links, so that their touch area doesn't extend to the right? I'm not convinced that this is right. The other thing I'm concerned about is if testers start doing this very precisely, the people getting test reports are not going to understand why in a link of lists that are equally spaced, some links pass while others fail.

I'm not sure what to suggest on this one, but I wanted to share my experience with implementing this SC into a tool. I'm finding issues with this all over the place. It's a real challenge, and it's giving me serious doubts about if we got the math right on this.

alastc:
Closes #2695

### DIFF
--- a/guidelines/sc/22/target-size-minimum.html
+++ b/guidelines/sc/22/target-size-minimum.html
@@ -7,7 +7,7 @@
     
     <p>The size of the <a>target</a> for <a>pointer inputs</a> is at least 24 by 24 <a>CSS pixels</a>, except where:</p>
     <ul>
-        <li><strong>Spacing:</strong> The <a>target offset</a> is at least 24 CSS pixels to every adjacent target;</li>
+        <li><strong>Spacing:</strong> The target is non-overlapping and has a <a>target offset</a> is at least 24 CSS pixels to every adjacent target;</li>
         <li><strong>Equivalent:</strong> The function can be achieved through a different control on the same page that meets this criterion.</li>
         <li><strong>Inline:</strong> The target is in a sentence or block of text;</li>
         <li><strong>User agent control:</strong> The size of the target is determined by the user agent and is not modified by the author;</li>
--- a/guidelines/sc/22/target-size-minimum.html
+++ b/guidelines/sc/22/target-size-minimum.html
@@ -7,7 +7,7 @@
     
     <p>The size of the <a>target</a> for <a>pointer inputs</a> is at least 24 by 24 <a>CSS pixels</a>, except where:</p>
     <ul>
-        <li><strong>Spacing:</strong> The target is non-overlapping and has a <a>target offset</a> of at least 24 CSS pixels to every adjacent target;</li>
+        <li><strong>Spacing:</strong> The target does not overlap any other target and has a <a>target offset</a> of at least 24 CSS pixels to every adjacent target;</li>
         <li><strong>Equivalent:</strong> The function can be achieved through a different control on the same page that meets this criterion.</li>
         <li><strong>Inline:</strong> The target is in a sentence or block of text;</li>
         <li><strong>User agent control:</strong> The size of the target is determined by the user agent and is not modified by the author;</li>
--- a/guidelines/sc/22/target-size-minimum.html
+++ b/guidelines/sc/22/target-size-minimum.html
@@ -7,7 +7,7 @@
     
     <p>The size of the <a>target</a> for <a>pointer inputs</a> is at least 24 by 24 <a>CSS pixels</a>, except where:</p>
     <ul>
-        <li><strong>Spacing:</strong> The target is non-overlapping and has a <a>target offset</a> is at least 24 CSS pixels to every adjacent target;</li>
+        <li><strong>Spacing:</strong> The target is non-overlapping and has a <a>target offset</a> of at least 24 CSS pixels to every adjacent target;</li>
         <li><strong>Equivalent:</strong> The function can be achieved through a different control on the same page that meets this criterion.</li>
         <li><strong>Inline:</strong> The target is in a sentence or block of text;</li>
         <li><strong>User agent control:</strong> The size of the target is determined by the user agent and is not modified by the author;</li>
--- a/guidelines/terms/22/target-offset.html
+++ b/guidelines/terms/22/target-offset.html
@@ -1,6 +1,11 @@
 <dt class="new"><dfn data-lt="target offsets">target offset</dfn></dt>
 <dd class="new">
 	<p class="change">New</p>	
-   <p>the distance measured from the farthest point of a target (A), to the nearest point of the second target (B). The offset includes the target and spacing around the target.</p>
+   <p>size of a target (A) plus the spacing to a second target (B).</p>
    <p class="note">The offset from A to B may be different than the offset from B to A, if the sizes of these targets differ.</p>
+   <div class="note">
+      <p>For horizontally aligned targets, target offset is the width of target A plus the horizontal space to the closest edge of target B. For vertically aligned targets, target offset is the the height of target A plus the vertical space to the closest edge of target B. For targets that are neither horizontally nor vertically aligned, the target offset is the longest measure from any point of A to the nearest point of B.</p>
+
+      <p>Two targets are horizontally aligned if a horizontal line can be drawn that goes through both targets, but no vertical line can be drawn that goes through both targets. Two targets are vertically aligned if a vertical line can be drawn that goes through both targets, but no horizontal line can be drawn that goes through both targets.</p>
+   </div>
 </dd>
--- a/guidelines/terms/22/target-offset.html
+++ b/guidelines/terms/22/target-offset.html
@@ -1,11 +1,7 @@
 <dt class="new"><dfn data-lt="target offsets">target offset</dfn></dt>
 <dd class="new">
 	<p class="change">New</p>	
-   <p>size of a target (A) plus the spacing to a second target (B).</p>
-   <p class="note">The offset from A to B may be different than the offset from B to A, if the sizes of these targets differ.</p>
-   <div class="note">
-      <p>For horizontally aligned targets, target offset is the width of target A plus the horizontal space to the closest edge of target B. For vertically aligned targets, target offset is the the height of target A plus the vertical space to the closest edge of target B. For targets that are neither horizontally nor vertically aligned, the target offset is the longest measure from any point of A to the nearest point of B.</p>
-
-      <p>Two targets are horizontally aligned if a horizontal line can be drawn that goes through both targets, but no vertical line can be drawn that goes through both targets. Two targets are vertically aligned if a vertical line can be drawn that goes through both targets, but no horizontal line can be drawn that goes through both targets.</p>
-   </div>
+   <p>length of the longest possible line that starts at an edge of a target (A), intersects a second edge of A, and ends at the closest edge of a second target (B). For horizontally aligned targets, target offset is measured with a horizontal line. For vertically aligned targets, target offset is measured with a vertical line. For targets that are neither, target offset is measured diagonally.</p>
+   <p>Two targets are horizontally aligned if a horizontal line can be drawn that goes through both targets, but no vertical line can be drawn that goes through both targets. Two targets are vertically aligned if a vertical line can be drawn that goes through both targets, but no horizontal line can be drawn that goes through both targets.</p>
+   <p class="note">The target offset from A to B may be different than the target offset from B to A, if the sizes of these targets differ.</p>
 </dd>


guidelines/terms/22/target-offset.html
```diff
@@ -1,6 +1,11 @@
 <dt class="new"><dfn data-lt="target offsets">target offset</dfn></dt>
 <dd class="new">
 	<p class="change">New</p>	
-   <p>the distance measured from the farthest point of a target (A), to the nearest point of the second target (B). The offset includes the target and spacing around the target.</p>
+   <p>size of a target (A) plus the spacing to a second target (B).</p>
    <p class="note">The offset from A to B may be different than the offset from B to A, if the sizes of these targets differ.</p>
+   <div class="note">
+      <p>For horizontally aligned targets, target offset is the width of target A plus the horizontal space to the closest edge of target B. For vertically aligned targets, target offset is the the height of target A plus the vertical space to the closest edge of target B. For targets that are neither horizontally nor vertically aligned, the target offset is the longest measure from any point of A to the nearest point of B.</p>
```
dbjorge, line 7: "the closest edge of target B" feels imprecisely worded for targets that aren't rectangular. How should this be interpreted if B is a circle?

dbjorge, line 7: I think this part of the definition should not be in a note - per 5.1, notes aren't normative, but the information here is *required* to precisely interpret the currently-normative part of the proposed definition.

guidelines/terms/22/target-offset.html
```diff
@@ -1,6 +1,11 @@
...
+
+      <p>Two targets are horizontally aligned if a horizontal line can be drawn that goes through both targets, but no vertical line can be drawn that goes through both targets. Two targets are vertically aligned if a vertical line can be drawn that goes through both targets, but no horizontal line can be drawn that goes through both targets.</p>
```
dbjorge, line 9: This definition allows for some types of overlapping elements to be "aligned". I think this is okay, but only if the reader interprets the "closest edge of target B" as possibly being "the boundary of A and B where the overlap ends", ie, there is 0 space between them. If we go with this example, it would be good to add examples of overlapping elements which are aligned vs not to the understanding doc to clarify this.

guidelines/terms/22/target-offset.html
```diff
@@ -1,6 +1,11 @@
 <dt class="new"><dfn data-lt="target offsets">target offset</dfn></dt>
 <dd class="new">
 	<p class="change">New</p>	
-   <p>the distance measured from the farthest point of a target (A), to the nearest point of the second target (B). The offset includes the target and spacing around the target.</p>
+   <p>size of a target (A) plus the spacing to a second target (B).</p>
    <p class="note">The offset from A to B may be different than the offset from B to A, if the sizes of these targets differ.</p>
+   <div class="note">
```
dbjorge, line 6: I think all of the text in this `<div>` should be part of the definition, not a note - per 5.1, notes aren't normative, but the information here is *required* to precisely interpret the currently-normative part of the proposed definition.

guidelines/terms/22/target-offset.html
dbjorge: (it's not as simple as just "the nearer of the left and right edges of target B's bounding box" because that would do the wrong thing in the case of targets which are both aligned and overlapping)

GreggVan:
+1 
If it is explaining the definition  - OK as note
If it is an example   - OK as note
If it is informational - but not critical to the definition — OK as note

If it is required information in order for the definition to be accurate — it has to be part of the definition. 

gregg

———————————

guidelines/terms/22/target-offset.html
alastc: If B is a circle, I think the closest edge would be the closest point.

In fact, I guess we could just say the closest point and it would work for all cases?

If a target is overlapping, target offset is not needed and you check for 24 by 24px available in each target.

alastc:
Sorry, I managed to edit Gregg's comment rather than add one, I think I've put it back to how it was.

> If it is required information in order for the definition to be accurate — it has to be part of the definition. 

As Wilco noted, the same can be said for the relative luminance definition where the calculations are in the 1st note. 

I think the key for this is that you can understand the core intent and method from the base definition, and then use notes for details most people don't need to use. (E.g. the tool-makers need the exact calculations.) 

As someone that mostly does manual testing, the easiest way is to have a browser script that adds a 24px square around your cursor and outlines all controls, which is then pretty straightforward for most scenarios.

bruce-usab (approved):
Nice edit that left/right (and top/bottom ) can be removed from note about horizontally (and vertically) aligned.

Also +1 for changing "to edge of B" into "to point of B"

guidelines/terms/22/target-offset.html
dbjorge: "Closest point" is less ambiguous than "closest edge" but still gets us back to the same ambiguity around what to interpret it relative to that we've discussed ad nauseam in #2695. What point in (A) do you measure from (particularly if A is non-rectangular)? What is "closest" relative to?

For example, here are two side-by-side circular targets which are "horizontally aligned" per this note, but with the left one' center slightly higher than the right one's:

image

In this scenario, what is the "horizontal space to the closest point of target B"? Should a reader interpret this as "for every possible tuple of one point in A and one in B, find the difference in those two points' X values, and take the minimum of all those differences"? Or "for every possible tuple of one point in A and one in B, find the distance between those points; take the tuple with the minimum distance, and then use the difference in X values of those two points"? Or maybe one of the first two definitions but restricted to "rightmost point in A" and "leftmost point in B", rather than "any point"? I can imagine reasonable readers interpreting it in any of these ways, and they result in 3 different values in this example.

dbjorge: > If a target is overlapping, target offset is not needed and you check for 24 by 24px available in each target.

I think this is a reasonable intent, but it doesn't match the writing in the current proposal. As currently written in this proposal, nothing prevents target offset from applying to overlapping targets. Here is an example with a 20x24px target B overlaid on top of a 20x40px target A such that A underhangs B on both the left and right sides.

rendering of the 20x24px over 20x40px scenario described above

Per the current proposal, these are "horizontally aligned", and since B is 24px wide, it passes the Spacing exception regardless of how its spacing to A is calculated. I think as currently written it's ambiguous whether A passes the Spacing exception - a reasonable reader might conclude that A is "40px wide" and so passes, but I think the intended behavior is to fail such a case.

bruce-usab: With this two circles example, calculating target offset is not too bad:  diameter plus length of any a horizontal line between -- which is not as intuitive as diameter plus measure of closest distance.  But measuring 'size' of oddly shaped targets (needed when target are not aligned) seems fickle.

dbjorge: > 



> With this two circles example, calculating target offset is not too bad: diameter plus length of any a horizontal line between -- which is a larger value than, and not as intuitive as, diameter plus spacing.

"any a horizontal line" is not specific enough; there are many different-length horizontal lines that you could draw in this example. You get different values for offset if you choose to start from "the rightmost point of A" vs "the point of A closest to some point of B" vs "the point of A which minimizes the horizontal distance to the closest point of B". The normative wording of the spec needs to define the term unambiguously if it's going to be the basis of an SC's pass/fail conditions.

> With your 'overlapping' rectangles, I think you would have to treat that as three targets, no?

The definition of "target" in the current draft is "region of the display that will accept a pointer action, such as the interactive area of a user interface component". This definition does not require that it be a *contiguous* region, so I don't think you would *have to* treat that as three targets, though I think you could if you wanted to. Practically speaking, the image I gave is from a bit of HTML that involves 2 buttons, not 3 buttons, so I think it's likely that at least some readers would interpret it as 2 targets in practice. That said, I do think it's likely that different readers will interpret that differently; it might be nice for the target definition to include a note clarifying.
 
> But consider if A is 2 pixels taller, so then A and B are no longer horizontally aligned. Now what is the 'size' of A?
> 
> These examples did make me realize that measuring 'size' of oddly shaped targets (needed when target are not aligned) seems fickle.
> 
> If ellipses instead of circles, using 'diameter' is no good. Bounding box then?

Yes, I agree that the terms "width" and "size" are problematic for oddly shaped targets. I think that's actually problematic for the primary clause of the 2.5.8 too. Is a 40x40px square button with a different 20x20px button overlaid on top of it "at least 24 by 24 CSS pixels"?

40x40px square target A with 20x20px square target B overlaid in the middle

I think we obviously want target A in this example **not** to be considered to meet the 24x24 requirement, but I think different readers of the current wording of the SC will disagree about whether such a button meets the requirement.

GreggVan:
This is why standards are so messy and why provisions sometimes are unreadable and incomprehensible to most authors creating simple web pages…..

So many ifs, ands, or buts. 

When we are done - just think about trying to write it in plain language…   uffduh    my head hurts. 


gregg

———————————

guidelines/terms/22/target-offset.html
bruce-usab: Yes, presuming overlap of A to be thin, I agree it must not be considered to meet the Spacing exception.

With the present wording -- "size plus spacing" -- I think were are okay.  One has to allow a bad-faith assertion about the size of A (e.g., length of an outside side) rather than inside thickness.  Size measurement taken in good faith can be  covered by Understanding.

Several SC have loopholes allowing for terrible design or malicious compliance, but that's okay if balanced against phrasing which is simpler to understand.

With your circles example, the measure for "horizonal width" is obviously the diameter.  If B is up high enough, yes I can imagine someone naively extending the horizontal diameter line from A to the top (or near top) of B, and that would be bad.

Right now, (as I understand it) the current sentence under consideration is:

> For horizontally aligned targets, target offset is the width of target A plus the horizontal space to the closest point of target B.

How about:

> For horizontally aligned targets, target offset is the width of target A plus the smallest horizontal measurement between A and the closest point of target B.

Ping to @WilcoFiers

dbjorge: I've broken out the conversation about "size" and "24 x 24" being ambiguous to a separate issue (#2803), since they're not really about these changes.

That said, I think the term "width" here has the same problems that #2803 describes, where reasonable interpretations of it behave poorly for overlapped targets. I'm struggling to come up with a reasonably understandable equivalent that is not ambiguous. The closest I've come up with is

> For horizontally aligned targets, target offset is the length of the longest horizontal line with one endpoint touching a boundary of A, the other endpoint touching the bounding rectangle of B, and which never intersects any non-A target except at the endpoints.

...but this is pretty rough to understand in plain language and feels really questionable in terms of how far off it is from the original intent of the SC.

WilcoFiers, line 7: How about:

```suggestion
      <p>Offset is the length of the longest possible line that starts at an edge of A, intersects a second edge of A, and ends at an edge of B. For horizontally aligned targets, offset is measured with a horizontal line. For vertically aligned targets, offset is measured with a vertical line. For targets that are not aligned the offset is measured horizontal, vertical or diagonal (whichever is longest).</p>
```

alastc: I think "and ends at an edge of B" needs to be "and ends at the closest edge of B", otherwise you could extend it to the furthest edge of B.

Also, if a target is not aligned vertically or horizontally, that will have to be diagonal won't it? So the last bit could be "For targets that are not aligned the offset is measured diagonally".

GreggVan:
>Offset is the length of the longest possible line that starts at an edge of A, intersects a second edge of A, and ends at an edge of B. For horizontally aligned targets, offset is measured with a horizontal line. For vertically aligned targets, offset is measured with a vertical line. For targets that are not aligned the offset is measured horizontal, vertical or diagonal (whichever is longest).

Boy I think this is unnecessarily complicated and hard to figure out. Also if they are not aligned -  you end using the diagonal across A so B could be quite close, touch, or even overlap, and still pass where there would be a gap if they were side by side or one over the other.   Is that intended? 

Also where B is inside of A —  how do you draw a line that goes through two sides of A before it hits B?  (We talked about this being a case we needed to address.)

alastc:
> Boy I think this is unnecessarily complicated and hard to figure out.

Yes, but remember this is a note explaining the more straightforward initial sentence of the definition:
"size of a target (A) plus the spacing to a second target (B)."

And that is used within "The target offset is at least 24 CSS pixels to every adjacent target;"

This note is if you need to work through the more detailed workings.

> Also if they are not aligned - you end using the diagonal across A so B could be quite close, touch, or even overlap,

Anything overlapping has no space, so doesn't come under the spacing exception. (Well, technically you could check it, but you'd be looking for 24px size not including the overlap as there's no spacing.)

Also, anything overlapping will be horizontal and/or vertical.

Picture of a square with dotted lines emanating from each edge, away from the square. It shows the areas that count as vertical, horizontal, and diagonal.

> Also where B is inside of A — how do you draw a line that goes through two sides of A before it hits B? (We talked about this being a case we needed to address.)

I think that discussion was from a previous version where we leaned more heavily on the spacing exception?

There is another issue where we're looking at updating the understanding document to show how you check nested targets based on the first line of the SC: "The size of the target for pointer inputs is at least 24 by 24 CSS pixels,"

I.e. your looking for a 24px square within each target.

guidelines/terms/22/target-offset.html
```diff
@@ -1,6 +1,7 @@
 <dt class="new"><dfn data-lt="target offsets">target offset</dfn></dt>
 <dd class="new">
 	<p class="change">New</p>	
-   <p>the distance measured from the farthest point of a target (A), to the nearest point of the second target (B). The offset includes the target and spacing around the target.</p>
-   <p class="note">The offset from A to B may be different than the offset from B to A, if the sizes of these targets differ.</p>
+   <p>length of the longest possible line that starts at an edge of a target (A), intersects a second edge of A, and ends at the closest edge of a second target (B). For horizontally aligned targets, target offset is measured with a horizontal line. For vertically aligned targets, target offset is measured with a vertical line. For targets that are neither, target offset is measured diagonally.</p>
```
detlevhfischer, line 4: I find the definition "the longest possible line that starts at an edge of a target (A), intersects a second edge of A, and ends at the closest edge of a second target (B)" **very hard** to understand. It may be necessary to put it in such terms if there is no other way to cope with the many scenarios of target juxtapositions, but people will balk at this and I can understand why. 

I also think the logic is not quite clear. It is not obvious that for starting the line for measurement, you should pick the the edge that is **furthest away** from the adjacent target and then move across to the closest edge of the adjacent target. It actually understood, the implicit algorithm of "longest possible line" works fine, but it is far from obvious to the general user base trying to work out what this means.

JAWS-test: especially since it is not true that the line must intersect the object again

awkawk: The longest line that starts at an edge of the triangle AND intersects with a second edge of the triangle AND connects with the square can't be the very tip of point A - one pixel in along it left edge would work however.

guidelines/sc/22/target-size-minimum.html
```diff
@@ -7,7 +7,7 @@ <h4>Target Size (Minimum)</h4>
     
     <p>The size of the <a>target</a> for <a>pointer inputs</a> is at least 24 by 24 <a>CSS pixels</a>, except where:</p>
     <ul>
-        <li><strong>Spacing:</strong> The <a>target offset</a> is at least 24 CSS pixels to every adjacent target;</li>
+        <li><strong>Spacing:</strong> The target is non-overlapping and has a <a>target offset</a> of at least 24 CSS pixels to every adjacent target;</li>
```
awkawk, line 10: I think that to avoid future "the target is overlapping the page background" comments we should specify what it is not overlapping. I think that we mean "The target does not overlap any other target and has a..."

guidelines/terms/22/target-offset.html
alastc: @detlevhfischer It isn't that different (in practice) from the furthest point of A to the nearest point of B. 
The difference is that it must touch two edges of A, so you wouldn't draw it from the very tip of the triangle (from JAWs-test's example). 
Also remember that it will be horizontal or vertical if the buttons are at all aligned.

JAWS-test: @awkawk I assumed that the new definition should be identical to the old one and was just an attempt to find a new wording. If the new definition isn't supposed to be identical, then you're right, but then I think the new definition is nonsensical.

dbjorge: I think this most recent example from @JAWS-test is a pretty convincing case that the "touch two edges" wording is confusing. I think different testers would interpret the new proposed wording differently - I expect some will assume that the measurement should still be from the point because that is the limit of what the wording would accept as the thickness of the line being used to evaluate the definition approaches zero, but other testers might assume that there needs to be some minimum line thickness (1 css px probably) where the line touches each of the two edges at two different pixels.

I think the wording is also confusing for targets that aren't simple/convex, where a "longest line" might leave and then reenter target A. It's not clear to me from the current wording whether a target offset measurement line would be allowed to intersect additional edges of A beyond a second one. For example, consider a target which is ring-shaped - can a target offset line pass through the "hole" in the middle? I think different testers would give different answers with this proposal.

mbgower: @dbjorge it says "a second". Would "another" be less possible to misinterpet as a specific number for you? I think the reality is that we are never going to arrive at a phrase that someone could not misinterpret. That's what examples and clarification in the Understanding document can do. This is usable. Incidentally, in your donut example, yes, if it passes through 2 lines (the exterior and interior of one side of the donut) it would meet this definition. It would mean the user was assessing the further point of one target from another. That is the intent. The offset is about spacing, not about the size of the donut. What is the problem you see with that?

dbjorge: @mbgower I agree that it's not realistic to come up with wording that could never be misinterpreted - my intent is to suggest a change to help make predictable misinterpretations less likely, not impossible, especially where there exists a fairly small change that achieves this. I think "a second edge" and "another edge" are equally good and agree that they're usable, but I think "at least one other edge" would be a bit harder to misinterpret, based on what I think you're describing as the intent.

I think the new wording that excludes overlapping targets from using the offset exception probably avoids most practical issues with shapes with holes in them; all the practical examples I can think of involving real cases using targets with holes in them come from other targets overlapping them. I think cases with pointed edges in different orientations like the triangle example @JAWS-test posted are more likely to come up in practice (shapes that are rotated slightly, star rating widgets, etc), though still not particularly common.

PR #2800, opened 2022-11-16
Remove F87

### LINKED ISSUE #1554: F87 and ARIA 24 seem to contradict in the use of CSS content  property
[ARIA24 example 3](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA24#example-3-email-icon-font-in-a-link-without-visible-text) (and maybe others) indicates the content property can be used while [F87](https://www.w3.org/WAI/WCAG21/Techniques/failures/F87.html) says it can't be relied upon.

### LINKED ISSUE #3781: 1.3.1 / F87 for input fields with mandatory visual indication via pseudo-element
Does the F87 fail test for the CSS override of a pseudo-element concern the case of an input field with aria-required attributes?  

What are your thoughts on this? 

/J

awkawk:
Removing F87 Failure technique. Resolves #1554. Closes #3781


JAWS-test:
[F3](https://www.w3.org/WAI/WCAG21/Techniques/failures/F3) may also be removed soon, because CSS graphics can also get an alternative text. Possibly we should still wait for sufficient screen reader support

patrickhlauke (approved):

techniques/index.html
```diff
@@ -74,6 +74,7 @@ <h2>Introduction to Techniques for WCAG 2.2</h2>
 			<h2>Change Log</h2>
 			<p>A list of new, removed or significantly updated techniques:</p>
 			<ol>
+        <li><time datetime="2020-07-15">Nov 16th 2022</time>: Removed F87 Failure of Success Criterion 1.3.1 due to inserting non-decorative content by using :before and :after pseudo-elements and the 'content' property in CSS</li>
```
scottaohara, line 77: minor nit that it should be `::before` and `::after` rather than the legacy single `:`

but also that the date for the changelog will need to be updated (both since this took a long time to merge, and also because the datetime attribute needs to match the visible date)

patrickhlauke: i just pushed a fix for the incorrect `datetime` attribute

mbgower:
I altered the description so it will close two different related issues on adoption.

patrickhlauke:
Closing this in favor of https://github.com/w3c/wcag/pull/3818

PR #2813, opened 2022-11-25
Re-naming accessible-authentication-no-exception

### LINKED ISSUE #2715: Normative wording inconsistencies for Accessible Authentication SCs
For both [3.3.7](https://www.w3.org/TR/WCAG22/#accessible-authentication) and [3.3.8](https://www.w3.org/TR/WCAG22/#accessible-authentication-no-exception), there are a number of trivial language considerations that could be advanced to improve consistency.
1. In other situations where a baseline minimum requirement is introduced and more pronounced requirement is added at another level, the convention is for the first to be followed by "Minimum" and the second is followed by "Enhanced". Here, it is following a different convention of appending "No exception". This is used only in Keyboard and Images of Text, where the language of the lower bar specifically uses "except" in the wording. That word is absent in 3.3.7. As well, 3.3.8 arguably continues to provide an 'exception' through its "unless" language. Rather than reword the SC, it is probably easier to just make it "(Minimum)" and "(Enhanced)".
2. In both SCs, the wording in the bullets is not consistent. The first bullet ("Another authentication method that does not rely on a cognitive function test.") is a sentence fragment and reads as a continuation of the pre-amble. Others, such as "A mechanism is available to assist the user in completing the cognitive function test", is a full sentence.   Easiest way is to either add "exists" to the first bullet (Another authentication method exists that..) OR remove the verbs from the others ("The cognitive function test recognizes objects").
Depending on which way we want to go, I'm happy to do a PR for these

alastc:
Closes #2715

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/pull/2813.html" title="Last updated on Dec 1, 2022, 9:51 PM UTC (ac38a86)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/2813/0a9b04a...ac38a86.html" title="Last updated on Dec 1, 2022, 9:51 PM UTC (ac38a86)">Diff</a>


bruce-usab:
I cannot tell if the SC name change was effective everywhere.  For example, 3.3.8 still shows "no exception" here:
https://pr-preview.s3.amazonaws.com/w3c/wcag/2813/0a9b04a...9ab5e90.html#new-features-in-wcag-2-2

alastc:
> I cannot tell if the SC name change was effective everywhere. 

I've done a search across the codebase and caught a few more, I think the only one left is appropriate (in the change log).

Issue #2792, opened 2022-11-15
Editorial: Remove blank Benefits section from Understanding Guideline 2.5


maryjom:
In WCAG 2.2: The guidelines typically do not have a benefits section. [Guideline 2.5: Input Modalities Understanding](https://www.w3.org/WAI/WCAG22/Understanding/input-modalities) has a blank Benefits section which should be removed.

kfranqueiro:
Fixed by #2822

PR #2822, opened 2022-12-01
Remove empty Benefits section

maryjom:
Per #2792 removing the empty section.


alastc:
Thanks @maryjom, I did restore one of the closing section tags, that was needed for the intent section.

Issue #2820, opened 2022-11-30
Do not remove 4.1.1 Parsing from WCAG 2.2


WilcoFiers:
Since I wasn't able to make it to the call in which this was discussed, I wanted to raise the issue here. I have strong concerns with removing SC 4.1.1 Parsing in WCAG 2.2. Removing this SC creates a major discrepancy with prior versions of WCAG 2. This results in a lot of (largely unnecessary) difficulties for adopting WCAG 2.2, and raises complicated questions for people still on WCAG 2.0 or 2.1.

Many organizations have resources that are used with different versions of WCAG 2. Think of training courses, templates, test tools, etc. The W3C have many such resources too. If SC 4.1.1 is removed, for all those thousands of resources a decision will now need to be made on how to handle this removed criterion. Do you remove it from the resource, and so removing it from WCAG 2.0 and 2.1? Do you leave it in, and so basically ignore it was removed from WCAG 2.2? Do you insist on readers/users selecting a version so you can switch things out? Do you try to add clarifying text? If so, what do you call a criterion that is removed in one version, but not others?

Because WCAG 2 is designed to be additive, nobody built / wrote things to allow success criteria to get removed. There are multiple projects I am personally involved with that are not able to handle this change without some significant rethinking and costly changes that need to be made. I won't be the only one, but here's how this is going to impact me:

So for [ACT rules](https://www.w3.org/WAI/standards-guidelines/act/rules/) we probably can't remove rules. We've been deprecating rules, so I imagine ACT will decide to list 4.1.1 rules as deprecated. But then is it clear that one place on the W3C says 4.1.1 is removed, and another says resources related to it are deprecated? For [axe-core](https://www.npmjs.com/package/axe-core) the way conformance levels is done simply doesn't allow for rules to be turned off when a certain level is set. Changing that would be very costly. Probably not worth the investment for a one-off scenario. No clue what we'll do instead though.

Then there's the challenge people who're still using 2.0 and 2.1 face. What are they going to do with a success criterion that is removed from WCAG 2.2? Can they stop testing it? Is that contractually, and legally allowed? Are they going to want to take the risk? Are they now stuck on WCAG 2.0 because the lawyers aren't sure they can use WCAG 2.2 for testing? 

What's being proposed is almost a textbook example of a breaking change. This is costly, will create a lot of confusion, will result in inconsistencies and splintering, etc.

[...]

PR #2823, opened 2022-12-01
List SC 4.1.1 as obsolete

WilcoFiers:
See #2820. Instead of removing the SC, I think we should have marked it as obsolete, with a note explaining not to fail it for modern technologies. I think this causes far fewer down-stream problems. Even if we don't do an errata for WCAG 2.0 and 2.1 (which I think we should), this note makes it clear that 4.1.1 should not be failed there either.

### DIFF
--- a/guidelines/sc/20/parsing.html
+++ b/guidelines/sc/20/parsing.html
@@ -1,8 +1,31 @@
 <section class="sc">
    
-   <h4>Parsing</h4>
+   <h4>Parsing (Obsolete)</h4>
 
-   
-   <p>This criterion has been removed.</p>
+   <p class="conformance-level">A</p>
+
+   <p>In content implemented using markup languages, elements have complete start and end
+      tags, elements are nested according to their specifications, elements do not contain
+      duplicate attributes, and any IDs are unique, except where the specifications allow
+      these features.
+   </p>
+
+   <div class="note">
+      <p>
+         Modern web technologies have standardized how user agents parse incorrect markup. 
+         Any invalid markup is therefore allowed under 4.1.1. Parsing for HTML 5 and CSS 3. 
+         When using such technologies this success criterion is always satisfied. 
+         This success criterion is therefore obsolete.
+      </p><p>
+         Issues such as incorrect states or names due to a duplicate ID, or missing roles 
+         due to nesting interactive elements are covered by different success criteria.
+      </p>
+   </div>
+
+   <p class="note">Start and end tags that are missing a critical character in their formation, such
+      as a closing angle bracket or a mismatched attribute value quotation mark are not
+      complete.
+   </p>
+
    
 </section>
--- a/guidelines/sc/20/parsing.html
+++ b/guidelines/sc/20/parsing.html
@@ -18,7 +18,7 @@
          This success criterion is therefore obsolete.
       </p><p>
          Issues such as incorrect states or names due to a duplicate ID, or missing roles 
-         due to nesting interactive elements are covered by different success criteria.
+         due to inappropriately nested elements are covered by different success criteria.
       </p>
    </div>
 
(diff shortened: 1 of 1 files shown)


guidelines/sc/20/parsing.html
```diff
@@ -1,8 +1,31 @@
 <section class="sc">
    
-   <h4>Parsing</h4>
+   <h4>Parsing (Obsolete)</h4>
```
WilcoFiers, line 3: This "(Obsolete)" is new.

guidelines/sc/20/parsing.html
```diff
@@ -1,8 +1,31 @@
...
+
+   <p class="conformance-level">A</p>
+
+   <p>In content implemented using markup languages, elements have complete start and end
+      tags, elements are nested according to their specifications, elements do not contain
+      duplicate attributes, and any IDs are unique, except where the specifications allow
+      these features.
+   </p>
+
+   <div class="note">
```
WilcoFiers, line 13: This note is also new. (Everything else is a revert from #2797)

guidelines/sc/20/parsing.html
```diff
@@ -1,8 +1,31 @@
...
+      <p>
+         Modern web technologies have standardized how user agents parse incorrect markup. 
+         Any invalid markup is therefore allowed under 4.1.1. Parsing for HTML 5 and CSS 3. 
+         When using such technologies this success criterion is always satisfied. 
+         This success criterion is therefore obsolete.
```
WilcoFiers, line 18: Slight rewording.

```suggestion
         Modern web technologies have standardized how user agents parse incorrect markup. 
         Any invalid markup is therefore allowed under 4.1.1. Parsing for technologies such as
         HTML 5 and CSS 3. This success criterion is always satisfied for these technologies. 
         This success criterion is therefore obsolete.
```

guidelines/sc/20/parsing.html
```diff
@@ -1,8 +1,31 @@
...
+
+   <div class="note">
+      <p>
+         Modern web technologies have standardized how user agents parse incorrect markup. 
+         Any invalid markup is therefore allowed under 4.1.1 Parsing for technologies such as
+         HTML 5 and CSS 3. This success criterion is always satisfied for these technologies. 
+         This success criterion is therefore obsolete.
+      </p><p>
+         Issues such as incorrect states or names due to a duplicate ID, or missing roles 
+         due to nesting interactive elements are covered by different success criteria.
```
scottaohara, line 21: not just due to nesting of interactive elements, but invalid nesting of elements in general can result in certain element's roles being dropped by the browser, or if not dropped, still not respected by AT due to their unexpected location in the a11y tree.

WilcoFiers, line 21: How about this then @scottaohara?

```suggestion
         Issues such as incorrect states or names due to a duplicate ID, or missing roles 
         due to inappropriately nested elements are covered by different success criteria.
```

guidelines/sc/20/parsing.html
```diff
@@ -1,8 +1,31 @@
...
+      as a closing angle bracket or a mismatched attribute value quotation mark are not
+      complete.
+   </p>
 
    
-   <p>This criterion has been removed.</p>
-   
-</section>
\ No newline at end of file
+</section>
```
WilcoFiers, line 31: Note I didn't update the understanding doc. If this goes through, the understanding doc needs to be tweaked. Happy to help with that if the group's open to this solution.

guidelines/sc/20/parsing.html
scottaohara: yup. that'll do.

ollie-iterators:
4.1.1 has already been changed to be labeled as obsolete since this PR (https://github.com/w3c/wcag/pull/2840) was merged

alastc:
Closing, decisions went a different way.

PR #2825, opened 2022-12-01
Update to spacing exception

### LINKED ISSUE #2695: Difficulties with inconsistency of Target Size (Minimum)

alastc:
Closes #2695

### DIFF
--- a/understanding/22/target-size-minimum.html
+++ b/understanding/22/target-size-minimum.html
@@ -25,11 +25,11 @@
              <p>The size of the <a>target</a> for <a>pointer inputs</a> is at least 24 by 24 CSS pixels, except where:</p>
 
             <ul>
-                <li><strong>Spacing:</strong> The <a>target offset</a> is at least 24 CSS pixels to every adjacent target;</li>
+                <li><strong>Spacing:</strong> There is at least one point in a target that is at least 12 CSS pixels from the closest edge of all other targets simultaneously;</li>
+                <li><strong>Equivalent:</strong> The function can be achieved through a different control on the same page that meets this criterion;</li>
                 <li><strong>Inline:</strong> The target is in a sentence or block of text;</li>
-                <li><strong>Essential:</strong> A particular presentation of the target is <a>essential</a> or is legally required for the information being conveyed;</li>
                 <li><strong>User agent control:</strong> The size of the target is determined by the user agent and is not modified by the author;</li>
-                <li><strong>Equivalent:</strong> The function can be achieved through a different control on the same page that has an area of at least 24 by 24 CSS pixels.</li>
+                <li><strong>Essential:</strong> A particular presentation of the target is <a>essential</a> or is legally required for the information being conveyed.</li>
             </ul>
             <p class="note">Targets that allow for values to be selected spatially based on position within the target are considered one target for the purpose of the Success Criterion. Examples include sliders with granular values, color pickers displaying a gradient of colors, or editable areas where you position the cursor.</p>
             <p class="note">Are there issues with internationalization when describing inline links?</p>
--- a/guidelines/sc/22/target-size-minimum.html
+++ b/guidelines/sc/22/target-size-minimum.html
@@ -7,11 +7,11 @@
     
     <p>The size of the <a>target</a> for <a>pointer inputs</a> is at least 24 by 24 <a>CSS pixels</a>, except where:</p>
     <ul>
-        <li><strong>Spacing:</strong> The <a>target offset</a> is at least 24 CSS pixels to every adjacent target;</li>
-        <li><strong>Equivalent:</strong> The function can be achieved through a different control on the same page that meets this criterion.</li>
+        <li><strong>Spacing:</strong> There is at least one point in a target that is at least 12 CSS pixels from the closest edge of all other targets simultaneously;</li>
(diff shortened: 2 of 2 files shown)


bruce-usab (approved):
Minor nit pick: 
> There is at least one point in _a_ target that... 

Should be 
> There is at least one point in _the_ target that...

GreggVan:
Good catch.

But if conformance is by page  then it should be every target 
 - no?

gregg

understanding/22/target-size-minimum.html
```diff
@@ -52,9 +52,9 @@ <h2>Intent of Target Size (Minimum)</h2>
...
-            <li>If the spacing of the targets is essential to the information being conveyed, the "Essential" exception applies. For example, in digital maps, the position of pins is analogous to the position of places shown on the map. If there are many pins close together, the offset between pins and neighboring pins will often be below 24 CSS pixels. It is essential to show the pins at the correct map location, therefore the Essential exception applies. As well, some jurisdictions legally require online forms to replicate paper forms, which can impose constraints on the size of targets. In such jurisdictions, any legal requirement to replicate small targets can be considered essential.</li>
+            <li>If a target is smaller than 24 by 24 CSS pixels but has sufficient spacing, it can pass. The spacing is sufficient if there is a 24 CSS pixel circle that you can draw without overlapping another target. A point in the target that has 12 pixels in each direction creates that circle. For example, if a target is 16 CSS pixels wide, but there is no other target within 4 pixels, it would pass.</li>
```
dbjorge, line 55:
```suggestion
            <li>If a target is smaller than 24 by 24 CSS pixels but has sufficient spacing, it can pass. The spacing is sufficient if there is a 24 CSS pixel diameter circle that you can draw without overlapping another target. A point in the target that has 12 pixels in each direction creates that circle. For example, if a target is 16 CSS pixels wide, but there is no other target within 4 pixels, it would pass.</li>
```

understanding/22/target-size-minimum.html
```diff
@@ -25,11 +25,11 @@ <h4>Target Size (Minimum)</h4>
              <p>The size of the <a>target</a> for <a>pointer inputs</a> is at least 24 by 24 CSS pixels, except where:</p>
 
             <ul>
-                <li><strong>Spacing:</strong> The <a>target offset</a> is at least 24 CSS pixels to every adjacent target;</li>
+                <li><strong>Spacing:</strong> There is at least one point in a target that is at least 12 CSS pixels from the closest edge of all other targets simultaneously;</li>
```
dbjorge, line 28:
```suggestion
                <li><strong>Spacing:</strong> There is at least one point in the target that is at least 12 CSS pixels from the closest edge of all other targets simultaneously;</li>
```

dbjorge:
Some of the examples towards the bottom of the understanding document need to be updated to account for this change. In particular:

* The example in Figure 3 passes in the updated spacing exception. I think the most straightforward update would be to update the image to include a third row of buttons and update the text to clarify that the middle row of buttons fails the spacing exception in such an example. The text should be adjusted to avoid referring to the term "offset"
* The examples in Figures 5, 8, and 9 are still using offset-based language and diagrams; it would be helpful to update them to demonstrate pass/fail by presenting an example of a point in the passing targets that are sufficiently far from others, and to avoid using the term "offset"

dbjorge:
Can the "target offset" definition be removed, since the exception no longer references it

alastc:
This proposal ran into issues in the [meeting](https://www.w3.org/2022/12/06-ag-minutes#t03). Closing.

Issue #2736, opened 2022-10-17
Potential unintentional failure for Focus Not Obscured (Enhanced), due to wording and definition


mbgower:
The wording of [2.4.13 Focus Not Obscured (Enhanced)](https://www.w3.org/TR/WCAG22/#focus-not-obscured-enhanced) reads:
> When a [user interface component](https://www.w3.org/TR/WCAG22/#dfn-user-interface-components) receives keyboard focus, no part of the [focus indicator](https://www.w3.org/TR/WCAG22/#dfn-focus-indicator) is hidden by author-created content.

This causes a potential problem where an item with a compliant focus indicator (per 2.4.11 Focus Appearance) could fail this AAA even though the component and the area of the focus indicator that passes Focus Appearance are not obscured at all.

This is caused because the wording of Focus Not Obscured talks about "no part of the focus indicator" instead of "no part of the sufficient focus indicator area" or words to that effect. Essentially, the second set of bullets in Focus Appearance allow for part of the focus indicator to be below contrast requirements (for things like halos, shadow, fades, etc) so long as an area of sufficient size and contrast exists, whereas 2.4.13 just points to the definition, which reads:
> pixels that are changed to visually indicate when a user interface component is in a focused state

The following two images, which are Figure 17 and 18 in the  Focus Appearance Understanding document, may help. The first uses a shadow to indicate focus around the middle of three buttons. The entire shadow is not needed to meet the contrast requirements. 
![focus-indicator-box-shadow](https://user-images.githubusercontent.com/14298245/196245618-65367180-79a1-4c66-bb08-4cd38ae58ff5.png)
The second image shows the parts of the focus indicator that do meet 3:1. 
![image](https://user-images.githubusercontent.com/14298245/196245746-e642fdeb-482c-4715-99de-e542a2bc6459.png)
If something obscured the very light pixels that 'smooth out' the shadow in the first image, it would not affect any of the pixels that are meeting the focus area. 

It follows logically that an inconsequential part of the focus indicator should not lead to a failure of Focus Not Obscured, even the enhanced version.

---
If the rephrasing suggested above does not work, another approach is to fall back to just making the enhanced version not obscure the component as opposed to its focus indicator. For instance:
> When a [user interface component](https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-enhanced#dfn-user-interface-component) receives keyboard focus, no part of the component is hidden by author-created content.

[...]

PR #2838, opened 2022-12-09
Update normative text

mbgower:
Addresses #2736 with the preferred approach from Tuesday's AGWG meeting.
The requirement is reworded to focus on the obscuring of the component, not its focus indicator


mbgower:
@alastc I can't remember if this was a duplicate or if it got missed, so just flagging you on it.

alastc:
No objections from here: https://lists.w3.org/Archives/Public/w3c-wai-gl/2023JanMar/0013.html

So merging...

PR #2840, opened 2022-12-13
4.1.1 updates

alastc:


GreggVan:
suggest adding   (after "It is also that case that assistive technology used to do their own parsing of markup, but now rely on the browser.")   "As a result, the problem that this SC was meant to solve no longer exists, yet it is one of the most common failures on web sites.  That is, the site fails this SC yet there is no negative impact on the use of the site by people with disabilities because all browsers auto-repair such issues before the information is used by AT. Since there is no impact on accessibility anymore this item is now obsolete and has been removed."

bruce-usab (changes requested):
Current:
> This was originally adopted to address problems that Assistive Technology had directly parsing HTML. This is no longer true so this criterion no longer solves that problem and is removed.

Proposed:
> This criterion was originally adopted to address problems that Assistive Technology had directly parsing HTML.  Assistive Technology no longer has any need to directly parse HTML and, consequently, this problem no longer exists.  Accessibility errors failed by this criterion also fail other criteria.  This criteria no longer has utility and is removed.

bruce-usab:
Understanding might be updated to reflect that correcting well-formedness (essentially what 4.1.1 required) can be a useful technique with root cause analysis of some accessibly errors.

GreggVan:
+1    very good and clear

gregg

GreggVan:
Hi Bruce.  Not sure that is understandable by most people.  (For example - I don’t know what it means.) it also seems to contradict the other text in that this says it is useful while the other text you proposed says it doesnt have any effect on accessibility.    Can you explain? 

gregg

PR #2858, opened 2022-12-19
Align AAA target-size inline definition with the newer AA version

### LINKED ISSUE #2857: Errata for 2.5.5 Target Size (Enhanced) to align with  2.5.8
There is a proposed update in #2856 for Target Size (min) to the "inline" exception:
"The target is in a sentence, or the size of the target is otherwise constrained by the line-height of non-target text;"

I think this is a clearer and more testable version, and would like to update SC 2.5.5 in WCAG 2.1 on that basis.

alastc:

---

This PR aligns the older AAA target-size with the newer AA version, for the "inline exception".

The old definition is essentially: The target is in a sentence, or multiple sentences. (Because the definition of "block of text" is one or more sentences.)

The new one is better, but slightly different.

This would be a normative change. As things stand, _IF_ this were approved, it would go into the errata page for WCAG 2.2 and 2.1, and the editors draft, but would not appear on the front of WCAG unless it was re-published.

Closes #2857

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/pull/2858.html" title="Last updated on Oct 28, 2024, 4:17 PM UTC (465e5d7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/2858/bfa7f4c...465e5d7.html" title="Last updated on Oct 28, 2024, 4:17 PM UTC (465e5d7)">Diff</a>


bruce-usab:
Discussed on TF call 11/22.  Concurrence on edit, and already highlighted in [email to AG](https://lists.w3.org/Archives/Public/w3c-wai-gl/2024OctDec/0114.html).

dbjorge:
-1 from me. This is a backwards-incompatible normative change; this erratum would mean that some cases will pass in 2.2+errata which would have failed under 2.2 without errata (ie, in 2.2 as referenced in some legislation).

I agree that it's frustrating that they're inconsistent, but I think backwards compatibility is more important than consistency.

mbgower:
This may be a good example for allowing us to identify the scope of the TF's work.

alastc:
> (ie, in 2.2 as referenced in some legislation).

Hi @dbjorge, which legislation references the AAA guidelines? At AA I'd agree, but (as per the advice in conformance) I don't think anyone requires AAA criteria.

> this erratum would mean that some cases will pass in 2.2+errata which would have failed under 2.2 without errata

I'd argue this is a disambiguation, where the original text is more ambiguous than the update. I.e. people aren't arguing whether something is a sentence or not. Therefore cross-tester consistency is improved and some proportion of things people failed before, and others didn't, would be reduced.

bruce-usab:
Discussed on TF call 2/7.  Deferred because of some concern that this is a Class 2 change.   Moved to Future version updates.

PR #2862, opened 2022-12-20
Updates to index and about pages

iadawn:
Seems the header information for the Understanding and Techniques index and about pages wasn't included. I have added this in. There is also a minor change to the base.css. Finally, I have created a relative link to the base.css. I am not 100% sure on how this is normally linked but it wasn't being pulled in.


michael-n-cooper:
OBE, merged into redesign and then into main.

PR #2876, opened 2022-12-23
Adding analytics code to about and index pages

iadawn:


shawna-slh:
Hey @iadawn , this doesn't seem right. I expected a change to one XSLT file, but this had load of changes -- and conflicts. I'm sure Michael can't address it in this state.

michael-n-cooper:
OBE, merged into redesign then main

PR #2899, opened 2023-01-05
Remove reference to "black/white color blindness"

jkshapiro:
How about this?

Could also replace "black/white color blindness" with "achromatopsia" as suggested at https://github.com/w3c/wcag/pull/2895#issuecomment-1372290891, but the guidance applies equally for anyone in low-contrast environments e.g. on their cellphones in bright sunlight. The proposed text is disability-agnostic.


JAWS-test:
@jkshapiro Please note that the [current version of G183](https://www.w3.org/WAI/WCAG21/Techniques/general/G183) that belongs to WCAG 2.1 is different from the version edited by that belongs to WCAG 2.0. I wonder 
* if it is a technical error that different Understandings are available (@alastc: please comment) and
* if it would make sense to adapt G183 of 2.0 to the version of 2.1, if the differences between understandings are there on purpose, but are not needed in G183

patrickhlauke:
superseded by https://github.com/w3c/wcag/pull/4887

Issue #744, opened 2019-05-19
SC Understanding update: "What it is about" Vs. "What it isn't about"


jake-abma:
At GAAD I was at a conference and attended an almost 2 hour presentation of WCAG 2.1 with focus on the new SC.  The presenter was a professional from an (well known) accessibility company doing audits.

It was astonishing to see and hear how the new SC were not really well understood. Many corrections were needed to keep the presentation on track, lacking all the nuances we spend so much time on.

After the presentation I had a talk with the presenter and he told me the company as well as many other people in our country discussed the new SC multiple times within a group and this was the result how they understood the SC.

I would like to suggest / propose we add a specific section in ALL the Understanding docs what is IS about and specifically what it is NOT about. This would really, really help lots of people.

2 simpel / not elaborated examples:
------------------------------------------------
**1.3.4 Orientation:**

Is about:
- Only not locking the orientation from within the content

Is **NOT** about:
- No loss of information or functionality
- Preventing horizontal scroll
------------------------------------------------
**1.3.5 Identify Input Purpose**

Is about:
- Only providing the purpose via an attribute
- The attributes may come from different techniques like microformats / autocomplete or others...
- Purely the user using the web page at that moment, not other users or puposes NOT defined in the WCAG list

Is **NOT** about:
- Filling in data so it's easy for the user to use forms
- Only about / requiring autocomplete
- All possible names, emails, etc. on a webpage
------------------------------------------------

This really seems very much needed, they had so much nuances "just a bit wrong" and all agreed with each other if there was not someone (in this case me) to tell them otherwise.

Wondering what you all think of this?!

jake-abma:
This Issue is related to the one from @DavidMacDonald https://github.com/w3c/wcag/issues/742

Some comments already made:

Jake
https://github.com/w3c/wcag/issues/742#issuecomment-493669066

Patrick
https://github.com/w3c/wcag/issues/742#issuecomment-493669211
https://github.com/w3c/wcag/issues/742#issuecomment-493669252

Jake
https://github.com/w3c/wcag/issues/742#issuecomment-493670035

patrickhlauke:
yes ... i've taken a similar stance in internal documentation about how to test WCAG SCs, giving some really basic common examples of what the SC does and doesn't cover. this is the sort of thing that should be very clear and evident in our understanding docs.

[...]

PR #2905, opened 2023-01-06
Add 'In brief' section at start of 2.2 SCs

mbgower:
Addresses https://github.com/w3c/wcag/issues/744

Update: I consistently added an "In brief" section directly before the Intent section for all 2.2

Previews for:

- [Dragging movements](https://raw.githack.com/w3c/wcag/WCAGQuickInfo/understanding/22/dragging-movements.html)


shawna-slh:
We planned to add near the top of the Understanding documents the persona quotes from [What’s New in WCAG 2.1](https://www.w3.org/WAI/standards-guidelines/wcag/new-in-21/) and  [What’s New in WCAG 2.2](https://www.w3.org/WAI/standards-guidelines/wcag/new-in-22/) (and add other persona quotes as we got to writing and reviewing them).

+1 to integrating the brief info **_with_ the persona quotes** in the Understanding docs &mdash; and probably adding the brief info to the [What’s New in WCAG 2.2](https://www.w3.org/WAI/standards-guidelines/wcag/new-in-22/) page.

bruce-usab (approved):

understanding/22/accessible-authentication-enhanced.html
```diff
@@ -45,6 +45,16 @@ <h2>Cognitive function test definition</h2>
         </blockquote>
     </section>
    
+    <section id="brief">
+        <h2>In brief</h2>
+        <dl>
+            <dt>Objective</dt><dd>Make logins possible with less mental effort</dd>
+            <dt>Author task</dt><dd>Don’t make people recognize objects or user-supplied information in order to login</dd>
+            <dt>Key beneficiaries</dt><dd>Users with some cognitive disabilities</dd>
```
patrickhlauke, line 53: should probably also include the author task from AA?

```html
<dt>Author task</dt><dd>Don’t make people memorize or transcribe something in order to log in</dd>
```

ljoakley: "Don’t make people recognize objects", isn't this going to cause a vast majority of web sites to be in violation? It's quite common to use the "select the pictures with XXX" for user verification. This component includes an alternate way for the user to be verified.

mbgower: Especially for brevity, I've thought of these as listing the differences, so am trying_to list the new ask at the AAA level. The assumption is they've done already met the AA.

patrickhlauke: i would recommend making these "standalone" to be honest. never good to assume that devs are aware of this, particularly if they just landed on the page in isolation (after a google search, for instance, or just because they went directly to it from the main spec)

patrickhlauke: alternatively, explicitly add a sentence that says something like "all the AA ones, and then these..."

alastc: We discussed this in the meeting. We agreed to keep them short and focused for now, but we'll revisit that after doing the WCAG 2.1 items. There were some concerns about how long they get if you have to include all the A/AA requirements at AAA.

Issue #2924, opened 2023-01-15
Where do technique ARIA5's example 1's and 2's snippets come from?


ShamarYarde:
I tried accessing the site `oaa-accessibility.org` where the image is supposed to come from, but it directs me to "DRAGON99BET: Daftar Situs Judi Slot Online Gacor Mudah Menang" unrelated resource's URL.  The same domain name is used at the end of each example.

[Patrick: edited to remove the live link]

patrickhlauke:
Well spotted, seems the domain's expired and been snapped up by some shady online gambling outfit. I'll do a PR that neutralises any links to the old oaa-accessibility.org

patrickhlauke:
Ah, luckily, seems that ARIA5 is the only place where this is linked to.

PR #2925, opened 2023-01-15
Update ARIA5 technique

patrickhlauke:
* remove the link and reference to the old oaa-accessibility.org Open Ajax Alliance site, which has since lapsed and been replaced with an online gambling site
* update the examples to use the new APG pattern examples

It doesn't look like there's any other references/links to oaa-accessibility.org throughout the site/examples.

Closes https://github.com/w3c/wcag/issues/2924


alastc:
Merging to remove dodgy links.

PR #2929, opened 2023-01-16
Update G223.html

### LINKED ISSUE #2922: Expected results of the G223 technique should be changed.
The Expected Results section of the [G223: Using an author-supplied, highly visible focus indicator](https://www.w3.org/WAI/WCAG22/Techniques/general/G223) technique should be changed to: "All three checks are true."

detlevhfischer:
Correcting an error in the testing section (it requires that checks 2, 4, and 5 are true, but there are only checks 1, 2, and 3, and all need to be true).

Closes #2922

### DIFF
--- a/techniques/general/G223.html
+++ b/techniques/general/G223.html
@@ -32,7 +32,7 @@
     </section>
     <section class="results"><h3>Expected Results</h3>
        <ul>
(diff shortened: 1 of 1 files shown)


patrickhlauke (approved):

alastc:
Looks good, thanks.

PR #2940, opened 2023-01-17
C20-corrections

detlevhfischer:
* Resources: Correcting Maxdesign URL


alastc:
I added a replacement link for the 456Bereastreet one.

Issue #1077, opened 2020-03-09
Understanding doc contradicts SC2.4.5 Multiple ways


guyhickling:
The second paragraph of the Understanding document says "For a three or four page site, with all pages linked from the home page, it may be sufficient simply to provide links from and to the home page where the links on the home page can also serve as a site map." The way I read this is that in this case each page will have only one way to locate it, which directly contradicts the SC, unless I'm reading the paragraph wrong. What does everybody think?

And yes, I agree this might seem a trivial case to raise, but it sets a precedent for ignoring the SC. Suppose you have a 20 page site, but it is the same simple structure as the example in the Understanding, i.e. all 20 pages can be reached from a link somewhere on the Home page and all have a link back to the Home page.

While seemingly trivial, this question takes on more meaning when you consider the tablet and mobile phone responsive layouts. Even a 4 option menu can become a drop down hamburger menu in the phone layout, and we all know how prone they are to not work for all kinds of users. (And the reason I am raising this query is I just had a case of exactly that. It was 4 pages and, yes, the phone version and 400% zoom layouts did not work for keyboard users. If there had been a second menu in the page footer that wrapped lines instead of turning into a drop down menu on the zoomed layout, the site would still have been usable by those people. As it was, it wasn't.) I agree the prime remedy in that case is to fix that failure, but it does highlight that this issue of multiple ways for small sites is not as trivial as it first looks. 

So, should there be this apparent contradiction in the Understanding document?

patrickhlauke:
> The way I read this is that in this case each page will have only one way to locate it, which directly contradicts the SC, unless I'm reading the paragraph wrong. What does everybody think?

I'd agree that that particular paragraph does contradict the SC.

But it brings me to a different/related question: if the paragraph instead said "and on each page there is always link to any other page" (i.e. there's a standard navigation, and every page can be reached from any other page), does that pass? Because if you have N page, there are N-1 different ways to reach that page (from any of the other pages).

The way I've interpreted (loosely) the SC is to mainly be "user doesn't have to remember 'oh, i can only reach THAT page if i'm on THAT OTHER page and there's a link there to it". that basically they can always work out how to reach a page easily, and they don't have to start remembering complex paths/journeys that are the only way to get to a particular page they had in mind.

detlevhfischer:
I think the inaccessibility of the menu button in your example is a different issue covered by 2.1.1 (possibly also 1.3.1 / 4.1.2) and should not enter the discussion of 2.4.5.
[...]

PR #2970, opened 2023-01-26
Remove technique H59

### LINKED ISSUE #1234: Retire technique H59
[Using the link element and navigation tools](https://www.w3.org/WAI/WCAG21/Techniques/html/H59) may really not be all that useful/relevant anymore. To my knowledge, no browser actually exposes those link elements to users in any way. Old Opera browsers used to have an optional toolbar that would take advantage of these links and show some navigation buttons, but I seem to remember that even back then this was the exception. I'm not aware of any commonly-used extensions for modern browsers that take advantage of these links/expose them to the user either. x-ref #1077

patrickhlauke:
the technique is out of date and has no real-world usefulness, since no major browser these days has any kind of `<link rel="...">` navigation toolbar.

Closes #1234

Also x-ref https://github.com/w3c/wcag/issues/1077


patrickhlauke:
@alastc @mbgower any chance this could be looked at? it's fairly non-controversial in my view, because...no browser has any meaningful support for these `<link rel="..." ...>` constructs these days (and even back in the old days, only Opera really did something with them, for other browsers you needed extensions)

wcag20/sources/misc/wcag2.json
```diff
@@ -4074,14 +4070,6 @@
                 {
                   "id": "TECH:G128",
                   "title": "Indicating current location within navigation bars"
-                },
-                {
-                  "id": "TECH:G127",
```
dbjorge, line 4079: This looks off to me - G127 is still a sufficient technique for 2.4.8, no?

dbjorge (changes requested):

dbjorge (approved):

ljoakley:
Refers to the use of the rel attribute using next/previous in a link tag. Google removed support for next/previous/ in 2019. 77% of all web traffic as of November 2023 uses google, advise marking it as deprecated as rel has many other uses besides next/previous.

ljoakley (approved):

bruce-usab:
On deck for call 1/26, no objections to deleting so far...

PR #2987, opened 2023-01-31
Syntax highlighting for client side script techniques

fstrr:
- Update client-side scripting techniques to allow for syntax highlighting in the new design.
- Out of date resource links either replaced or removed.
- Updated SCR14 to use aria instead of JavaScript alerts.


fstrr:
I also went through the working examples for the techniques and, where necessary, added `lang` attributes, updated the doctype, and updated the meta charset tag.

PR #2993, opened 2023-02-01
Fix padding issue in nav bar

hidde:
There is some whitespace underneath the navigation that looks quite out of place:

navigation with some whitespace at the bottom, indicated with dev tools, the nav is clearly larger than the space that is meant for it

This PR fixes it by removing that extra padding, so that the navigation is the same height as the items in it, this makes it looks more tidy, like this: 

same as above but with extra space removed

### DIFF
--- a/css/base.css
+++ b/css/base.css
@@ -289,10 +289,6 @@ width: 2em;
 height: 2em;
 }
 
-.standalone-resource-pager {
-padding: .7em 0;
-}
-
 .standalone-resource-pager ul {
 display: block;
 }


hidde:
@michael-n-cooper @alastc could this please be considered? I hope this is a non controversial change and the fixing of what most would consider a visual bug.

iadawn (approved):

hidde:
@michael-n-cooper @alastc the updated CSS was not deployed yet for WCAG 2.1 docs, would you please consider deploying it, as these are the documents people are looking at today? many thanks!

Issue #2378, opened 2022-05-11
Expect difference in Sufficient Techniques 3.3.7. and 3.3.8, at least one


jake-abma:
https://w3c.github.io/wcag/understanding/accessible-authentication.html
https://w3c.github.io/wcag/understanding/accessible-authentication-no-exception.html

Sufficient Techniques are exactly the same for both 3.3.7 and 3.3.8, would be good to see a difference as they are 2 SC where one will be sufficient for 3.3.7 but not 3.3.8

alastc:
Sure, just need to find someone with time to write such a technique...

mbgower:
@jake-abma 
I agree it seems rather odd when even the future techniques are essentially the same for both the AA and AAA.

Examining the differences between the SCs, I think the thing to do is add a technique to 3.3.7 (the AA, not the AAA) that covers the use of the CAPTCHA 'select objects' technique, since as far as I can tell that is what the exception is there for. That provides a clear difference between the two SCs in regard to implementation.

Of course, focusing on the CAPTCHA highlights a problem with this SC since it's not an authentication method. It's a human validation process that for whatever reason pops up after an authentication process in some circumstances. Since we don't actually define authentication, I guess we can just let it pass...

alastc:
> Of course, focusing on the CAPTCHA highlights a problem with this SC since it's not an authentication method. It's a human validation process that for whatever reason pops up after an authentication process in some circumstances. 

If it isn't part of the authentication process then it's out of scope, but there are instances when it pops up _during_ an authentication process. Those are the ones it is intended to catch.

awkawk:
Seems like "authentication process" is the parent and it includes the human validation/verification process that is the basis of the exception. I agree with @mbgower that the "click on all of the cars" challenges are not authentication but do thing that they are part of some authentication processes.

alastc:
> It would be good to understand if this really is the only intended difference between the AA and AAA requirements. 

That aspect used to be part of the cognitive function test (CFT) definition, but the separation came about because we had two things in conflict: 

- Folks in COGA objected to object-recognition not being covered by CFT. 
- There are instances where those types of CAPTCHAs are used as part of an authentication flow to prevent automated abuse.

I didn't think we could ban people from using something that prevents that type of abuse, they are often triggered by too much traffic from a particular IP address ranges, or sometimes just have an empty cookie. 

So we put in an exception at the AA level (with text pointing out it's still not good to do), and removed it from the AAA level.

It is a niche but important case.

[...]

PR #2997, opened 2023-02-03
Make restriction on captcha clearer

mbgower:
addresses #2378 
Altered the bullets in the AAA understanding and added a step and expected outcome to G218


understanding/22/accessible-authentication-enhanced.html
```diff
@@ -50,13 +50,11 @@ <h2>Intent of Accessible Authentication (Enhanced)</h2>
 
         <p>The purpose of this Success Criterion is to ensure there is an accessible, easy-to-use, and secure method to log in, access content, and undertake tasks. This criterion is the same as <a href="accessible-authentication">Accessible Authentication</a> but without the exceptions for objects and user-provided content.</p>
 
-        <p>The scenarios where the two exceptions might apply are authentication mechanisms which:</p>
+        <p> Any step of the authentication process:</p>
         <ul>
-            <li>display a selection of images, and the user must choose which image they provided;</li>
-            <li>display a selection of items as text, and the user must choose which they had provided;</li>
-            <li>display a selection of images, and the user must choose the images which contain an object such as a car.</li>
+            <li>cannot display a selection of images, videos, or audio clips, where users must choose which image they provided;</li>
```
dbjorge, line 55: I agree that the original wording is a bit confusing, but I think this new wording is still problematic; as I understand it, a step of an authentication process **can** display the stuff in these two bullet points **if** it provides an alternative or mechanism.

I think the quickest fix would be to change L53 to something like `Any **required** step of the authentication process without an assistance mechanism:`. Ideally, I'd like to see the understanding doc provide a more details understanding about what exactly qualifies as "a mechanism... to assist the user in completing the cognitive function test", but that should probably be its own PR

PR #3008, opened 2023-02-09
Examples of 'set of web pages'

### LINKED ISSUE #2298: Clarification sought on "set of web pages"
The definition of a set of web pages indicates a collection of web pages with a common purpose - it would be helpful to have more examples beyond the publication example give.  

For example - within a financial institution - are the web pages related to auto loans considered as part of the same set as pages on home loans?  

With use of this term in the Findable Help SC it means we need to have more consistency in how we collectively agree it should be understood.

alastc:
Addresses #2298, but will leave that open for the errata for early versions.

Adding another example of 'set of web pages', but looking for another one.

### DIFF
--- a/guidelines/terms/21/set-of-web-pages.html
+++ b/guidelines/terms/21/set-of-web-pages.html
@@ -4,11 +4,12 @@
    <p>collection of <a>web pages</a> that share a common purpose and that are created by the same author, group or organization
    </p>
    
-   <aside class="example"><p>Examples include a publication which is split across multiple Web pages, where each page contains
-      one chapter or other significant section of the work. The publication is logically
-      a single contiguous unit, and contains navigation features that enable access to the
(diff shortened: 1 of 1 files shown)


fstrr:
I added some examples on [the related issue](https://github.com/w3c/wcag/issues/2298).

bruce-usab:
Is there disagreement that the term has some subjectively?  I will know it when I see it...

- Set of pages not in the primary language.  Example:  https://www.ssa.gov/es
- Set of pages which are topic specific and don't match the parent theme.  Example:  ANDI landing pages and very thing linked from there:  https://www.ssa.gov/accessibility/andi

 


I do not think we will be able to come up with a tight definition.  Some blogs will

alastc:
Approved: https://www.w3.org/2023/02/28-ag-minutes#t12

PR #3012, opened 2023-02-09
Overhaul, and removing some examples

### LINKED ISSUE #2429: Adding of exceptions at explaining them in Understanding
In Understanding / Intent there is an explanation for all exceptions:
https://w3c.github.io/wcag/understanding/target-size-minimum.html

> The requirement is for targets to be at least 24 by 24 CSS pixels in size. There are five exceptions...:

The exception itself is not mentioned at the beginning, it would be recommended to do so for clearance on what exception is talked about

Add all 5 to the beginning of the explanation

> 1. Spacing...
> 2. Inline...
> 3. Essentail...
> 4. User agent control...
> 5. Equivalent...

### LINKED ISSUE #2431: First sentence Intent example 2 not correct 2.5.8 Target Size (Minimum)
First sentence Intent example 2 not correct 2.5.8 Target Size (Minimum)
https://w3c.github.io/wcag/understanding/target-size-minimum.html

> The Success Criterion does not apply to inline targets in blocks of text – for example, text links inside a sentence or paragraph.

The first part of the sentence only talks about a sentence(s) **"in blocks of text"** and not about situations where the target is in a sentence BUT NOT in a block of text, thus mentioning / explaining only half of the exception.

Also a target in a sentence must be explained / added.

> blocks of text
> more than one sentence of text

Exception:

> Inline: The target is in a sentence **OR** block of text;

### LINKED ISSUE #2432: Intent explanation nr. 5 is 'area' based, must be 'size' based 2.5.8: Target Size (Minimum)
Intent explanation nr. 5 is 'area' based, must be 'size' based 2.5.8: Target Size (Minimum)
https://w3c.github.io/wcag/understanding/target-size-minimum.html

> In cases where a target does not have an area equivalent to 24 by 24 CSS pixels, but there is another control that can achieve the underlying function that does meet the minimum area, the target can be excepted based on the "Equivalent" exception.

See also: https://github.com/w3c/wcag/issues/2426

alastc:
Closes #2431
Closes #2429
Closes #2432


understanding/22/target-size-minimum.html
```diff
@@ -22,18 +22,16 @@ <h4>Target Size (Minimum)</h4>
...
-            <p class="note">Are there issues with internationalization when describing inline links?</p>
-            <p class="note">Are there issues with pop-over content overlapping targets triggering failures?</p>
+            <p class="note">Targets that allow for values to be selected spatially based on position within the target are considered one target for the purpose of the success criterion. Examples include sliders with granular values, color pickers displaying a gradient of colors, or editable areas where you position the cursor.</p>
+            <p class="note">For inline targets the line-height should be interpreted as perpendicular to the flow of text. For example, in a language displayed top to bottom, the line-height would be horizontal.</p>
             <dl><dt class="new"><dfn data-lt="target offsets">target offset</dfn></dt>
             <dd class="new">
                 <p class="change">New</p>
```
dbjorge, line 37: The Target offset definition has also changed since understanding was written and needs an update

understanding/22/target-size-minimum.html
```diff
@@ -94,25 +96,11 @@ <h3>Spacing versus size</h3>
...
 
-        <p>The aim of this Success Criterion is that targets are ideally at least 24 by 24 CSS pixels. However, the offset exception allows for smaller targets if there is sufficient spacing around them. Authors should be aware that using smaller targets can have an impact on layouts, as the combined space of the smaller target and the required spacing around it will generally require an overall area that is larger than 24 by 24 CSS pixels.</p>
+        <p>The aim of this Success Criterion is that targets are ideally at least 24 by 24 CSS pixels. However, the offset exception allows for smaller targets if there is sufficient spacing around them. Authors should be aware that using smaller targets can have an impact on layouts, as the combined space of the smaller target and the required spacing around it will generally require an overall size that is larger than 24 by 24 CSS pixels.</p>
```
dbjorge, line 103:
```suggestion
        <p>The aim of this Success Criterion is that targets are ideally at least 24 by 24 CSS pixels. However, the Spacing exception allows for smaller targets if there is sufficient spacing around them. Authors should be aware that using smaller targets can have an impact on layouts, as the combined space of the smaller target and the required spacing around it will generally require an overall size that is larger than 24 by 24 CSS pixels.</p>
```

understanding/22/target-size-minimum.html
```diff
@@ -52,27 +50,31 @@ <h2>Intent of Target Size (Minimum)</h2>
...
+            <li><strong>Spacing:</strong> If a target is smaller than 24 by 24 CSS pixels but has sufficient spacing, it passes. The spacing is sufficient if the offset between targets is at least 24 CSS pixels. <a href="#dfn-target-offset">Target offset</a> measures the distance between targets, measured from the farthest side of one target to the nearest side of the adjacent target. It assessed for each target, to each nearby target. What edge of the target is <em>farthest</em> depends on the relative position of the neighboring target. For example, if considering distance above the target, the farthest edge would be the bottom of the target, and the offset would be measured from this bottom edge to the bottom edge of the neighboring target above. Note that where the targets differ in size or shape, the offset between targets may also differ. The offset distance must be at least 24 CSS pixels in all cases to fall under this exception.</li>
```
dbjorge, line 53: We should be consistent about using the full term "target offset" instead of "offset" or "offset distance" - it has a very specific meaning which is different from what most people would think of as an "offset". I think it might also be a good idea to avoid using "farthest" and "nearest" language here, based on the confusion we saw in #2695. Maybe something like this?

```suggestion
            <li><strong>Spacing:</strong> If a target is smaller than 24 by 24 CSS pixels but has sufficient spacing, it passes. The spacing is sufficient if the <a href="#dfn-target-offset">target offset</a> between targets is at least 24 CSS pixels. Target offset is the length of the longest possible line that starts at an edge of a target (A), intersects a second edge of A, and ends at the closest edge of a second target (B). It is assessed <em>separately</em> from a target to each other nearby target; the target offset must be at least 24 CSS pixels to <em>all</em> nearby targets to fall under this exception. The line used to measure target offset depends on the sizes, shapes, and relative positions of the two targets. For example, if considering the target offset from a rectangular target (A) to an equally-sized rectangular neighbor target (B) directly to A's right, target offset would be measured with a horizontal line starting from the left edge of A, passing through the right edge of A, and ending at the left edge of B. Note that the target offset from A to B may be different than the target offset from B to A.</li>
```

dbjorge, line 53: The `#dfn-target-offset` link doesn't seem to work in the rendered githack page - does there need to be an explicit anchor added to the quoted definition near the top?

dbjorge: (maybe also s/offset/target offset in the caption that begins with "Fail: two rows of buttons")

alastc: The links get 'built' when the page is published, the preview is just a quick render of the raw HTML.

alastc: That looks good, apart from the "separately" bit. I don't think it is clear what is separate in that sentence.

I think this would mean what you intended?
`It is assessed from a target to <em>each</em> other nearby target;`

understanding/22/target-size-minimum.html
alastc: I've copied it in from the normative text. (Noting that this version doesn't appear in the published docs.)

Issue #2706, opened 2022-10-05
[WCAG 2.2 Draft Feedback] Success Criterion 2.4.11 Focus Appearance (Level AA)


dshoukry:
“Success Criterion 2.4.11 Focus Appearance (Level AA): When the keyboard focus indicator is visible, one or both of the following are true:

The entire focus indicator meets all the following:
- encloses the user interface component or sub-component that is focused, and
- has a contrast ratio of at least 3:1 between the same pixels in the focused and unfocused states, and
- has a contrast ratio of at least 3:1 against adjacent non-focus-indicator colors.

An area of the focus indicator meets all the following:
- is at least as large as the area of a 1 CSS pixel thick perimeter of the unfocused component or sub-component, or is at least as large as a 4 CSS pixel thick line along the shortest side of the minimum bounding box of the unfocused component or sub-component, and
- has a contrast ratio of at least 3:1 between the same pixels in the focused and unfocused states, and
- has a contrast ratio of at least 3:1 against adjacent non-focus-indicator colors, or is no thinner than 2 CSS pixels.

Exceptions:

- The focus indicator is determined by the user agent and cannot be adjusted by the author, or
- The focus indicator and the indicator's background color are not modified by the author.
- What is perceived as the user interface component or sub-component (to determine enclosure or size) depends on its visual presentation. The visual presentation includes the component's visible content, border, and component-specific background. It does not include shadow and glow effects outside the component's content, background, or border.”

Recommendation:
Similar to our previous rounds of feedback, we kindly request that the group remove this SC until research is completed for the specifics around contrast, shape, and thickness of the indicator as specified in the SC. We are also concerned about the overall understandability and testability. 

From the last responses in [our Github issue](https://github.com/w3c/wcag/issues/1896): 
> "Surface area of the indicator, bigger is better. The contrast of that surface area is compared the un-focused state, more is better. The contrast of the surface area compared to it's adjacent colours, more is better." 

This is all likely so, but there should be research to quantify the pixel and contrast minimums. 

> "We could, but then typical (and reasonable) indicators that people use now would not pass. It would make a lot of websites have to do extra work in order to pass, when their current indicators are reasonably clear to users and pass the current SC text." 

If the current indicators are reasonably clear and pass the current SC text, why is this SC needed and what is the issue we are trying to solve/improve?

[...]

PR #3017, opened 2023-02-10
Update focus-appearance.html

mbgower:
attempts to address comments circa https://github.com/w3c/wcag/issues/2706#issuecomment-1422857287


alastc:
I'm not sure about the `target="blank"` additions, if some do have that, and some don't, we should probably mark those that open a new window.

mbgower:
> I'm not sure about the target="blank" additions, if some do have that, and some don't, we should probably mark those that open a new window.

I'm fine with how you want to proceed, but in a document this big, when your existing location is overwritten by a new page as you active a link, it makes it a lot harder to reorient yourself when you reload the page.

mbgower:
> we should probably mark those that open a new window.

I would have thought that is an automatic thing across all of w3c if that's the case? maybe there's not

understanding/22/focus-appearance.html
```diff
@@ -434,15 +434,16 @@ <h3>First exception: the focus indicator cannot be adjusted by the author</h3>
...
+        <p style="note">Altering the <code>body</code> element's <code>background-color</code> attribute <em>is</em> one way of altering the pixels directly adjacent to the indicator in most implementations. However, specifying a value of white (<code>#FFFFFF</code>) does not nullify this exception since, as established in the third note of the <a href="https://www.w3.org/TR/WCAG22/#dfn-contrast-ratio" target="_blank">contrast ratio definition</a>, the default ("unspecified") color is assumed to be white.</p>
         
-            <p>[Future image set: a button with a default light blue focus indicator; the same buton and focus indicator on a light blue page background]</p>
+            <p>[Future image set: a button with a default light blue focus indicator on a white page; the same buton and focus indicator on a light blue page background]</p>
```
patrickhlauke, line 444: there's an existing typo worth fixing: "buton" > "button"

patrickhlauke:
i know this is closed, but...looks reasonable to me either way

alastc:
Approved: https://www.w3.org/2023/02/28-ag-minutes#r08

Issue #1705, opened 2021-03-28
1.4.3 Minimum Contrast, 1.5 multiplication factor — original research paper?


huayunh:
I was checking [1.4.3 Contrast (Minimum)](https://www.w3.org/WAI/WCAG22/Understanding/contrast-minimum.html)'s page to gain a better understanding of where the 4.5:1 minimum contrast ratio requirement comes from. It cited the paper 

> Arditi, A. and Faye, E. (2004). Monocular and binocular letter contrast sensitivity and letter acuity in a diverse ophthalmologic practice. Supplement to Optometry and Vision Science, 81 (12S), 287.

I checked the American Academy of Optometry's website. According to [the abstract provided](https://www.aaopt.org/detail/knowledge-base-article/monocular-and-binocular-letter-contrast-sensitivity-and-letter-acuity-diverse-ophthalmologic), the only place they ever mentioned the number "1.5" is this sentence: 

> The regression of log CS on log MAR resulted in an intercept of 1.50 and a slope of -0.63.

Which, for me, seems unrelated to the topic. 

My question is, is this Arditi-Faye 2004 paper the correct paper to refer to? If not, where does the 1.5 multiplication factor come from?

jlukosky:
The NIH has more full version of the [Arditi and Faye paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3964624/). It is more technical than I can comprehend.

huayunh:
Appreciate the help! However, it appears that the NIH paper is different from what WCAG is citing...

Myndex:
Hi @huayunh 

> _Appreciate the help! However, it appears that the NIH paper is different from what WCAG is citing..._

Correct, The paper he linked to is not relevant to WCAG 2 contrast, in fact it's from 2013, about 6-8 years after WCAG2 contrast was created. Nevertheless, I do have answers to your questions, in detail below.


## _An Air Or Occured_
> ....._If not, where does the 1.5 multiplication factor come from?_

It comes from thin air, and is not supported by any science. And I've combed through all the related papers, email chains, and much more.

I looked into this back in 2019. If you're not familiar with it, I started thread number #695 in April 2019, while investigating the WCAG 2 contrast issue. 

[...]

PR #3020, opened 2023-02-12
Changes/corrections to 1.4.3 and 1.4.6 understanding

patrickhlauke:
The understanding documents for 1.4.3 Contrast (Minimum) and 1.4.6 Contrast (Enhanced) currently contain various small errors, incorrect links/references, a few throwaway non-sequiturs, some unnecessarily duplicated content, and are not harmonised. Key changes include:

* add a new note (inspired by the note that was in 1.4.6) about thin/unusual fonts and antialiasing and what to actually *do* as an author
* remove the weird inconsistency of talking about "thin or unusual", and in the next part of the sentence "fancy or thin" (also... "fancy"? really?)
* harmonise capitalisation of "Success Criterion"
* in 1.4.3, removes the whole start of the "Rationale" section which is an exact copy of the end of the "Intent" section? reorders that section to be more logical
* some code cleanup (removing excessive line breaks, sorting out some excessively long single lines
* use of `ppi` (as a unit of measure) lowercase and without preceding space
* use of `<code>` for code-y things (measurements, formulae)
* harmonise 1.4.3 and 1.4.6 understanding

in the process, this also removes some broken and incorrect links in 1.4.6 understanding (e.g. link to "resources" that was borked)

### DIFF
--- a/understanding/20/contrast-enhanced.html
+++ b/understanding/20/contrast-enhanced.html
@@ -8,10 +8,8 @@
 <body>
    <h1>Understanding Contrast (Enhanced)</h1>
    
-   
    <section id="intent">
       <h2>Intent of Contrast (Enhanced)</h2>
-      
       
       <p>The intent of this Success Criterion is to provide enough contrast between text and
          its background so that it can be read by people with moderately low vision (who do
@@ -27,7 +25,6 @@
          words are used to create a background and the words could be rearranged or substituted
          without changing meaning, then it would be decorative and would not need to meet this
          criterion. 
-         
       </p>
       
       <p>Text that is larger and has wider character strokes is easier to read at lower contrast.
@@ -37,53 +34,52 @@
          enough to require a lower contrast ratio. (See The American Printing House for the
          Blind Guidelines for Large Printing and The Library of Congress Guidelines for Large
          Print under 
-         <a href="#visual-audio-contrast7-resources-head">Resources</a>). "18 point" and "bold" can both have different meanings in different fonts but,
-         except for very thin or unusual fonts, they should be sufficient. Since there are
-         so many different fonts, the general measures are used and a note regarding fancy
-         or thin fonts is included.
+         <a href="#resources">Resources</a>). "18 point" and "bold" can both have different meanings in
+         different fonts but, except for very thin or unusual fonts, they should be sufficient. Since there
+         are so many different fonts, the general measures are used and a note regarding thin or unusual
+         fonts is included if the definition for <a>large-scale</a> text.
       </p>
       
       <div class="note">
-         
-         <p>The point size should be obtained from the user agent, or calculated based on font
-            metrics as the user agent does when evaluating this success criterion. Point sizes
-            are based on the CSS pt size as defined in 
-            <a href="https://www.w3.org/TR/css-values-3/#reference-pixel">CSS3 Values</a>. The ratio between sizes in points and CSS pixels is 1pt = 1.333px, therefore 14pt
-            and 18pt are equivalent to approximately 18.5px and 24px.
+         <p>When evaluating this Success Criterion, the font size in points should be obtained
+            from the user agent or calculated on font metrics in the way that user agents do.
+            Point sizes are based on the CSS <code>pt</code> size as defined in 
+            <a href="https://www.w3.org/TR/css-values-3/#reference-pixel">CSS3 Values</a>. The ratio between
+            sizes in points and CSS pixels is <code>1pt = 1.333px</code>, therefore <code>14pt</code>
+            and <code>18pt</code> are equivalent to approximately <code>18.5px</code> and <code>24px</code>.
          </p>
-         
-         <p>When fonts have anti-aliasing applied to make them look smoother, they can lose darkness
-            or lightness.  Thus, the actual contrast can be reduced. Thicker stem widths will
-            reduce this effect (thin fonts could have the full stem lightened rather than just
-            the ends). Using larger fonts and testing for legibility in user agents with font
-            smoothing turned on is recommended.
-            
-         </p>
-         
          <p>Because different image editing applications default to different pixel densities
-            (ex. 72 PPI or 96 PPI), specifying point sizes for fonts from within an image editing
-            application can be unreliable when it comes to presenting text at a specific size.
+            (e.g. <code>72ppi</code> or <code>96ppi</code>), specifying point sizes for fonts from within an
+            image editing application can be unreliable when it comes to presenting text at a specific size.
             When creating images of large-scale text, authors should ensure that the text in the
             resulting image is roughly equivalent to 1.2 and 1.5 em or to 120% or 150% of the
-            default size for body text as rendered by the browser.
+            default size for body text. For example, for a <code>72ppi</code> image, an author would need
+            to use approximately 19pt and 24pt font sizes in order to successfully present images
+            of large-scale text to a user. 
          </p>
          <p>The 7:1 and 4.5:1 contrast ratios referenced in this Success Criterion are intended to be
             treated as threshold values. When comparing the computed contrast ratio to the Success Criterion
             ratio, the computed values should not be rounded (e.g. 4.499:1 would not meet the 4.5:1 threshold).</p>
-         
+      </div>
+
+      <div class="note">
+         <p>Due to anti-aliasing, particularly thin or unusual fonts may be rendered by user agents with a much fainter
+            color than the actual text color defined in the underlying CSS. This can lead to situations where text has
+            a contrast ratio that nominally passes the Success Criterion, but has a much lower contrast in practice.
+            For these cases, we recommend as a best practice that authors choose a font with stronger/thicker lines,
+            or aim for a text color that exceeds the baseline requirements of this Success Criterion.
+         </p>
       </div>
       
-      <p>The previously-mentioned contrast requirements for text also apply to
-         images of text (text that has been rendered into pixels and then stored in an image
-         format) as stated in Success Criterion 1.4.5
-         
-         
+      <p>The contrast requirements for text also apply to images of text
+         (text that has been rendered into pixels and then stored in an image format) - see
+         <a href="images-of-text">Success Criterion 1.4.5: Images of Text</a>.
       </p>
       
       <p>This requirement applies to situations in which images of text were intended to be
          understood as text. Incidental text, such as in photographs that happen to include
          a street sign, are not included. Nor is text that for some reason is designed to be
-         invisible to all users. Stylized text, such as in corporate logos, should be treated
+         invisible to all viewers. Stylized text, such as in corporate logos, should be treated
          in terms of its function on the page, which may or may not warrant including the content
          in the text alternative. Corporate visual guidelines beyond logo and logotype are
          not included in the exception.
@@ -96,14 +92,31 @@
          in order to get a particular look. 
       </p>
       
-      <p>Although this Success Criterion only applies to text, similar issues occur for content presented in charts, graphs, diagrams, and other non-text-based information which is covered by 
-         <a href="../21/non-text-contrast.html">Success Criterion 1.4.11 Non-Text Contrast</a>.         
+      <div class="note">
+         
+         <p>Images of text do not scale as well as text because they tend to pixelate. It is also
+            harder to change foreground and background contrast and color combinations for images
+            of text, which is necessary for some users. Therefore, we suggest using text wherever
+            possible, and when not, consider supplying an image of higher resolution.
+            
+         </p>
+         
+      </div>
+      
+      <p>The minimum contrast Success Criterion (1.4.3) applies to text in the page, including
+         placeholder text and text that is shown when a pointer is hovering over an object
+         or when an object has keyboard focus. If any of these are used in a page, the text
+         needs to provide sufficient contrast.
+      </p>
+      
+      <p>Although this Success Criterion only applies to text, similar issues occur for content presented
+         in charts, graphs, diagrams, and other non-text-based information which is covered by
+         <a href="non-text-contrast">Success Criterion 1.4.11 Non-Text Contrast</a>.         
       </p>
       
       <section>
          
          <h3>Rationale for the Ratios Chosen</h3>
-         
          
          <p>A contrast ratio of 3:1 is the minimum level recommended by [[ISO-9241-3]] and [[ANSI-HFES-100-1988]] for standard text and vision. The 4.5:1 ratio is used in Success Criterion 1.4.3
             to account for the loss in contrast that results from moderately low visual acuity,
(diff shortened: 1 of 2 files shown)


patrickhlauke:
would be nice to squeeze these updates in (assuming they're accepted) in time for 2.2 final

Myndex:
Hi Patrick @patrickhlauke 

> _The understanding documents for 1.4.3 Contrast (Minimum) and 1.4.6 Contrast (Enhanced) currently contain **various small errors**_


Those understanding documents contain some **very large errors** as well. Not to mention [unsupportable misunderstandings](https://github.com/w3c/wcag/issues/1705#issuecomment-1027058976).

I'm not gonna make any jokes about painting a burning house _(or am I?)_ but one question I have relates to commit number [d1ef550](https://github.com/w3c/wcag/pull/3020/commits/d1ef550716ce56b8f54a4037ad3c1d5a55fcc816) which seems to be a substantive change, and I bring this up because I was informed that we can't make substantive changes? Or can we? Because the understanding needs a massive ton of substantive changes to make it relevant and actually truthful, which presently it is not.

For instance the statement made in this particular commit d1ef550 is essentially negated by the note that states that ["authors may turn off anti-aliasing when testing"](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html#dfn-contrast-ratio)

Note Because authors do not have control over user settings as to how text is rendered (for example font smoothing or anti-aliasing), the contrast ratio for text can be evaluated with anti-aliasing turned off.

Thing is, as I'm sure you are well aware, **I agree with you with this change** that you suggest making. After all I've been writing about this problem (anti-aliasing and thin fonts) as one of the key readability problems for the last four years.

And anti-aliasing is definitely something that's within control of the author through the related CSS properties, and is frequently misused and abused.... Along with far too thin fonts, which were not common in 2007 when everyone was just using plain web fonts and black `#000000` text.

## _Can we do substantive changes?_
Perhaps this is a better question for Alastair @alastc ??? 

Since this pull request of yours (Patrick) includes some substantive changes relating to meaning, intent, and understanding, I'd like to know to what extent substantive changes can be made to understanding documents here?

This document could be re-written substantially, while keeping the SC 1.4.3 as it stands, and at least improve understanding and help get rid of the poor misunderstandings that appear to stem from this document. Now, I understand there will be no change to the actual math or thresholds at this time—but getting rid of misunderstandings, hand-wavy pseudoscience 'splaining, and the inherent _"reciting facts not in evidence"_, would be a step in the right direction.

While I have no interest in spending the time and effort to do so if I'm ultimately going to be told that it's simply not possible to make such changes, nevertheless in reviewing this PR, I reread that document and I can best describe it as listening to nails on a chalkboard to me, and it makes me immediately think of [**basic changes that are needed.**](https://www.smashingmagazine.com/2022/09/realities-myths-contrast-color/) 


Thank you for reading.

patrickhlauke:
@Myndex I had a strong feeling that, just like saying "Candyman" 5 times while looking in the mirror, this thread would raise you...  🍬 

> I'm not gonna make any jokes about painting a burning house

We've had this discussion many times before, but: until WCAG 3 comes out in 20XX, for better or worse, WCAG 2.x will still be what authors have to care about (section 508 in the US still references WCAG 2.0, the european EN 301 549 still references WCAG 2.1, and even if they adopt WCAG 2.2 coming out soon at some point, this will *still* be referencing 1.4.3 and 1.4.6 for a good few years coming). So while I appreciate that for you it's all about APCA ... where the rubber meets the road, in the here and now of accessibility auditing and legal requirement, *this* is what authors need to work with, so any clarification of our currently valid documentation will have immediate benefits right here, right now.

> but one question I have relates to commit number [d1ef550](https://github.com/w3c/wcag/pull/3020/commits/d1ef550716ce56b8f54a4037ad3c1d5a55fcc816) which seems to be a substantive change, and I bring this up because I was informed that we can't make substantive changes?

i'm sure @alastc can confirm, but ... this is NOT a "substantive change" for three reasons:

1) it's in the understanding document, which is information/non-normative
2) it does not define any change in how authors would evaluate a pass or fail. it suggests that *as a best practice* (i.e. not normatively, not required to pass the SC, but clearly a suggestion of going beyond what the normative requirement is) authors should consider what to do if they do have particularly thin/unusual fonts that result in a visually lower contrast than what the pure numbers evaluation would lead you to believe
3) the subject matter of that note was already present in both 1.4.3 and 1.4.6 understanding, and in 1.4.6's it already hinted at this suggestion more directly... so it's not a new aspect. this change here just makes it clearer what this actually means for authors, makes a best practice suggestion, and then harmonises this across both 1.4.3 and 1.4.6

patrickhlauke:
> getting rid of misunderstandings, hand-wavy pseudoscience 'splaining, and the inherent "reciting facts not in evidence", would be a step in the right direction.

feel free to submit your own PR that fixes this, without changing any of the normative requirements,

detlevhfischer (commented):
All good stuff. 
The only thing I wasn't sure about was the use of "we": 
"we recommend as a best practice", "we suggest" etc. Can't rember whether that kind of "we" is common in understanding texts.

patrickhlauke:
@detlevhfischer 

> The only thing I wasn't sure about was the use of "we"

ah, good catch. yeah, this is probably coming more from the sort of thing I (we) write in my audits for clients...I tend to use this phrasing for best practices in those cases, and it bubbled over into this 😄 

Having a quick peek at existing parts of WCAG, this is the way best practices are generally introduced:

* `A best practice is to ...` (from 2.5.3 Label in Name)
* `Best practice is to ...` (from F24)
* `The best practice is to ...` (from G83)
* `The best practice would be to ...` (from 3.2.4 Consistent Identification understanding)
* `It is a best practice ...` (from 2.4.4 Link Purpose (In Context) understanding)

I'll make an edit to the PR here to adopt this style instead.

patrickhlauke:
@detlevhfischer fixed/tweaked

mbgower (approved):
Nice adjustments. Improves existing non-normative guidance.

patrickhlauke:
@mbgower made a further addition to the note about anti-aliasing https://github.com/w3c/wcag/pull/3020/commits/03427af4c4a16d749603b7373e57752cdedda708 as this aspect (how to actually evaluate regardless of antialiasing) was not clear,  and the note https://www.w3.org/TR/WCAG21/#dfn-contrast-ratio "the contrast ratio for text can be evaluated with anti-aliasing turned off" makes things potentially more confusing/misleading - *can* rather than *should* of *must*?

Myndex:
> _the contrast ratio for text can be evaluated with anti-aliasing turned off" makes things potentially more confusing/misleading_ - _can_ rather than _should_ of _must_?

Hi Patrick @patrickhlauke 

I'd rather not say "**must** turn off anti-aliasing", because while that might be _permitted_ it is definitely **_not best practice_**.


The problem that I have with the way it was originally written, is that turning off anti-aliasing is ** almost a way of cheating** (except that I think it's unlikely that people use the eyedropper method). But the point is, turning off antialiasing before using an eyedropper does not serve the user need for accessibility, because it allows for a situation where the ultimately rendered font with anti-aliasing on, is at a much lower perceived contrast than the low contrast that's already permitted WCAG 2.

From the draft guidance in the readability guidelines we're working on, when using an eyedropper tool to sample:

> - _The eye dropper tool must report the native CSS value for the color space (nominally sRGB)._
> - _The eye dropper sample should be a pixel in the middle of a major stroke of the thinnest weight/smallest size font to be tested, and with the content at the default size (no zoom or scaling)._  
> - how to eyedropper sample letter      
> - _For small and thin fonts, where the major stroke is inconsistent due to antialiasing_, **sample at least three to five different letters and average the result**.

### Also:
The statement (which is also in the original):

> _"...Because authors do not have control over user settings for font smoothing/anti-aliasing...."_

_This is not true._ There is CSS ` -webkit-font-smoothing: ` for macOS, which allows direct modification of anti-aliasing features, and ` text-rendering: ` for Windows and Linux which affects anti-aliasing control to a degree.

A problem with saying that authors don't have control is that it removes those considerations from being tested for, but IMO should be. 

In particular, ` -webkit-font-smoothing: ` is often abused to a very poor effect, meaning unreadable text. If somebody is using ` -webkit-font-smoothing: `, the font weight needs to be substantial enough to handle the blending to the background that ` -webkit-font-smoothing: ` causes.

patrickhlauke:
> I'd rather not say "must turn off anti-aliasing", because while that might be permitted it is definitely not best practice.

it is in terms of testing against the terms of the SC as it was created and as it stands now. the SC never took into account anti-aliasing, smoothing, etc. for evaluation purposes, it relies on checking the color values as defined in the code. which yes, we know is not actually useful in the real world, but that's what we have to work with. given *that* reality, then having a handwavy "can" in a definition is even more confusing.

> This is not true. There is CSS -webkit-font-smoothing: for macOS, which allows direct modification of anti-aliasing features, and text-rendering: for Windows and Linux which affects anti-aliasing control to a degree.

to a degree, perhaps, but not consistently. and that's the main point: since what *I* as an evaluator may see if i color-pick on *my* computer with *my* browser/OS/user settings may be different from what *another* evaluator sees on their machine, we may end up with different results where one of us passes something, and the other person fails it.  which is not ideal for consistency.

also, that phrasing is in the glossary/definitions, which *are* normative, so changing that will be a world of back-and-forth pain to approve and push through a normative change to WCAG...so I left it as is, despite its slight inaccuracy.

> A problem with saying that authors don't have control is that it removes those considerations from being tested for, but IMO should be.

in the context of these two SCs in WCAG 2.x, smoothing and antialiasing were never part of the testing considerations, and that's not going to change.

understanding/20/contrast-enhanced.html
```diff
@@ -37,53 +34,56 @@ <h2>Intent of Contrast (Enhanced)</h2>
...
          Blind Guidelines for Large Printing and The Library of Congress Guidelines for Large
          Print under 
-         <a href="#visual-audio-contrast7-resources-head">Resources</a>). "18 point" and "bold" can both have different meanings in different fonts but,
-         except for very thin or unusual fonts, they should be sufficient. Since there are
-         so many different fonts, the general measures are used and a note regarding fancy
-         or thin fonts is included.
+         <a href="#resources">Resources</a>). "18 point" and "bold" can both have different meanings in
+         different fonts but, except for very thin or unusual fonts, they should be sufficient. Since there
+         are so many different fonts, the general measures are used and a note regarding thin or unusual
+         fonts is included if the definition for <a>large-scale</a> text.
```
mitchellevan, line 40:
```suggestion
         fonts is included in the definition for <a>large-scale</a> text.
```

understanding/20/contrast-enhanced.html
```diff
@@ -136,11 +153,22 @@ <h3>Rationale for the Ratios Chosen</h3>
...
          </p>
          
+         <div class="note">
+            
+            <p>Some people with cognitive disabilities require color combinations or hues that have
+               low contrast, and therefore we allow and encourage authors to provide mechanisms to
+               adjust the foreground and background colors of the content. Some of the combinations
+               that could be chosen may have contrast levels that will be lower than those found
+               in the Success Criteria. This is not a violation of this Success Criterion, provided
+               there is a mechanism that will return to the default values set out here.
```
mitchellevan, line 163:
```suggestion
               there is a mechanism that will return to the required values set out here.
```

understanding/20/contrast-minimum.html
```diff
@@ -37,40 +34,50 @@ <h2>Intent of Contrast (Minimum)</h2>
...
          Blind Guidelines for Large Printing and The Library of Congress Guidelines for Large
          Print under 
-         <a href="#resources">Resources</a>). "18 point" and "bold" can both have different meanings in different fonts but,
-         except for very thin or unusual fonts, they should be sufficient. Since there are
-         so many different fonts, the general measures are used and a note regarding fancy
-         or thin fonts is included.
+         <a href="#resources">Resources</a>). "18 point" and "bold" can both have different meanings in
+         different fonts but, except for very thin or unusual fonts, they should be sufficient. Since there
+         are so many different fonts, the general measures are used and a note regarding thin or unusual
+         fonts is included if the definition for <a>large-scale</a> text.
```
mitchellevan, line 40:
```suggestion
         fonts is included in the definition for <a>large-scale</a> text.
```

understanding/20/contrast-minimum.html
```diff
@@ -173,33 +164,16 @@ <h3>Rationale for the Ratios Chosen</h3>
                low contrast, and therefore we allow and encourage authors to provide mechanisms to
                adjust the foreground and background colors of the content. Some of the combinations
                that could be chosen may have contrast levels that will be lower than those found
-               in the Success Criteria. This is not a violation of this Success Criterion provided
-               there is a mechanism that will return to the default values set out in the Success
-               Criteria.
-               
+               in the Success Criteria. This is not a violation of this Success Criterion, provided
+               there is a mechanism that will return to the default values set out here.
```
mitchellevan, line 168:
```suggestion
               there is a mechanism that will return to the required values set out here.
```

mitchellevan (commented):
(1) Typo. (2) The required values might not be "default" values if there is a conforming alternate version.

understanding/20/contrast-enhanced.html
patrickhlauke: good spot, thanks

understanding/20/contrast-enhanced.html
patrickhlauke: sure, makes sense. tweaked it a bit further, for better readability.

Myndex:
> _in the context of these two SCs in WCAG 2.x, smoothing and antialiasing were never part of the testing considerations, and that's not going to change._

Hi @patrickhlauke 

Hi Patrick I just want you to know that I do appreciate you taking the extra time to reply to me, and point out where I misinterpret various aspects of these documents. Thank you.

understanding/20/contrast-enhanced.html
```diff
@@ -96,15 +96,32 @@ <h2>Intent of Contrast (Enhanced)</h2>
          in order to get a particular look. 
       </p>
       
-      <p>Although this Success Criterion only applies to text, similar issues occur for content presented in charts, graphs, diagrams, and other non-text-based information which is covered by 
-         <a href="../21/non-text-contrast.html">Success Criterion 1.4.11 Non-Text Contrast</a>.         
+      <div class="note">
+         
+         <p>Images of text do not scale as well as text because they tend to pixelate. It is also
+            harder to change foreground and background contrast and color combinations for images
+            of text, which is necessary for some users. Therefore, we suggest using text wherever
```
mbgower, line 103: It's more than a suggestion; it's a requirement of https://www.w3.org/WAI/WCAG21/Understanding/images-of-text.html


```suggestion
            of text, which is necessary for some users. Therefore, WCAG requires using text wherever
```

understanding/20/contrast-enhanced.html
```diff
@@ -96,15 +96,32 @@ <h2>Intent of Contrast (Enhanced)</h2>
...
+            possible, and when not, consider supplying an image of higher resolution.
```
mbgower, line 104:
```suggestion
            possible (see <a href="images-of-text">1.4.5: Images of Text</a>, and when not, consider supplying an image of higher resolution.
```

dbjorge: ed: missing closing paren in the suggestion

PR #3024, opened 2023-02-14
Rewrite SCR37 as an HTML technique

fstrr:
- ARIA APG covers non-HTML modal dialogs, so we don’t need to repeat all of that content here.
- SCR37 was very out of date
- HTML `dialog` element has cross-browser support now
- SCR37 and working example deleted.
- note: I couldn’t find any HTML working examples, do didn’t create one for this.

Here's a [githack preview of the page](https://raw.githack.com/w3c/wcag/re-write-scr37-as-html-technique/techniques/html/H-creating-dialogs-with-the-html-dialog-element.html)


techniques/html/H-creating-dialogs-with-the-html-dialog-element.html
```diff
@@ -0,0 +1,215 @@
...
+});</code></pre>
+         
+      </section>
+   </section><section id="tests"><h2>Tests</h2>
+      <section class="procedure"><h3>Procedure</h3>
+         <ol>
+            <li>Find the components on the page that trigger modal <code class="language-html">dialog</code> elements.</li>
+            <li>Check that each <code class="language-html">dialog</code> can be opened using the keyboard to activate the triggering element (for example: activate a <code class="language-html">button</code> by pressing the <kbd>Enter</kbd> key or <kbd>Space Bar</kbd>.</li>
+            <li>When the <code class="language-html">dialog</code> is opened, check that focus is placed inside it.</li>
+            <li>While the <code class="language-html">dialog</code> is open, check that focus is trapped inside the <code class="language-html">dialog</code>.</li>
```
scottaohara, line 176: this should be changed to rather state that focus does not return to elements of the primary document.  Focus is not entirely trapped in an HTML dialog element, after all, as focus can leave and access the browser chrome.  So, this wording might confuse people into think this expected behavior is a bug.

fstrr: Thanks for that feedback. I thought I had that in there to start with, but apparently I hadn't. I've updated the text.

fstrr:
[Rawgit preview if that's useful](https://raw.githack.com/w3c/wcag/re-write-scr37-as-html-technique/techniques/html/H-creating-dialogs-with-the-html-dialog-element.html).

techniques/html/H-creating-dialogs-with-the-html-dialog-element.html
scottaohara: looks like this change wasn't made.  
```suggestion
            <li>While the <code class="language-html">dialog</code> is open, check that keyboard focus cannot move to elements of the primary document.</li>
```

techniques/html/H-creating-dialogs-with-the-html-dialog-element.html
```diff
@@ -0,0 +1,215 @@
...
+   </section><section id="examples"><h2>Examples</h2>
+      <section class="example">
+         <h3>A dialog to sign up to a mailing list</h3>
+				 <p>This is an example of using a modal <code class="language-html">dialog</code> element to show a mailing-list sign-up form to a user. The main part of the page contains a <code class="language-html">button</code> element that, when activated, triggers the appearance of the modal <code class="language-html">dialog</code>. The <code class="language-html">button</code> uses the <code class="language-html">aria-controls</code> property to create a programmatic connection between the <code class="language-html">button</code> and the <code class="language-html">dialog</code>; the <code class="language-html">aria-haspopup</code> property, set to <code class="language-html">dialog</code>, to tell assistive technology that the <code class="language-html">button</code> controls a <code class="language-html">dialog</code>, and the <code class="language-html">type</code> attribute to tell the browser that this isn't a <code class="language-html">submit</code> <code class="language-html">button</code>.</p>
```
scottaohara, line 32: I'm not sure this technique should be mentioning either aria-haspopup or aria-controls.  

see note about limiting unnecessary use of `aria-haspopup` in ARIA 1.2 - https://www.w3.org/TR/wai-aria-1.2/#issue-container-generatedID-57   but without a visual design, i'm assuming that based on the markup, there would be no visual indicator that the button would cause a 'popup'.  I'm indifferent to the attribute staying so long as it's prefaced that it's _optionally_ added.

re: aria-controls, there's really no value in the attribute being here, both for users or for UA, since the browser will handle focus of the dialog element - any benefit that aria-controls could have had is negated by the default focus management.   I think this one should be removed.

fstrr: thanks for that feedback—I hadn't seen the ARIA 1.2 note on `aria-haspopup`. I removed `aria-controls` as well.

scottaohara:
thank you @fstrr !

scottaohara (approved):

techniques/html/H-creating-dialogs-with-the-html-dialog-element.html
```diff
@@ -0,0 +1,212 @@
+<!DOCTYPE html>
+<html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml">
+	<head>
+		<title>Creating Modal Dialogs With The HTML Dialog Element</title>
+		<link rel="stylesheet" type="text/css" href="../../css/editors.css" class="remove"></link>
+   </head>
+	 <body>
+		<h1>Creating Modal Dialogs With The <abbr title="HyperText Markup Language">HTML</abbr> <code class="language-html">dialog</code> Element</h1>
```
mbgower, line 8:
```suggestion
		<h1>Creating modal dialogs with the <abbr title="HyperText Markup Language">HTML</abbr> <code class="language-html">dialog</code> Element</h1>
```

techniques/html/H-creating-dialogs-with-the-html-dialog-element.html
```diff
@@ -0,0 +1,212 @@
+<!DOCTYPE html>
+<html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml">
+	<head>
+		<title>Creating Modal Dialogs With The HTML Dialog Element</title>
```
mbgower, line 4:
```suggestion
		<title>Creating modal dialogs with the HTML dialog element</title>
```

techniques/html/H-creating-dialogs-with-the-html-dialog-element.html
```diff
@@ -0,0 +1,212 @@
...
+        <li>with no extra code, the underlying page is hidden from assistive technology when the modal <code class="language-html">dialog</code> is open;</li>
+        <li>the use of the <kbd>Escape</kbd> key to close the modal <code class="language-html">dialog</code>.</li>
+      </ul>
+      <p>This technique uses the HTML <code class="language-html">dialog</code> element rather than the ARIA equivalent.</p>
+   </section><section id="examples"><h2>Examples</h2>
+      <section class="example">
+         <h3>A dialog to sign up to a mailing list</h3>
+				 <p>This is an example of using a modal <code class="language-html">dialog</code> element to show a mailing-list sign-up form to a user. The main part of the page contains a <code class="language-html">button</code> element that, when activated, triggers the appearance of the modal <code class="language-html">dialog</code>. The <code class="language-html">button</code> uses the the <code class="language-html">type</code> attribute to tell the browser that this isn't a <code class="language-html">submit</code> <code class="language-html">button</code>.</p>
```
mbgower, line 32:
```suggestion
				 <p>This is an example of using a modal <code class="language-html">dialog</code> element to show a mailing-list sign-up form to a user. The main part of the page contains a <code class="language-html">button</code> element that, when activated, triggers the appearance of the modal <code class="language-html">dialog</code>. The <code class="language-html">button</code> uses the <code class="language-html">type</code> attribute to tell the browser that this isn't a <code class="language-html">submit</code> <code class="language-html">button</code>.</p>
```

techniques/html/H-creating-dialogs-with-the-html-dialog-element.html
```diff
@@ -0,0 +1,212 @@
...
...browser will hide all of the content on the underlying page from assistive technology. For example: a screen reader will not announce any of the content. Additionally, focusing with the keyboard is constrained to the the <code class="language-html">dialog</code> element's focusable content and the browser's controls. The browser takes care of focus management for us by putting the focus onto the first focusable element in the dialog's <abbr title="Document Object Model">DOM</abbr>, which in this demo is the <code class="language-html">h1</code> element because it has a <code class="language-html">tabindex="-1"</code> attribute, making it focusable by scripts. Note that, although the <code class="language-html">dialog</code>'s close <code class="language-html">button</code> is visibly before the <code class="language-html">h1</code>, it is the last item in the <code class="language-html">dialog</code>'s DOM. If the <code class="language-html">button</code> was first, it would receive focus when dialog was opened, which could result in the <code class="language-html">dialog</code> being closed by mistake if the <code class="language-html">button</code> was accidentally activated.</p>
```
mbgower, line 34:
```suggestion
				 <p>When the modal <code class="language-html">dialog</code> is opened, the browser will hide all of the content on the underlying page from assistive technology. For example: a screen reader will not announce any of the content. Additionally, focusing with the keyboard is constrained to the <code class="language-html">dialog</code> element's focusable content and the browser's controls. The browser takes care of focus management for us by putting the focus onto the first focusable element in the dialog's <abbr title="Document Object Model">DOM</abbr>, which in this demo is the <code class="language-html">h1</code> element because it has a <code class="language-html">tabindex="-1"</code> attribute, making it focusable by scripts. Note that, although the <code class="language-html">dialog</code>'s Close <code class="language-html">button</code> is visibly before the <code class="language-html">h1</code>, it is the last item in the <code class="language-html">dialog</code>'s DOM. If the <code class="language-html">button</code> was first, it would receive focus when dialog was opened, which could result in the <code class="language-html">dialog</code> being closed by mistake if the <code class="language-html">button</code> was accidentally activated.</p>
```

techniques/html/H-creating-dialogs-with-the-html-dialog-element.html
```diff
@@ -0,0 +1,212 @@
...
...the browser will hide all of the content on the underlying page from assistive technology. For example: a screen reader will not announce any of the content. Additionally, focusing with the keyboard is constrained to the <code class="language-html">dialog</code> element's focusable content and the browser's controls. The browser takes care of focus management for us by putting the focus onto the first focusable element in the dialog's <abbr title="Document Object Model">DOM</abbr>, which in this demo is the <code class="language-html">h1</code> element because it has a <code class="language-html">tabindex="-1"</code> attribute, making it focusable by scripts. Note that, although the <code class="language-html">dialog</code>'s Close <code class="language-html">button</code> is visibly before the <code class="language-html">h1</code>, it is the last item in the <code class="language-html">dialog</code>'s DOM. If the <code class="language-html">button</code> was first, it would receive focus when dialog was opened, which could result in the <code class="language-html">dialog</code> being closed by mistake if the <code class="language-html">button</code> was accidentally activated.</p>
```
mbgower, line 34:
```suggestion
				 <p>When the modal <code class="language-html">dialog</code> is opened, the browser will hide all of the content on the underlying page from assistive technology. For example, a screen reader will not announce any of the content. Additionally, focusing with the keyboard is constrained to the <code class="language-html">dialog</code> element's focusable content and the browser's controls. The browser takes care of focus management for us by putting the focus onto the first focusable element in the dialog's <abbr title="Document Object Model">DOM</abbr>, which in this demo is the <code class="language-html">h1</code> element because it has a <code class="language-html">tabindex="-1"</code> attribute, making it focusable by scripts. Note that, although the <code class="language-html">dialog</code>'s Close <code class="language-html">button</code> is visibly before the <code class="language-html">h1</code>, it is the last item in the <code class="language-html">dialog</code>'s DOM. If the <code class="language-html">button</code> was first, it would receive focus when dialog was opened, which could result in the <code class="language-html">dialog</code> being closed by mistake if the <code class="language-html">button</code> was accidentally activated.</p>
```

techniques/html/H-creating-dialogs-with-the-html-dialog-element.html
```diff
@@ -0,0 +1,212 @@
...
+    d.close();
+  }, false);
+});</code></pre>
+         
+      </section>
+   </section><section id="tests"><h2>Tests</h2>
+      <section class="procedure"><h3>Procedure</h3>
+         <ol>
+            <li>Find the components on the page that trigger modal <code class="language-html">dialog</code> elements.</li>
+            <li>Check that each <code class="language-html">dialog</code> can be opened using the keyboard to activate the triggering element (for example: activate a <code class="language-html">button</code> by pressing the <kbd>Enter</kbd> key or <kbd>Space Bar</kbd>.</li>
```
mbgower, line 171:
```suggestion
            <li>Check that each <code class="language-html">dialog</code> can be opened using the keyboard to activate the triggering element (for example, activate a <code class="language-html">button</code> by pressing the <kbd>Enter</kbd> key or <kbd>Space Bar</kbd>.</li>
```

techniques/html/H-creating-dialogs-with-the-html-dialog-element.html
```diff
@@ -0,0 +1,212 @@
...
...the browser will hide all of the content on the underlying page from assistive technology. For example, a screen reader will not announce any of the content. Additionally, focusing with the keyboard is constrained to the <code class="language-html">dialog</code> element's focusable content and the browser's controls. The browser takes care of focus management for us by putting the focus onto the first focusable element in the dialog's <abbr title="Document Object Model">DOM</abbr>, which in this demo is the <code class="language-html">h1</code> element because it has a <code class="language-html">tabindex="-1"</code> attribute, making it focusable by scripts. Note that, although the <code class="language-html">dialog</code>'s Close <code class="language-html">button</code> is visibly before the <code class="language-html">h1</code>, it is the last item in the <code class="language-html">dialog</code>'s DOM. If the <code class="language-html">button</code> was first, it would receive focus when dialog was opened, which could result in the <code class="language-html">dialog</code> being closed by mistake if the <code class="language-html">button</code> was accidentally activated.</p>
```
mbgower, line 34:
```suggestion
				 <p>When the modal <code class="language-html">dialog</code> is opened, the browser will hide all of the content on the underlying page from assistive technology. For example, a screen reader will not announce any of the content. Additionally, focusing with the keyboard is constrained to the <code class="language-html">dialog</code> element's focusable content and the browser's controls. The browser takes care of focus management for us by putting the focus onto the first focusable element in the dialog's <abbr title="Document Object Model">DOM</abbr>, which in this demo is the <code class="language-html">h1</code> element because it has a <code class="language-html">tabindex="-1"</code> attribute, making it focusable by scripts. Note that, although the <code class="language-html">dialog</code>'s close <code class="language-html">button</code> is visibly before the <code class="language-html">h1</code>, it is the last item in the <code class="language-html">dialog</code>'s DOM. If the <code class="language-html">button</code> was first, it would receive focus when dialog was opened, which could result in the <code class="language-html">dialog</code> being closed by mistake if the <code class="language-html">button</code> was accidentally activated.</p>
```

techniques/html/H-creating-dialogs-with-the-html-dialog-element.html
```diff
@@ -0,0 +1,212 @@
...
+    d.close();
+  }, false);
+});</code></pre>
+         
+      </section>
+   </section><section id="tests"><h2>Tests</h2>
+      <section class="procedure"><h3>Procedure</h3>
+         <ol>
+            <li>Find the components on the page that trigger modal <code class="language-html">dialog</code> elements.</li>
+            <li>Check that each <code class="language-html">dialog</code> can be opened using the keyboard to activate the triggering element (for example, activate a <code class="language-html">button</code> by pressing the <kbd>Enter</kbd> key or <kbd>Space Bar</kbd>.</li>
```
mbgower, line 171:
```suggestion
            <li>Check that each <code class="language-html">dialog</code> can be opened using the keyboard to activate the triggering element (for example, activate a <code class="language-html">button</code> by pressing the <kbd>Enter</kbd> key or <kbd>Spacebar</kbd>.</li>
```

techniques/html/H-creating-dialogs-with-the-html-dialog-element.html
```diff
@@ -0,0 +1,212 @@
...
+  background:#fff;
+  border:1px solid hsl(0, 0%, 50%);
+  border-radius:3px;
+  color:inherit;
+  font:inherit;
+  margin:0;
+}</code></pre>
+     
+		<h4>The JavaScript</h4>
+		<p>The script toggles the display of the <code class="language-html">dialog</code>. The HTML <code class="language-html">dialog</code> element can be opened using two different commands: <code class="language-javascript">show()</code> (for non-modal <code class="language-html">dialog</code>s), and  <code class="language-javascript">showModal()</code> (for modal <code class="language-html">dialog</code>s).</p>
```
scottaohara, line 150:
```suggestion
		<p>The script is necessary to display the <code class="language-html">dialog</code> when invoked. The HTML <code class="language-html">dialog</code> element can be opened using two different commands: <code class="language-javascript">show()</code> (for non-modal <code class="language-html">dialog</code>s), and  <code class="language-javascript">showModal()</code> (for modal <code class="language-html">dialog</code>s) - used in this example.</p>
		 <p><strong>Note:</strong> instead of writing a function to close the <code class="language-html">dialog</code>, the close button is nested within a separate <code class="language-html">form</code> element with a <code class="language-html">method="dialog"</code> attribute. This method allows the dialog to be closed when activating this button, and the browser will handle returning keyboard focus to the invoking element, </p>
```

techniques/html/H-creating-dialogs-with-the-html-dialog-element.html
```diff
@@ -0,0 +1,212 @@
...
+  const d = document.querySelector("dialog");
+  const btnOpen = document.querySelector(".open-modal");
+  const btnClose = document.querySelector(".close-modal");
+
+  btnOpen.addEventListener("click", function(){
+    d.showModal();
+  }, false);
+
+  btnClose.addEventListener("click", function(){
+    d.close();
```
scottaohara, line 162:
```suggestion
```

techniques/html/H-creating-dialogs-with-the-html-dialog-element.html
```diff
@@ -0,0 +1,212 @@
...
+  margin:0;
+}</code></pre>
+     
+		<h4>The JavaScript</h4>
+		<p>The script toggles the display of the <code class="language-html">dialog</code>. The HTML <code class="language-html">dialog</code> element can be opened using two different commands: <code class="language-javascript">show()</code> (for non-modal <code class="language-html">dialog</code>s), and  <code class="language-javascript">showModal()</code> (for modal <code class="language-html">dialog</code>s).</p>
+         
+<pre xml:space="preserve"><code class="language-javascript">document.addEventListener("DOMContentLoaded", function(e){
+  const d = document.querySelector("dialog");
+  const btnOpen = document.querySelector(".open-modal");
+  const btnClose = document.querySelector(".close-modal");
```
scottaohara, line 155:
```suggestion
```

techniques/html/H-creating-dialogs-with-the-html-dialog-element.html
```diff
@@ -0,0 +1,212 @@
...
+  btnClose.addEventListener("click", function(){
+    d.close();
+  }, false);
+});</code></pre>
+         
+      </section>
+   </section><section id="tests"><h2>Tests</h2>
+      <section class="procedure"><h3>Procedure</h3>
+         <ol>
+            <li>Find the components on the page that trigger modal <code class="language-html">dialog</code> elements.</li>
```
scottaohara, line 170:
```suggestion
            <li>Find the components on the page that invoke modal <code class="language-html">dialog</code> elements.</li>
```

techniques/html/H-creating-dialogs-with-the-html-dialog-element.html
```diff
@@ -0,0 +1,212 @@
...
+            <li>Check that each <code class="language-html">dialog</code> can be opened using the keyboard to activate the triggering element (for example, activate a <code class="language-html">button</code> by pressing the <kbd>Enter</kbd> key or <kbd>Spacebar</kbd>.</li>
```
scottaohara, line 171:
```suggestion
            <li>Check that each <code class="language-html">dialog</code> can be opened using the keyboard to activate the invoking element (for example, activate a <code class="language-html">button</code> by pressing the <kbd>Enter</kbd> key or <kbd>Spacebar</kbd>.</li>
```

techniques/html/H-creating-dialogs-with-the-html-dialog-element.html
```diff
@@ -0,0 +1,212 @@
...
+            <li>When the <code class="language-html">dialog</code> is opened, check that focus is placed inside it.</li>
```
scottaohara, line 172:
```suggestion
            <li>When the modal <code class="language-html">dialog</code> is opened, check that focus has moved to it / one of its focusable descendants.</li>
```

techniques/html/H-creating-dialogs-with-the-html-dialog-element.html
```diff
@@ -0,0 +1,212 @@
...
+            <li>While the <code class="language-html">dialog</code> is open, check that keyboard focus cannot move to elements of the primary document.</li>
```
scottaohara, line 173:
```suggestion
            <li>While the modal <code class="language-html">dialog</code> is open, check that keyboard focus cannot move to elements of the primary document.</li>
```

techniques/html/H-creating-dialogs-with-the-html-dialog-element.html
```diff
@@ -0,0 +1,212 @@
...
+						<li>When the <code class="language-html">dialog</code> is closed, check that focus is placed back onto the triggering element, if that is still on the page.</li>
```
scottaohara, line 174:
```suggestion
						<li>When the modal <code class="language-html">dialog</code> is closed, check that focus is placed back onto the invoking element, if the element still exists on the page.</li>
```

scottaohara (commented):
my previous approving review still stands.  My suggeted edits here are largely editoral nits / clarifying things further and swapping out some language to allude to other conepts - such as 'inert', or current work happening to implement "invokers" into HTML - setting up vocabulary parity for when that concept/method becomes available.

techniques/html/H-creating-dialogs-with-the-html-dialog-element.html
```diff
@@ -0,0 +1,212 @@
...
+			<h2>When to Use</h2>
+      <p>HTML</p>
+		</section>
+		<section id="description">
+			<h2>Description</h2>
+			<p>Site designers often want to create modal dialogs to get users to focus on a task-related activity outside of the rest of the page's content. Although this can be accomplished using <a href="https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/">ARIA to create a modal dialog</a>, the HTML <code class="language-html">dialog</code> element can also be used. As well as meeting <a href="https://www.w3.org/TR/using-aria/#rule1">the first rule of ARIA</a>, the HTML <code class="language-html">dialog</code> element offers several advantages to its ARIA equivalent, with the browser handling these features:</p>
+      <ul>
+        <li>focus management into the newly-opened modal <code class="language-html">dialog</code>;</li>
+        <li>focus management back onto the triggering element (assuming the trigger is still on the page) when the modal <code class="language-html">dialog</code> is closed;</li>
+        <li>focus is constrained to the contents of the modal <code class="language-html">dialog</code> and the browser's chrome;</li>
```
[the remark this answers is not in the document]
fstrr: thanks for the comment and suggested edit. I agree with Scott that adding a description of what browser chrome is, is the way to go.

alastc:
Bring the [email conversation](https://lists.w3.org/Archives/Public/w3c-wai-gl/2023OctDec/thread.html#msg43) to github, Wilco wrote:

> I'm happy with this as an advisory technique. I think as a sufficient technique this is confusing. The part of SC37 that made it directly relevant to SC 2.4.3 was test 3 "Check that, once opened, the dialog is next in the tab order.". In the rewrite, that part was replaced with a test that is no longer about the order in which elements are in sequential focus navigation. "When the dialog is opened, check that focus is placed inside it.". 
> I support that change. That is a better way to do this, but it removed the only thing in SC37 that was about the order in which elements appear when a user navigates the page with tab focus. 

I replied:
> If I understand your perspective, is it that: The opening of the model does not impact the focus order, so it is orthogonal to the SC?
> 
> I think others (including me) have a slightly wider interpretation. The SC doesn’t say the tab-order, it’s about the focus order.
> 
> If you open a dialogue, the best way to preserve meaning and operability is to do what the new technique says (i.e. move the focus to the dialog).
> 
> It can apply when you use a button which moves the focus, not just tab order. In that way dynamic interfaces can be covered.
>
> As a sufficient technique it isn’t the *only* way to do it, you could use content-ordering rather than focus management, but this is *a* way to meet it.

@WilcoFiers - Does that resolve things, or should we discuss this async?

mbgower:
@WilcoFiers making sure you saw the comment above

WilcoFiers:
@mbgower If the groups thinks this is the best way to go I can live with it. I would caution that it seems to me most people visiting these pages don't know the difference between normative and informative, and don't know that there are things these documents say need to be tested / or don't allow in the tests, that mismatch the normative text.

In this particular example, the previous technique allowed the modal to be the next focusable element, whereas the updated one requires focus to be moved into/onto the modal. I think that matters. Similarly it seems to me that if the focus didn't move back to the original trigger, but to the element after the trigger that would be an acceptable way to build a modal.

mbgower:
Thanks @WilcoFiers.
> I would caution that it seems to me most people visiting these pages don't know the difference between normative and informative, and don't know that there are things these documents say need to be tested / or don't allow in the tests, that mismatch the normative text.

That seems like an overall concern with the construction of sufficient techniques in WCAG 2, since none of them are normative, nor are they prescriptive (except to the degree that if an author meets the test, they can be confident they have used a technique the WG felt met the requirement). I've always thought of the sufficient techniques as in-the-moment attempts to document best practices. As such, it makes sense that some will need tweaking or even fall out of relevance or favour over the years.

> In this particular example, the previous technique allowed the modal to be the next focusable element, whereas the updated one requires focus to be moved into/onto the modal. I think that matters. Similarly it seems to me that if the focus didn't move back to the original trigger, but to the element after the trigger that would be an acceptable way to build a modal.

As you know, the existing SCR37 never represented the only way to create a dialog, merely one sufficient way. However, I think most people would agree that its test for acceptable is a far inferior user experience to that provided in the new technique, which closely matches that in the [ARIA Authoring Practices dialog pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/) and of course is predominantly the default behaviour built into HTML5.

alastc:
I know we've looked at this before (quite a long time ago), but perhaps (in a separate issue) we should tweak the standard text above techniques?

Current:
> Each numbered item in this section represents a technique or combination of techniques that the WCAG Working Group deems sufficient for meeting this Success Criterion. However, it is not necessary to use these particular techniques. For information on using other techniques, see [Understanding Techniques for WCAG Success Criteria](https://www.w3.org/WAI/WCAG22/Understanding/understanding-techniques), particularly the "Other Techniques" section.

Suggestion (change in bold):
> Each numbered item in this section represents a technique or combination of techniques that the WCAG Working Group deems sufficient for meeting this Success Criterion. **The technique may go beyond the minimum requirement of the criterion. There may other ways of meeting the criterion not covered by techniques.** For information on using other techniques, see [Understanding Techniques for WCAG Success Criteria](https://www.w3.org/WAI/WCAG22/Understanding/understanding-techniques), particularly the "Other Techniques" section.

If that seems reasonable (e.g. thumbs up) I'll open something up separately.

bruce-usab (approved):

PR #3027, opened 2023-02-15
Example of content overlay

alastc:


understanding/22/target-size-minimum.html
```diff
@@ -65,12 +65,16 @@ <h2>Intent of Target Size (Minimum)</h2>
...
+          <li>interacting with a menu that shows a dropdown of options, partially obscuring targets below</li>
           <li>activating a button displays a modal dialog</li>
           <li>content displays a cookie banner after page load</li>
           <li>content displays a "Take a survey" dialog after some period of user inactivity</li>
-        </ul>    
+        </ul>
         <p>The requirement does however apply to targets in any new content that appears on top of other content.</p>
+        <figure id="target-overlays">
+            <img src="img/target-size-overlaps.jpg" alt="A menu showing a drop-down that overlaps targets underneath, partially obscuring one." />
+            <figcaption>Four rows of targets, illustrating three ways of meeting this Success Criterion and one way of failing it.</figcaption>
```
dbjorge, line 76: It looks like there might be a copy-paste error in the caption

mbgower, line 76: I think this is a more appropriate caption for the image
```suggestion
            <figcaption>Target size requirements do not apply when a target is obscured as a result of user action, such as opening a menu.</figcaption>
```

alastc: Thanks, Michael updated it.

Issue #1385, opened 2020-09-04
Does the Understanding Doc for 1.4.11: Non-text Contrast (try to) exempt authors from conforming to 2.4.7 Focus Visible when relying on default UA focus styles?


yatil:
In the [Understanding for 1.4.11: Non-text Contrast](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html), one of the examples says:

> **Default focus style** 
> Links are required to have a focus indicator by 2.4.7 Focus Visible. Where the focus style of the user-agent is not adjusted on interactive controls (such as links, form fields or buttons), the default focus style is sufficient.

1.4.11 has an exemption of contrast for unmodified user agent, but 2.4.7 has not.

While that (to my understanding) means that that there is no 1.4.11 3:1 contrast requirement, it seems still insufficient to have an invisible/imperceivable focus style even when relying on the UA style. So if the color of the focus outline of the UA is black it would not conform on a black background as it is not visible.

It feels that the non-normative claim “Where the focus style of the user-agent is not adjusted on interactive controls (such as links, form fields or buttons), the default focus style is sufficient.” is not substantiated by the actual normative text of 2.4.7.

This section in the understanding should be reflected to address this, alternatively an exception should be added to 2.4.7.

(See also this discussion on Twitter: https://twitter.com/stommepoes/status/1301813234585014274)

JAWS-test:
2.4.7 requires a visible focus indicator, even if only the standard focus of the browser is used (see https://www.w3.org/WAI/WCAG21/Techniques/failures/F78).
2.4.7 does not require a minimum contrast, however.

1.4.11 requires a minimum contrast, but only for focus indicators that are not from the browser.

This is indeed somewhat confusing, but in itself not a contradiction, because it means that a browser focus must be visible, but not necessarily meet the 3:1 contrast.

I'm not sure if anything needs to be adjusted here. Soon WCAG 2.2 will be released and with that there will be 2 new SCs for focus which will clarify the problem.

yatil:
The main issue, again, here is that it says 

> Links are required to have a focus indicator by 2.4.7 Focus Visible. Where the focus style of the user-agent is not adjusted on interactive controls (such as links, form fields or buttons), the default focus style is sufficient.

The sentences in conjunction indicate that they need to have a focus indicator but, if I is the UA default, that UA default is sufficient.

One way to clarify that is:

[...]

PR #3030, opened 2023-02-16
"Not modified by the author" does not exempt 2.4.7 Focus Visible

mitchellevan:
Address this aspect of #1385:

> 1.4.11 has an exemption of contrast for unmodified user agent, but 2.4.7 has not.
> 
> While that (to my understanding) means that that there is no 1.4.11 3:1 contrast requirement, it seems still insufficient to have an invisible/imperceivable focus style even when relying on the UA style. So if the color of the focus outline of the UA is black it would not conform on a black background as it is not visible.

### DIFF
--- a/understanding/21/non-text-contrast.html
+++ b/understanding/21/non-text-contrast.html
@@ -90,7 +90,7 @@
 
 				<h4 id="related-focus">Relationship with Focus Visible</h4>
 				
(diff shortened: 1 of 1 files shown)


mitchellevan:
The 2.4.7 clarification was covered by #2146, so my PR is not needed.

PR #3031, opened 2023-02-16
Adding heading structure and size requirement

### LINKED ISSUE #2803: 2.5.8 Target Size (Minimum) is ambiguous for non-rectangular targets
#### Problem

The current wording of the main clause of 2.5.8 (before exceptions) is:

> The size of the [target](https://w3c.github.io/wcag/guidelines/22/#dfn-targets) for [pointer inputs](https://w3c.github.io/wcag/guidelines/22/#dfn-pointer-inputs) is at least 24 by 24 [CSS pixels](https://w3c.github.io/wcag/guidelines/22/#dfn-css-pixels)

For non-rectangular targets, this is ambiguous about how a target's "size" should be compared to the "24 by 24" requirement. The understanding doc doesn't help clarify this; every single example is for a rectangular target. Targets which are rounded rectangles (eg, via a CSS `border-radius` property) are extremely common in practice, and targets which are circles are also reasonably common - the SC should be unambiguous about how to determine if those cases are conformant, and ideally the understanding doc would contain examples.

Ideally, the SC would also be unambiguous about how to apply the rule to irregularly shaped targets (eg, targets with holes in them due to other targets being overlaid on top of them). For example, search controls are commonly targets which contain overlaid buttons, such as this example from google.com with "search by voice" and "search by image" buttons overlaid on top of a search text input:

screenshot of google.com search bar, a text input with two buttons overlaying its right side

#### Suggested solutions

My preference would be something like this, which removes the ambiguity for both rounded and irregular cases:

> The [target](https://w3c.github.io/wcag/guidelines/22/#dfn-targets) for [pointer inputs](https://w3c.github.io/wcag/guidelines/22/#dfn-pointer-inputs) completely overlaps at least one 24 [CSS pixel](https://w3c.github.io/wcag/guidelines/22/#dfn-css-pixels) diameter circle

Something like this is slightly simpler to understand and would still clarify the requirements for rounded/circular cases, but it incorrectly passes many cases involving targets with overlapping bounding rectangles that ought to fail:

> The size of the bounding rectangle of the [target](https://w3c.github.io/wcag/guidelines/22/#dfn-targets) for [pointer inputs](https://w3c.github.io/wcag/guidelines/22/#dfn-pointer-inputs) is at least 24 by 24 [CSS pixels](https://w3c.github.io/wcag/guidelines/22/#dfn-css-pixels)

Another consideration which is unambiguous even for irregular shapes might be something like:

> The area of the [target](https://w3c.github.io/wcag/guidelines/22/#dfn-targets) for [pointer inputs](https://w3c.github.io/wcag/guidelines/22/#dfn-pointer-inputs) is at least 450 square [CSS pixels](https://w3c.github.io/wcag/guidelines/22/#dfn-css-pixels)
>
> *Note: 450 square CSS pixels is very slightly less than the area of a 24 CSS pixel diameter circle*

Regardless of if/how the normative text is updated, it would be helpful to add examples for these cases to the understanding doc.

#### Context

This started as [a conversation with @bruce-usab in the review of #2798](https://github.com/w3c/wcag/pull/2798#discussion_r1026599540) - I broke it out to a separate issue since it was applicable to parts of the SC not changed by that PR.

alastc:
Closes #2803 
The new content is lines 67-88, the rest is making the indents uniform.


understanding/22/target-size-minimum.html
```diff
@@ -50,74 +50,104 @@ <h2>Intent of Target Size (Minimum)</h2>
...
+          </figure>
+
+          <p>The square does not have to be aligned with the page, it could be skewed.</p>
+
+          <figure id="basic-target-size-skewed">
+            <img src="img/target-size-requirement3.png" alt="A skewed rectangle that includes a 24 by 24px square within it." />
+            <figcaption>So long as there is a 24 by 24px square within the target it meets the first line of the criterion and passes.</figcaption>
+          </figure>
+
+          <p>If a target does not have a 24 by 25px square but does have space around it, check the spacing exception (below).</p>
```
dbjorge, line 83:
```suggestion
          <p>If a target does not encompass a 24 by 24px square but does have space around it, check the spacing exception (below).</p>
```

understanding/22/target-size-minimum.html
```diff
@@ -50,74 +50,104 @@ <h2>Intent of Target Size (Minimum)</h2>
...
+
+          <figure id="basic-target-size-circle">
+            <img src="img/target-size-requirement2.png" alt="A circle with a diameter of 24px, but not filling up a 24px square." />
+            <figcaption>The rounded corners of the circle do not fill the 24 by 24px square so does not pass the first line of the criterion. However, depending on the spacing to other targets it may pass by the exception.</figcaption>
```
dbjorge, line 87:
```suggestion
            <figcaption>The rounded corners of the circle do not fill the 24 by 24px square, so the circle does not pass the first line of the criterion. However, depending on the spacing to other targets it may pass by the Spacing exception.</figcaption>
```

understanding/22/target-size-minimum.html
```diff
@@ -50,74 +50,104 @@ <h2>Intent of Target Size (Minimum)</h2>
...
-        <p>Spacing, alone and in conjunction with size, can improve user experience. There is less chance of accidentally activating a neighboring target if a target is not immediately adjacent to another. Touchscreen devices and user agents generally have internal heuristics to identify which link or control is closest to a user's touch interaction - this means that sufficient spacing around a target can work as effectively as a larger target size itself.</p>
+          <figure id="basic-target-size">
+            <img src="img/target-size-requirement.png" alt="Three square boxes next to each other, each 24px wide." />
+            <figcaption>Where targets are a 24 by 24px square (and larger is better), they meet the first line of the criterion and pass.</figcaption>
+          </figure>
+
+          <p>The square does not have to be aligned with the page, it could be skewed.</p>
+
+          <figure id="basic-target-size-skewed">
+            <img src="img/target-size-requirement3.png" alt="A skewed rectangle that includes a 24 by 24px square within it." />
```
dbjorge, line 79: I think the alt text needs a little bit more detail to communicate the same meaning that it would to a visual user:

```suggestion
            <img src="img/target-size-requirement3.png" alt="A sheared rectangle that could not include a page-aligned 24 by 24px square within it but that does include a skewed 24 by 24px square within it." />
```

dbjorge:
I think the content you've added looks good (modulo the minor editorial suggestions above). However, I would really like to see an example of a target with a hole in it due to overlap with a different element; I think that's going to be more common in practice than a sheared rectangle (for example, https://google.com's main search bar is such a target), so it feels off to have an example covering rotation but not shapes with holes.

understanding/22/target-size-minimum.html
mbgower: "skew" is more in line with descriptions of geometric shapes. https://en.wikipedia.org/wiki/Skew_polygon When I hear "shear" I think of something that is cut off. I get what you're saying, but it implies to me that part of the 24x24 has been cut off.

understanding/22/target-size-minimum.html
```diff
@@ -50,74 +50,104 @@ <h2>Intent of Target Size (Minimum)</h2>
...
-        <ul>
-          <li>interacting with a combobox shows a dropdown list of suggestions</li>
-          <li>activating a button displays a modal dialog</li>
-          <li>content displays a cookie banner after page load</li>
-          <li>content displays a "Take a survey" dialog after some period of user inactivity</li>
-        </ul>    
-        <p>The requirement does however apply to targets in any new content that appears on top of other content.</p>
-        
-        <p>While the Success Criterion primarily helps touch users by providing target sizing to prevent accidental triggering of adjacent targets, it is also useful for mouse or pen users. It reduces the chances of erroneous activation due to either a tremor or reduced precision, whether because of reduced fine motor control or input imprecision.</p>
+          <p>For a target to be "24 by 24 CSS pixels", it must be possible to draw a 24 by 24 CSS pixel (px) square such that the square is completely within the target.</p>
```
mbgower, line 69: I get worried someone may read that "within" as needing to be inside the outer edge of the target, so...
```suggestion
          <p>For a target to be "24 by 24 CSS pixels", it must be possible to draw a 24 by 24 CSS pixel (px) square such that the square is completely within the target (does not extend outside the target's area).</p>
```

understanding/22/target-size-minimum.html
```diff
@@ -50,74 +50,104 @@ <h2>Intent of Target Size (Minimum)</h2>
...
+          <figure id="basic-target-size">
+            <img src="img/target-size-requirement.png" alt="Three square boxes next to each other, each 24px wide." />
+            <figcaption>Where targets are a 24 by 24px square (and larger is better), they meet the first line of the criterion and pass.</figcaption>
+          </figure>
+
+          <p>The square does not have to be aligned with the page, it could be skewed.</p>
+
+          <figure id="basic-target-size-skewed">
+            <img src="img/target-size-requirement3.png" alt="A skewed rectangle that includes a 24 by 24px square within it." />
+            <figcaption>So long as there is a 24 by 24px square within the target it meets the first line of the criterion and passes.</figcaption>
```
mbgower, line 80: Adding the word "solid" to avoid anyone interpreting that a hollow (or partially hollow) 24x24 outline would pass this...
```suggestion
            <figcaption>So long as there is a solid 24 by 24px square within the target, it meets the first line of the criterion and passes.</figcaption>
```

understanding/22/target-size-minimum.html
alastc: This is outdated, and updated with different text.

PR #3035, opened 2023-02-17
Editorial: Match terms spelling and add links

maryjom:
This definition uses other defined terms "down-event" and "up-event" introduced in WCAG 2.1. This edit adds in the hyphen in the spelling. It also adds links to the definitions of those terms.

### DIFF
--- a/guidelines/terms/22/dragging-movement.html
+++ b/guidelines/terms/22/dragging-movement.html
@@ -2,7 +2,7 @@
 <dd class="new">
 	<p class="change">New</p>
    					
-   <p>an operation where the pointer engages with an element on the down event and the element (or a representation of its position) follows the pointer until an up event</p>
+  <p>an operation where the pointer engages with an element on the <a>down-event</a> and the element (or a representation of its position) follows the pointer until an <a>up-event</a></p>
    
    <p class="note">Examples of draggable elements include list items, text elements, and images.</p>
 </dd>


detlevhfischer (commented):
Looks good to me.

guidelines/terms/22/dragging-movement.html
```diff
@@ -2,7 +2,7 @@
 <dd class="new">
 	<p class="change">New</p>
    					
-   <p>an operation where the pointer engages with an element on the down event and the element (or a representation of its position) follows the pointer until an up event</p>
+  <p>an operation where the pointer engages with an element on the <a>down-event</a> and the element (or a representation of its position) follows the pointer until an <a>up-event</a></p>
```
mbgower, line 5: Maybe these links automagically point to the defs? Otherwise, I've added in targets to the defs...
```suggestion
  <p>an operation where the pointer engages with an element on the <a href="https://www.w3.org/TR/WCAG22/#dfn-down-event">down-event</a> and the element (or a representation of its position) follows the pointer until an <a href="https://www.w3.org/TR/WCAG22/#dfn-up-event">up-event</a></p>
```

maryjom: @mbgower The full URL is not needed on definition references. I think that there are scripts or ReSpec magic that goes on to resolve to the actual URL when these types of definition link references are used.

alastc: Indeed, it's best not to so that it goes to the correct version of the definition (e.g. the 2.2 version from 2.2 etc).

PR #3038, opened 2023-02-20
Editorial suggestions for consistent links to definitions and capitalization

maryjom:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/pull/3038.html" title="Last updated on Nov 5, 2024, 8:12 PM UTC (ad93d52)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/3038/f8efc41...ad93d52.html" title="Last updated on Nov 5, 2024, 8:12 PM UTC (ad93d52)">Diff</a>


patrickhlauke (approved):

fstrr:
Those merge conflicts weren't fun.

kfranqueiro (approved):
Fixed a couple of leftover unresolved terms, and restored one change that probably got lost in the merge conflicts Francis mentioned.

alastc:
@wilco wrote on list:
> Several of the added use of the "text" definition seems wrong to me. The biggest problem is linking images of text to text. The a WCAG "text" is programmatically determined, whereas images of text seems to be explicitly for text that is not programmatically determined. That's a contradiction. Several of the other new uses of the definition of "text" also seem to change the success criteria in incorrect ways. The change of 3.3.1 Error identification seems to suggest an image of text are no longer allowed, even if all requirements of 1.4.5 Images of text are met. For 3.1.5 the change seems to suggest that images of text are exempt from the success criterion. And for 1.4.8 it would seem that justified text actually is allowed in images.

I responded:
> For the images of text definition links, I assume you mean this bit:
> https://github.com/w3c/wcag/pull/3038/files#diff-5992e3ead3ec31ca61641bf8fa29895aa0dce67d9d6408168124cbec88575672
> 
> I thought that was appropriate because it added the definition to the top part (which is “_text_ that has been rendered in a non-text form”).
> And it took away the current link from “text that is part of a picture that contains significant other visual content.”, which appears to align with your argument?
> 
> For error identification:
> https://github.com/w3c/wcag/pull/3038/files#diff-b9e2005d006c77d296e7e6a40d722a6ff1281ef5df93355821be06214554717c
> I had always interpreted that as text, not an image of text, even without the definition link. It probably hasn’t come up for me as an image of text  would fail the 1.4.5 SC anyway (I can’t think of a reasonable instance which would fall into the exceptions), and I don’t recall anyone trying!
> 
> For 3.1.5 Reading Level, again, I think it depends on whether you think adding the link changes the meaning. I had not, so whether it was linked or not, I’d read that as (non-image) text. 

Would be interested in whether other people perceive the link to add/change meaning to the text.

mraccess77:
I agree with Wilco that updating the word "text" to point to the definition which focus only on programmatically determined human language may have some unintended consequences removing some situations from scope.  While some of those would also fail SC 1.4.5 only obscures the the other issues.

dbjorge:
I think this is highlighting the existing issues with WCAG's use of the term "text", where sometimes it means "text" as it's defined in the glossary ("programmatically determined"), but other times it is almost certainly *intended* to mean text in a broader sense (eg, I think the *intent* of 3.1.5 Reading Level is that there's no reason an author should be exempted from that SC just because they rendered text as an image).

I think these cases are already problematic - I think a reasonable reader looking at an SC that uses the term "text" would be likely to use the glossary definition of "text" regardless of whether it's linked or not, but I think that mismatches our intent in some of the cases Wilco highlighted.

I'd lean towards avoiding linking "text" in the cases where a broader meaning is intended; I don't think just "not linking the definition" fully resolves the problem, but I think it's what makes sense for this PR. Longer term, I think this is good feedback for WCAG 3; it'd be good for us to have distinct glossary terms for the "programmatically determined text" vs "text in a more general sense" cases.

alastc:
I certainly agree with disambiguating this in WCAG 3.

As a slight tangent, when I'd read the definition:
> sequence of characters that can be [programmatically determined](https://www.w3.org/TR/WCAG22/#dfn-programmatically-determinable), where the sequence is expressing something in [human language](https://www.w3.org/TR/WCAG22/#dfn-human-language-s)

I did wonder if an image of text, with identical alt-text, would meet this definition... so it could be tightened in future as well.

bruce-usab:
For highlighting on AG call 9/17.  TF discussed, may need another review.

bruce-usab (approved):

bruce-usab:
Firstly, I want to express my appreciation for this editorial work.  Over the years, defined terms being used (but not hyperlinked) has caused more friction than I would have ever imagined.

> I'd lean towards avoiding linking "text" in the cases where a broader meaning is intended; I don't think just "not linking the definition" fully resolves the problem, but I think it's what makes sense for this PR.

+1 to both points.

Having re-reviewed, I agree with @dbjorge that for 3.1.5 Reading Level "text" could just as easily be "text or image of text" which argues for not adding the hyperlink in that case.  But IMHO it is okay either way.  The other hyperlinks to the defined term seem unambiguously correct (and helpfully clarifying).

mbgower:
Discussed on TF call. I'm doing a careful read through of the proposed changes, keeping comments in the PR in mind -- particularly in regard to linking (or not linking) to the definition of "text".
I will make suggestions in the PR, and where changes to text are proposed, I will specifically comment as well.

guidelines/sc/20/error-identification.html
mbgower, line 1: If this was left without the link to the definition, it's hard to anticipate it leading to a problem. Any authors using  images of text to convey error identification would need to meet both Images of Text _and_ Non-text Content. That is, they would need to justify why they didn't use text _and_ provide a text alternative to the image. I note that the definition of [Text Alternative](https://www.w3.org/TR/WCAG20/#text-altdef) already provides a link to "text", so the end result should be the same. There is some noise in the "text alternative" definition about "location" and "programmatically associated", but I haven't figured out how any scenario/interpretation would lead to a problem if we made this a link to definition of text. @WilcoFiers and @dbjorge, were you okay with this one? I can go either way.

guidelines/sc/20/focus-visible.html
mbgower, line 1: Focus indicator is a definition added in WCAG 2.2. So in 2.2 a link to "focus indicator" can exist, but the 2.1 (and 2.0) version of the same file cannot have that link since the definition doesn't exist. @kfranqueiro and @alastc can our model now support this?

guidelines/sc/20/language-of-parts.html
mbgower, line 1: I personally don't see any "win" by adding the link, but it doesn't seem to create a problem changing it either.

guidelines/sc/20/name-role-value.html
mbgower, line 1: It looks like there was no definition of "state" in 2.0; it's there in 2.1, but not marked as "New". @alastc I'm assuming that as long as this link exists in 2.1 and 2.2, we're okay, since we wouldn't be touching 2.0.

guidelines/sc/20/reading-level.html
```diff
@@ -4,7 +4,7 @@ <h4>Reading Level</h4>
    
    <p class="conformance-level">AAA</p>
    
-   <p>When text requires reading ability more advanced than the <a>lower secondary education level</a> after removal of proper names and titles, <a>supplemental content</a>, or a version that does not require reading ability more advanced than the lower
+  <p>When <a>text</a> requires reading ability more advanced than the <a>lower secondary education level</a> after removal of proper names and titles, <a>supplemental content</a>, or a version that does not require reading ability more advanced than the lower
```
mbgower, line 7: I agree that linking this to the definition of text has the potential to except something conveyed through an Image of Text from having its reading level assessed.

```suggestion
  <p>When text requires reading ability more advanced than the <a>lower secondary education level</a> after removal of proper names and titles, <a>supplemental content</a>, or a version that does not require reading ability more advanced than the lower
```

guidelines/sc/20/focus-visible.html
kfranqueiro: As long as this is only happening _within_ the `guidelines` folder I think we're okay; the plan for building 2.1 is to use a snapshot of the 2.1 TR so it won't be looking at that folder in the main branch (which contains 2.2-specific things).

If you discover anything like this directly within the source of Understanding pages, that would definitely be worth flagging because we would need to conditionalize it somehow.

guidelines/sc/20/visual-presentation.html
```diff
@@ -9,9 +9,9 @@ <h4>Visual Presentation</h4>
    <ul>
      <li>Foreground and background colors can be selected by the user.</li>   
      <li>Width is no more than 80 characters or glyphs (40 if CJK).</li>   
-     <li>Text is not justified (aligned to both the left and the right margins).</li>   
+     <li><a>Text</a> is not justified (aligned to both the left and the right margins).</li>
```
mbgower, line 12: Since there is an included change in this PR to include a link to the definition of text in "Block of text" (which is a defined term in the preamble of this SC) , it seems redundant and unnecessary to also link to the definition of text in a bullet of the same SC.
```suggestion
     <li>Text is not justified (aligned to both the left and the right margins).</li>   
```

guidelines/terms/20/blocks-of-text.html
mbgower, line 1: This seems appropriate. All uses of "block of text" or "blocks of text" work when considering this as programmatic text.

guidelines/terms/20/image-of-text.html
```diff
@@ -1,11 +1,11 @@
 <dt><dfn id="dfn-images-of-text" data-lt="images of text">image of text</dfn></dt>
 <dd>
    
-   <p>text that has been rendered in a non-text form (e.g., an image) in order to achieve
+  <p><a>text</a> that has been rendered in a non-text form (e.g., an image) in order to achieve
```
mbgower, line 4: Making this a link to programmatic text seems contradictory to me. If the text has been rendered as non-text, it cannot meet our definition of text

```suggestion
  <p>text that has been rendered in a non-text form (e.g., an image) in order to achieve
```

guidelines/terms/20/image-of-text.html
```diff
@@ -1,11 +1,11 @@
...
       a particular visual effect
    </p>
    
-   <p class="note">This does not include <a>text</a> that is part of a picture that contains significant other visual content.
+   <p class="note">This does not include text that is part of a picture that contains significant other visual content.
```
mbgower, line 8: I agree this link to the definition should be removed.

mbgower (changes requested):
I've made  suggested changes (rejecting the proposed link to the definition) in a few of the files. I'd like those discussed. Otherwise, looks good.

guidelines/sc/20/visual-presentation.html
bruce-usab: Images of text could use full justification (and the other features listed in 1.4.8 Visual Presentation might be achieved using images of text) -- so I agree this use should not link to the defined term.

guidelines/sc/20/error-identification.html
```diff
@@ -5,7 +5,7 @@ <h4>Error Identification</h4>
    <p class="conformance-level">A</p>
    
    <p>If an <a>input error</a> is automatically detected, the item that is in error is identified and the error
-      is described to the user in text.
+     is described to the user in <a>text</a>.
```
mbgower, line 8:
```suggestion
     is described to the user in text.
```

guidelines/sc/20/error-identification.html
```diff
@@ -5,7 +5,7 @@ <h4>Error Identification</h4>
    <p class="conformance-level">A</p>
    
    <p>If an <a>input error</a> is automatically detected, the item that is in error is identified and the error
-      is described to the user in text.
+     is described to the user in text.
```
mbgower, line 8:
```suggestion
      is described to the user in text.
```

mbgower (approved):

mbgower:
Reviewed and finalized on Sep 20 TF call

guidelines/terms/20/blocks-of-text.html
```diff
@@ -1,6 +1,6 @@
-<dt><dfn id="dfn-blocks-of-text">blocks of text</dfn></dt>
+<dt><dfn id="dfn-blocks-of-text" data-lt="block of text">blocks of text</dfn></dt>
 <dd>
    
-   <p>more than one sentence of text</p>
+  <p>more than one sentence of <a>text</a></p>
```
WilcoFiers, line 4: I've reported this before. This is a mistake. WCAG's "text" definition refers to programmatic text, whereas the dictionary meaning of "text" also covers images of text. Connecting "blocks of text" to WCAG's definition of text means blocks of text no longer cover images of text. That then means 1.4.8, which is the only SC using blocks of text, now no longer applies to images of text. That seems like a mistake to me, and a significant normative change.

There are still plenty of non-programmatic texts on the web. Anything rendered on canvas is effectively an image of text, and that's also the exact scenario where reflow can't be taken for granted. This is not a good change IMO.

guidelines/terms/22/minimum-bounding-box.html
```diff
@@ -1,6 +1,6 @@
 <dt class="new"><dfn id="dfn-bounding-boxes" data-lt="bounding boxes|bounding box">minimum bounding box</dfn></dt>
 <dd class="new">
 	<p class="change">New</p>	
-   <p>the smallest enclosing rectangle aligned to the horizontal axis within which all the points of a shape lie. For components which wrap onto multiple lines as part of a sentence or block of text (such as hypertext links), the bounding box is based on how the component would appear on a single line.</p>
+   <p>the smallest enclosing rectangle aligned to the horizontal axis within which all the points of a shape lie. For components which wrap onto multiple lines as part of a sentence or <a>block of text</a> (such as hypertext links), the bounding box is based on how the component would appear on a single line.</p>
```
WilcoFiers, line 4: There is no "block of text" definition. Looking at the code, it doesn't seem like respec is going to understand that this needs to go to "blocks of text". You'll have to adjust this.

alastc: There is a definition, including the singular: https://github.com/w3c/wcag/blob/mjm-wcag-editorial-suggestions/guidelines/terms/20/blocks-of-text.html

mbgower: Given your thumbs up, @WilcoFiers I'm marking this as resolved. If you consider it still unresolved, please add another comment.

guidelines/terms/20/blocks-of-text.html
mbgower: @WilcoFiers prior feedback from you and otherson these changes was reviewed and incorporated by the TF. For instance, the link to the definition of text has been removed from both versions of Images of Text, for the reason you cite.

If you examine the criteria specified in 1.4.8, including the four "requirements" specified in the techniques, this AAA criterion is clearly targetted at text that can be manipulated, including:
- text and background colors altered
- width is restricted to a certain number of characters
- the ability to increase text size without scrolling horizontally

And of course there is another criterion which effectively rules out the use of images of text at AAA. Given these two considerations, amongst others, it was felt appropriate and clarifying to include the link to the definition in 1.48, just as it was considered clarifying to remove it in Images of Text (both 1.4.5 and 1.4.9).

dbjorge: > If you examine the criteria specified in 1.4.8, including the four "requirements" specified in the techniques, this AAA criterion is clearly targetted at text that can be manipulated via a mechansim

The point of the criteria is specifically to forbid techniques that would break the ability to manipulate text. Techniques that prevent the text from being programmatically manipulable are *exactly what this criterion needs to be able to target*. I agree with Wilco that we should omit the link from this specific case.

mbgower:
```suggestion
  <p>more than one sentence of text</p>
```

gundulaniemann:
Concering definition of 'block of text':
Well, in some contexts one sentence can cover a whole book page. It is typical for German literature written around 100-150 years ago, for example.
So should we stick the the rather grammatical concept of 'one sentence'?

bruce-usab:
Discussed on TF call 11/8.

> Concerning definition of 'block of text'

@gundulaniemann, I think your example reinforces the change Mike made to _not_ linking to the defined term in the one instance.

> So should we stick the rather grammatical concept of 'one sentence'?

Please raise this as a separate issue.  Of course, you are well of the reluctance to touch normative text!  :-)

kfranqueiro:
Note: While syncing this PR against main, the change to single-pointer was effectively obsoleted by #4070. Whereas this PR had expanded "an input" to "a pointer input" with "pointer input" linking to its definition, #4070 replaced that phrase with "an input modality".

PR #3042, opened 2023-02-20
Remove 'steps', add show-hide component

### LINKED ISSUE #2805: Does SC 3.3.9 Redundant Entry require the data available for the user to select available in the current step if the other steps are on the same page?
Does SC 3.3.9 Redundant Entry require data available for the user to select be available with in the current step if the other steps are on the same page?  In addition, if the data is available but hidden behind an expand widget - is that a failure?

alastc:
closes #2805

### DIFF
--- a/understanding/22/redundant-entry.html
+++ b/understanding/22/redundant-entry.html
@@ -35,11 +35,11 @@
 
       <h2>Intent of Redundant Entry</h2>
 
-        <p>The intent of this Success Criterion is to ensure that users can successfully navigate multi-step processes. It reduces cognitive effort where  information is asked for more than once during steps in a process. It also reduces the need to recall information provided in a previous  step.</p>
+        <p>The intent of this Success Criterion is to ensure that users can successfully navigate multi-step processes. It reduces cognitive effort where  information is asked for more than once during a process. It also reduces the need to recall information provided in a previous  step.</p>
         
         <p>Information that is required to be remembered for input can pose a significant barrier to users with cognitive or memory difficulties. All users experience a natural gradual mental fatigue as they proceed through steps in a process. This fatigue is accelerated by the stress of recalling information from short-term working memory. Users with learning, and cognitive disabilities are highly susceptible to mental fatigue.</p>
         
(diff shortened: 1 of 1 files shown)


understanding/22/redundant-entry.html
```diff
@@ -35,11 +35,11 @@ <h2>Redundant Entry Success Criteria text</h2>
...
         <p>Information that is required to be remembered for input can pose a significant barrier to users with cognitive or memory difficulties. All users experience a natural gradual mental fatigue as they proceed through steps in a process. This fatigue is accelerated by the stress of recalling information from short-term working memory. Users with learning, and cognitive disabilities are highly susceptible to mental fatigue.</p>
         
-        <p>Requiring people to recall information already entered in the previous steps can cause them to give up or enter incorrect information. The autocomplete feature of browsers is not considered sufficient because it is the content (the web site) that needs to provide the stored information for a redundant entry, or avoid asking for the same information again.</p>
+        <p>Requiring people to recall information already entered in the previously can cause them to give up or enter incorrect information. The autocomplete feature of browsers is not considered sufficient because it is the content (the web site) that needs to provide the stored information for a redundant entry, or avoid asking for the same information again.</p>
```
mbgower, line 42:
```suggestion
        <p>Requiring people to recall information previously entered can cause them to give up or enter incorrect information. The autocomplete feature of browsers is not considered sufficient because it is the content (the web site) that needs to provide the stored information for a redundant entry, or avoid asking for the same information again.</p>
```

understanding/22/redundant-entry.html
```diff
@@ -51,6 +51,8 @@ <h2>Intent of Redundant Entry</h2>
             <li>tick a checkbox to populate inputs with the same values as previously entered (e.g., my billing address is the same as my shipping address).</li>
         </ul>
 
+        <p>The location of the data "available to select" is evaluated by the Web page, so it needs to be on the same page. Ideally, it would be visible by default and closely associated with the input where the data is required. However, it could be elsewhere on a page, including within a show/hide component.</p>
```
alastc, line 54:
```suggestion
        <p>Data which is "available to select" would need to be on the same page. Ideally, it would be visible by default and closely associated with the input where the data is required. However, it could be elsewhere on a page, including within a show/hide component.</p>
```

PR #3044, opened 2023-02-20
Updating objects bit

### LINKED ISSUE #2850: Change wording on 3.3.7 Accessible Authentication's exceptions
The wording of the third exception:
"Object Recognition The cognitive function test is to recognize objects.", [on this page](https://www.w3.org/WAI/WCAG22/Understanding/accessible-authentication), is a little confusing. Recently I encountered a reCAPTCHA which required me click the squares of all horses facing left, my first and only animal-based reCAPTCHA to date.

Given that reCAPTCHA appears to be introducing living things, which are not objects. the wording here should be amended to include recognition of non-objects, such as animals.

Suggestion:

"Recognition The cognitive function test is to recognize objects, animals or other visual information, such as shapes or patterns"

Or words to that effect.

I added shapes and patterns for those tests where we are required to type the characters, within an image, which are arguable not objects either and going forward, we may be required to identify circles or triangles etc.

alastc:
New line is 92

Closes #2850


alastc:
Approved in the survey: https://www.w3.org/2002/09/wbs/35422/wcag22-misc5/results#xq18

PR #3046, opened 2023-02-20
Adding to copy-paste

### LINKED ISSUE #2866: Accessible Authentication - one-time-passcodes and the test for when an activity requires 'transcription'
Working through Accessible Authentication, as an auditor I see the possibility for the question of 'cross-device copying' raised by clients who don't want to overhaul their current 2FA practices:

- a time-sensitive one-time password generated by an authenticator app can be copied and emailed or air-dropped across devices, but it is highly unreasonable to expect someone to do so within the standard limit of 30 seconds to 1 minute, and so the likelihood of transcription here is so high that I can confidently say this method _requires_ transcription for most who would try it
- however, what about a code sent via text that has, say, a 15 or even 30 minute window of viability? I can say there is some relative likelihood of transcription occurring there, and so is that the bar? That if there is a necessary time limit (for example), then it's forseeable that transcription could be a required option for someone?

I notice in the Understanding section for Accessible Authentication that copying and pasting such that it mitigates transcription is only referenced in the context of the action being performed locally:

> Copy and paste can be relied on to avoid transcription. Users can copy their login credentials from a **local source**...

Is the implicit intention that the value of the field requiring input, if it is being pasted, must also be sourced from the same device, so as to significantly reduce the likelihood of transcription due to cross-device complexity (which includes things like mandatory time-limits and the general some-user-unfriendliness of this approach)? 

I'm just wondering how this argument might be countered (providing others also belive it to be viable/I'm not misreading the success criteria).

Edit: some further hunting has found a similar discussion resolved here: https://github.com/w3c/wcag/issues/1359, however, I wonder if clarity around both this and the second example above (around copying and pasting between multiple devices with constraints of ambiguous impact) might be beneficial if made more explicit in the SC?

alastc:
Closes #2866


patrickhlauke:
x-ref https://github.com/w3c/wcag/pull/2618

understanding/22/accessible-authentication.html
```diff
@@ -86,6 +86,8 @@ <h2>Intent of Accessible Authentication</h2>
...
         <p>Copy and paste can be relied on to avoid transcription. Users can copy their login credentials from a local source (such as a standalone third-party password manager) and paste it into the username and password fields on a login form, or into a web-based command line interfaces asking for a password. Blocking people from pasting into authentication fields, or using a different format between the copied text and the input field (for example, "Enter the 3rd, 4th, and 6th character of your password"), would force the user to transcribe information and therefore fail this criterion, unless another method is available.</p>
 
+        <p>If the copy and paste requires the user to have and use multiple devices (e.g., copy code from phone, email to yourself, open on laptop) copying multiple times, then it would still be considered a Cognitive Function Test. Some systems do allow for simple copy and paste across devices (e.g., tap code on phone, paste on laptop), which do not involve copying multiple times, but you would need all users to have easy access to that for it to be considered <a href="#dfn-accessibility-supported">accessibility supported</a>.</p>
```
alastc, line 89:
```suggestion
        <p>If the copy and paste requires the user to have and use multiple devices (e.g., copy code from phone, email to yourself, open on laptop) copying multiple times, then it would still be considered a Cognitive Function Test and fail this criterion. Some systems do allow for simple copy and paste across devices (e.g., tap code on phone, paste on laptop), which do not involve copying multiple times, but you would need all users to have easy access to that for it to be considered <a href="#dfn-accessibility-supported">accessibility supported</a>.</p>
```

PR #3047, opened 2023-02-20
Richard's suggestion

### LINKED ISSUE #2996: "in a language displayed top to bottom" -> "in vertically set text"
Success Criterion 2.5.8 Target Size (Minimum)
https://www.w3.org/TR/2023/CR-WCAG22-20230124/#target-size-minimum (note)

> For example, in a language displayed top to bottom, the line-height would be horizontal.

Editorial: Arabic language in vertically set text is displayed bottom to top, but is still relevant for this guidance.  I suggest using something like:

"For example, in a language displayed vertically, the line-height would be horizontal."

This would also cover the cases where a language is displayed bottom to top, such as setting English with `writing-mode:sideways-lr`.

Screenshot 2023-01-30 at 15 15 47

Screenshot 2023-01-30 at 15 26 33

alastc:
Closes #2996


alastc:
[Approved today](https://www.w3.org/2023/03/07-ag-minutes.html#t12).

PR #3049, opened 2023-02-21
Updating example 2 in prog determined

### LINKED ISSUE #3001: Programmatically Determined example describes outdated direct parsing of markup
## Currently

For reasons explained in 4.1.1 removal (#2525, #2797), the first example for "programmatically determined" is also outdated. Assistive technologies should no longer access markup directly.

https://github.com/w3c/wcag/blob/main/guidelines/terms/20/programmatically-determined.html

> EXAMPLE:
Determined in a markup language from elements and attributes that are accessed directly by commonly available assistive technology.

## Proposed

Is it possible to edit a non-normative example?

> EXAMPLE:
Determined in a markup language from elements and attributes that are exposed via an accessibility API to commonly available assistive technology.

alastc:
Closes #3001

### DIFF
--- a/guidelines/terms/20/programmatically-determined.html
+++ b/guidelines/terms/20/programmatically-determined.html
@@ -6,8 +6,7 @@
       <a>user agents</a>, including <a>assistive technologies</a>, can extract and present this information to users in different modalities
    </p>
    
-   <aside class="example"><p>Determined in a markup language from elements and attributes that are accessed directly
-      by commonly available assistive technology.
+   <aside class="example"><p>Determined in a markup language from elements and attributes that are exposed in the accessibility tree.
    </p></aside>
    
       <aside class="example"><p>Determined from technology-specific data structures in a non-markup language and exposed


guidelines/terms/20/programmatically-determined.html
```diff
@@ -6,8 +6,7 @@
       <a>user agents</a>, including <a>assistive technologies</a>, can extract and present this information to users in different modalities
    </p>
    
-   <aside class="example"><p>Determined in a markup language from elements and attributes that are accessed directly
-      by commonly available assistive technology.
+   <aside class="example"><p>Determined in a markup language from elements and attributes that are exposed in the accessibility tree.
```
mitchellevan, line 9:
```suggestion
   <aside class="example"><p>Determined in a markup language from elements and attributes that are accessed via the accessibility tree by commonly available assistive technology.
```
As I mentioned in a [comment](https://github.com/w3c/wcag/issues/3001#issuecomment-1452535712) today, I'd prefer to keep "commonly available assistive technology" in both examples, consistent with [5.2.4](https://www.w3.org/TR/WCAG/#cc4) and "[accessibility supported](https://www.w3.org/TR/WCAG/#dfn-accessibility-supported)."

patrickhlauke: i'd keep the aspect of whether something is in the accessibility tree, and whether or not AT then accesses/uses it for anything, separate

gundulaniemann: I prefer 'exposed', as it is rather about clean html instead of depending on abilities of existing technologies.

alastc:
Note that we did discuss this previously and had [resolved](https://www.w3.org/2023/04/11-ag-minutes#item09) to change the understanding document instead.

We can revisit, particularly with a new version, but we should look at updating the understanding doc.

An updated version shouldn't include term "accessibility tree".

patrickhlauke:
x-ref this was discussed back in April 2023 https://us02web.zoom.us/j/9880060302?pwd=Rmk1MUV2b253RE9tVnBJUkZJZ2QzUT09

PR #3065, opened 2023-02-25
fix link to Statement of Partial Conformance

luce-carevic:
I'm not very familiar with the architecture of the project so I don't know if it's the right way to fix the broken link to the Statement of Partial Conformance section.


alastc:
HI @luce-carevic,

We try not to hard-code links to particular versions. I think this one is due to the missing "Key terms" section which Michael needs to resolve, so I'll close this.

PR #3073, opened 2023-02-28
fix broken link to on-focus document

fstrr:


mbgower:
Not working across all versions of WCAG. 2.2 works but not 2.1 -- and likely not main. Some magic happening somewhere...

patrickhlauke:
for further context: the `#` relative link is needed when the SC is in the spec, but this PR seems to try and fix it for when the SC is pulled into understanding. understanding for 2.2 works (seems to have some hardcoded fix to add in the absolute url reference back to spec), but doesn't seem to work currently in 2.1

but fundamentally we don't want to change this in the SC itself, as it still needs to be the `#` relative in-page link for when it's in the spec.

kfranqueiro:
Seeing this PR for the first time, and confirming this should not be changed - it currently works as intended both within `guidelines/index.html` and within built understanding pages.

RE 2.1, this will be resolved once we get that rebuilt (which #4007 is a big step towards).

PR #3083, opened 2023-03-03
Update focus-not-obscured-minimum.html

### LINKED ISSUE #2751: 2.4.12 Focus not Obscured (Minimum) and user opened / controlled content

mbgower:
Attempt to add an additional section on considerations to address #2751 

Closes #2751

Preview of the [understanding doc for the AA version.](https://raw.githack.com/w3c/wcag/2751-FocusObscured-movable/understanding/22/focus-not-obscured-minimum.html)


understanding/22/focus-not-obscured-minimum.html
```diff
@@ -41,12 +41,30 @@ <h2>Intent of Focus Not Obscured</h2>
...
+         <p>A dialog-like overlay that does not take focus on appearance and does not either constrain interaction to the overlay or dismiss itself on loss of focus (thus allowing focus to exit into the content behind it) will be at risk of failing this SC, where it is positioned such that it can obscure other content.</p>
+       </section>
+       <section>
+         <h4>Persistence of temporarily overlayed information</h4>
+         <p>A number of components on the web disclose additional content (on activation or on focus) intended for immediate user interaction or information. These disclosures often overlay other content. Examples of such overlays are menu items, select element items, combobox lists (and other dropdown items), date picker calendars, and tooltips. The comnmon trait of all these disclosed overlays is that they are only expected to persist until either acted on or no longer the primary point of user interaction. If such overlay content persists after the user has 1) activated one of the disclosed items or 2) moved the focus away from the triggering item and the overlay, the overlay is at risk of failing this SC.</p>
```
mbgower, line 59:
```suggestion
         <p>A number of components on the web disclose additional content (on activation or on focus) intended for immediate user interaction or information. These disclosures often overlay other content. Examples of such overlays are menu items, select element items, combobox lists (and other dropdown items), date picker calendars, and tooltips. The common trait of all these disclosed overlays is that they are only expected to persist until either acted on or no longer the primary point of user interaction. If such overlay content persists after the user has 1) activated one of the disclosed items or 2) moved the focus away from the triggering item and the overlay, the overlay is at risk of failing this SC.</p>
```

understanding/22/focus-not-obscured-minimum.html
```diff
@@ -41,12 +41,30 @@ <h2>Intent of Focus Not Obscured</h2>
...
+       </section>
+</section>
 
     <section id="benefits">
       <h2>Benefits of Focus Not Obscured</h2>
       
       <ul>
-         li>Sighted users who rely on a keyboard interface to operate the page will be able to see the component which gets keyboard focus. Such users include those who rely on devices which use the keyboard interface, including speech input, sip-and-puff software, onscreen keyboards, scanning software, and a variety of assistive technologies and alternate keyboards.</li>
+         <li>Sighted users who rely on a keyboard interface to operate the page will be able to see the component which gets keyboard focus. Such users include those who rely on devices which use the keyboard interface, including speech input, sip-and-puff software, onscreen keyboards, scanning software, and a variety of assistive technologies and alternate keyboards.</li>
```
mbgower, line 67:
```suggestion
         <li>Sighted users who rely on a keyboard interface to operate the page will be able to see the component which gets keyboard focus. Such users include those who rely on a keyboard or on devices which use the keyboard interface, including speech input, sip-and-puff software, onscreen keyboards, scanning software, and a variety of assistive technologies and alternate keyboards.</li>
```

understanding/22/focus-not-obscured-minimum.html
```diff
@@ -41,12 +41,30 @@ <h2>Intent of Focus Not Obscured</h2>
...
+     
+       </section>
+       <section>
+         <h4>Modal dialogs versus overlays</h4>
+         <p>A properly constructed modal dialog will always pass this SC. Even if it appears directly on top of an item with focus, the dialog takes focus on appearance, and thus the item receiving focus -- the dialog or one of its components -- is visible. A properly constructed modal maintains that focus and prevents interaction outside the modal until it is dismissed.</p>
+         <p>A dialog-like overlay that does not take focus on appearance and does not either constrain interaction to the overlay or dismiss itself on loss of focus (thus allowing focus to exit into the content behind it) will be at risk of failing this SC, where it is positioned such that it can obscure other content.</p>
```
alastc, line 55:
```suggestion
         <p>A dialog-like overlay that does not take focus on appearance and does not either constrain interaction to the overlay or dismiss itself on loss of focus (thus allowing focus to exit into the content behind it) will be at risk of failing this SC, where it is positioned such that it can obscure other focusable items.</p>
```

understanding/22/focus-not-obscured-minimum.html
```diff
@@ -41,12 +41,44 @@ <h2>Intent of Focus Not Obscured</h2>
...
+
+         <ul>
+            <li><strong>When the additional content appears, it displaces existing content.</strong> An accordion is an example of this. When an accordion is opened, the disclosed content shifts existing content further down the page. Since the new content does not obscure existing content, it cannot obscure the item with focus.</li>
+            <li><strong>When the additional content appears, existing content reflows.</strong> The popout sidebar on the <a href="https://www.w3.org/TR/WCAG22/">WCAG standard</a> is an example of this pattern. When the side menu is activated, it discloses a new section of information along the left side of the page. The main content area is reduced horizontally to accommodate this, and the existing main content reflows to fit in the thinner space. As a result, there is no overlapping content between the two sections; the item receiving focus, whether in the left navigation or in the main content, will not be obscured by the other section.</li>
```
dbjorge, line 64: `The popout sidebar on the <a href="https://www.w3.org/TR/WCAG22/">WCAG standard</a> is an example of this pattern.` - are we so confident that the UX of this page won't change for the lifetime of this understanding document that we're comfortable having the text take a dependency on its behavior? Might be safer to omit the example.

understanding/22/focus-not-obscured-minimum.html
```diff
@@ -41,12 +41,44 @@ <h2>Intent of Focus Not Obscured</h2>
...
+       </section>
+       <section>
+         <h4>Non-persistent disclosed information</h4>
+         <p>A number of components on the web disclose additional content (on activation or on focus) intended for immediate user interaction or information. These disclosures often open on top of other content, obscuring it. Examples of such disclosures are menu items, select element items, combobox lists (and other dropdown items), date picker calendars, and tooltips. The common trait of all these disclosures is that they are only expected to persist until either acted on or no longer the primary point of user interaction. Such non-persistent disclosures should not fail this SC since they don't obscure the item with focus. However, if an author allows such disclosures to persist after the user has 1) activated one of the disclosed items or 2) moved the focus away from the triggering item and the disclosure, the additional content is at risk of failing this SC by obscuring the item with focus.</p>
```
alastc, line 73:
```suggestion
         <p>A number of components on the web disclose additional content (on activation or on focus) intended for immediate user interaction or information. These disclosures often open on top of other content, obscuring it. Examples of such disclosures are menu items, select element items, combobox lists (and other dropdown items), date picker calendars, and tooltips. The common trait of all these disclosures is that they are only expected to persist until either acted on or no longer the primary point of user interaction. Such non-persistent disclosures do not fail this SC so long as they don't obscure the item with focus. However, if an author allows such disclosures to persist after the user has 1) activated one of the disclosed items or 2) moved the focus away from the triggering item and the disclosure, the additional content is at risk of failing this SC by obscuring the item with focus.</p>
```

understanding/22/focus-not-obscured-minimum.html
```diff
@@ -56,12 +56,21 @@ <h4>Modal dialogs</h4>
...
+            <li><strong>The disclosure expands into an area of the page containing no other content.</strong> Many pages are designed with wide margins, providing significant white space into which new content can be disclosed. Many chatbots and toasters are designed to 'slide up' into the right unpopulated side of a page. Where authors are careful to ensure content is not obscured at each breakpoint in a response design, no obscuring of other operable content need occur.</li>
+            <li><strong>When focus leaves the additional content, it is hidden or collapsed.</strong> This is very similar to patterns discussed next under Non-persistent disclosed information. A distinguishing factor can be that the user's last point of interaction in the disclosure is preserved (it persists) even though it may be hidden until a user returns. Some trees and left navigation pattterns behave this way. </li>
+         </ul>
+         <p>Where disclosed content persists and causes existing content to be obscured, it will be at risk of failing this criterion if it prevents users from seeing the focus indicator.</p>
```
scottaohara: what about disclosures that persist/overlay other content, and allow keyboard focus to leave the disclosure, but allow click and tap outside of the disclosure content, or esc key to dismiss the disclosure content at any point that it is open?   

the persistent disclosure that traps focus but allows outside clicks is mentioned, but not this variant where outside clicks/taps or esc key would dismiss the overlaying disclosure.

understanding/22/focus-not-obscured-minimum.html
```diff
@@ -56,12 +56,21 @@ <h4>Modal dialogs</h4>
...
        </section>
        <section>
          <h4>Non-persistent disclosed information</h4>
```
scottaohara, line 72: in the intro to this doc, non-persistent disclosures aren't mentioned, but get quite a bit written about them.  maybe worth alluding to?
>Typical types of content that can overlap focused items are sticky footers, sticky headers, and non-modal dialogs. As a user tabs through the page, these layers of content can obscure the item receiving focus, along with its focus indicator.

scottaohara (commented):
just a couple of things, but particularly interested in getting clarity for the disclosure ux i mentioned.

understanding/22/focus-not-obscured-minimum.html
mbgower: @scottaohara , I'd like to explore wording that mirrors some of the behaviour required in On Hover or Focus to try to allow for such an interaction mechanism. Something like:

> Where content disclosed by the user can obscure other content, if the user can reveal the item with focus without moving keyboard focus (for example, by pressing Esc), such disclosed content is not considered to fully obscure the item receiving focus.

scottaohara: i think that's on the right track.  Thank you!  That'd definitely alleviate my concerns if that sort of text was added

understanding/22/focus-not-obscured-minimum.html
mbgower: I'd like to have techniques that illustrate these. In the interim, especially when the SC is new, I think linking to a live site is useful.

guidelines/sc/22/focus-not-obscured-minimum.html
```diff
@@ -7,5 +7,7 @@ <h4>Focus Not Obscured (Minimum)</h4>
     
     <p>When a <a>user interface component</a> receives keyboard focus, the component is not entirely hidden due to author-created content.</p>
 
-    <p class="note">If the interface is configurable so that the user can reposition content such as toolbars and non-modal dialogs, then only the initial positions of user-movable content would be considered for testing and conformance of this Success Criterion.</p>
+    <p class="note">Where content in a configurable interface can be repositioned by the user, then only the initial positions of user-movable content is considered for testing and conformance of this Success Criterion.</p>
+
+    <p class="note">Where content disclosed by the <em>user</em> can obscure other content, if the user can reveal the item with focus without moving keyboard focus (for example, by dismissing with Esc), such disclosed content is not considered to fully obscure the item receiving focus.</p>
```
bruce-usab, line 12: @mbgower thanks for making this two notes.  I do not think "disclosed" is a great word to use this paragraph.  My suggestion:

Where user interaction can cause new content to obscure other content, if the user can reveal the item with focus without moving keyboard focus (for example, by dismissing with <kbd>Esc</kbd>), that new content is not considered to be fully obscuring the item receiving focus.

mbgower: Thanks, @bruce-usab I'm concerned that the suggested text "Where user interaction can cause new content to obscure other content..." may be a little broad. We want to 'except' content the user directly opens. We don't want this to be interpreted as meaning that anything that occurs as a result of any user interaction (e.g., tabbing through the UI, populating a user input, answering a prompt, etc) could lead to a loss of focus. Obviously if the user isn't aware an 'interaction' of theirs has potentially obscured focus, they won't know to 'uncover' it.
Other suggestions welcome!

alastc: I agree that the term "disclose" is odd in an interface context. How about "opens", or "reveals"?

bruce-usab: @mbgower I concede that my phrasing could be too broad.  I was trying to introduce as few new words as possible, and make the exception a little more active voice.  "Reveals" seems better (more generic) to me than "opens"..

understanding/22/focus-not-obscured-minimum.html
```diff
@@ -21,7 +21,9 @@ <h2>Focus Not Obscured Success Criterion text</h2>
         <blockquote>
             <p>When a <a>user interface component</a> receives keyboard focus, the component is not entirely hidden due to author-created content.</p>
         </blockquote>
-        <p class="note">If the interface is configurable so that the user can reposition content such as toolbars and non-modal dialogs, then only the initial positions of user-movable content would be considered for testing and conformance of this Success Criterion.</p>
+        <p class="note">Where content in a configurable interface can be repositioned by the user, then only the initial positions of user-movable content is considered for testing and conformance of this Success Criterion.</p>
+
+        <p class="note">Where content disclosed by the <em>user</em> can obscure other content, if the user can reveal the item with focus without moving keyboard focus (for example, by dismissing with Esc), such disclosed content is not considered to fully obscure the item receiving focus.</p>
```
alastc, line 26:
```suggestion
        <p class="note">Content opened by the <em>user</em> may obscure the component receiving focus. If the user can reveal the focused component without advancing the keyboard focus, the component with focus is not considered hidden due to author-created content.</p>
```

alastc:
Approved: https://www.w3.org/2023/04/25-ag-minutes#item06

melaniephilipp:
@alastc Should this content be visible in the Understanding document: https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-minimum.html?

melaniephilipp:
@mbgower Should this content be visible in the Understanding document: https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-minimum.html ? This is referenced as closing my issue #2751, but it isn't appearing in the WCAG 2.2 CR?

patrickhlauke:
it's been merged into https://github.com/w3c/wcag/tree/FocusObscured-suggested-changes but not the main branch, which is why that's not showing up on the published version

melaniephilipp:
Yup, yup, I see that, hence my question about "should" it be visible in the Understanding doc. That is the intention, yes?

alastc:
It should have been, but the branching got confused, I've added this as the first question for today's meeting (WCAG 2.x wise) to check everyone's happy with that merge.

PR #3058, opened 2023-02-24
Better diagram examples for 2.5.8 Target Size (Minimum)

patrickhlauke:
UPDATE: obviously, do not merge this if https://github.com/w3c/wcag/pull/3089 goes through. we'll need a replacement PR to update the understanding for the new circle-based approach altogether, and this here will be obsolete anyway.

* replaces first spacing figure with a tweaked SVG version to scale (with link to open original SVG to zoom into)
* replaces last two spacing figures (for the "telescoping spacing" aspect) with a single, hopefully clearer and more real-world, example (also as SVG with link to original)
* make note about users with different disabilities an actual `class="note"`
* change the `<ol>` in the "Exceptions" section to `<ul>` - it's unordered in the normative SC wording, so why make it a numbered list here?

RawGit preview: https://raw.githack.com/w3c/wcag/patrickhlauke-target-size-minimum-issue2755/understanding/22/target-size-minimum.html

Closes https://github.com/w3c/wcag/issues/2755

PR #3089, opened 2023-03-06
Spacing clause update

### LINKED ISSUE #3050: 2.5.8 Target Size "overlapping any other target" does not work
The latest CR draft introduced this wording in 2.5.8 Target Size:

> Spacing: The target does not overlap any other target 

I think this wording mixes "targets" and "user interface components". While UICs can often be stacked, a target is simply whatever region of the display responds to a pointer event. If the UIC is partially obscured, it reduces the target size, it isn't "overlapped". That's important, because we don't want people claiming a target passes because its 24x24 in size, while most of it is obscured.

The reason "overlapping" was added I think, was to address my H4 example, which feels like it shouldn't pass, but in the previous definition it did:
https://codepen.io/wilcofiers/full/abZxPow

Screenshot of codepen with overlapping elements

The problem though, even if we ignore that it's the UICs that overlap, not the target, is that the element that passes but shouldn't is not actually the element that is "**overlapping**". The element is "**overlapped**", it has things on top of it, but it isn't overlapping other UICs. 

So the problem that this change in wording tried to solve isn't solved at all.

alastc:
Closes #3050

### DIFF
--- a/guidelines/sc/22/target-size-minimum.html
+++ b/guidelines/sc/22/target-size-minimum.html
@@ -7,7 +7,7 @@
     
     <p>The size of the <a>target</a> for <a>pointer inputs</a> is at least 24 by 24 <a>CSS pixels</a>, except where:</p>
     <ul>
-        <li><strong>Spacing:</strong> The target does not overlap any other target and has a <a>target offset</a> of at least 24 CSS pixels to every adjacent target;</li>
+        <li><strong>Spacing:</strong> Targets are spaced such that if 24 CSS pixel diameter circles are centered on each target, none of the circles intersect another circle or target.</li>
         <li><strong>Equivalent:</strong> The function can be achieved through a different control on the same page that meets this criterion.</li>
         <li><strong>Inline:</strong> The target is in a sentence, or is in a bulleted or numbered list, or its size is otherwise constrained by the line-height of non-target text;</li>
         <li><strong>User agent control:</strong> The size of the target is determined by the user agent and is not modified by the author;</li>


patrickhlauke:
as mentioned in the survey, this seems to be quite a fundamental change to something that's been laboured over for years...and will also require a whole change to the related understanding document / testing approaches / etc. suggest it's worth bundling all of that together in the same PR here, to make clear what the impact/change proposed here actually is in practice

JAWS-test:
> Targets are spaced such that if 24 CSS pixel diameter circles are **centered** on each target, none of the circles intersect another circle or target

Think the new spacing exception is much better than the previous one. But think that the "centered" should be removed because it excludes certain well clickable targets, like e.g.
* Circle with 20 px diameter and circular ring with thickness of 20 px and diameter of 100 px. 
* Two bars (20 px wide and 100 px long) arranged as in a plus or x sign.

In both cases, a centered circle of the two targets would be identical because both objects have the same center, but the click area is perfectly adequate. 

If desired, I can make corresponding illustrations for my two examples.

guidelines/sc/22/target-size-minimum.html
```diff
@@ -7,7 +7,7 @@ <h4>Target Size (Minimum)</h4>
     
     <p>The size of the <a>target</a> for <a>pointer inputs</a> is at least 24 by 24 <a>CSS pixels</a>, except where:</p>
     <ul>
-        <li><strong>Spacing:</strong> The target does not overlap any other target and has a <a>target offset</a> of at least 24 CSS pixels to every adjacent target;</li>
+        <li><strong>Spacing:</strong> Targets are spaced such that if 24 CSS pixel diameter circles are centered on each target, none of the circles intersect another circle or target.</li>
```
mbgower, line 10: I think this scans a bit better
```suggestion
        <li><strong>Spacing:</strong> Targets are spaced such that if a 24 CSS pixel diameter circle is centered on each target, none of the circles intersect another circle or target.</li>
```

mbgower: I also think that we can make a small addition to take care of an edge case.

```suggestion
        <li><strong>Spacing:</strong> Targets are spaced such that if a 24 CSS pixel diameter circle is centered on each undersized target, none of the circles intersect another circle or target.</li>
```

JAWS-test: @mbgower Which edge case should make it necessary to insert "undersized"? My examples are not.

patrickhlauke:
It's a side conversation, but I'd still be interested to know what has changed now that makes us explore the idea of circles, when that same idea was shot down years ago when I floated it (see my mention here, but I'm certain I had that discussion about circles even well before then: https://lists.w3.org/Archives/Public/w3c-wai-gl/2021JanMar/0132.html) for having gaps/problems.

Speaking of the understanding document that will also need to be updated for this fundamental change: if somebody can provide me just with a rough sketch, I can work on getting some clean SVGs done in the same style as https://github.com/w3c/wcag/pull/3058 (which I assume will now be obsolete work, which is a shame since it took me forever to get those just right and accurate)

guidelines/sc/22/target-size-minimum.html
alastc: @JAWS-test in the case of a large target with small target inside (with some spacing)  in the centre, their circles will overlap. Assuming the large target passes the primary line of the SC, we don't need to worry that it's circle overlaps the smaller target's circle.

JAWS-test: @alastc My examples were such that both targets are not large enough (see above)

mbgower: @JAWS-test please provide visuals, as offered. Your '2 bars' examples sounds to me like it's actually 3 or 4 targets, not two, and all of them seem to pass.

mbgower: For the tiny circle inside a big thin donut example, I can see a way to word this to cover this edge case, but it would begin to get pretty cumbersome ('centered within' or something like that that makes it clear that the center point has to actually be part of the target). As far as I can tell, the only odd scenario left with this wording is the one you mentioned -- a small target centered inside another small target with sufficient spacing such that risk of hitting the wrong one isn't increased. But I personally have no issue telling the owner of such a designer that they should just make their targets big enough. It's not a common pattern; the new wording makes common patterns much easier to assess (with understandable outcomes).

patrickhlauke:
@mbgower 

> So, first, do you agree that what we're proposing here is closer to that vision? We get to toss out the whole cumbersome definition of [offset](https://w3c.github.io/wcag/guidelines/22/#dfn-target-offsets), for one thing.
> [...]
> I hope you'll agree that with a few illustrations, this new wording should be much easier to convey to users, while not actually changing the basic goal you were intoning in that old list.

oh, for sure. target offset was a weird and cumbersome neologism to start with, and the concept that it tried to convey was cumbersome in how it was trying to square the circle (hah) of "we can't really define a true minimum size, so we'll try and lump size and spacing together somehow"

> To me, what has changed is that the language of the SC got firmed up [...]

Sure, to be clear, I was just trying to work out what led us to 2+ years of noodling around with the concept of target offset and trying to make it work, with much energy being expended on it...

guidelines/sc/22/target-size-minimum.html
JAWS-test: @mbgower I have added the two figures above. The question is why the circle must be centered? In the example with the "radio button" (inner circle, circular ring) the strange thing is that it forbids the variant with the best click area, but allows if the inner circle is moved closer to the circular ring (see figure 3), although this is harder to click

Figure 3:

target3

mbgower: @JAWS-test 
>  The question is why the circle must be centered?

We've already gone into that exploration before. If you remove the centered notion, you immediately end up with odd allowed edge cases too, not to mention the whole thing is a _lot_ more fiddly to test. 
As stated in other threads, it's really useful to look at these things at scale. I encourage you to draw these things at scale and ask yourself whether any of those are good designs.

JAWS-test: @mbgower 

> ask yourself whether any of those are good designs

I don't think that's the question, at least not for me as a WCAG tester. As a WCAG tester, I wonder much more why example 1 fails and example 3 passes, even though 1 is better than 3.

detlevhfischer (commented):
What I find odd is the bit "another circle or target". Why "or"? Don't target and circle belong together always and are just one unit for purposes of measurement?  Should this read instead "none of the circles intersect a circle on any other target"?

patrickhlauke:
> Don't target and circle belong together always and are just one unit for purposes of measurement? Should this read instead "none of the circles intersect a circle on any other target"?

@detlevhfischer but with your suggested rewrite, you're *just* looking at the circle, whereas the original wants you to also check the other targets themselves, which I *think* is the correct approach here (otherwise a 1x1 target could be right next to a large 150x150 target, and it would be fine because the first target's circle centered on that single pixel would not overlap the circle centered on the large target and be valid)

detlevhfischer:
@patrickhlauke yes, I see, point taken.

patrickhlauke:
note that, unless i'm getting confused, the new circle approach removes that weird "telescoping spacing" quirk that I documented ages ago... so at least there's one less counterintuitive aspect

patrickhlauke:
heads-up that i'm working on an update for the understanding, including a clean (hopefully) diagram for the circle idea applied to the toolbar example. here's what I have so far...

target-spacing-toolbar

(this here is a PNG converted version of a proper SVG)

patrickhlauke:
work in progress on the companion understanding update https://github.com/w3c/wcag/pull/3103

JAWS-test:
@patrickhlauke I would remove the circles in the top line of the graphic, because the targets are 24x24 px and therefore not measured with circles. Your approach suggests that the circles must always work, which is not true. In my example 1 and 2, the circle measurement would result in a violation, even if the targets were sufficiently large and fulfilled the SC.

patrickhlauke:
@JAWS-test fair point yes, the top should not have the circles version at all to drive home that at 24x24 is actually quite small (at 1 to 1 scale) but you don't have to worry about the circle bits, and only if you then go even smaller, you have to start engaging with it.

patrickhlauke:
tweaked it on #3103

target spacing toolbar diagram

guidelines/sc/22/target-size-minimum.html
mbgower: > wonder much more why example 1 fails and example 3 passes, even though 1 is better than 3.

I think a better way to look at this is to phrase that as "even though 3 is an even worse design than 1". They are both pretty bad.

It is possible to come up with hypothetical edge cases that demonstrate shortcomings of pretty much every requirement. Whether _anyone_ would be well-served by someone implementing such a design is a valid question, IMO. Do we create SCs that seek to guide real-world design to a better outcome, or do we focus on preventing a bunch of unlikely edge cases by creating a wordy SC that is not easy to understand or implement? My firm belief is we should not allow the perfect to be the enemy of the good.

BTW, the metrics are such that for the uncentered circle in the ring to pass, it has to be big enough that the circle is about 20 pixels in diameter. It's nothing to celebrate, but since it's the same spacing needed for an undersized circle target close to a full-sized target, it is at least consistent, from that perspective.

a dot off-center inside a ring at roughly actual size

patrickhlauke:
i'm assuming this PR would be superseded by https://github.com/w3c/wcag/pull/3123/ if that gets merged, right?

alastc:
Superseded by #3123

PR #3093, opened 2023-03-07
Removing parsing from 2.0 / 2.1

alastc:
Subject to CFC, not agreed yet.

We'll need a separate PR for the understanding docs.


alastc:
The agreed approach was to add notes, so closing this PR.

Issue #2755, opened 2022-10-28
Can we get a better example?


mbgower:
During a review of the [Understanding document](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html), it was pointed out that the Figure 9 diagram on a hypothetical arrangement of 5 buttons in a cross, was not the best set of examples on  the advantages of the criterion, since it was difficult to understand how the second example, where the middle target is reduced in size to 10x10, but the spacing is maintained for the other 4 targets, is really more difficult to operate than the first example, where it is in the same position at 24x24.
![image](https://user-images.githubusercontent.com/14298245/198733034-1285a0d8-d945-4dfd-9ec7-094489eb27af.png)

Someone argued that there is no higher likelihood of the user mis-pressing the center control in the second example than the first, and in fact in the second one, the reduced size of the target may result in less  presses of the neighbouring controls.

One could argue that the example illustrates some of the curiosities of the measuring outcomes. Is it better to remove it, or to explain more fully what makes B a failure?

patrickhlauke:
I intended that figure to be an example of the curious aspects/ side effects of the SC ... so I'd be in favour of keeping it and explaining it furtherOn 28 Oct 2022, at 23:41, Mike Gower ***@***.***> wrote:﻿
During a review, it was pointed out that the Figure 9 diagram on a hypothetical arrangement of 5 buttons in a cross, was not the best set of examples on  the advantages of the criterion, since it was difficult to understand how the second example, where the middle target is reduced in size to 10x10, but the spacing is maintained for the other 4 targets, is really more difficult to operate than the first example, where it is in the same position at 24x24.

Someone argued that there is no higher likelihood of the user mis-pressing the center control in the second example than the first, and in fact in the second one, the reduced size of the target may result in less  presses of the neighbouring controls.
One could argue that the example illustrates some of the curiosities of the measuring outcomes. Is it better to remove it, or to explain more fully what makes B a failure?

GreggVan:
Yes - can someone explain again (or point to the explanation) of how 1 passes but 2 fails? 

gregg

------------------------------
Gregg Vanderheiden
***@***.***

patrickhlauke:
On 29/10/2022 21:31, Gregg Vanderheiden wrote:
> Yes - can someone explain again (or point to the explanation) of how 1 
> passes but 2 fails?

Just measure it. When the controls are all touching, the central one has 
a width/height of 24x24 CSS px. when they're not, and you calculate the 
[...]

PR #3097, opened 2023-03-08
Aligning target size SCs

alastc:
Closes #3086

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/pull/3097.html" title="Last updated on May 29, 2024, 2:38 PM UTC (f84e036)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/3097/f3e9113...f84e036.html" title="Last updated on May 29, 2024, 2:38 PM UTC (f84e036)">Diff</a>

PR #3103, opened 2023-03-15
Update 2.5.8 Target Size (Minimum) understanding for new spacing approach

### LINKED ISSUE #3045: Loophole in 2.5.8 Target Size (Minimum)?
The spacing exception has been designed with tightly packed controls in mind, but as soon as controls are far enough apart, (or, in extreme cases, there *are* no other controls around), it effectively loses any normative power to define a minimum size.

Taking the extreme opposite case: imagine a page that only has a single button, and no other actionable/clickable controls. Per the normative wording, using the spacing exception, it would appear that it's "legal" for that single button to be sized at 1 CSS px by 1 CSS px.

Just wondering if we want to acknowledge this loophole, or hope that nobody twigs? EDIT: acknowledge as in "at least mention it in the understanding, but also emphasise that while it may normatively pass, as a *best practice* authors really shouldn't do it"

patrickhlauke:
Companion for <s>#3089</s> https://github.com/w3c/wcag/pull/3123

* remove outdated target offset language and examples
* reinforce the fact that this requires *either* minimum size *or* spacing in prose
* add new spacing diagram as SVG, with link to scalable version directly
* change the `<ol>` in the "Exceptions" section to `<ul>` - it's unordered in the normative SC wording, so why make it a numbered list here?
* markup cleanup
* delete unused images
* harmonised with https://github.com/w3c/wcag/pull/3123
* update the two examples with text-based buttons in a single row / two rows (to also use the spacing circle approach and language)
* update the dropdown example for consistency and to introduce the circle concept there as well
* update the basic, skewed, and rounded corners (not circle, to avoid confusion with the spacing circles concept later on) size examples for consistency
* add explanation and examples for the how to center the 24 CSS pixel diameter circle on the bounding box of an undersized target (for rectangular, convex, and concave targets)
* expand/reinforce the idea (in the intent) that this SC does still allow exceedingly small targets that pass as long as there's no adjacent targets that are too close, and adds best practice recommendation to at least just stick with the minimum size (regardless of spacing)

Supersedes https://github.com/w3c/wcag/pull/3058 if #3089 gets accepted and merged

Closes https://github.com/w3c/wcag/issues/2755
Closes #3045 

RawGit preview: https://raw.githack.com/w3c/wcag/patrickhlauke-target-size-minimum-issue3089-companion/understanding/22/target-size-minimum.html


patrickhlauke:
Set to draft for now, as I want to work on an additional SVG to illustrate the "circle must not overlap an adjacent target itself" aspect

detlevhfischer (approved):
Looks good to me. 
Spacing exception: "..any other other adjacent targets, nor the adjacent targets themselves" is more explicit, better. 
Great graphics!

patrickhlauke:
suggestion from @graemecoleman to split the one large diagram for the spacing exception into three separate diagrams. makes sense, at it will be much more readable and avoid the giant front-loaded explanation. will work on this if agreed in principle here that this is the approach we want to take

patrickhlauke:
@detlevhfischer hah, spotted a typo (duplicate "other") thanks to your comment too :)

patrickhlauke:
thoughts on my additional diagram I wanted to make to show the "overlaps the adjacent target" aspect? this also covers the same ground as the old boy/girl/magnifying glass image

image

(this is only a rough at this stage, can tweak it to be pixel-perfect and write some prose for it if this is deemed ok)

/cc @detlevhfischer

detlevhfischer:
@patrickhlauke I have a view suggestions:

* Group images in a pass row and two fail rows (words PASS and FAIL go after the rows)
* The first colum covers target overlapping the image, the second placement outside to the right, the third lie the second but with a smaller target
* Align cases so the pass case sits above the fail case
* No. 4 , the FAIL example 16x16 that abuts the image, is without circle - is that on purpose? I thought only targets bigger than 24 x 24 get no circle?
* All FAIl examples shoud reinforce FAIL with pink tint (incl. No. 4) all pass examples have green tint
* It may be easier to see the image as an image if it was an actual image rather than an empty space witth an img icon floating in the middle. At least I would give it a grey tint - or use a stock photo of some street scene which warrants magnification (my fuzzy portrait was not ideal in that sense)

Numbering your current images from the top left to bottom right in normal reading order, you'd have this structure:

```
1  7  6  PASS
3  5  4  FAIL
2      FAIL
```

patrickhlauke:
> No. 4, the FAIL example 16x16 that abuts the image, is without circle - is that on purpose? I thought only targets bigger than 24 x 24 get no circle?

there's two columns on that image. on the left side, the targets and their measurements. on the right side, the same one but with the circle overlaid. this is to make sure that you can actually see the actual thing on the left properly, not covered by the circle already that obfuscates things. and the first row, as it's >= 24x24, doesn't need a circle to begin with

detlevhfischer:
sorry I didn't get that logic just from the illustration, at least it is not easy to understand - I saw these as separate cases. I only now see they are actually the same! I would simplify and draw the circle whenever the target is below 24 x 24... and this will reduce the examples needed.

patrickhlauke:
i'll noodle around some more with it in any case. btw, as i've lost track...how likely does it seem that we'll be adopting the simpler circles model versus the *target offset* weirdness?

mbgower:
> i'll noodle around some more with it in any case. btw, as i've lost track...how likely does it seem that we'll be adopting the simpler circles model versus the target offset weirdness?

@patrickhlauke I suspect the odds are very high now that a new CR is in order, and that it will be discussed in a couple of minutes on the backlog call, if you can attend

patrickhlauke:
@mbgower unfortunately friday afternoons are the worst time in the week for me to attend live (otherwise i'd have loved to be more active in the backlog stuff directly myself, rather than just sniping from the sides here on github)

patrickhlauke:
just added a slightly cleaned-up version of my proposed "overlapping and touching/abutting targets" diagram, with prose

detlevhfischer:
@patrickhlauke Yes, the dividing line have done it. Good stuff. Now let's hope the circle exception will make it into 2.2.

patrickhlauke:
Updated to harmonise with https://github.com/w3c/wcag/pull/3123, to replace "overlap" with "intersect" in both prose and SVG illustrations (cc @mbgower), and to split out the second large SVG into two smaller ones (one for the clipping example, one for the touching/abutting example)

patrickhlauke:
Need to see how https://github.com/w3c/wcag/pull/3097 will affect this too

patrickhlauke:
Refreshed the examples for text-based buttons (to use the spacing circles language, and to visualise that consistently)

mbgower:
@patrickhlauke it keeps improving!
Not that there have been updates to the Spacing normative text, to be incorporated.

patrickhlauke:
@mbgower

> Not that there have been updates to the Spacing normative text, to be incorporated.

if the change to circles doesn't happen, i'll publish my own WCAG 2.2 (Ronin Edition) with my lovely diagrams ;)

mbgower:
That should have been "**note** that there have been updates", not "not". :)
 As of tomorrow, there hopefully will be new wording for the SC.

patrickhlauke:
> **note** that there have been updates

even beyond https://github.com/w3c/wcag/pull/3123 ? is that the one you meant? if the latter, already incorporated here

patrickhlauke:
The last parts that I want to add/modify in the understanding document (in the next day or so) are:

* adding a short preamble/explanation, at the start of the "Spacing" section, that reiterates the idea of the 24 CSS pixel circles, and gives examples of concave, convex, complex undersized targets to show how circle is centered on the bounding box, regardless
* redo the dropdown menu example in the same style as the other new diagrams and the redone text button examples, for visual consistency

detlevhfischer (commented):
Great stuff - thanks for improving the examples!

patrickhlauke:
thanks @detlevhfischer ... slowly getting there :)

patrickhlauke:
@detlevhfischer @mbgower added a whole bit at the start of the "Spacing" section now to explain the concept of the 24 CSS diameter circle and how to center it (for square/rectangular, convex, and concave targets) ... I think, from my point of view, this was the last piece of the puzzle. happy to wordsmith further (as the explanation there was admittedly a bit rushed), but otherwise will make this PR ready for review when/if the actual SC wording changes to use the spacing approach

mbgower:
Great work, @patrickhlauke ! A vast improvement, and the whole thing reads very well.
I directly committed one trivial editorial change for readability, which you can review in the diff of that commit.

detlevhfischer:
@patrickhlauke It's really come a long way. Beautiful work. Good to see also graphics for irregular shapes.  I am not sure if @dbjorge's proposal to swap "be able to draw a 24x24 rectangle inside the target" for a 24 x 24 circle is still a option but it looks like the impact on this understanding text would be minimal (one small change in section "Size requirement").

patrickhlauke:
> I am not sure if @dbjorge's proposal to swap "be able to draw a 24x24 rectangle inside the target" for a 24 x 24 circle is still a option but it looks like the impact on this understanding text would be minimal

While I was thinking about that as well while revising this, that would essentially change the whole nature of the SC (as then we're not talking about "The size of the target for pointer inputs is at least 24 by 24 CSS pixels" since the size is understood as "a width of 24 CSS pixels and a height of 24 CSS pixels", and that thing with the rectangle/square helps in determining *that* part

patrickhlauke:
to more directly address #3045 I've reinforced the second paragraph in the intent a bit in the latest commit https://github.com/w3c/wcag/pull/3103/commits/78d1253162cd22f2300d80ab627843c2048d72f4

patrickhlauke:
@alastc now that #3123 has been merged, it would be good to see this companion merged soon as well (will just look at the merge conflict now to make sure it's kosher)

PR #3107, opened 2023-03-17
Fix reflow issue in 1.4.10 Reflow Understanding

yatil:
@alastc please merge.

The video breaks out of the grid and creates horizontal scrolling. It’s bad. Added a max-width: 100% style to fix it.

(Contributing this change was only possible due to [my supporters](https://steadyhq.com/en/yatil/about).)


yatil:
I hereby irrevocably donate all possible IP related to this Pull Request to W3C.org Inc.

PR #3112, opened 2023-03-20
Rewrite technique C40 and associated example per discussion in #3026

### LINKED ISSUE #2469: C40 Example 1 "Test link" and "Test 2" fail 2.4.11 and 2.4.13 (thus NOT sufficient)
C40 Example 1 "Test link" and "Test 2" fail 2.4.11 and 2.4.13 (thus NOT sufficient)
https://w3c.github.io/wcag/techniques/css/C40

See: Example 1: A light and dark dotted indicator
https://w3c.github.io/wcag/working-examples/css-two-focus-colors/dotted.html

"Test link" and "Test 2" have dotted lines, failing 2.4.11 and 2.4.13 and so not applicable as Sufficient

(box shadow not seen)

### LINKED ISSUE #2470: Procedure C40 needs to be evaluated for 2.4.11: Focus Appearance (Minimum) and 2.4.13 Focus Appearance (Enhanced)
Procedure C40 needs to be evaluated for 2.4.11: Focus Appearance (Minimum) and 2.4.13 Focus Appearance (Enhanced)
https://w3c.github.io/wcag/techniques/css/C40

> Procedure
> For each focusable user interface element:
> 
> 1. Check that the two-colors in the focus indicator are adjacent and have a contrast ratio that is 3:1 or greater with each other.
> 2. Check that the colors used contrast with the adjacent colors, or are at least 2 CSS pixels thick.

Change 
> For each focusable user interface element:

To:
> For each [user interface component](https://w3c.github.io/wcag/understanding/focus-appearance-minimum.html#dfn-user-interface-component) 

And: **The 2 bullet point must be changed to make sure they follow the 3 base requirements for 2.4.1. AND 2.4.13**

### LINKED ISSUE #2575: Should C40 be also sufficient for 1.4.11 Non-Text Contrast?
Hi everyone!

The reason why I ask if [C40: Creating a two-color focus indicator to ensure sufficient contrast with all components](https://www.w3.org/WAI/WCAG21/Techniques/css/C40) is because I am trying to determine if a focus indicator that fails the contrast requirement fails both 1.4.11 and 2.4.7.

Let's say we have the situation like in figure 12 of 1.4.11. See below:
Screenshot of figure 12 where the focus fails non-text contrast inside of a button

In the image there is a focus but it fails non-text contrast. Therefore, would it be a failure of 1.4.11 Non-text Contrast and a pass of 2.4.7 Focus visible because the focus is there? Would it fail both because we can pretend that any non-text element that fails the contrast requirement is "invisible" to the user.

That said, the sufficient technique C40 is about ensuring proper contrast. Considering how 1.4.11 Non-text Contrast speaks about focus indicators, C40 sounds like it would also fix non-text contrast requirements.

Restatement of the questions:

1.  Is a failure of non-text contrast for a focus indicator also mean a failure of focus visible?
2. Should [C40: Creating a two-color focus indicator to ensure sufficient contrast with all components](https://www.w3.org/WAI/WCAG21/Techniques/css/C40) be also sufficient to pass 1.4.11 Non-text Contrast for focus indicators?

### LINKED ISSUE #3026: Does a two-color focus indicator pass SC 2.4.11 if it contrasts itself?
Success Criterion [2.4.11 Focus Appearance (Minimum)](https://www.w3.org/TR/WCAG22/#focus-appearance-minimum) (Level AA): When a [user interface component](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance-minimum#dfn-user-interface-component) has keyboard focus, the [focus indicator](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance-minimum#dfn-focus-indicator):
- [Encloses](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance-minimum#dfn-encloses) the visual presentation of the user interface component;
- Has a contrast ratio of at least 3:1 between its pixels in the focused and unfocused states;
- Has a contrast ratio of at least 3:1 against adjacent colors.

In the screen below you can see a button with an outer focus indictor composed of:
- 2px width blue #3E95FA
- 1 px width white

image

 

One of the requirement is: has a contrast ratio of at least 3:1 between the same pixels in the focused and unfocused states.

With the grey background example the contrast of the blue or white rings is not high enough. However the blue and white constituent parts of the indicator contrast themselves which ensures this state change is always highly distinguishable for users. 

Does this combination of contrasting colors for the focus indicator allow this UI component to pass the requirements for success criteria 2.4.11 regardless of the background color?

### LINKED ISSUE #3055: Technique C40: description and links mention text-shadow but examples use box-shadow
In [Technique C40](https://www.w3.org/WAI/WCAG21/Techniques/css/C40q) the description text (and example links' text) both reference `text-shadow` but the examples actually use `box-shadow`:

screenshot showing the issue

dbjorge:
This PR updates C40 and its associated examples per the discussion in #3026. Particularly:

* It specifies explicitly that the requirement for the two colors is 9:1, and updates the colors in the examples accordingly
* It disclaims that the technique only guarantees that the SC will be satisfied if the entire indicator is overtop one single, solid background color
* It removes use of dotted/dashed outlines in favor of a single example that uses 2px solid color bands
* It adds a note warning of a gotcha related to `outline: none` that can come up when trying to implement CSS similar to the recommended example

I explored writing a second example using two 1px color bands offset slightly from the focused component. This type of indicator would pass in all the same scenarios as the first example, since the "adjacent contrast" test would be comparing against the same pixels that would have been background pixels for the first example. However, I didn't include an example because I didn't come up with a good technique for implementing it generally. The closest option I explored was along the lines of this:

```css
*:focus {
  /* required to support correct positioning in the ::after style */
  position: relative;
  /* outer outline */
  outline: solid 2px #330000;
  outline-offset: 1px;
}
*:focus::after {
  content: "";
  position: absolute;
  top: 0; left: 0; right: 0; bottom: 0;
  /* inner outline */
  outline: solid 1px #E2E2F2;
  outline-offset: 1px;
}
```

...but I didn't feel comfortable with this as a general technique since it has issues with elements that can't accept `::after` content (eg, many `input` types) and it can cause issues to override `position` so broadly like this.

If someone has a bright idea for a more generally appropriate implementation, I'd be happy to include it as a second example.

This is the first time I've made changes this substantial to a technique, please feel free to have at me with any suggestions to help make the style more consistent with other techniques :)

Closes #3026
Closes #3055 
Closes #2469 
Closes #2575
Closes #2470

[Preview C40](https://raw.githack.com/dbjorge/wcag/fix-3026-c40/techniques/css/C40.html), [preview C41](https://raw.githack.com/dbjorge/wcag/fix-3026-c40/techniques/css/C41.html)


techniques/css/C40.html
```diff
@@ -18,63 +18,51 @@ <h2>When to Use</h2>
...
 
-			<p class="note">If it can be determined that the two-color focus indicator CSS takes precedence then the test can be applied programmatically rather than by manually focusing on each interface component.</p>
-		
+			<p>In CSS, two-color indicators can be implemented by combining the <code>outline</code> and <code>box-shadow</code> properties with the <code>:focus</code> or <code>:focus-visible</code> pseudo-classes.</p>
+
+			<p class="note">Avoid setting <code>outline: none</code> to use <code>box-shadow</code> on its own. User agents commonly suppress the <code>box-shadow</code> property in forced-color modes, so authors should avoid relying on <code>box-shadow</code> alone to implement focus indicators.</p>
```
alastc, line 31:
```suggestion
			<p class="note">Avoid setting <code>outline: none</code> to use <code>box-shadow</code> on its own. User agents commonly suppress the <code>box-shadow</code> property in forced-color modes, so authors should avoid relying on <code>box-shadow</code> alone to implement focus indicators. Setting the outline to transparent does not prevent it appearing in forced-color modes.</p>
```

PR #3116, opened 2023-03-23
4.1.1 adding note.

alastc:


guidelines/sc/20/parsing.html
```diff
@@ -9,6 +9,8 @@ <h4>Parsing</h4>
       duplicate attributes, and any IDs are unique, except where the specifications allow
       these features.
    </p>
+
+   <p class="note">This SC should be considered as automatically met for any content using HTML.  Modern browsers all have automatic error correction for parsing errors, and issues such as incorrect states or names due to a duplicate ID, or missing roles due to inappropriately nested elements are covered by different success criteria. This criterion can therefore be ignored as being redundant.  It no longer provides any benefit to people with disabilities in itself and should not be enforced or required for accessibility.</p>
```
mbgower, line 13: This is entirely editorial, just to match established style, as I understand it:
```suggestion
   <p class="note">This Success Criterion should be considered as automatically met for any content using HTML.  Modern browsers all have automatic error correction for parsing errors, and issues such as incorrect states or names due to a duplicate ID, or missing roles due to inappropriately nested elements are covered by different Success Criteria. This criterion can therefore be ignored as being redundant.  It no longer provides any benefit to people with disabilities in itself and should not be enforced or required for accessibility.</p>
```

understanding/20/parsing.html
```diff
@@ -12,27 +12,29 @@ <h1>Understanding Parsing</h1>
...
+      <p>This criterion has been removed from WCAG 2.2. It remains in place for WCAG 2.0 and 2.1, with the following note:</p>
 
-         <p>The following content is left for historical purposes to show the original intent.</p>  
-            
-         <hr />
+      <p class="note">This SC should be considered as automatically met for any content using HTML.  Modern browsers all have automatic error correction for parsing errors, and issues such as incorrect states or names due to a duplicate ID, or missing roles due to inappropriately nested elements are covered by different success criteria. This criterion can therefore be ignored as being redundant.  It no longer provides any benefit to people with disabilities in itself and should not be enforced or required for accessibility.</p>
```
mbgower, line 18:
```suggestion
      <p class="note">This Success Criterion should be considered as automatically met for any content using HTML.  Modern browsers all have automatic error correction for parsing errors, and issues such as incorrect states or names due to a duplicate ID, or missing roles due to inappropriately nested elements are covered by different Success Criteria. This criterion can therefore be ignored as being redundant.  It no longer provides any benefit to people with disabilities in itself and should not be enforced or required for accessibility.</p>
```

guidelines/sc/20/parsing.html
```diff
@@ -9,6 +9,8 @@ <h4>Parsing</h4>
       duplicate attributes, and any IDs are unique, except where the specifications allow
       these features.
    </p>
+
+   <p class="note">This Success Criterion should be considered as automatically met for any content using HTML.  Modern browsers all have automatic error correction for parsing errors, and issues such as incorrect states or names due to a duplicate ID, or missing roles due to inappropriately nested elements are covered by different Success Criteria. This criterion can therefore be ignored as being redundant.  It no longer provides any benefit to people with disabilities in itself and should not be enforced or required for accessibility.</p>
```
dbjorge, line 13: In the interest of using the note to minimize drive-by non-conformance claims, I think it would be helpful to reword this somewhat to make it very clear that we are **intepreting the existing normative text** (specifically, the "except where the specifications allow these features" part) as being automatically met for HTML documents, and not using an informative note to ignore or override the normative text. Particularly, I think the sentence "This criterion can therefore be ignored as being redundant" is problematic; the normative text of the conformance model doesn't allow authors to ignore criteria just because they are redundant.

I think an update along these lines would stay more firmly within the boundaries of the conformance model, but achieve the same intent:

```suggestion
   <p class="note">This Success Criterion should be considered as automatically met for any content using HTML. Since this criterion was written, the HTML Standard has adopted specific requirements governing how user agents must handle incomplete tags, incorrect element nesting, duplicate attributes, and non-unique IDs. Although the HTML Standard treats some of these cases as non-conforming for authors, it is considered to "allow these features" for the purposes of this Success Criterion because the specification requires that user agents support handling these cases consistently. In practice, this criterion no longer provides any benefit to people with disabilities in itself; issues such as missing roles due to inappropriately nested elements or incorrect states or names due to a duplicate ID are covered by different Success Criteria.</p>
```

mbgower: Since we're getting more explanatory, I've heard some folks mention historical AT behaviour no longer in place which also has lessened the need for 4.1.1. We might want to consider adding that, if it was felt to be helpful. (On the other hand, a cleaner/simpler is typically better for this kind of thing.)

guidelines/sc/20/parsing.html
dd8, line 1: The F70 and F77 failures referenced here also need updated since folks may get to them via Google without seeing the understanding document. Would also be good being explicit in the failures section of the understanding document that they are now non-failures.

dd8: Sorry - comment on wrong file - this refers to understanding/20/parsing.html

alastc:
This was replaced with #3152, closing.

PR #3123, opened 2023-03-28
WCAG 2.2 CR 3 updates

### LINKED ISSUE #3037: Microsoft WCAG 2.2 CR2 Feedback: 2.4.11 Focus Appearance
I met with Microsoft CELA's Accessibility Technology group over the last two weeks to gather collective feedback on CR2 changes. Their feedback is aligned with the Microsoft's Accessibility Insights team's feedback from the perspective of a tool developer.

The parts we had specific new feedback on were the removal of 4.1.1 Parsing (separate issue #3034) and whether we supported keeping vs removing the at-risk 2.4.11 (this issue).

On 2.4.11 Focus Appearance:

> * The draft/language of 2.4.11 Focus Appearance is very difficult to implement. The current text with 2.4.11 Focus Appearance is making it harder and should be providing clarity. With this said, because of the dynamic nature of the web, web pages are becoming denser and therefore guidance is needed to help with this. WCAG guidance needs to be written in a much easier to digest fashion. 
> * 2.4.11 needs to be postponed until cleaner, easier to read and understand text is written. 

We are very supportive of the intent of the SC, and there was substantial internal debate around whether the merits of the intended outcome of the SC outweighed the understandability and implementation difficulty concerns.

However, in the end, we felt that the SC as currently written is too complicated to understand and too difficult to implement for us to feel comfortable supporting it as a recommendation in its current state. We support dropping the at-risk SC from WCAG 2.2 unless/until it can be substantially simplified.

Some of our experiences that we felt justified this conclusion were:

* We attempted to write up step-by-step instructions for manually validating an element in simple, clear language, in the same style that we currently use to write similar instructions for other WCAG criteria as part of Accessibility Insights for Web's Assessment. Our most successful draft of this was approximately three times as long as the currently-most-complex requirement (24.2 Contrast > State changes). This was despite the draft being incomplete; it would have needed to become more complex to handle certain edge cases (for example, the special case in the "minimum bounding box" definition for line-wrapped links)
* We gathered subjective feedback from of employees in different roles (engineers, PMs, legal, compliance, accessibility domain experts) and with different levels of accessibility expertise. Folks across all roles and expertise levels found it to be subjectively more confusing than other comparable SCs.
* We spent approximately 2 engineer-weeks in Accessibility Insights researching and prototyping how we might automate a test for the requirement (in whole or part). We didn't come up with anything very satisfactory. We concluded that in practice it's probably infeasibly complicated to try to approach this by comparing browser-reported stylings of different components, and that to approach it reliably you'd need to do image-diffs of each component of a page in a focused vs unfocused state as-rendered by the browser. But we think an image-diff-based solution would *also* be unreliable for indicators that are heavy anti-aliased. Additionally, automating a test for this is complicated by the lack of browser support for tools like ours to simulate focus states without user interaction.
* I completed the exercises in [Testing Focus Appearance #2901](https://github.com/w3c/wcag/issues/2901). In general, I felt that there was an uncomfortable number of real-world cases from this exercise that required subjective tester judgement and/or which required onerous pixel-by-pixel analysis to assess accurately. Full results and notes were shared privately with the AG chairs per the instructions in that issue.

### LINKED ISSUE #3050: 2.5.8 Target Size "overlapping any other target" does not work

### LINKED ISSUE #3051: 2.5.8 target size should not have a "lists" exception
This change was slipped in during the last-minute scramble to get to CR. I only noticed it recently. The current inline exception reads as follows:

> Inline: The target is in a sentence, or is in a bulleted or numbered list, or its size is otherwise constrained by the line-height of non-target text;

I think there is a legitimate reason to exempt controls inside blocks of text, as adjusting the height of those controls would increase the line-height, potentially making the text difficult to read. That argument hold for link lists though, and I feel in adding this to the exception we're doing people who need these larger targets a major disservice by this exception.

You can tell this wasn't thought out well, from how "numbered lists" are mentioned, but "lettered lists" are not. Why does this exception apply to 1. 2. 3. but not to A. B. C.?

### LINKED ISSUE #3086: Further alignment of structure/wording of 2.5.5 Target Size (Enhanced) and 2.5.8 Target Size (Minimum)
There's currently already an issue about aligning the "Inline" definition https://github.com/w3c/wcag/issues/2857 but I'd suggest going further and making sure the structure and wording of the two SCs are not unnecessarily different.

* 2.5.8 Target Size (Minimum) introduces the exceptions with "except *where*:" and then has an *unordered list* for the exceptions; 2.5.5 Target Size (Enhanced) introduces the exceptions with "except *when*:" and then uses a *description list* for the exceptions
* 2.5.8 has the following definition/exception: "Equivalent: The function can be achieved through a different control on the same page that meets this criterion."; 2.5.5 uses different wording to mean the same? "Equivalent: The target is available through an equivalent link or control on the same page that is at least 44 by 44 CSS pixels;"
* 2.5.8 has the following definition/exception: "Essential: A particular presentation of the target is [essential](https://w3c.github.io/wcag/guidelines/22/#dfn-essential) or is legally required for the information being conveyed;"; 2.5.5 has "Essential: A particular presentation of the target is [essential](https://w3c.github.io/wcag/guidelines/22/#dfn-essential) to the information being conveyed." ... so, is the "legally required" aspect of no concern for AAA ? And if so, why? Or just an oversight? (this leaves me with a weird "can't do AAA and be legal" feeling.

I'd say pick one of these as the authoritative wording and format, and stick with it consistently. And then harmonise the exceptions that are meant to be the same, and use the same consistent wording for both the AA and AAA version

alastc:
(PHL: added these) closes #3050 (?) and possibly closes #3086 (?)

Closes #3037
Closes #3050
Closes #3051

[Preview focus appearance](https://raw.githack.com/w3c/wcag/wcag22-cr3-last-updates/understanding/22/focus-appearance.html).
[Preview target size](https://raw.githack.com/w3c/wcag/wcag22-cr3-last-updates/understanding/22/target-size-minimum.html).

(PHL: updated raw githack links - protip: make the raw githack links point to the PR branch, rather than individual commit versions, so the links here always remain in sync with any further changes to the PR)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/pull/3123.html" title="Last updated on Apr 20, 2023, 4:03 PM UTC (358097e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/3123/c560987...358097e.html" title="Last updated on Apr 20, 2023, 4:03 PM UTC (358097e)">Diff</a>


guidelines/sc/22/target-size-minimum.html
```diff
@@ -7,11 +7,11 @@ <h4>Target Size (Minimum)</h4>
     
     <p>The size of the <a>target</a> for <a>pointer inputs</a> is at least 24 by 24 <a>CSS pixels</a>, except where:</p>
     <ul>
-        <li><strong>Spacing:</strong> The target does not overlap any other target and has a <a>target offset</a> of at least 24 CSS pixels to every adjacent target;</li>
+        <li><strong>Spacing:</strong> Undersized targets (those less than 24 by 24 CSS pixels) are positioned so that if a 24 CSS pixel diameter circle is centered on the <a>bounding box</a> of each, the circles do not intersect another target or the circle for another undersized target;</li>
         <li><strong>Equivalent:</strong> The function can be achieved through a different control on the same page that meets this criterion.</li>
```
patrickhlauke, line 11: change the full stop `.` at the end of the list item to `;`

bruce-usab:
+1 for this PR.  @alastc missing is the update to AAA Success Criterion 2.5.5 Target Size (Enhanced) inline exception (to match the inline exception of the new Target Size (Minimum) SC).

patrickhlauke:
assume this PR supersedes https://github.com/w3c/wcag/pull/3089

patrickhlauke:
need to work out relationship to https://github.com/w3c/wcag/pull/3097 as we now have variations around the same area all making slightly different updates

understanding/22/target-size-minimum.html
```diff
@@ -23,13 +23,13 @@ <h4>Target Size (Minimum)</h4>
              <p class="change">New</p>
 
              <p>The size of the <a>target</a> for <a>pointer inputs</a> is at least 24 by 24 <a>CSS pixels</a>, except where:</p>
-            <ul>
-                <li><strong>Spacing:</strong> The target does not overlap any other target and has a <a>target offset</a> of at least 24 CSS pixels to every adjacent target;</li>
+             <ul>
+                <li><strong>Spacing:</strong> Undersized targets (those less than 24 by 24 CSS pixels) are positioned so that if a 24 CSS pixel diameter circle is centered on the <a>bounding box</a> of each, the circles do not intersect another target or the circle for another undersized target;</li>
```
patrickhlauke, line 27: does this need a new definition somewhere for "bounding box"?

understanding/22/target-size-minimum.html
```diff
@@ -23,13 +23,13 @@ <h4>Target Size (Minimum)</h4>
...
-            </ul>
+                 <li><strong>Inline:</strong> The target is in a sentence or its size is otherwise constrained by the line-height of non-target text;</li>
+                 <li><strong>User agent control:</strong> The size of the target is determined by the user agent and is not modified by the author;</li>
+                 <li><strong>Essential:</strong> A particular presentation of the target is <a>essential</a> or is legally required for the information being conveyed.</li>
+             </ul>
             <p class="note">Targets that allow for values to be selected spatially based on position within the target are considered one target for the purpose of the success criterion. Examples include sliders with granular values, color pickers displaying a gradient of colors, or editable areas where you position the cursor.</p>
             <p class="note">For inline targets the line-height should be interpreted as perpendicular to the flow of text. For example, in a language displayed top to bottom, the line-height would be horizontal.</p>
             <dl><dt class="new"><dfn data-lt="target offsets">target offset</dfn></dt>
```
patrickhlauke, line 35: need to remove the target offset definition which is now not needed anymore

understanding/22/target-size-minimum.html
patrickhlauke: probably something very high-level like https://developer.mozilla.org/en-US/docs/Glossary/Bounding_box (rather than that definition of "minimum bounding box" that used to be used for focus appearance)

understanding/22/target-size-minimum.html
alastc: Already done :-)

understanding/22/target-size-minimum.html
alastc: Well, seeing as focus-appearance doesn't use that definition any more, I think we can re-use it here. The latest push includes that change.

understanding/22/target-size-minimum.html
patrickhlauke: so why do I still see it here in the file?

understanding/22/target-size-minimum.html
patrickhlauke: I do think the definition on https://developer.mozilla.org/en-US/docs/Glossary/Bounding_box is quite nice and concise. also, not quite sure what "For components which wrap onto multiple lines as part of a sentence or block of text (such as hypertext links), the bounding box is based on how the component would appear on a single line." part of the minimum bounding box definition actually means/shakes out to, so if we ARE going with that definition, this may need some clarification

alastc:
> need to work out relationship to #3097 as we now have variations around the same area all making slightly different updates

I'll take the text from this, and later apply the formatting from 3097.

guidelines/terms/22/minimum-bounding-box.html
```diff
@@ -1,6 +1,6 @@
-<dt class="new"><dfn data-lt="bounding boxes">bounding box</dfn></dt>
+<dt class="new"><dfn data-lt="bounding boxes">minimim bounding box</dfn></dt>
 <dd class="new">
 	<p class="change">New</p>	
-   <p>the smallest enclosing rectangle aligned to the horizontal axis within which all the points of a shape lie.</p>
+   <p>the smallest enclosing rectangle aligned to the horizontal axis within which all the points of a shape lie. For components which wrap onto multiple lines as part of a sentence or block of text (such as hypertext links), the bounding box is based on how the component would appear on a single line.</p>
```
patrickhlauke, line 4: i'm still unclear what that second part actually means in practice...how would I imagine a link on a single line? it's a bit cryptic

patrickhlauke: this also makes it confusing when considering targets that *are* fragmented on purpose somehow...would the "pretend it's all together" approach apply to them too?

patrickhlauke: ah, as you were, I see you answered this in the email

> The second sentence about wrapping targets should not apply to target size because, by the definition of target, it is “a region” rather than multiple regions.
>
> It should also not come up due to the inline exception.

patrickhlauke: however, are you re-applying this second sentence here in the PR? or am I getting confused?

alastc: It is applicable to focus-appearance, and yes, if the link were not wrapped, that's the size.
It is not applicable to target-size, both due to the inline exception, and because a target is defined as a region (not multiple regions).

alastc:
I just want to get down the rational for the perimeter example, mostly based on an email from @dbjorge.

<hr>

Whether you include corner pixels depends on whether we think that the 2px thickness of the perimeter should be drawn such that:
 
a. The thickness of the perimeter line is centered on the boundary of the control, ie, 1px of the line is drawn overtop the shape and 1px of the line is drawn outside the shape, or
b. both of the 2px is drawn outside of the shape
 
The definition is a bit ambiguous about this. It currently reads “continuous line forming the boundary of a shape not including shared pixels”. We have been interpreting that using “a”, but I could see someone interpreting the “not including shared pixels” part as implying “b”.
 
Using interpretation “a”, like we’d discussed last week:
 
- Rectangle case:
  - The inner pixel’s worth of thickness is 2w + 2h - 4 (ie, what the current example suggests) and the outer pixel is 2w + 2h + 4, for a total of 4w + 4h
  - Second equivalent way to calculate it: It is the area of a (w+2) by (h+2) rectangle minus the area of a (w-2) by (h-2) rectangle, ie, (w+2)(h+2) - (w-2)(h-2) = 4w + 4h
- Circle case:
  - It is the area of a (r+1) radius circle minus the area of a (r-1) radius circle, ie, 𝜋(r+1)^2 - 𝜋(r-1)^2 = 4𝜋r
 
Using interpretation “b”, like your diagram below:
 

- Rectangle case:
  - As you counted in your diagram, there are 16px of “extra” pixels to count, and counting pixels gets you to 4w + 4h + 16
  - Second equivalent way to calculate it: It is the area of a (w+4) by (h+4) rectangle minus the area of a w by h rectangle, ie, (w+4)(h+4) - wh = 4w + 4h + 16
- Circle case:
  - It is the area of a (r+2) radius circle minus the area of a r radius circle, ie, 𝜋(r+2)^2 - 𝜋r^2 = 4𝜋(r+1)
 
Really, this comes down to what we think the pass/fail boundary of Focus Appearance should be. Do we think that an author using a well-contrasting, 2px thick CSS border property (which gets drawn on this inside of the component) should pass the requirement? If so, we should actually be going with a third option “c” that draws the entire perimeter inside the shape and ends up arriving at 4w + 4h – 16 and 4𝜋(r-1), respectively.

<hr>

My thought was: keep it simple and just go with 4 * W + 4 * H, and 4𝜋r. 

Also, as this is to create an area, measuring the outside 1px of the component and doubling it is reasonable, the actual size of a 2px thick outline will be bigger by a few pixels.

understanding/22/target-size-minimum.html
bruce-usab: I agree that the Mozilla bounding box is definition seems sufficient.  We have the wrinkle that _our_ bounding box for hypertext is to based on the bounding box formed when linked text is unspooled to a single line.

bruce-usab:
Where are we with definition for [perimeter](https://w3c.github.io/wcag/guidelines/22/#dfn-perimeter)?  I think it is (emphasis added):

> continuous line forming the boundary of a shape **not including shared pixels**, or the [minimum bounding box](https://w3c.github.io/wcag/guidelines/22/#dfn-minimum-bounding-box), whichever is shortest.

I am fairly certain we are much better off deleting <q>not including shared pixels</q>.  I get that we are talking about a rendering of a line that  has some thickness to it (in pixels).  But the common understanding of the the term is that it is a zero-thickness line, so a measure of length and not an area.  We are also, with this definition, making perimeter a synonym to [bounding-box](https://developer.mozilla.org/en-US/docs/Glossary/Bounding_box) which is okay since the boundary of a bounding box also defined as having zero thickness.

The [example](https://w3c.github.io/wcag/guidelines/22/#example-16) hangs a lampshade on the problem:

> The perimeter calculation for a rectangle is 2h+2w -4, where h is the height and w is the width and the corners are not counted twice. The perimeter of a circle is 2𝜋r.

Please move the detail of subtracting off the corners elsewhere.  It could be a note to the SC or to Understanding.  It does not belong in a definition for perimeter.  An alternative is for the note to go into detail for counting pixels making up the rendered perimeter, but in that case, the px thickness of the line must be mentioned.

guidelines/terms/22/perimeter.html
```diff
@@ -2,6 +2,6 @@
 <dd class="new">
 	<p class="change">New</p>	
    <p>continuous line forming the boundary of a shape not including shared pixels, or the <a>minimum bounding box</a>, whichever is shortest.</p>
-  <aside class="example"><p>The perimeter calculation for a rectangle is 2<em>h</em>+2<em>w</em> -4, where <em>h</em> is the height and <em>w</em> is the width and the corners are not counted twice. The perimeter of a circle is 2𝜋<em>r</em>.</p></aside>
+  <aside class="example"><p>The perimeter calculation for a 2 CSS pixel perimeter around a rectangle is 4<em>h</em>+4<em>w</em>, where <em>h</em> is the height and <em>w</em> is the width. For a 2 CSS pixel perimeter around a circle it is 4𝜋<em>r</em>.</p></aside>
```
WilcoFiers, line 5: This note is wrong. A 2px border around the perimeter isn't 2*width + 2*height. You're missing the corners in that. Instead its 2 * (width+2) + 2 * (height+2).  And for a circle it's not 4𝜋*radius. It's 2𝜋*(radius+1) + 2𝜋*(radius+2).

JAWS-test: In my opinion, it makes no sense to count the actual pixels for a rectangle and therefore subtract the corner pixels, but for a circle take the theoretical pixels and subtract nothing. The actual pixels for a circle can deviate considerably from 4*PI*radius, especially for small circles.

PR #3130, opened 2023-03-31
Moved note from Understanding to SC

### LINKED ISSUE #3129: For 1.4.13 Content on Hover or Focus, note at bottom of Intent section should be moved up
The Understanding document for [Content on Hover or Focus](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html) has 2 notes underneath the normative text clarifying scope. These are prominent and useful. There is another note at the bottom of the Intent section which is easier to overlook, yet is just as important for context
>This SC covers content that appears in addition to the triggering component itself. Therefore, a non-visible component, like a Skip to Main link, that is made visible on keyboard focus (with no additional content beyond the trigger becoming visible) is not covered by the SC.

There was a recent report on some failures from a tester (sorry, can't locate) where someone had wrongly failed Skip to Main links, arguing they failed 1.4.13. I was pretty sure we'd covered this section, but when I went back to quickly check, I was shocked to see it wasn't covered in the notes. This is because if looking at the SC text only, only the first two note are listed. So both of us failed to look at the bottom of the intent section.

I think this makes sense to address by making the third note be at the bottom of the SC text, not the bottom of the Intent section in the Understanding document.

mbgower:
Resolves #3129 
Moved note at bottom of Understanding document's Intent section to the bottom of the SC notes.

### DIFF
--- a/guidelines/sc/21/content-on-hover-or-focus.html
+++ b/guidelines/sc/21/content-on-hover-or-focus.html
@@ -23,6 +23,5 @@
 
     	<p class="note">Examples of additional content controlled by the user agent include browser tooltips created through use of the HTML <a href="https://www.w3.org/TR/html/dom.html#the-title-attribute"><code>title</code> attribute</a>.</p>
      <p class="note">Custom tooltips, sub-menus, and other nonmodal popups that display on hover and focus are examples of additional content covered by this criterion.</p>
-     <p class="note">This SC covers content that appears in addition to the triggering component itself. Therefore, a non-visible component, like a Skip to Main link, that is made visible on keyboard focus (with no <em>additional</em> content beyond the trigger becoming visible) is not covered by the SC.
-     </p>
+     <p class="note">This criterion applies to content that appears in addition to the triggering component itself. Since hidden components that are made visible on keyboard focus (such as links used to skip to another part of a page) do not present additional content they are not covered by this criterion.</p>
   </section>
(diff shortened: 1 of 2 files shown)


bruce-usab:
I am a solid +1 on this.  For my own edification, has adding the 3rd note to the SC already gone to AGWG survey?

guidelines/sc/21/content-on-hover-or-focus.html
```diff
@@ -23,5 +23,6 @@ <h4>Content on Hover or Focus</h4>
 
     	<p class="note">Examples of additional content controlled by the user agent include browser tooltips created through use of the HTML <a href="https://www.w3.org/TR/html/dom.html#the-title-attribute"><code>title</code> attribute</a>.</p>
      <p class="note">Custom tooltips, sub-menus, and other nonmodal popups that display on hover and focus are examples of additional content covered by this criterion.</p>
-
+     <p class="note">This SC covers content that appears in addition to the triggering component itself. Therefore, a non-visible component, like a Skip to Main link, that is made visible on keyboard focus (with no <em>additional</em> content beyond the trigger becoming visible) is not covered by the SC.
```
WilcoFiers, line 26:
```suggestion
     <p class="note">This criterion applies to content that appears in addition to the triggering component itself. Skip links and other hidden component that is made visible on keyboard focus is not covered by the SC, unless <em>additional</em> content beyond the trigger becomes visible.
```

Issue #3075, opened 2023-03-01
2.5.5 & 2.5.8: How useful is target size when click area is not visible?


JAWS-test:
The [intent of SCs for target size](https://www.w3.org/WAI/WCAG22/Understanding/target-size-enhanced.html) is:

> ... to help users who may have trouble activating a small target because of hand tremors, limited dexterity or other reasons.

However, the SCs only define a size for the target, but not visibility. In [my example](https://codepen.io/jaws-test/pen/bGxgyxY), the targets are large enough (50x50 px) to satisfy both SCs, but the visible targets are 
* too small (approx. 8x12 px)
* too close together

to meet the intent of the criterion.

Adjusting the SC 2.5.5 is too late. This is probably also true for 2.5.8, but I suggest to add at least in the Understanding a note:

> The click area of the targets should be visually visible (e.g. via their content, a border or a background color). If this is not possible, the visual content of the target should at least be in the middle of the click area.

alastc:
There's an argument that, as a user, you'd probably aim for the right-hand side of the right-hand arrow, so ensure you don't hit the left one. Having the expanded hit area makes that more likely to work.

There are certainly cases where you can usefully expand the hit area beyond the visible part of the controls, that isn't something we would warn against in general.

GreggVan:
What is the right hand side of the right hand arrow.

I think people usually point with the tip of the pointer   — or the tip of the finger when the hand cursor is used.   Is that not correct? 
(Note - my comment is ONLY from personal experience — not any research.  Hence it is a question to find out what research has found ) 


gregg

patrickhlauke:
> What is the right hand side of the right hand arrow.

the example in the OP has two arrow icons. that's the arrows we're talking about, not the mouse pointer

JAWS-test:
> There's an argument that, as a user, you'd probably aim for the right-hand side of the right-hand arrow, so ensure you don't hit the left one. Having the expanded hit area makes that more likely to work.

I think that most would try to click one of the two icons, because they assume that only these are clickable. Since the two icons are small and close together, there is a risk of accidentally hitting the other one.

alastc:
Perhaps, but I think restricting hit areas to visible boundaries would be a net-negative.

rachaelbradley:
I've added a short paragraph to https://github.com/w3c/wcag/pull/3147

mbgower:
To answer the title of this issue:
>How useful is target size when click area is not visible?

[...]

PR #3147, opened 2023-04-12
Proposed updates to target size enhanced and target size minimum to a…

rachaelbradley:
…ddress #3075


understanding/21/target-size-enhanced.html
```diff
@@ -19,6 +19,7 @@ <h2>Intent</h2>
 			  <li>the control is positioned where it will be difficult to reach, or is near the edge of the screen;</li>
 			  <li>the control is part of a sequential task.</li>
 	  </ul>
+	  <p>This SC does not require that the visible indicator of a target, such as a button or icon, be the same size as the target. It is recommended that when the visible indicator of the target is notably smaller than the target, that the visible indicator be placed at the center of the target.</p>
```
mbgower, line 22: As per my comments in the issue, I'd prefer we not get into this space. But if we're going to add some content, I don't see why it is only germane to small indicators. We'd want big indicators to be centered on the target too.
```suggestion
	  <p>This SC does not require that the visible indicator of a target, such as a button or icon, be the same size or shape as the target. It is recommended that when the visible indicator of the target differs from the target, that the visible indicator be centered on the target.</p>
```

patrickhlauke, line 22: I believe we always expand "SC" to "Success Criterion" (here and in the other file change)

understanding/22/target-size-minimum.html
```diff
@@ -101,6 +101,8 @@ <h3>Size requirement</h3>
           
           <p>While the Success Criterion primarily helps touch users by providing target sizing to prevent accidental triggering of adjacent targets, it is also useful for mouse or pen users. It reduces the chances of erroneous activation due to either a tremor or reduced precision, whether because of reduced fine motor control or input imprecision.</p>
 
+          <p>This SC does not require that the visible indicator of a target, such as a button or icon, be the same size as the target. It is recommended that when the visible indicator of the target is notably smaller than the target, that the visible indicator be placed at the center of the target.</p>
```
mbgower, line 104:
```suggestion
          <p>This Success Criterion does not require that the visible indicator of a target, such as a button or icon, be the same size or shape as the target. It is recommended that when the visible indicator of the target differs from the target, the visible indicator be centered on the target.</p>
```

patrickhlauke:
As the original issue that this came from has been closed, and there was strong opposition in that issue, wondering if this PR should also be closed

mbgower:
Given the closure of the issue by the issue opener, we are not proceeding with this PR.

gundulaniemann:
There are reasonable invisible targets, like
whitespace inside a card, where the card also contains visible targets like action buttons.

PR #3148, opened 2023-04-14
2.5.1 understanding note on sliders

detlevhfischer:
Just added this note for clarification regarding the frequent image sliders where the current SC text does not clearly say whether their operation falls under 2.5.1 Pointer Gestures or the future 2.5.7 Dragging Movements. In my current formulation, horizontal product sliders engaging in verious directions (i.e. no specific intermediate point in one direction) would not be considered under 2.5.1 Pointer Gestures (but fall under 2.5.7 Dragging Movements in WCAG 2.2)

I have no specific interest in exempting this type of slider from 2.5.1, I just want more clarity here...


patrickhlauke:
I'll admit that the new illustration and wording confused me, more than anything else. It might be better to use much simpler language to explain what's happening with a slider/carousel of this nature...something like "The user clicks/taps on the carousel, and moves the pointer (mouse or finger), with the carousel following the movement. The user is free to move the pointer arbitrarily, rather than being constrained to move it along a very strict path. Once they release the mouse button/raise their finger from the touchscreen, the carousel 'settles' on a particular slide/position. This is not path-based, but a dragging motion..." or words to that effect. I'd avoid the whole "moving up/down scrolls" and similar.

detlevhfischer:
@patrickhlauke I see your point, and the illustration may indeed not be useful. Your suggeston seems a bit too specific though, focusing on 'carousel'. I've used that term for those constructs where slides indeed snap to a particular position - and often in those cases there are additional controls (dot lists or similar) that do provide an alternative. The text should be general enough to cover any area where content is parly partly visible and can be panned along an axis, with or without snap-to-position behaviour, including areas allowing vertical movement (rare as that may be). So you are right to avoid a specific "moving up/down to scroll/pan the view" language. Must come up with better text (sigh)...

patrickhlauke:
I *think* that this PR is made redundant by the changes i'm proposing in https://github.com/w3c/wcag/pull/4843 - **if** *that* PR is accepted, it's probably worth closing this PR (but to perhaps make a change/addition to the dragging movement understanding doc to more explicitly mention carousels and similar widgets, if it's not clear *there* already that dragging applies to them)

patrickhlauke:
Worked some more on https://github.com/w3c/wcag/pull/4843 - I believe that PR now supersedes this one

PR #3150, opened 2023-04-16
Rewrite OTP section

### LINKED ISSUE #3095: Accessible authentication and copying between devices
There seems to be an open question when it comes to one-time codes (OTCs) and copying between systems to authenticate. If for example I receive an OTC via text message, I could copy it to authenticate if 1. I'm authenticating on my phone, or 2. I have my devices linked such that I have a shared clipboard.

The SC doesn't provide guidance on how to decide when copying OTPs is allowed. If we take a strict reading you could argue that it's never allowed, because you can't know for sure that copying is possible. That doesn't seem right, as it basically prohibits OTPs altogether. A more generous reading would be that WCAG is about web content alone. How the OTP gets into the clipboard is out of scope and the only thing that matters for WCAG 2.2 conformance is that pasting isn't prevented by the page itself.

I'm not keen on either of these, and would hope there's some kind of middle ground I'm not thinking of. If I had to pick between the two I think the "WCAG is not about how the OTP gets into the clipboard" reading of this seems the more defensible to me. That does contradict some of the wording in the current understanding document. If we can't provide clearer guidance I think we should think of that more as advisory work, rather than required by the SC.

patrickhlauke:
* more tightly scope the SC to just the "paste" part, out-of-scoping how the temporary code/OTP gets to the user's clipboard in the first place

Closes #3095


understanding/22/accessible-authentication-minimum.html
```diff
@@ -86,18 +86,19 @@ <h3>Cognitive Function Tests</h3>
...
+            <p>Note that two-factor systems that do not rely on codes or passwords – including hardware authentication devices (such as YubiKey), secondary applications (either on the same primary device, or on a secondary device) that expect the user to confirm that it is indeed them trying to log in, and authentication methods provided by the user's operating system (such as Windows Hello, or Touch ID/Face ID on macOS and iOS) - are <em>not</em> <a>cognitive function test</a>.</p>
```
mbgower, line 101:
```suggestion
            <p>Note that two-factor systems that do not rely on codes or passwords – including hardware authentication devices (such as YubiKey), secondary applications (either on the same primary device, or on a secondary device) that expect the user to confirm that it is indeed them trying to log in, and authentication methods provided by the user's operating system (such as Windows Hello, or Touch ID/Face ID on macOS and iOS) - are <em>not</em><a> a cognitive function test</a>.</p>
```

ljoakley:
In the following sentence you can't use 2FA by itself until it's defined, You need to alter the first sentence as follows.

Beyond usernames and passwords, some sites may use two-factor authentication, **(2FA)**, asking the user to enter a temporary code or One-Time Password, OTP. A service that requires manual transcription of a 2FA code is not compliant.

source: https://developers.google.com/tech-writing/one/words

PR #3158, opened 2023-04-25
Change 'non-text graphics' to the normative SC term 'non-text content'

mitchellevan:
The "Testing Principles" section covers the two kinds of non-text content in scope for this SC: components and graphical objects. However, the introductory sentence said "non-text graphics" — this could be construed as limiting the scope of the testing just to the graphics (not the controls), and possibly further limiting it to just the "non-text" kind of graphics (if such things exist). So I changed "non-text graphics" to "non-text content" as in the normative SC.

### DIFF
--- a/understanding/21/non-text-contrast.html
+++ b/understanding/21/non-text-contrast.html
@@ -411,7 +411,7 @@
 
 			<section id="testing-principles">
 					<h3>Testing Principles</h3>
-					<p>A summary of the high-level process for finding and assessing non-text graphics on a web page:</p>
+					<p>A summary of the high-level process for finding and assessing non-text content on a web page:</p>
 					<ul>
 						<li>Identify each user-interface component (link, button, form control) on the page and:
 							<ul>


mitchellevan:
Maybe relevant to @philljenkins' [question](https://lists.w3.org/Archives/Public/w3c-wai-ig/2023AprJun/0004.html). I agree with @patrickhlauke's answer there. 'Non-text content' includes ASCII art and font icons used as graphics, because these things do not express something "in a human language" per the [definition of 'text'](https://www.w3.org/TR/WCAG21/#dfn-text).

srinivasuchakravarthula:
Perhaps wording it as "non-text elements" will help? This will avoid any confusion.

Regards,
Srinivasu

Srinivasu Chakravarthula
Director - Product Accessibility
Freshworks Inc
________________________________

awkawk:
I'm not sure that ASCII art isn't text. To be "text" it needs to:
1) be a "sequence of characters that can be programmatically determined" (which it is), and
2) "the sequence is expressing something in human language".

The key here is the definition for human language ("language that is spoken, written or signed (through visual or tactile means) to communicate with humans"). Is the ASCII art sequence expressing something that is spoken, written, or signed? 

I'm inclined to think that it is, which would mean that 1.4.11 (Non-text Contrast) wouldn't apply to ASCII art, but 1.4.3 (Contrast Minimum) would.

I don't think that the 1.4.11 Understanding document should reference "non-text content" here as reflected in this change since we are only talking about graphics, not other types of non-text content such as audio and video.

patrickhlauke:
> The key here is the definition for human language ("language that is spoken, written or signed (through visual or tactile means) to communicate with humans"). Is the ASCII art sequence expressing something that is spoken, written, or signed?
>
> I'm inclined to think that it is, which would mean that 1.4.11 (Non-text Contrast) wouldn't apply to ASCII art, but 1.4.3 (Contrast Minimum) would.

I think this is stretching the definition's intent a bit. ASCII art uses the visual shape of characters as building blocks for an image. yes, the image itself then "expresses something", but it's not spoken (it's not audio), written (as in, it is not forming words that can be seen), nor signed (visual).

awkawk:
I don't think that it is a good look for WCAG to be saying that ASCII text isn't text. Is it a problem to meet 4.5:1 for ASCII art?

awkawk:
Although the note in the definition of Non-text content is pretty clear on this... :)
NOTE
This includes [ASCII Art](https://www.w3.org/TR/WCAG21/#dfn-ascii-art) (which is a pattern of characters), emoticons, leetspeak (which uses character substitution), and images representing text

That said, I still don't think that the text in understanding should be changed for 1.4.11 to indicate non-text content since that is too broad for the context.

fstrr:
Did we end up making a decision on this?

patrickhlauke:
i'd be in favour of this change, as it's unnecessarily confusing (and leads to unnecessary thin-slicing / trying to divine deeper meaning of why "graphics" was used instead of "content", like we're seeing in this very thread).

just a late rejoinder on the "but what about images of text? they need 4.5:1 not just 3:1". sure, but i'd say it's fine here because yes, based purely on THIS criterion, they may be subject to a minimum of 3:1, but then the more specific SC for (text) contrast also applies and says it must be 4.5:1, so to satisfy THAT criterion authors need to go for the higher ratio anyway.

giacomo-petri (approved):

Issue #2809, opened 2022-11-23
Does 2.4.12 Focus not obscured encourage a keyboard anti-pattern


WilcoFiers:
In yesterday's AGWG call the group seems to have decided against adding an exception to SC 2.4.12 Focus not obscured regarding things like cookie popups which can be closed. I raised this last week on the cal too, but I think it's worth repeating.

In my opinion one of the easiest way to solve a SC 2.4.12 Focus not obscured violation with something like a cookie popup is to trap the keyboard focus in the popup, and force keyboard users to deal with the cookie popup before allowing them access to the rest of the content. You can do that without blocking the page for mouse / touch screen users.

It was suggested that this may be a violation of SC 2.1.1 Keyboard. That feels like stretching the meaning of 2.1.1 a bit. Especially if closing the cookie popup is a single-button action. What it does create is a very confusing experience for screen reader users. They'll be able to read the page with the virtual cursor, but the actual keyboard focus will be stuck in the cookie popup.

Just having an exception for things that can be closed would have given an easy "out" for these popups. I appreciate why AGWG didn't want to go with that solution. An exception for this is not ideal, but then I do think something is needed to discourage trapping the keyboard focus to solve for SC 2.4.12 Focus not obscured

alastc:
> force keyboard users to deal with the cookie popup before allowing them access to the rest of the content. You can do that without blocking the page for mouse / touch screen users.

In which case that content isn't available via the keyboard when it is for pointer users, therefore a 2.1.1 issue. 

> What it does create is a very confusing experience for screen reader users. They'll be able to read the page with the virtual cursor, but the actual keyboard focus will be stuck in the cookie popup.

That sounds like many modals which are almost right but don't work to lock access to the modal. I think we usually fail that aspect under 1.3.1 as the main content area is dimmed, but that isn't programmatically available to AT. Why is this type of modal different?

I'm not seeing how this SC is an incentive to create poor cookie banners (which generally fail many SCs anyway), and if it did, I don't see that creating an exception would help...

alastc:
Notes from the issues meeting:
- It is an in-equitable experience. 
- Unfortunately the negative scenario Wilco outlined probably doesn't fail 2.1.1 (or 2.1.2).
- There are good implementations, both modal and not (see iicsa.org.uk, gov.uk).
- We should add positive examples to the understanding document, which is our main place to influence people implementing this SC.
[...]

PR #3163, opened 2023-04-28
Update Understanding Focus Not Obscured

bruce-usab:
Closes [Issue #2809](https://github.com/w3c/wcag/issues/2809)

WRT to **_diff_** &mdash; look at **(Minimum) version first** 

- Understanding Focus Not Obscured (Minimum) has relatively few changes as compared to the previous version.
- Understanding Focus Not Obscured (Enhanced) started from (Minimum) version in this PR, so diff looks like a complete rewrite.


understanding/22/focus-not-obscured-minimum.html
```diff
@@ -40,43 +40,43 @@ <h2>Intent of Focus Not Obscured</h2>
...
+<section id="examples">
+  <h2>Examples of Focus Not Obscured (Minimum)</h2>
 
-      <ul>
-         <li>A page has a sticky footer (attached to the bottom of the viewport). When tabbing down the page the focused item is not hidden by the footer.</li>
-         <li>A page has a large (30% wide) cookie approval dialogue. The dialogue is modal, preventing access to the other controls in the page until it has been dismissed.</li>
-         <li>A notification is implemented as a sticky header and the keyboard focus is moved to the notification. The notification is dismissed when it loses focus so it does not obscure any other controls.</li>
-      </ul>
+   <ul>
+      <li>A page has a sticky footer (attached to the bottom of the viewport). When tabbing down the page the focused item is not hidden by the footer because content in the viewport scrolls up to always display the item with keyboard focus using <a href="https://www.w3.org/TR/css-scroll-snap/#propdef-scroll-padding" rel="nofollow">scroll padding</a>.</li>
```
alastc, line 58:
```suggestion
      <li>A page has a sticky footer (attached to the bottom of the viewport). When tabbing down the page the focused item is not completely hidden by the footer because content in the viewport scrolls up to always display the item with keyboard focus using <a href="https://www.w3.org/TR/css-scroll-snap/#propdef-scroll-padding" rel="nofollow">scroll padding</a>.</li>
```

bruce-usab:
After backlog call 5/5, I noticed another place parenthetical needed.
I added (Enhanced) version to this PR.

bruce-usab:
Added @mbgower to review since he was feeling caught up.

Should (Enhanced) version be so tolerant of opaque overlay?

bruce-usab:
Edit to (Enhanced) version after AGWG call 5/9.

understanding/22/focus-not-obscured-enhanced.html
```diff
@@ -59,6 +59,7 @@ <h3>Additional Techniques (Advisory)</h3>
     <h3>Failures</h3>
     <ul>
       <li>An interaction that causes content to appear over the component with keyboard focus, visually covering part of the focus indicator. This behavior might be encountered with advertising or promotional material meant to provide more information about a product as the user navigates through a catalogue.</li>
+      <li>A light box or other semi-opaque effect that overlaps the focus indicator.  The obscuring may triggered by a component receiving focus or from other interation, but is a failure regardless.</li>
```
mbgower, line 62: @bruce-usab for some reason I cannot make suggestions in this change. I suggest the following:

> A light box or other semi-opaque effect that diminishes the visibility of the focus indicator. This form of  obscuring may be triggered by a component receiving focus or from another interaction, but where the item receiving focus is partially obscured due to a semi-opaque overlay, it is a failure regardless.

mbgower:
> Should (Enhanced) version be so tolerant of opaque overlay?

Actually, @bruce-usab, on re-reading the Enhanced normative text, I do not believe that a lightbox should be considered to fail this criteria. The language does not use "obscured", it uses "hidden".

> no part of the component is hidden by author-created content.

I don't think we can make the case that dimming something hides it. Even in the minimum version, we don't say the lightbox is failing Focus Not Obscured. We say it is likely to fail 2.4.11 Focus Appearance.

On rereading that part of the Minimum understanding document, I have some changes to suggest.
1. Given that 2.4.11 is now a AAA, my feeling is that the language in the minimum should actually point to Non-text contrast, and 
2. the Enhanced version should point to Focus Appearance.

That gives us something appropriate to assess at each of the levels. If you agree, can you alter this PR to support that?

understanding/22/focus-not-obscured-enhanced.html
mbgower: Withdrawing this suggestion. Please see my comments in the Conversation. I think the lightbox language should be adjusted.

bruce-usab:
@mbgower I think the root problem is that there is a bit of a disconnect between the SC name and text. because <q>hidden</q> is a little bit stronger than <q>obscured</q>.  A light box certainly obscures.   But does it <q>hide</q>.

Ping to @alastc since I think we might actually need a survey to resolve this.

Also maybe the _text_ of Focus Not Obscured **(Enhanced)** AAA &mdash; and _only_ Enhanced _not_ (Minimum), AA 
> When a user interface component receives keyboard focus, no part of the component is <ins>obscured</ins> <del>hidden</del> by author-created content.

bruce-usab:
@mbgower wrote:

> That gives us something appropriate to assess at each of the levels. If you agree, can you alter this PR to support that?
....
> As part of that, you will need to remove this commit: `61dcd0c`

I am doing a new PR (and separate issue) for these details.

alastc:
Approved: https://www.w3.org/2023/05/16-ag-minutes#item06

PR #3165, opened 2023-05-01
Addressing issue #2479

Babysedaghati:
Added language to clarify that The language of the success criterion is an OR statement between the requirement for an alternative for time-based media OR an audio description. Therefore, if an audio description cannot be provided because there are insufficient spaces in the dialog to insert the Audio Description, then an alternative for time-based media would be required.

### DIFF
--- a/understanding/20/audio-description-or-media-alternative-prerecorded.html
+++ b/understanding/20/audio-description-or-media-alternative-prerecorded.html
@@ -39,12 +39,14 @@
          more complete representation of the synchronized media content than audio description
          alone.
       </p>
+      <p>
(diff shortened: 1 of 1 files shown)


patrickhlauke:
Not sure this is needed, as this is not the only "OR" type SC. would be strange to make an explicit clarification *here* but not elsewhere.

bruce-usab:

1atAlcone:
[]()

patrickhlauke:
> Not sure this is needed, as this is not the only "OR" type SC. would be strange to make an explicit clarification _here_ but not elsewhere. Suggest closing this PR.

actually, in the context of the "other" debate we've been having about videos that don't have sufficient pauses for AD ... this may still be relevant. retracting my thumbs-down, but this may be ground already covered by other more recent PRs that try to more fundamentally address the problem (but have been mired in lengthy discussions) https://github.com/w3c/wcag/issues/4460

patrickhlauke:
this PR *may* be superseded by #1790 if that gets accepted/merged

patrickhlauke:
actually...this PR seems to be a blatant copy of https://github.com/w3c/wcag/pull/2517 ... closing

PR #2990, opened 2023-02-01
New Failure Technique F109 for Issue #1918 Accessible Authentication needs better techniques

dan-hw:
- Created new failure technique F109 for forcing transcription of individual password characters, including some examples of failures I've seen in the wild
- Linked out to this new failure technique from the Understanding pages for SC3.3.8 and 3.3.9

#1916

PR #3174, opened 2023-05-04
Failure technique for accessible auth - Transfering from dan-hw:F109

### LINKED ISSUE #1916: Accessible Authentication needs better techniques
[G128 Email link authentication](https://w3c.github.io/wcag/understanding/accessible-authentication.html) is the only existing technique for [Accessible Authentication](https://w3c.github.io/wcag/understanding/accessible-authentication.html). I don't believe anyone would recommend it as the optimal way of meeting this criterion, especially because it just 'kicks authentication down the road' by sending the password to an email account, where the user is going to have to authenticate to _that_ website, which in turn must meet the same criterion for accessible authentication.
Here are other candidates:
1. There are some obvious missing techniques that match each of the Examples: properly marking up inputs and not preventing paste, WebAuthn, etc.
1. A possible Advisory technique is: **Providing a mechanism for the user to reveal the masked password string** (something that allows users to confirm the populated field matches the password they expect)
1. A possible failure technique: **Failure of Success Criterion 3.3.7 due to restricting text input options**

It could focus on this kind of construction:
` <input type="text" onpaste="return false;" ondrop="return false;" autocomplete="off" />
myElement.addEventListener('paste', e => e.preventDefault());`

alastc:
I can't edit the source, so I'm re-creating this PR:
https://github.com/w3c/wcag/pull/2990/

[Preview link](https://raw.githack.com/w3c/wcag/F109-accessible-auth/techniques/failures/F109.html)

Closes #1916


alastc:
@awkawk I've tried to address your comments on the title and procedure, see what you think.

PR #3176, opened 2023-05-05
Added info on inactive components to contrast Understandings

### LINKED ISSUE #481: Include details on user interface components in 1.4.3 & 1.4.6 Understanding

mbgower:
Resolves #481


mbgower:
I had to redo as https://github.com/w3c/wcag/pull/3177 because since I branched off Patrick's changes to #3020 , those were all included.
To keep the changes cleaner, I rebranched from main

PR #3178, opened 2023-05-05
Failure for focus-not-obscured

alastc:
[Preview of technique](https://raw.githack.com/w3c/wcag/focus-not-obscured-failure-technique/techniques/failures/focus-not-obscured.html)


patrickhlauke:
@alastc based on the conflict here...is this overwriting another new failure technique and need renumbering?

alastc:
Approved: https://www.w3.org/2023/05/16-ag-minutes#item04

Minor modification to make it specific to the AA version, and remove the link from the AAA version, which needs a separate failure technique.

PR #3179, opened 2023-05-08
Creating possible changes for Reflow

mbgower:
Addresses part of [discussion](https://github.com/w3c/wcag2ict/issues/98#issuecomment-1538596192) on WCAG2ICT https://github.com/w3c/wcag2ict/issues/98

### DIFF
--- a/guidelines/sc/21/reflow.html
+++ b/guidelines/sc/21/reflow.html
@@ -9,7 +9,7 @@
        <li>Vertical scrolling content at a width equivalent to 320 <a>CSS pixels</a>;</li>
        <li>Horizontal scrolling content at a height equivalent to 256 <a>CSS pixels</a>.</li>
    </ul>
-   <p>Except for parts of the content which require two-dimensional layout for usage or meaning.</p>
+   <p>Except for parts of the content with <a>two-dimensional layout</a> where scrolling in two dimensions is <a>essential</a> for usage or meaning.</p>
   
    <p class="note">320 CSS pixels is equivalent to a starting viewport width of 1280 CSS pixels wide at 400% zoom. For web content which is designed to scroll horizontally (e.g., with vertical text), 256 CSS pixels is equivalent to a starting viewport height of 1024 CSS pixels at 400% zoom.</p>
    
--- a/guidelines/terms/21/two-dimensional-layout.html
+++ b/guidelines/terms/21/two-dimensional-layout.html
@@ -1,0 +1,9 @@
+<dt><dfn data-lt="two-dimensional layout">two-dimensional layout</dfn></dt>
+<dd>
+   					
+   <p>layout with a prescribed vertical and horizontal size or aspect ratio
+   </p>
+   <p class="note">See the <a href="https://www.w3.org/TR/pointerevents/#dfn-pointer">Pointer Events
+      definition for "pointer"</a> [[!pointerevents]].</p>
+   				
+</dd>


detlevhfischer (approved):
Would it help to point out that in "requiring two-dimensional layout" the requirement is actually to be able to visually process or read the information in that content, where available? In diagrams with text, text will soon become too small under reflow to be discernable even with perfect vision. On the contrary, many photographs retain their meaning to some degree (I think you made that point in the other discussion over at WCAG2ICT), except for things that are 'read' to be identified (like a particular face in a group photo).

guidelines/terms/21/two-dimensional-layout.html
```diff
@@ -0,0 +1,9 @@
+<dt><dfn data-lt="two-dimensional layout">two-dimensional layout</dfn></dt>
+<dd>
+   					
+   <p>layout with a prescribed vertical and horizontal size or aspect ratio
+   </p>
+   <p class="note">See the <a href="https://www.w3.org/TR/pointerevents/#dfn-pointer">Pointer Events
+      definition for "pointer"</a> [[!pointerevents]].</p>
```
patrickhlauke, line 7: what the reason for referencing PE here?

mbgower: oops! I think that was the result of a copy from an existing definition I started from. Removed

mbgower:
> Would it help to point out that in "requiring two-dimensional layout" the requirement is actually to be able to visually process or read information in that content, where available? In diagrams with text, text will soon become too small under reflow to be discernable even with perfect vision. On the contrary, many photographs retain their meaning to some degree (I think you made that point in the other discussion over at WCAG2ICT), except for things that are 'read' to be identified (like a particular face in a group photo).

@detlevhfischer yes, this is one of the tricker aspect to the existing wording. The current use of "required" doesn't really say what it's referring to. In the Understanding document changes I'll be adding to this, I refer back to the "without loss of information" preamble to discuss how a lower resolution must be losing information -- but whether that violates essential  of course needs to be assessed on a case by case basis, which is why the 'archival' idea has been introduced. When you need to preserve something at its original dimensions or at its original resolution, you cannot easily put it in a reduced space. 

The counterargument to that I'm still working through is that one cannot simply provide the ability to download at the original size. What about _displaying_ it at the original size is essential?

guidelines/terms/21/two-dimensional-layout.html
mitchellevan, line 1: I agree it's valuable to define the term "two-dimensional layout," but "layout with a prescribed vertical and horizontal size or aspect ratio" doesn't cover enough. Consider a `<table>` with no width or height values specified, but a lot of content that makes the table both taller and wider than most viewports.

Unfortunately I can't think of a better definition. I feel it would have more to do with prescribing vertical and horizontal relationships within the content.

guidelines/sc/21/reflow.html
```diff
@@ -9,7 +9,7 @@ <h4>Reflow</h4>
        <li>Vertical scrolling content at a width equivalent to 320 <a>CSS pixels</a>;</li>
        <li>Horizontal scrolling content at a height equivalent to 256 <a>CSS pixels</a>.</li>
    </ul>
-   <p>Except for parts of the content which require two-dimensional layout for usage or meaning.</p>
+   <p>Except for parts of the content with <a>two-dimensional layout</a> where scrolling in two dimensions is <a>essential</a> for usage or meaning.</p>
```
mitchellevan, line 12: I like the clarity, but it would make the criterion more stringent than precedent established in the Understanding. Something is "essential" when "information and functionality cannot be achieved in another way that would conform." Two-dimensional scrolling of most data tables is not "essential" because the information could be achieved through a [responsive table](https://medium.com/allenhwkim/mobile-friendly-table-b0cb066dbc0e) technique. Two-dimensional scrolling of most toolbars could be avoided with responsive menus.

With this PR do you intend to establish a more stringent precedent for WCAG 2.1-2.2, or just steer the ship in a clearer direction for 3.0? If the former — I might support such a clarification, if it gets enough attention and support from the community. It would need to happen in the Understanding doc as a clarification of what "required" means, not as a change to normative WCAG.

mbgower:
@scottaohara I just thought I'd flag this to you. It was an exploration, and if you do not think it is worth pursuing, I'll close. LMK.

scottaohara:
thanks mike.  i'm slowly going through every issue / PR labelled with or mentioning Reflow.  this'll help make sure i don't miss this one.

PR #3185, opened 2023-05-10
updating language to use 'code'

### LINKED ISSUE #3180: Accessible authentication - consider other phrase than one time password
In discussion around approving a PR to address https://github.com/w3c/wcag/issues/3095
there was considerable discussion around passcode and password, and then someone suggested verification code

I dug in to usage trends and the latter looks more solid. I will continue investigation and create a small PR
https://trends.google.com/trends/explore?date=all&q=one%20time%20passcode,one%20time%20password,verification%20code

mbgower:
Closes #3180 
Changes to the understanding document and failure technique now bias towards use of "verification code" and "code", alongside "password" when discussing the pasting of information.


understanding/22/accessible-authentication-minimum.html
```diff
@@ -105,10 +105,10 @@ <h3>Copy and paste</h3>
...
-            <p>Note that two-factor systems that do not rely on codes or passwords – including hardware authentication devices (such as YubiKey), secondary applications (either on the same primary device, or on a secondary device) that expect the user to confirm that it is indeed them trying to log in, and authentication methods provided by the user's operating system (such as Windows Hello, or Touch ID/Face ID on macOS and iOS) - are <em>not</em><a> a cognitive function test</a>.</p>
+            <h3>Two-factor authentication systems (verifcation codes)</h3>
```
patrickhlauke, line 108: noting though that TOTP (time-based one-time password) / OTP (one-time password) are actual acronyms/terms that are used in practice when talking about these...codes

patrickhlauke: ah, I see you're then mentioning/namechecking it in the first sentence. I'd still love to see the acronym itself also presented/mentioned, but not too precious about it. Also might be nice to, parenthetically, mention 2FA / MFA? or is that too technical?

understanding/22/accessible-authentication-minimum.html
```diff
@@ -105,10 +105,10 @@ <h3>Copy and paste</h3>
...
+            <p>Beyond usernames and passwords, some sites may use two-factor authentication, asking the user to enter a verification code (also called a passcode or one-time password). A service that requires <em>manual</em> transcription of a verification code is not compliant. As with usernames and passwords, it must be possible for a user to at least paste the code (such as through a standalone third-party password manager, text message application, or software-based security key), or to allow user agents to fill in the fields automatically.</p>
```
patrickhlauke, line 109: maybe instead of "through", more like "from"?

mbgower, line 109:
```suggestion
            <p>Beyond usernames and passwords, some sites may use two-factor authentication, asking the user to enter a verification code (also called a passcode or one-time password). A service that requires <em>manual</em> transcription of a verification code is not compliant. As with usernames and passwords, it must be possible for a user to at least paste the code (such as from a standalone third-party password manager, text message application, or software-based security key), or to allow user agents to fill in the fields automatically.</p>
```

understanding/22/accessible-authentication-minimum.html
mbgower: Normally one only uses an acronym if the intent is to reuse later, and it's not reused here. Since I'm using the phrases in parentheses already, it's a bit disruptive to try to work in the acronym too.

understanding/22/accessible-authentication-minimum.html
```diff
@@ -105,10 +105,10 @@ <h3>Copy and paste</h3>
...
+            <p>Note that two-factor systems that do not rely on codes -– including hardware authentication devices (such as YubiKey), secondary applications (either on the same primary device, or on a secondary device) that expect the user to confirm that it is indeed them trying to log in, and authentication methods provided by the user's operating system (such as Windows Hello, or Touch ID/Face ID on macOS and iOS) -- are <em>not</em><a> a cognitive function test</a>.</p>
```
mbgower, line 111:
```suggestion
            <p>Note that two-factor systems that do not rely on codes &#8212; including hardware authentication devices (such as YubiKey), secondary applications (either on the same primary device, or on a secondary device) that expect the user to confirm that it is indeed them trying to log in, and authentication methods provided by the user's operating system (such as Windows Hello, or Touch ID/Face ID on macOS and iOS) &#8212; are <em>not</em><a> a cognitive function test</a>.</p>
```

understanding/22/accessible-authentication-minimum.html
dbjorge, line 108: typo:

```suggestion
            <h3>Two-factor authentication systems (verification codes)</h3>
```

understanding/22/accessible-authentication-minimum.html
```diff
@@ -105,10 +105,10 @@ <h3>Copy and paste</h3>
...
```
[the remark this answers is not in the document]
patrickhlauke: ah, ignore me, github was showing me the outdated diff here

PR #3188, opened 2023-05-12
Added examples to Understanding Target Size (Minimum)

bruce-usab:
[Closes #3901](../issues/3091) Does Target Size (Minimum) result in widespread failure of common dataviz?


understanding/22/target-size-minimum.html
```diff
@@ -40,15 +40,15 @@ <h2>In brief</h2>
...
-
+        <p>The intent of this Success Criterion is to help ensure targets can be easily activated without accidentally activating an adjacent target.  Users with dexterity limitations and those who have difficulty with fine motor movement find it difficult to accurately activate small targets when there are other targets that are too close.  Providing sufficient size, or sufficient spacing between targets, will reduce the likelihood of accidentally activating the wrong control.</p>
+        <p>Disabilities addressed by this requirement include hand tremors, spasticity, and quadriplegia.  Some people with disabilities use specialized input devices instead of a computer mouse or trackpad.  Typically these types of input device do not provide as much accuracy as mainstream pointing devices.  Meeting this requirement also ensures that touchscreen interfaces are easier to use.</p>
```
patrickhlauke, line 51: i'd remove the "two spaces after a full stop" typewriter convention here. it's not rendered in web content, and unnecessary in the age of non-monospaced fonts

patrickhlauke (approved):

understanding/22/target-size-minimum.html
bruce-usab: @patrickhlauke the double space after full stop was for my own benefit -- since I am editing in monospace.  Yes, HTML ignores those regardless.

But, as an aside, I am of the opinion that "unnecessary in the age of non-monospaced fonts" is not entirely correct.  Can you point me to a web browser that provides a little extra space between sentences (and after colons), as compared to periods ending abbreviations?  That bit of extra whitespace is very common in print publications and its typography tooling chain, and it really helps with readability.

understanding/22/target-size-minimum.html
```diff
@@ -40,15 +40,15 @@ <h2>In brief</h2>
       <dl>
           <dt>Objective</dt><dd>Make controls easier to activate</dd>
           <dt>Author task</dt><dd>Ensure targets meet a minimum size or have sufficient spacing around them</dd>
-          <dt>Key beneficiaries</dt><dd>Users with some physical disabilities</dd>
+          <dt>Key beneficiaries</dt><dd>Users with limitations to fine motor input</dd>
```
mbgower, line 43: @bruce-usab I tried to keep the beneficiaries quite generic in the 'In brief' material, both for speed (there is a fairly small set of  'canned' beneficiaries to draw from) and so that anyone reading would likely understand what it meant.
I am mildly concerned that if we start to be more specific for this SC, we set ourselves up for a lot of nitpicking on the precision of other Key beneficiary statements. To me, the main Understanding document is where more particulars for different users can be detailed (which you have done). I'd advocate to leave this line as is for now.

understanding/22/target-size-minimum.html
patrickhlauke: > Can you point me to a web browser that provides a little extra space between sentences (and after colons), as compared to periods ending abbreviations?

not sure i understand the question. all browsers, per spec, collapse multiple spaces into a single space (unless it's inside a `<pre>` preformatted element) when displaying/rendering HTML content.

mbgower: I'm going to chime in here.
 @patrickhlauke, agreed that the extras space after the period when published will disappear, as will all the potentially weird line spaces, indents, etc., that some of us use to more easily work with the info in our code viewers. 
But that doesn't mean that _while working in a monospace code viewer_, some of us don't derive benefit from that 'unpreserved' formatting.
I think you two are at risk of sending this comment thread into the weeds :) but I think it might be an idea to work out between you 'offline' when it's okay to not trim spaces.

patrickhlauke: not precious about the double spaces. leave them, remove them. just for consistency, nowhere else in this same file do we have double spaces, so purely from a code consistency point of view...

understanding/22/target-size-minimum.html
mbgower: @bruce-usab you had marked this as resolved without responding. It might help to have the context of #3191 
```suggestion
          <dt>Key beneficiaries</dt><dd>Users with some physical disabilities</dd>
```

PR #3190, opened 2023-05-15
Update text alternative for target size (minimum) understanding illustrations

patrickhlauke:
* corrects the woefully incomplete alt for the target-spacing-toolbar example
* expands the alt for a few of the following illustrations


detlevhfischer (changes requested):
The `alt` value seems overly long, IMO. I'd say it would better to keep the `alt` content itself short and referring to the `figcaption` text directly underneath.

patrickhlauke:
@detlevhfischer assuming you just mean that first one, i'm easy either adding it to the alt, or to the figcaption. but currently neither of them mention specifically what's going on there. i'll see if i can rejig it...

detlevhfischer:
I'd treat all of them the same way: short `alt` with reference on the `figcaption` text underneath - which also helps LV and other users...

patrickhlauke:
i've generally tried to keep the idea along the lines of: alt describes what the image *is*, figcaption describes what the image represents/tries to show... but as said, i'll mull over how to possibly rejig this.

understanding/22/target-size-minimum.html
```diff
@@ -120,22 +120,22 @@ <h3>Spacing</h3>
             <p>The following example shows a horizontal row of icon-based buttons. In the top row, the dimensions of each target are 24 by 24 CSS pixels, passing this Success Criterion. In the second row, the same targets are only 20 by 20 CSS pixels, but have a 4 CSS pixel space between them – as the target size is below 24 by 24 CSS pixels, these need to be evaluated against the Success Criterion's spacing exception, and they pass. In the last row, the targets are again 20 by 20 CSS pixels, but have no space between them – these fail the spacing exception, because when drawing the 24 CSS pixel diameter circles over the targets, the circles intersect.</p>
```
mbgower, line 120:
```suggestion
            <p>The following example shows three versions of a horizontal row of icon-based buttons. In the top row, the dimensions of each target are 24 by 24 CSS pixels, passing this Success Criterion. In the second row, the same targets are only 20 by 20 CSS pixels, but have a 4 CSS pixel space between them – as the target size is below 24 by 24 CSS pixels, these need to be evaluated against the Success Criterion's spacing exception, and they pass. In the last row, the targets are again 20 by 20 CSS pixels, but have no space between them – these fail the spacing exception, because when drawing the 24 CSS pixel diameter circles over the targets, the circles intersect.</p>
```

understanding/22/target-size-minimum.html
```diff
@@ -120,22 +120,22 @@ <h3>Spacing</h3>
...
+                <img src="img/target-spacing-toolbar.svg" width="560" height="265" alt="Three rows showing a horizontal toolbar of icon-based controls with measurements. In the first row, the targets have a dimension of 24 by 24 CSS pixels, so pass. In the second row, the toolbar controls have a dimension of 20 by 20 CSS pixels, with 4 pixel spacing between them. The same toolbar is shown again on the same row, but overlaid with the 24 CSS pixel circles – the circles don't intersect, so this case passes. In the last row, the toolbar controls have a dimension of 20 by 20 CSS pixels, but there is no gap between them. Again, the toolbar is shown on the same row overlaid with the 24 CSS pixel circles – this time, the circles all overlap/intersect, so this case fails." />
```
mbgower, line 123:
```suggestion
                <img src="img/target-spacing-toolbar.svg" width="560" height="265" alt="The second toolbar, with smaller targets, shows 24 CSS pixel circles drawn on each target for assessment. The circles do not intersect and so the targets have enough space to pass. The third shows the same circles on the targets, but they overlap on tightly spaced controls that are only 20 by 20." />
```

patrickhlauke:
ok, tweaked these a little bit. thoughts? (keeping in mind that currently some of the alts/descriptions are plain wrong, so we may want to prioritise correcting them now and refining this...at some point in the future)

understanding/22/target-size-minimum.html
```diff
@@ -117,33 +117,33 @@ <h3>Spacing</h3>
...
 
             <figure id="target-spacing-toolbar">
-                <img src="img/target-spacing-toolbar.svg" width="560" height="265" alt="Three rows showing a horizontal toolbar of icon-based controls with measurements. In the first row, the targets have a dimension of 24 by 24 CSS pixels, so pass. In the second row, the same toolbar is shown again, but overlaid with the 24 CSS pixel circles – the circles don't intersect, so pass. In the last row, the toolbar is shown again with the circles, but the circles intersect, so fail." />
+                <img src="img/target-spacing-toolbar.svg" width="560" height="265" alt="The first toolbar's targets have a dimension of 24 by 24 CSS pixels, so pass. The second toolbar, with smaller targets, shows 24 CSS pixel circles drawn on each target for assessment. The circles do not intersect and so the targets have enough space to pass. The third toolbar shows similar circles on the targets, but the circles intersect due to the lack of spacing between targets, so the toolbar fails." />
```
detlevhfischer, line 123: Would it make sense to have the verdict PASS or FAIL at a consistent place inside `fig caption`? On the image `id="target-text-buttons-two-rows"` the `figcaption`text starts with `<strong>Fail:</strong>` but not on the others.. Those that Pass might start with `<strong>Fail:</strong>`.

detlevhfischer (approved):

understanding/22/target-size-minimum.html
patrickhlauke: As that one with the FAIL was the only one that had it, and most other examples have a mix of both pass and fail, it might make more sense to remove that one `<strong>Fail:</strong>`

detlevhfischer:
@patrickhlauke I see the point about the illustrations where both pass and fail occur so leaving it out in the `figcaption` is probably more consistent. I Looked at the text more properly now. Good stuff.

Some points for consideration:
Once we have settled for a final number of illustrations, it would help numbering them to facilitate reference.

I notice the equivalent exception is not covered in any detail in the Understanding text. The sentence "there is another control that can achieve the underlying function" seems to leave a lot of room for interpretation. Would enlisting all functions in a 6pt text block in the footer with inline links (themselves exempt via the inline exception) qualify as equivalent exception? Does it make a difference if the equivalent function is 

1. on the same page?
2. in a disclosure on the same page?
3. indirectly accessible via a link to an alternative version on another page?

The equivalent exception does not spell this out, and perhaps cannot. Not sure whether this has been discussed before at any length.

The inline exception probably also deserves some more text and illustrations too, likely focussing on helping to determine whether list structures are exempt or fall under the SC (there used to be some which I cobbled together less nicely as Patrick's work which are now gone, possibly because they no longer fitted). 

The essential exception could use a small text section with an image of a map with many location markers in one area.

patrickhlauke:
@detlevhfischer for all other points not directly related to the core "tweak the text alternatives/explanations" of this PR, I'd suggest opening a new issue/PR for them (keeping this one small and more easily mergeable, so that the inaccurate/incomplete alt we currently have can get quickly fixed first before we go down longer rabbithole discussions on the other aspects)

mbgower (approved):
Nice changes, thanks!

patrickhlauke:
@mbgower @alastc any chance this could be considered/merged soon?

PR #3222, opened 2023-05-29
Update use-of-color.html

### LINKED ISSUE #905: Visited and Unvisited link color a failure of Use of Color?
I had a developer ask me how it's possible to meet Use of Color for visited and unvisited links, since they have no other convention than color for differentiation.
We discussed that if we follow the 3:1 relative luminance rule (as per [G183](https://www.w3.org/TR/WCAG20-TECHS/G183.html), if the body, visited and unvisited text are all to be 4.5:1 against the background (assuming a white background), it is virtually impossible to get colors that meet this ratio. For example, I used the [websafe colors posted by the w3c ](https://www.w3.org/WAI/WCAG20/Techniques/working-examples/G183/link-contrast.html) which can be used to meet 3:1 against black to overcome a lack of underline for distinguishing links. None of those combinations rendered anything approaching 3:1.

One could argue that IF underlines were used on all links and IF the unvisited links were uncolored (e.g., remaining black where body text is also black), quite a few colors would then become available which could meet 3:1 against the black (all the colors given in the w3c link above, for example). Thus any link with color would be visited. But I suspect there would be understandable concerns with people about being unable to distinguish between underlined text and links, and that this would in fact be a bigger problem.

Is there any guidance on this? I couldn't find anything.

mbgower:
Closes #905

### DIFF
--- a/understanding/20/use-of-color.html
+++ b/understanding/20/use-of-color.html
@@ -84,7 +84,7 @@
 
       <div class="note">
 
-         <p>Under certain conditions, most user agents provide users with a color-only cue that a link has been previously activated by them ("visited"). However, several technical constraints result in authors having very limited control over these color-only indications of visited links.  The technical constraints are as follows:
+         <p>Most user agents provide users with a color-only cue that a link has been previously activated by them ("visited"). However, several technical constraints result in authors having very limited control over these color-only indications of visited links.  The technical constraints are as follows:
             <ol>
                <li> 
                   User agents constrain the exposure of a link's visited state due to <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/Privacy_and_the_:visited_selector">privacy concerns</a>. Author queries to user agents will indicate all links have not been visited.
--- a/understanding/20/use-of-color.html
+++ b/understanding/20/use-of-color.html
@@ -102,7 +102,7 @@
             </ol>
          </p>
 
-         <p>For these reasons, setting or conveying a link's visited status is not an author responsibility, and does not result in a failure of Use of Color</p>
+         <p>For these reasons, setting or conveying a link's visited status is not an author responsibility, and does not result in a failure of this Success Criterion</p>
       </div>
       
    </section>
--- a/understanding/20/use-of-color.html
+++ b/understanding/20/use-of-color.html
@@ -87,7 +87,7 @@
          <p>Under certain conditions, most user agents provide users with a color-only cue that a link has been visited. However, due to several technical constraints, authors should not be held accountable for this use of color, and color-only indications of visited links should not result in a failure of Use of Color (or a number of other criteria). The technical constraints are as follows:
             <ol>
                <li> 
-                  User agents constrain the exposure of a link's visited state due to <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/Privacy_and_the_:visited_selector">security concerns</a>. Author queries to user agents will indicate all links have not been visited.
+                  User agents constrain the exposure of a link's visited state due to <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/Privacy_and_the_:visited_selector">privacy concerns</a>. Author queries to user agents will indicate all links have not been visited.
                </li>
                <li>
                   Any available information on the visited state of a link would be inacurrate since it is browser-dependent. Even if an author could accurately get information on previously activitated links, the author would be constrained to the current browser's preserved history, and would be unable to determine if the user had visited the page using a different browser (or if the history was not preserved, either from cache clearing or use of private sessions).
--- a/understanding/20/use-of-color.html
+++ b/understanding/20/use-of-color.html
@@ -102,7 +102,7 @@
             </ol>
          </p>
 
-         <p>For these reasons, setting or conveying a link's visited status is not an author responsibility, and does not result in a failure of this Success Criterion</p>
+         <p>For these reasons, setting or conveying a link's visited status is not an author responsibility. Where color alone distinguishes between visited and unvisited links, it does not result in a failure of this Success Criterion, even where the contrast between the two link colors is below 3:1. Note that authors must continue to ensure that all text links meet contrast minimums against the page background (SC 1.4.3).</p>
       </div>
       
    </section>
--- a/understanding/20/use-of-color.html
+++ b/understanding/20/use-of-color.html
@@ -93,7 +93,7 @@
                   Any available information on the visited state of a link would be inacurrate since it is browser-dependent. Even if an author could accurately get information on previously activitated links, the author would be constrained to the current browser's preserved history, and would be unable to determine if the user had visited the page using a different browser (or if the history was not preserved, either from cache clearing or use of private sessions).
                </li>
                <li>
-                  Authors can <em>only</em> use color to modify the <code>:visited</code> CSS psuedoclass style. The technology constrains any other non-color styling. Due to palette limitations, an author cannot achieve 4.5:1 contrast for link and non-link text while also achieving a 3:1 contrast between visited and unvisited links.
+                  Authors can <em>only</em> use color to modify the <code>:visited</code> CSS psuedoclass style. The technology constrains any non-color styling. Due to palette limitations, an author cannot achieve 4.5:1 contrast for link and non-link text while also achieving a 3:1 contrast between visited and unvisited links.
                </li>
                <li>
                   Authors also cannot set the visited state of links. The anchor element does not include a "visited" attribute; therefore the author has no ability to alter the state through an attribute setting. As such, authors cannot achieve <a href="info-and-relationships">1.3.1 Info and Relationships</a> or
--- a/understanding/20/use-of-color.html
+++ b/understanding/20/use-of-color.html
@@ -81,7 +81,29 @@
          </p>
 
       </div>
-      
+
+      <div class="note">
+
+         <p>Under certain conditions, most user agents provide users with a color-only cue that a link has been visited. However, due to several technical constraints, authors should not be held accountable for this use of color, and color-only indications of visited links should not result in a failure of Use of Color (or a number of other criteria). The technical constraints are as follows:
+            <ol>
+               <li> 
+                  User agents constrain the exposure of a link's visited state due to <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/Privacy_and_the_:visited_selector">security concerns</a>. Author queries to user agents will indicate all links have not been visited.
+               </li>
+               <li>
+                  Any available information on the visited state of a link would be inacurrate since it is browser-dependent. Even if an author could accurately get information on previously activitated links, the author would be constrained to the current browser's preserved history, and would be unable to determine if the user had visited the page using a different browser (or if the history was not preserved, either from cache clearing or use of private sessions).
+               </li>
+               <li>
+                  Authors can <em>only</em> use color to modify the <code>:visited</code> CSS psuedoclass style. The technology constrains any other non-color styling. Due to palette limitations, an author cannot achieve 4.5:1 contrast for link and non-link text while also achieving a 3:1 contrast between visited and unvisited links.
+               </li>
+               <li>
+                  Authors also cannot set the visited state of links. The anchor element does not include a "visited" attribute; therefore the author has no ability to alter the state through an attribute setting. As such, authors cannot achieve <a href="info-and-relationships">1.3.1 Info and Relationships</a> or
+                  <a href="name-role-value">4.1.2 Name, Role, Value</a> in regard to visited links.
+               </li>            
+            </ol>
+         </p>
+
+         <p>For these reasons, setting or conveying a link's visited status is not an author responsibility, and does not result in a failure of Use of Color</p>
+      </div>
       
    </section>
    <section id="benefits">
--- a/understanding/20/use-of-color.html
+++ b/understanding/20/use-of-color.html
@@ -84,7 +84,7 @@
 
       <div class="note">
 
-         <p>Under certain conditions, most user agents provide users with a color-only cue that a link has been activated by them ("visited"). However, several technical constraints result in authors having very limited control over this use of color, and color-only indications of visited links.  The technical constraints are as follows:
+         <p>Under certain conditions, most user agents provide users with a color-only cue that a link has been activated by them ("visited"). However, several technical constraints result in authors having very limited control over this these color-only indications of visited links.  The technical constraints are as follows:
             <ol>
                <li> 
                   User agents constrain the exposure of a link's visited state due to <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/Privacy_and_the_:visited_selector">privacy concerns</a>. Author queries to user agents will indicate all links have not been visited.
--- a/understanding/20/use-of-color.html
+++ b/understanding/20/use-of-color.html
@@ -93,7 +93,7 @@
                   Any available information on the visited state of a link would be inaccurate since it is both user and browser-dependent. Even if an author could accurately get information on previously activated links by a certain user, the author would be constrained to the current browser's preserved history, and would be unable to determine if the user had visited the page using a different browser (or if the history was not preserved, either from cache clearing or use of private sessions).
                </li>
                <li>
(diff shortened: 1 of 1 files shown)


understanding/20/use-of-color.html
```diff
@@ -81,7 +81,29 @@ <h2>Intent of Use of Color</h2>
          </p>
 
       </div>
-      
+
+      <div class="note">
+
+         <p>Under certain conditions, most user agents provide users with a color-only cue that a link has been visited. However, due to several technical constraints, authors should not be held accountable for this use of color, and color-only indications of visited links should not result in a failure of Use of Color (or a number of other criteria). The technical constraints are as follows:
```
patrickhlauke, line 87: possibly instead of the `authors should not be held accountable for ...` angle, it may be a bit more "neutral" (rather than emphasising / suggesting WCAG is a way to assign culpability) to rephrase this as `authors have limited control over ...`

mbgower, line 87:
```suggestion
         <p>Under certain conditions, most user agents provide users with a color-only cue that a link has been visited. However, several technical constraints result in authors having very limited control over this use of color, and color-only indications of visited links  The technical constraints are as follows:
```

mbgower: Made some changes. LMK if they suffice

understanding/20/use-of-color.html
```diff
@@ -81,7 +81,29 @@ <h2>Intent of Use of Color</h2>
...
+            <ol>
+               <li> 
+                  User agents constrain the exposure of a link's visited state due to <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/Privacy_and_the_:visited_selector">security concerns</a>. Author queries to user agents will indicate all links have not been visited.
```
mbgower, line 90:
```suggestion
                  User agents constrain the exposure of a link's visited state due to <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/Privacy_and_the_:visited_selector">privacy concerns</a>. Author queries to user agents will indicate all links have not been visited.
```

understanding/20/use-of-color.html
```diff
@@ -81,7 +81,29 @@ <h2>Intent of Use of Color</h2>
...
+            <ol>
+               <li> 
+                  User agents constrain the exposure of a link's visited state due to <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/Privacy_and_the_:visited_selector">privacy concerns</a>. Author queries to user agents will indicate all links have not been visited.
+               </li>
+               <li>
+                  Any available information on the visited state of a link would be inacurrate since it is browser-dependent. Even if an author could accurately get information on previously activitated links, the author would be constrained to the current browser's preserved history, and would be unable to determine if the user had visited the page using a different browser (or if the history was not preserved, either from cache clearing or use of private sessions).
```
mbgower, line 93:
```suggestion
                  Any available information on the visited state of a link would be inaccurate since it is both user and browser-dependent. Even if an author could accurately get information on previously activated links by a certain user, the author would be constrained to the current browser's preserved history, and would be unable to determine if the user had visited the page using a different browser (or if the history was not preserved, either from cache clearing or use of private sessions).
```

understanding/20/use-of-color.html
```diff
@@ -81,7 +81,29 @@ <h2>Intent of Use of Color</h2>
          </p>
 
       </div>
-      
+
+      <div class="note">
+
+         <p>Under certain conditions, most user agents provide users with a color-only cue that a link has been visited. However, several technical constraints result in authors having very limited control over this use of color, and color-only indications of visited links.  The technical constraints are as follows:
```
mbgower, line 87:
```suggestion
         <p>Under certain conditions, most user agents provide users with a color-only cue that a link has been activated by them ("visited"). However, several technical constraints result in authors having very limited control over this use of color, and color-only indications of visited links.  The technical constraints are as follows:
```

understanding/20/use-of-color.html
```diff
@@ -81,7 +81,29 @@ <h2>Intent of Use of Color</h2>
          </p>
 
       </div>
-      
+
+      <div class="note">
+
+         <p>Under certain conditions, most user agents provide users with a color-only cue that a link has been activated by them ("visited"). However, several technical constraints result in authors having very limited control over this use of color, and color-only indications of visited links.  The technical constraints are as follows:
```
mbgower, line 87:
```suggestion
         <p>Under certain conditions, most user agents provide users with a color-only cue that a link has been activated by them ("visited"). However, several technical constraints result in authors having very limited control over this these color-only indications of visited links.  The technical constraints are as follows:
```

understanding/20/use-of-color.html
```diff
@@ -81,7 +81,29 @@ <h2>Intent of Use of Color</h2>
...
+                  User agents constrain the exposure of a link's visited state due to <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/Privacy_and_the_:visited_selector">privacy concerns</a>. Author queries to user agents will indicate all links have not been visited.
+               </li>
+               <li>
+                  Any available information on the visited state of a link would be inaccurate since it is both user and browser-dependent. Even if an author could accurately get information on previously activated links by a certain user, the author would be constrained to the current browser's preserved history, and would be unable to determine if the user had visited the page using a different browser (or if the history was not preserved, either from cache clearing or use of private sessions).
+               </li>
+               <li>
+                  Authors can <em>only</em> use color to modify the <code>:visited</code> CSS psuedoclass style. The technology constrains any non-color styling. Due to palette limitations, an author cannot achieve 4.5:1 contrast for link and non-link text while also achieving a 3:1 contrast between visited and unvisited links.
```
mbgower, line 96:
```suggestion
                  Authors can <em>only</em> use color to modify the <code>:visited</code> CSS psuedoclass style. The technology constrains any non-color styling. Due to palette limitations, an author cannot use color alone to achieve 3:1 contrast between link and non-line text as well as between visited and unvisited links while also achieving 4.5:1 contrast for all link and non-link text.
```

understanding/20/use-of-color.html
```diff
@@ -81,7 +81,29 @@ <h2>Intent of Use of Color</h2>
          </p>
 
       </div>
-      
+
+      <div class="note">
+
+         <p>Under certain conditions, most user agents provide users with a color-only cue that a link has been activated by them ("visited"). However, several technical constraints result in authors having very limited control over this these color-only indications of visited links.  The technical constraints are as follows:
```
patrickhlauke, line 87: `over this these` > `over these`

understanding/20/use-of-color.html
```diff
@@ -81,7 +81,29 @@ <h2>Intent of Use of Color</h2>
...
+                  Authors can <em>only</em> use color to modify the <code>:visited</code> CSS psuedoclass style. The technology constrains any non-color styling. Due to palette limitations, an author cannot use color alone to achieve 3:1 contrast between link and non-line text as well as between visited and unvisited links while also achieving 4.5:1 contrast for all link and non-link text.
+               </li>
+               <li>
+                  Authors also cannot set the visited state of links. The anchor element does not include a "visited" attribute; therefore the author has no ability to alter the state through an attribute setting. As such, authors cannot achieve <a href="info-and-relationships">1.3.1 Info and Relationships</a> or
+                  <a href="name-role-value">4.1.2 Name, Role, Value</a> in regard to visited links.
+               </li>            
+            </ol>
+         </p>
+
+         <p>For these reasons, setting or conveying a link's visited status is not an author responsibility, and does not result in a failure of Use of Color</p>
```
patrickhlauke, line 105: maybe change `failure of Use of Color` to `failure of this Success Criterion` (similar to how many other SC understandings self-reference)

understanding/20/use-of-color.html
patrickhlauke: lovely, much better

understanding/20/use-of-color.html
mbgower, line 87:
```suggestion
         <p>Under certain conditions, most user agents provide users with a color-only cue that a link has been activated by them ("visited"). However, several technical constraints result in authors having very limited control over these color-only indications of visited links.  The technical constraints are as follows:
```

understanding/20/use-of-color.html
```diff
@@ -81,7 +81,29 @@ <h2>Intent of Use of Color</h2>
          </p>
 
       </div>
-      
+
+      <div class="note">
+
+         <p>Under certain conditions, most user agents provide users with a color-only cue that a link has been activated by them ("visited"). However, several technical constraints result in authors having very limited control over these color-only indications of visited links.  The technical constraints are as follows:
```
patrickhlauke, line 87: `a color-only cue that a link has been activated`

re-reading this, wonder if this may be dangerously close to the idea of `:active`? maybe better to be even more explicit, a la "...that a links points to a resource that has previously been visited"

mbgower: > that a links points to a resource that has previously been visited

But is that really what visited link state indicates? Isn't it more accurate to say that it represents that a link has been previously activated? After all,  the target might have timed out -- or the target may even have been modified. Doesn't the link get changed to 'visited' when it has been activated, not when the target is visited? In other words, there is no auditing of a successful connection AFAIK.

patrickhlauke: well then, perhaps "previously activated"? just something to avoid confounding matters with the actual momentary `:active` state...

mbgower, line 87:
```suggestion
         <p>Under certain conditions, most user agents provide users with a color-only cue that a link has been previously activated by them ("visited"). However, several technical constraints result in authors having very limited control over these color-only indications of visited links.  The technical constraints are as follows:
```

understanding/20/use-of-color.html
```diff
@@ -81,7 +81,29 @@ <h2>Intent of Use of Color</h2>
...
+               </li>
+               <li>
+                  Any available information on the visited state of a link would be inaccurate since it is both user and browser-dependent. Even if an author could accurately get information on previously activated links by a certain user, the author would be constrained to the current browser's preserved history, and would be unable to determine if the user had visited the page using a different browser (or if the history was not preserved, either from cache clearing or use of private sessions).
+               </li>
+               <li>
+                  Authors can <em>only</em> use color to modify the <code>:visited</code> CSS psuedoclass style. The technology constrains any non-color styling. Due to palette limitations, an author cannot use color alone to achieve 3:1 contrast between link and non-line text as well as between visited and unvisited links while also achieving 4.5:1 contrast for all link and non-link text.
```
fstrr, line 96: psuedoclass should be pseudo-class

bruce-usab (dismissed):
> Under certain conditions, most user agents...

This opening line seems a bit contradictory (most vs only certain).  Maybe:

> Under default conditions, most user agents...

or
 
> It is usually the case that most user agents...

understanding/20/use-of-color.html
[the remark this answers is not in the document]
bruce-usab: Good catch!

understanding/20/use-of-color.html
```diff
@@ -81,7 +81,29 @@ <h2>Intent of Use of Color</h2>
...
+               </li>
+               <li>
+                  Authors also cannot set the visited state of links. The anchor element does not include a "visited" attribute; therefore the author has no ability to alter the state through an attribute setting. As such, authors cannot achieve <a href="info-and-relationships">1.3.1 Info and Relationships</a> or
+                  <a href="name-role-value">4.1.2 Name, Role, Value</a> in regard to visited links.
+               </li>            
+            </ol>
+         </p>
+
+         <p>For these reasons, setting or conveying a link's visited status is not an author responsibility, and does not result in a failure of this Success Criterion</p>
```
dbjorge, line 105: I worry that this concluding sentence might lead a reader to think that a link in the visited state can never be a use-of-color failure. I think our intent is that:

* Authors **do not** need to ensure that visited and unvisited links have 3:1 contrast with one another
* Authors **do** need to ensure that **both** visited and unvisited links have 3:1 contrast with surrounding non-link text

Have I understood correctly? If so, could we make this distinction more clear in the conclusion?

mbgower: Yes, I confirm that is the intent. i will undertake to incorporate that change

understanding/20/use-of-color.html
```diff
@@ -81,7 +81,29 @@ <h2>Intent of Use of Color</h2>
...
+               </li>
+               <li>
+                  Any available information on the visited state of a link would be inaccurate since it is both user and browser-dependent. Even if an author could accurately get information on previously activated links by a certain user, the author would be constrained to the current browser's preserved history, and would be unable to determine if the user had visited the page using a different browser (or if the history was not preserved, either from cache clearing or use of private sessions).
+               </li>
+               <li>
+                  Authors can <em>only</em> use color to modify the <code>:visited</code> CSS psuedoclass style. The technology constrains any non-color styling. Due to palette limitations, an author cannot use color alone to achieve 3:1 contrast between link and non-link text as well as between visited and unvisited links while also achieving 4.5:1 contrast for all link and non-link text.
```
mbgower, line 96:
```suggestion
                  Authors can <em>only</em> use color to modify the <code>:visited</code> CSS pseudoclass style. The technology constrains any non-color styling. Due to palette limitations, an author cannot use color alone to achieve 3:1 contrast between link and non-link text as well as between visited and unvisited links while also achieving 4.5:1 contrast for all link and non-link text.
```

understanding/20/use-of-color.html
```diff
@@ -81,7 +81,29 @@ <h2>Intent of Use of Color</h2>
...
+               </li>
+               <li>
+                  Authors also cannot set the visited state of links. The anchor element does not include a "visited" attribute; therefore the author has no ability to alter the state through an attribute setting. As such, authors cannot achieve <a href="info-and-relationships">1.3.1 Info and Relationships</a> or
+                  <a href="name-role-value">4.1.2 Name, Role, Value</a> in regard to visited links.
+               </li>            
+            </ol>
+         </p>
+
+         <p>For these reasons, setting or conveying a link's visited status is not an author responsibility, and does not result in a failure of this Success Criterion</p>
```
alastc, line 105:
```suggestion
         <p>For these reasons, setting or conveying a link's visited status (compared to unvisited links) is not an author responsibility, and does not result in a failure of this Success Criterion</p>
```

understanding/20/use-of-color.html
```diff
@@ -81,7 +81,29 @@ <h2>Intent of Use of Color</h2>
...
+                  Authors can <em>only</em> use color to modify the <code>:visited</code> CSS pseudoclass style. The technology constrains any non-color styling. Due to palette limitations, an author cannot use color alone to achieve 3:1 contrast between link and non-link text as well as between visited and unvisited links while also achieving 4.5:1 contrast for all link and non-link text.
+               </li>
+               <li>
+                  Authors also cannot set the visited state of links. The anchor element does not include a "visited" attribute; therefore the author has no ability to alter the state through an attribute setting. As such, authors cannot achieve <a href="info-and-relationships">1.3.1 Info and Relationships</a> or
+                  <a href="name-role-value">4.1.2 Name, Role, Value</a> in regard to visited links.
+               </li>            
+            </ol>
+         </p>
+
+         <p>For these reasons, setting or conveying a link's visited status is not an author responsibility, and does not result in a failure of this Success Criterion</p>
```
mbgower, line 105:
```suggestion
         <p>For these reasons, setting or conveying a link's visited status is not an author responsibility. While authors must ensure that all links meet text contrast (SC 1.4.3), using color alone to distinguish between visited and unvisited links does not result in a failure of this Success Criterion</p>
```

understanding/20/use-of-color.html
```diff
@@ -81,7 +81,29 @@ <h2>Intent of Use of Color</h2>
          </p>
 
       </div>
-      
+
+      <div class="note">
+
+         <p>Under certain conditions, most user agents provide users with a color-only cue that a link has been previously activated by them ("visited"). However, several technical constraints result in authors having very limited control over these color-only indications of visited links.  The technical constraints are as follows:
```
mbgower, line 87:
```suggestion
         <p>Most user agents provide users with a color-only cue that a link has been previously activated by them ("visited"). However, several technical constraints result in authors having very limited control over these color-only indications of visited links.  The technical constraints are as follows:
```

understanding/20/use-of-color.html
mbgower: @alastc I did something a bit more pronounced in the below suggestion. I'd appreciate your feedback.

alastc (approved):

PR #3224, opened 2023-05-31
Decision policy adjustments

alastc:


patrickhlauke:
@AbdullahWins well done for your very low effort, low gain way of spamming lots of issues here on w3c with your pointless approvals...

PR #3242, opened 2023-06-14
Remove custom decision policy

rachaelbradley:
Changes remove the link to the custom decision policy and revise the time limit for CFCs.


fstrr:
@rachaelbradley This is from 2022—does it still need to be merged in or is it out of date now?

PR #3252, opened 2023-06-26
Wcag22inbrief eow

mbgower:
Including the "Why" phrase from EOWG, and repositioning above the normative text.

Previews of each:
- [Accessible Authentication Min](https://raw.githack.com/w3c/wcag/wcag22inbriefEOW/understanding/22/accessible-authentication-minimum.html)
- [Accessible Authentication enh](https://raw.githack.com/w3c/wcag/wcag22inbriefEOW/understanding/22/accessible-authentication-enhanced.html)
- [Consistent help](https://raw.githack.com/w3c/wcag/wcag22inbriefEOW/understanding/22/consistent-help.html)
- [Dragging movements](https://raw.githack.com/w3c/wcag/wcag22inbriefEOW/understanding/22/dragging-movements.html)
- [Focus appearance](https://raw.githack.com/w3c/wcag/wcag22inbriefEOW/understanding/22/focus-appearance.html)
- [Focus not obscured enh](https://raw.githack.com/w3c/wcag/wcag22inbriefEOW/understanding/22/focus-not-obscured-enhanced.html)
- [Focus not obscured min](https://raw.githack.com/w3c/wcag/wcag22inbriefEOW/understanding/22/focus-not-obscured-minimum.html)
- [Redundant entry](https://raw.githack.com/w3c/wcag/wcag22inbriefEOW/understanding/22/redundant-entry.html)
- [Target size min](https://raw.githack.com/w3c/wcag/wcag22inbriefEOW/understanding/22/target-size-minimum.html)


understanding/22/accessible-authentication-enhanced.html
```diff
@@ -12,6 +12,16 @@ <h1>Understanding Accessible Authentication (Enhanced)</h1>
         <h2>Status</h2>
         <p>This understanding document is part of the <a href="https://w3c.github.io/wcag/guidelines/22/"><strong>draft</strong> WCAG 2.2 content</a>. It may change or be removed before the final WCAG 2.2 is published.</p>
     </section> 
+   
+    <section id="brief">
+        <h2>In brief</h2>
+        <dl>
+            <dt>Goal</dt><dd>Make logins possible with less mental effort.</dd>
+            <dt>What to do</dt><dd>Don’t make people recognize objects or personal information to login.</dd>
```
mbgower, line 20:
```suggestion
            <dt>What to do</dt><dd>Don’t make people recognize objects or recall personal information to login.</dd>
```

understanding/22/accessible-authentication-enhanced.html
```diff
@@ -12,6 +12,16 @@ <h1>Understanding Accessible Authentication (Enhanced)</h1>
         <h2>Status</h2>
         <p>This understanding document is part of the <a href="https://w3c.github.io/wcag/guidelines/22/"><strong>draft</strong> WCAG 2.2 content</a>. It may change or be removed before the final WCAG 2.2 is published.</p>
     </section> 
+   
+    <section id="brief">
+        <h2>In brief</h2>
+        <dl>
+            <dt>Goal</dt><dd>Make logins possible with less mental effort.</dd>
+            <dt>What to do</dt><dd>Don’t make people recognize objects or recall personal information to login.</dd>
```
dbjorge, line 20: "personal information" makes me think of the name/email/phone-number that we explicitly allow recall tests for - the understanding document allows them specifically *because* they are "personal to the user". Maybe "memorize information" instead of "recall personal information"?

understanding/22/accessible-authentication-enhanced.html
```diff
@@ -12,6 +12,16 @@ <h1>Understanding Accessible Authentication (Enhanced)</h1>
...
+            <dt>Why it's important</dt><dd>Some people with cognitive disabilities cannot identify objects.</dd>
```
melaniephilipp, line 21: WE know that this SC builds on 3.3.8, but taken in isolation the WHY for this SC only addresses part of the WHAT and appears incomplete. Maybe something like: "Some people with cognitive disabilities cannot memorize a username and password, or type them, nor identify objects."

patrickhlauke: yes, I'm sure I made a similar point previously. the "enhanced" ones should be additive here, not just point out the delta to the "minimum"

mbgower: I  want to set the context that for almost all these SCs, the language of the first two lines is not being changed. It's being repositioned, and a third line added.
So, while I think this round gives us a chance to revisit existing wording -- _especially_ in a case like this where the wording _has_ been changed for this PR -- I want to emphasize that the third line (and overall repositioning) is the focus of the PR (and I get that the repositioning may motivate some of these comments).

Just trying to manage expectations :)

understanding/22/accessible-authentication-enhanced.html
mbgower: I  want to set the context that for almost all these SCs, the language of the first two lines is not being changed. It's being repositioned, and a third line added.
So, while I think this round gives us a chance to revisit existing wording -- _especially_ in a case like this where the wording _has_ been changed for this PR -- I want to emphasize that the third line (and overall repositioning) is the focus of the PR (and I get that the repositioning may motivate some of these comments).

Just trying to manage expectations :)

understanding/22/accessible-authentication-enhanced.html
mbgower: Remember that they're going to pass AA with those other details.
I'm not convinced there's a sufficient win by making these cumulative. They become harder to 'spot' the difference, placing more demand on the user to do comparisons.  But I'm also willing to explore. :)
I suspect the same overall response from the WG (as per Patrick's and others earlier comments) will apply here: it's more valuable to emphasize the addition or otherwise make the difference obviously, where possible.

patrickhlauke: then make it explicit with something like "in addition to its 'Minimum' counterpart, ..."

unfortunately, the way our 'Enhanced' SCs themselves have been worded *is* cumulative, so the brief overview should really reflect that. yes, *in practice* the whole SC only brings an extra aspect to the table, but the SC itself is written to include the AA norms as well

understanding/22/accessible-authentication-minimum.html
```diff
@@ -13,6 +13,16 @@ <h2>Status</h2>
         <p>This understanding document is part of the <a href="https://w3c.github.io/wcag/guidelines/22/"><strong>draft</strong> WCAG 2.2 content</a>. It may change or be removed before the final WCAG 2.2 is published.</p>
     </section> 
 
+    <section id="brief">
+        <h2>In brief</h2>
+        <dl>
+            <dt>Goal</dt><dd>Make logins possible with less mental effort.</dd>
+            <dt>What to do</dt><dd>Don’t make people solve, memorize, or transcribe something to log in.</dd>
+            <dt>Why it's important</dt><dd>Some people with cognitive disabilities cannot memorize a username and password, or type them.</dd>
```
patrickhlauke, line 21:
```suggestion
            <dt>Why it's important</dt><dd>Some people with cognitive disabilities cannot solve puzzles, memorize a username and password, or manually type them.</dd>
```

patrickhlauke: "manually" in my suggestion is perhaps redundant, but could do with something that conveys the idea of transcribing. maybe "retyping"?

understanding/22/focus-appearance.html
```diff
@@ -13,7 +13,16 @@ <h2>Status</h2>
         <p>This understanding document is part of the <a href="https://w3c.github.io/wcag/guidelines/22/"><strong>draft</strong> WCAG 2.2 content</a>. It may change or be removed before the final WCAG 2.2 is published.</p>
     </section>
 
+    <section id="brief">
+        <h2>In brief</h2>
+        <dl>
+            <dt>Goal</dt><dd>Make it easier to spot the keyboard focus.</dd>
+            <dt>What to do</dt><dd>Use a focus indicator of sufficient size and contrast.</dd>
+            <dt>Why it's important</dt><dd>Many people can't see small contrast differences, including older people.</dd>
```
patrickhlauke, line 21:
```suggestion
            <dt>Why it's important</dt><dd>Many people can't see small or low contrast differences, including older people.</dd>
```

understanding/22/redundant-entry.html
```diff
@@ -13,6 +13,16 @@ <h2>Status</h2>
       <p>This understanding document is part of the <a href="https://w3c.github.io/wcag/guidelines/22/"><strong>draft</strong> WCAG 2.2 content</a>. It may change or be removed before the final WCAG 2.2 is published.</p>
   </section>
 
+  <section id="brief">
+   <h2>In brief</h2>
+   <dl>
+      <dt>Goal</dt><dd>Make it easier for users to complete multi-step processes.</dd>
+      <dt>What to do</dt><dd>Don't ask for the same information twice in the same session.</dd>
+      <dt>Why it's important</dt><dd>Some people with cognitive disabilities have difficulty remember what they entered before.</dd>
```
patrickhlauke, line 21:
```suggestion
      <dt>Why it's important</dt><dd>Some people with cognitive disabilities have difficulty remembering what they entered before, or may make a mistake when re-entering the same information more than once.</dd>
```

understanding/22/target-size-minimum.html
```diff
@@ -13,6 +13,15 @@ <h2>Status</h2>
         <p>This understanding document is part of the <a href="https://w3c.github.io/wcag/guidelines/22/"><strong>draft</strong> WCAG 2.2 content</a>. It may change or be removed before the final WCAG 2.2 is published.</p>
     </section>
 
+    <section id="brief">
+        <h2>In brief</h2>
+        <dl>
+            <dt>Goal</dt><dd>Make controls easier to activate.</dd>
+            <dt>What to do</dt><dd>Ensure targets meet a minimum size or have sufficient spacing around them.</dd>
+            <dt>Why it's important</dt><dd>Some people with physical impairments cannot click small buttons that are close together.</dd>
```
patrickhlauke, line 21:
```suggestion
            <dt>Why it's important</dt><dd>Some people with physical impairments find it hard to click small buttons that are close together.</dd>
```

understanding/22/focus-appearance.html
mbgower: Hmm, I don't think they meant "small contrast" here to mean small size... How about
```suggestion
            <dt>Why it's important</dt><dd>Many people can't see small changes in contrast, including older people.</dd>
```

understanding/22/redundant-entry.html
mbgower: How about 
```suggestion
      <dt>Why it's important</dt><dd>Some people with cognitive disabilities have difficulty remember or re-entering what they entered before.</dd>
```

mbgower, line 21:
```suggestion
      <dt>Why it's important</dt><dd>Some people with cognitive disabilities have difficulty remembering what they entered before.</dd>
```

understanding/22/target-size-minimum.html
mbgower: I believe the EOWG would like to emphasize inability over difficulty, and there are some people for whom this more of a 'cannot' do rather than a 'not easy'.

understanding/22/focus-appearance.html
patrickhlauke: ah, but then...the "sufficient size" part of the "what to do" isn't really touched on

understanding/22/redundant-entry.html
patrickhlauke: fair enough, i mis-remembered (hah) that the understanding touched on the fact that if you ask the same thing twice, you're doubling the chances of somebody getting it accidentally wrong.

however, the change from "remember" to "remembering" is still needed though

understanding/22/accessible-authentication-enhanced.html
mbgower: We previously had "user-supplied information" approved when we adopted the In Brief, and we can just fall back to that -- although this modification was based on user feedback AND arguably the prior text has the same potential for confusion

The wording of the SC is 
> Personal Content
The cognitive function test is to identify non-text content the user provided to the website.

with a note that non-text content "may be represented by images, video, or audio."

How about 
```suggestion
            <dt>What to do</dt><dd>Don’t make people recognize objects or user-provided images and videos to login.</dd>
```

understanding/22/accessible-authentication-minimum.html
```diff
@@ -13,6 +13,16 @@ <h2>Status</h2>
         <p>This understanding document is part of the <a href="https://w3c.github.io/wcag/guidelines/22/"><strong>draft</strong> WCAG 2.2 content</a>. It may change or be removed before the final WCAG 2.2 is published.</p>
     </section> 
 
+    <section id="brief">
+        <h2>In brief</h2>
+        <dl>
+            <dt>Goal</dt><dd>Make logins possible with less mental effort.</dd>
+            <dt>What to do</dt><dd>Don’t make people solve, memorize, or transcribe something to log in.</dd>
```
mbgower, line 20:
```suggestion
            <dt>What to do</dt><dd>Don’t make people solve, recall, or transcribe something to log in.</dd>
```

understanding/22/accessible-authentication-minimum.html
mbgower: I think we can be a bit more specific to address the main use case
```suggestion
            <dt>Why it's important</dt><dd>Some people with cognitive disabilities cannot solve puzzles, memorize a username and password, or transcribe one-time passcodes.</dd>
```

understanding/22/focus-appearance.html
mbgower: why not? A smaller change is more subtle. The Why is supposed to be illustrative, not a 1:1 response to each point in the wording of the goal or normative text.

understanding/22/accessible-authentication-enhanced.html
mbgower:
```suggestion
            <dt>Why it's important</dt><dd>Some people with cognitive disabilities are excluded by cognitive function test, including object identification.</dd>
```

understanding/22/accessible-authentication-enhanced.html
shawna-slh: "personal content" isn't clear to me.

understanding/22/accessible-authentication-enhanced.html
shawna-slh: I'd like to keep the wording simple and direct. maybe:
> Some people with cognitive disabilities can't do puzzles and tests, and can't identify objects.

understanding/22/accessible-authentication-minimum.html
shawna-slh: hummm - is there something simpler than "transcribe"?

understanding/22/focus-appearance.html
shawna-slh: I think "subtle changes in visual appearance" is not specific enough and also uses bigger words and more abstract concept

mbgower: changing to "small changes", which I think addresses both concerns.

understanding/22/accessible-authentication-enhanced.html
mbgower:
```suggestion
            <dt>Why it's important</dt><dd>Some people with cognitive disabilities can't do puzzles and tests, or identify objects and media.</dd>
```

understanding/22/accessible-authentication-enhanced.html
mbgower: Okay, falling back to previously approved language, let's make it
```suggestion
            <dt>What to do</dt><dd>Don’t make people recognize objects or user-supplied images to login.</dd>
```

understanding/22/accessible-authentication-enhanced.html
```diff
@@ -12,6 +12,16 @@ <h1>Understanding Accessible Authentication (Enhanced)</h1>
         <h2>Status</h2>
         <p>This understanding document is part of the <a href="https://w3c.github.io/wcag/guidelines/22/"><strong>draft</strong> WCAG 2.2 content</a>. It may change or be removed before the final WCAG 2.2 is published.</p>
     </section> 
+   
+    <section id="brief">
+        <h2>In brief</h2>
+        <dl>
+            <dt>Goal</dt><dd>Make logins possible with less mental effort.</dd>
+            <dt>What to do</dt><dd>Don’t make people recognize objects or user-supplied non-text information to login.</dd>
+            <dt>Why it's important</dt><dd>Some people with cognitive disabilities can't do puzzles and tests, including identify objects and non-text information they previously supplied.</dd>
```
mbgower, line 21:
```suggestion
            <dt>Why it's important</dt><dd>Some people with cognitive disabilities can't do puzzles and tests, including identifying objects and non-text information they previously supplied.</dd>
```

understanding/22/accessible-authentication-enhanced.html
```diff
@@ -12,6 +12,16 @@ <h1>Understanding Accessible Authentication (Enhanced)</h1>
         <h2>Status</h2>
         <p>This understanding document is part of the <a href="https://w3c.github.io/wcag/guidelines/22/"><strong>draft</strong> WCAG 2.2 content</a>. It may change or be removed before the final WCAG 2.2 is published.</p>
     </section> 
+   
+    <section id="brief">
+        <h2>In brief</h2>
+        <dl>
+            <dt>Goal</dt><dd>Make logins possible with less mental effort.</dd>
+            <dt>What to do</dt><dd>Don’t make people recognize objects or user-supplied non-text information to login.</dd>
+            <dt>Why it's important</dt><dd>Some people with cognitive disabilities can't do puzzles and tests, including identifying objects and non-text information they previously supplied.</dd>
```
mbgower, line 21:
```suggestion
            <dt>Why it's important</dt><dd>Some people with cognitive disabilities can't do puzzles, including identifying objects and non-text information they previously supplied.</dd>
```

mbgower:
Received the following comments from @gundulaniemann in survey. I am going to respond here for posterity:

>accessible authentication:
I am confused as to what is restricted with which accessible authentication level.
The what to do
"Don’t make people recognize objects or recall personal information to login"
and
"Don’t make people solve, memorize, or transcribe something to log in."
to me basically means the same. Rather the second restricts more, yet it was cited from the 'minimum' version.

You aren't alone in your observation that the SCs are cumulative but ALSO redundant. It make a challenge providing a succinct summary that stands alone. The WG previously voted _not_ to restate the earlier requirement as part of the enhanced 'in brief'. I've nonetheless _tried_ to add a tiny bit of connecting text, which has been reworded. Please review to see if it's any better.

>Dragging Movements:
The SC is about mouse and touch, and the point is that the user cannot perform a reliable and exact dragging movement. This should be reflected int he Why it's important section.
Suggestion:
Some people cannot perform an exact dragging movement.

The EO folks are interested in focusing on inability as opposed to difficulty, hence "Some people cannot use a mouse to drag items". As well, although you are correct that touchscreens are relevant, the use of "mouse" was felt to be more immediately recognizable than "pointer", which is nonetheless used in the what to do section -- so 'mouse' provides a recognizable synonym.  

>Focus not obscured enhanced:
The point is not, what people do not use nor whether they would be able to use a mouse, but what users in fact chose to use.
Why it's important: People who use a keyboard need to easily see what has keyboard focus.

>focus not obscured minimum:
The point is not, what people do not use nor whether they would be able to use a mouse, but what users in fact chose to use.
Why it's important: People who use a keyboard need to see what has keyboard focus.

Getting back to the EO perspective, many people not as familiar with accessibility will more quickly understand 'oh, it's for people who can't use a mouse'. The keyboard is obviously present in the same sentence and in the prior line. And of course the Understanding document provides in depth infor on the needs of those reliant on the keyboard API. This wording also gets us away from saying "keyboard" twice in one brief line!

>redundant entry:
not sure about the grammar as I not a native speaker.
Some people with cognitive disabilities have difficulty _to_ remember what they entered before.

Corrected

>target size minimum:
I prefer "cannot" instead of "find it hard to".

Adopted.

understanding/22/accessible-authentication-enhanced.html
```diff
@@ -12,6 +12,16 @@ <h1>Understanding Accessible Authentication (Enhanced)</h1>
         <h2>Status</h2>
         <p>This understanding document is part of the <a href="https://w3c.github.io/wcag/guidelines/22/"><strong>draft</strong> WCAG 2.2 content</a>. It may change or be removed before the final WCAG 2.2 is published.</p>
     </section> 
+   
+    <section id="brief">
+        <h2>In brief</h2>
+        <dl>
+            <dt>Goal</dt><dd>Make logins possible with less mental effort.</dd>
+            <dt>What to do</dt><dd>Don’t make people recognize objects or user-supplied non-text information to login.</dd>
```
mbgower, line 20: Since we are saying "non-text information" in the Why as well now, I wonder if we should swap it out here?
```suggestion
            <dt>What to do</dt><dd>Don’t make people recognize objects or user-supplied images and media to login.</dd>
```

mbgower: I've been toying with trying to make this _somehow_ reference the minimum, but it is VERY hard to do unless one assumes a user is a aware there is a minimum version, in which case, they theoretically already understand this is an additional requirement...
```suggestion
            <dt>What to do</dt><dd>In addition to the minimum requirements, don’t make people recognize objects or user-supplied images and media to login.</dd>
```

shawna-slh: (also wondering about "user-supplied"... )

patrickhlauke: > Were we going to replace "transcribe" -> "Type"?

that doesn't have the same connotation though. Transcribe is "read it in one place, then type it up in another place". Otherwise, just having "type" there might give the wrong impression that this is about just having autofill / QR codes / etc that don't require actual typing, which is not the point

mbgower: in the minimum, which was more concentrated, I've adjusted the 'transcribe' wording, and am fairly satisfied with this.
I do not think there's an elegant or easy way to crack the progression of these 2 criteria in a way that is always a 'win'.  My inclination is to stick pretty closely to the approved language for now, until the time that someone comes up with something that everyone responds YES!

PR #3255, opened 2023-06-28
Json related

michael-n-cooper:
Cleans up the JSON version of WCAG and adds features for WCAG2ICT. There are structure changes that might impact the quickref though.


AbdullahWins (approved):
well done

patrickhlauke:
@kfranqueiro @iadawn is this PR still valid/needed?

iadawn:
May not use in the long run but useful to leave for just now. We are going to look into JSON needs after TPAC

PR #3259, opened 2023-06-30
Understanding Focus Appearance CR3 updates

### LINKED ISSUE #3200: Re-write of Focus-Appearance at AAA
After the changes in the last CR version, it needs an overhaul.

Dan has taken this on, thank you!

dbjorge:
This PR substantially overhauls the Understanding document for Focus Appearance to bring it in line with the latest Candidate Recommendation, which moved it to AAA and changed it substantially. 

[Preview of the understanding doc](https://raw.githack.com/dbjorge/wcag/focus-appearance-understanding-cr3/understanding/22/focus-appearance.html).

Overall:

* I removed or rewrote several sections that had become obsolete, especially:
  * references to the removed adjacent contrast requirement
  * references to the "encloses or solidly bounds" text
* I restructured the document to consolidate the remaining parts of the minimum area and encloses solidly parts into one "Minimum Area" section
  * I moved the discussion of Inline Links here as well - it seemed to fit better here than with change-of-contrast
  * I mostly rewrote the "other shapes" section to account for the updated perimeter calculation math
* I expanded the "Relationship with Non-text Contrast" section slightly to make up for the loss of the adjacent contrast requirements in this SC itself
* I added some new examples to the change-of-contrast section to cover discussions from the WCAG 2.x working group meeting conversations on outset vs outline vs border vs inset counting as "perimeters"
* I audited all figures throughout the document and updated any that had become outdated/inaccurate (lots of replacing cases that previously passed with 1px thick borders to use 2px thick borders instead)
  * For cases where I needed to redo a figure that was only present as a png previously, I created an `html` file to act as the figure source and have included those in this PR to make it easier for future maintainers to update them.
* I audited for any cases that were still referring to the old numbering for 2.1.11-2.1.13 and updated them accordingly

Similar to when we overhauled Target Size, I recommend reviewing this change by reading the rendered document in its totality, rather than relying on the diff in GitHub.

This PR is already huge, so I've left out of scope any updates to the corresponding non-text contrast section. I also left off for a separate PR the section I had started drafting covering dotted/dashed outlines and anti-aliasing calculations, per discussions in the WCAG 2.x backlog meetings.

Closes #3200


understanding/22/focus-appearance.html
```diff
@@ -61,79 +65,265 @@ <h2>In brief</h2>
     </section>
 
     <section id="intent">
-        
+
         <h2>Intent of Focus Appearance</h2>
 
 
         <p>The purpose of this Success Criterion (SC) is to ensure a keyboard focus indicator is clearly visible and discernible. Focus Appearance is closely related to <a href="https://www.w3.org/WAI/WCAG22/Understanding/focus-visible.html">2.4.7 Focus Visible</a> and <a href="https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast.html">1.4.11 Non-text Contrast</a>. Focus Visible requires that a visible focus indicator exists while a component has keyboard focus; this SC defines a minimum level of visibility. Where Non-text Contrast requires a component to have adequate contrast against the background in each of its states, this SC requires sufficient contrast for the focus indicator itself.</p>
```
WilcoFiers, line 72: Are we sure we want to abbreviate success criterion?

dbjorge: There are a few other cases where an understanding doc does this, but most of them don't, so I agree it'd be better to avoid it for consistency. Dropped the abbreviation, replaced it with "Success Criterion" in most cases, replaced with "Focus Appearance" in cases where it was being directly compared with the titles of other related criteria.

PR #3266, opened 2023-07-07
Focus obscured suggested changes

alastc:
A previously approved branch.

[Preview of understanding doc](https://raw.githack.com/w3c/wcag/FocusObscured-suggested-changes/understanding/22/focus-not-obscured-minimum.html).


understanding/22/focus-not-obscured-minimum.html
```diff
@@ -48,14 +53,95 @@ <h2>Intent of Focus Not Obscured (Minimum)</h2>
...
+
+      <figure id="">
+         <img src="" alt="" />
+         <figcaption>Placeholder for future simplified illustration showing an abstract layout of an authoring tool UI with a user-movable toolbar</figcaption>
+     </figure>
+  
+    </section>
+    <section id="Disclosed"><h3>User-disclosed content</h3>
+   <p>This SC contains a note regarding content that is opened or disclosed by the user. One example of such content is a menu button opened by a user that discloses a list of choices over pre-existing content on the screen. Most such disclosed content can obscure other information on the screen, but it does not obscure an item receiving keyboard focus, because the new content doesn't persist through a change of focus. However, authors may create user-opened content that is intentionally designed to persist until closed by the user, such as a chat window. Such persistent content has the potential to fail Focus Not Obscured (Minimum). Various types are described in this section. All can be designed so that they pass this Success Criterion.</p>
```
scottaohara, line 67:
```suggestion
   <p>This SC contains a note regarding content that is opened or disclosed by the user. One example of such content is a menu button opened by a user that discloses a list of choices over pre-existing content on the screen. Such disclosed content can obscure other information on the screen, but it does not obscure an item receiving keyboard focus, because the new content doesn't persist through a change of focus. However, authors may create user-opened content that is intentionally designed to persist until closed by the user, such as a chat window. Such persistent content has the potential to fail Focus Not Obscured (Minimum). Various types are described in this section. All can be designed so that they pass this Success Criterion.</p>
```

understanding/22/focus-not-obscured-minimum.html
```diff
@@ -48,14 +53,95 @@ <h2>Intent of Focus Not Obscured (Minimum)</h2>
...
+         <figcaption>Placeholder for future illustration showing an open dropdown obscuring information underneath</figcaption>
+     </figure>
+
+    </section>
+   <section>
+      <h4>User openable, persistent disclosures</h4>
+      <p>Some disclosure patterns provide a mechanism for the user to open additional content that remains open until intentionally closed by the user. Accordions are a simple example of such a pattern. Chatbots and expandable side navigation are more complex examples. All of these patterns can be implemented so they are not at risk of failing this SC. Some possible approaches are:</p>
+
+       <ul>
+          <li><strong>When the additional content appears, it displaces existing content.</strong> An accordion is an example of this. When an accordion is opened, the disclosed content shifts existing content further down the page. Since the new content does not obscure existing content, it cannot obscure the item with focus.</ br>
```
scottaohara, line 87: is the `</ br>` necessary here?  if so, it should instead be `<br>`

understanding/22/focus-not-obscured-minimum.html
```diff
@@ -48,14 +53,95 @@ <h2>Intent of Focus Not Obscured (Minimum)</h2>
...
+
+          <figure id="">
+            <img src="" alt="" />
+            <figcaption>Placeholder for future illustrations showing an abstract representation of an accordion pushing content down when opened</figcaption>
+        </figure>
+</li>
+
+          <li><strong>When the additional content appears, existing content reflows.</strong> The popout sidebar on the <a href="https://www.w3.org/TR/WCAG22/">WCAG standard</a> is an example of this pattern. When the side menu is activated, it discloses a new section of information along the left side of the page. The main content area is reduced horizontally to accommodate the new content, and the existing content reflows to fit in the thinner space. As a result, there is no overlapping content between the two sections; the item receiving focus, whether in the left navigation or in the main content, will not be obscured by the other section.</ br>
```
scottaohara, line 95: is the `</ br>` necessary here?  if so, it should instead be `<br>`

understanding/22/focus-not-obscured-minimum.html
```diff
@@ -48,14 +53,95 @@ <h2>Intent of Focus Not Obscured (Minimum)</h2>
...
+
+            <figure id="">
+              <img src="" alt="" />
+              <figcaption>Placeholder for future illustrations showing an abstract representation of a side panel where the tab ring is restricted to the panel</figcaption>
+          </figure>
+         </li>
+          <li><strong>The disclosure expands into an area of the page containing no other content.</strong> Many pages are designed with wide margins, providing significant white space into which new content can be disclosed. Many chatbots and toasters are designed to 'slide up' into the right unpopulated side of a page. Where authors are careful to ensure content is not obscured at each breakpoint in a responsive design, no obscuring of other operable content need occur.</ br>
```
scottaohara, line 109: is the `</ br>` necessary here?  if so, it should instead be `<br>`

understanding/22/focus-not-obscured-minimum.html
```diff
@@ -48,14 +53,95 @@ <h2>Intent of Focus Not Obscured (Minimum)</h2>
...
+
+            <figure id="">
+              <img src="" alt="" />
+              <figcaption>Placeholder for future illustrations showing an abstract representation of a chat window expanding into empty space on the right side of the screen.</figcaption>
+          </figure>
+         </li>
+          <li><strong>When focus leaves the additional content, the additional content is hidden or collapsed.</strong> This is very similar to patterns discussed previously under Non-persistent disclosed information. A distinguishing factor can be that the user's last point of interaction in the disclosure is preserved (it persists) even though it may be hidden until a user returns. Some trees and left navigation pattterns behave this way. </ br>
```
scottaohara, line 116: is the `</ br>` necessary here?  if so, it should instead be `<br>`

understanding/22/focus-not-obscured-minimum.html
scottaohara: suggest "toast notifications" rather than "toasters"

re:
>...into the right unpopulated side of a page...

maybe remove "the right" and replace with "an".  I've seen it on both sides (i believe github even does lower left) - and i'd assume this not necessarily the typical location for right to left language pages.

understanding/22/focus-not-obscured-minimum.html
```diff
@@ -48,14 +53,95 @@ <h2>Intent of Focus Not Obscured (Minimum)</h2>
...
+
+   <p>It is important to emphasize two things. First, this note only applies to content that the user actively discloses. Content pre-positioned by the author (such as a sticky footer), or content that appears without direct user initiation, such as system warnings, must not prevent the item receiving focus from being immediately visible in the viewport. Second, this note is not intended to apply to disclosures that are by convention non-persistent. As discussed in the following sub-section, an open dropdown that does not close when no longer focused is not following convention. The note is not intended to excuse poor authoring practices.</p>
```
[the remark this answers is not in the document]
alastc: It should be saying that something which opens and does not dismiss could fail this SC.
What popovers should not auto-dismiss? I had thought that was the cross-over from modal to popover was that modals keep focus, popovers can be tabbed out of but should then close.

alastc, line 71:
```suggestion
   <p>This section only applies to content that the user actively discloses. Content pre-positioned by the author (such as a sticky footer), or content that appears without direct user initiation, such as system warnings, must not prevent the item receiving focus from being immediately visible in the viewport. Also, this note is not intended to apply to disclosures that are by convention non-persistent. As discussed in the following sub-section, an open dropdown that does not close when no longer focused is not following this convention.</p>
```

understanding/22/focus-not-obscured-minimum.html
alastc, line 109:
```suggestion
          <li><strong>The disclosure expands into an area of the page containing no other content.</strong> Many pages are designed with wide margins, providing significant white space into which new content can be disclosed. Many chatbots and toast notifications are designed to 'slide up' into the right unpopulated side of a page. Where authors are careful to ensure content is not obscured at each breakpoint in a responsive design, no obscuring of other operable content need occur.<br />
```

understanding/22/focus-not-obscured-minimum.html
scottaohara, line 87:
```suggestion
          <li><strong>When the additional content appears, it displaces existing content.</strong> An accordion is an example of this. When an accordion is opened, the disclosed content shifts existing content further down the page. Since the new content does not obscure existing content, it cannot obscure the item with focus.<br />
```

understanding/22/focus-not-obscured-minimum.html
scottaohara, line 95:
```suggestion
          <li><strong>When the additional content appears, existing content reflows.</strong> The popout sidebar on the <a href="https://www.w3.org/TR/WCAG22/">WCAG standard</a> is an example of this pattern. When the side menu is activated, it discloses a new section of information along the left side of the page. The main content area is reduced horizontally to accommodate the new content, and the existing content reflows to fit in the thinner space. As a result, there is no overlapping content between the two sections; the item receiving focus, whether in the left navigation or in the main content, will not be obscured by the other section.<br />
```

understanding/22/focus-not-obscured-minimum.html
scottaohara, line 109:
```suggestion
          <li><strong>The disclosure expands into an area of the page containing no other content.</strong> Many pages are designed with wide margins, providing significant white space into which new content can be disclosed. Many chatbots and toasters are designed to 'slide up' into the right unpopulated side of a page. Where authors are careful to ensure content is not obscured at each breakpoint in a responsive design, no obscuring of other operable content need occur.<br />
```

understanding/22/focus-not-obscured-minimum.html
scottaohara, line 116:
```suggestion
          <li><strong>When focus leaves the additional content, the additional content is hidden or collapsed.</strong> This is very similar to patterns discussed previously under Non-persistent disclosed information. A distinguishing factor can be that the user's last point of interaction in the disclosure is preserved (it persists) even though it may be hidden until a user returns. Some trees and left navigation patterns behave this way. <br />
```

scottaohara, line 116:
```suggestion
          <li><strong>When focus leaves the additional content, the additional content is hidden or collapsed.</strong> This is very similar to patterns discussed previously under Non-persistent disclosed information. A distinguishing factor can be that the user's last point of interaction in the disclosure is preserved (it persists) even though it may be hidden until a user returns. Some trees and left navigation patterns behave this way. <br />
```

understanding/22/focus-not-obscured-minimum.html
```diff
@@ -48,14 +53,100 @@ <h2>Intent of Focus Not Obscured (Minimum)</h2>
...
+            <!--
+            <figure id="">
+              <img src="" alt="" />
+              <figcaption>Placeholder for future illustrations showing a collapsible tree that when expanded shows the prior point of interaction</figcaption>
+          </figure>
+          -->
+         </li>
+       </ul>
+       <p>Where opened content persists and causes existing content to be obscured, it will be at risk of failing this criterion if it prevents users from seeing the item receiving focus. An example of this problem is described and captured in a short video in <a href="https://knowbility.org/blog/2020/accessible-slide-menus#slide-out-navigation-stays-open-while-focus-is-on-content-behind-it">Slide-out navigation stays open while focus is on content behind it</a>.</p>
```
scottaohara, line 129: i think what this section needs is an example of a persistent disclosure (for instance, a 'teaching tip' or 'info button popup' where the popup content can persist if focus leaves the content (if it can even enter in the first place), but a user can dismiss, or minimize the content by use of the Esc key.

alastc: Please do raise on Friday, or create a PR, or (if lacking time!) create a quick issue. I'm going to merge to make sure the rest gets in for the transition. Thanks.

PR #3267, opened 2023-07-07
Editorial fixes to term "assistive technology"

maryjom:
Capitalize first word of sentence, and have the example formatting include the list of examples so it shows in the block. This latter error was introduced in WCAG 2.1 when new document formatting was added but the bulleted list that comes after the sentence ending in ":" should be included inside the boundaries of the markdown for the example 
`<div  class="example"> All of the content including the bullets </div>`

### DIFF
--- a/guidelines/terms/20/assistive-technology.html
+++ b/guidelines/terms/20/assistive-technology.html
@@ -6,7 +6,7 @@
       of users with disabilities that go beyond those offered by mainstream user agents
    </p>
    
-   <p class="note">functionality provided by assistive technology includes alternative presentations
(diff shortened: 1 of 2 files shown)


scottaohara (approved):

awkawk:
Same issue for "idiom" - these are the only notes that have a capitalization issue.

awkawk:
These are Errata for WCAG 2.1 and WCAG 2.0, but should be integrated into WCAG 2.2. These are about as minimally editorial as I can imagine, but should go through the process.

alastc:
As it isn't an errata for WCAG 2.2 (yet), I'll merge into main as editorial, and add to updates/errata for WCAG 2.0/2.1

Issue #301, opened 2017-09-04
2.4.7 Focus Visible - explanation/understanding addition about "mode of operation"?


patrickhlauke:
The current wording of the SC

> 2.4.7 Focus Visible: Any keyboard operable user interface has a mode of operation where the keyboard focus indicator is visible

talks about a "mode of operation". However, the understanding document https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-focus-visible.html currently doesn't mention/explain this at all. What exactly is meant by "a mode of operation"? A user setting? Does the act of using a keyboard itself constitute a "mode of operation"? It would be good to get at least a mention of this in understanding.

alastc:
Going thorugh old issues, but I've been updating focus-visible so this is relevent.

For HTML-based web content I've generally ignored this phrase, assuming the default 'mode of operation' is the only mode of operation. I was assuming it accounted for flash or something, where there might be multiple modes of operation?

In the [new draft](https://docs.google.com/document/d/1g9_WBgfhViWAaRFIWWt10CP5rBsEVIWm3vT1vWqrHvI/edit?ts=5d124fb5#), how about a line such as:
> Where the success criterion says “mode of operation”, this is to account for platforms which may not always show a focus indicator. In most cases there is only one mode of operation so this success criteria applies.

patrickhlauke:
Context of this is question was also certain arguments for/against the use of `:focus-visible`, which allows you to style the focus state when the browser/UA/platform itself would show it. case in point buttons or links that when clicked with the mouse wouldn't show an outline/focus indicator by default, but when focused via keyboard they do. There have been various discussions saying this is bad/contravenes 2.4.7, whereas i'd argue that using `:focus-visible` is within the spirit of 2.4.7, as there IS a mode of operation where the keyboard focus indicator is visible...and that's when the user is actually using the keyboard to navigate. whether or not it shows the focus indicator when operating things purely with a mouse (or touchscreen, or whatever) is then not a failure, provided that once they switch to using keyboard it becomes visible

patrickhlauke:
in short, if the WG agrees with my above interpretation/assessment, i'd rather have the understanding doc explain this idea of "mode of operation" as in "a way in which the user controls/operates their UA / which input method they use", and perhaps a positive technique that shows the use of `:focus-visible` to satisfy 2.4.7

alastc:
[...]

PR #3272, opened 2023-07-11
Add sufficient technique for 2.4.7 using `:focus-visible`

patrickhlauke:
Two years late, but: here's the technique I intended to write for https://github.com/w3c/wcag/issues/301#issuecomment-624545525

(but a recent lengthy discussion on mailing list reminded me that back in #301 the WG had indeed agreed this interpretation)

[Preview of new techinque](https://raw.githack.com/w3c/wcag/patrickhlauke-focus-visible-technique/techniques/css/C45.html)


cstrobbe:
`<p>There may be situations where mouse/pointer users could also benefit from having a visible focus indication, even though they did not set focus to an element using the keyboard. As a best practice, consider still providing an explicit <code class="language-css">:focus</code> style for these cases.</p>`

1. Did you mean `:hover` instead of `:focus`? If not, does "not set focus ... using the keyboard" refer to script-triggered focus?
2. In the second sentence, do you mean a single style rule that covers both `:focus-visible` and `:focus` or separate styles for both pseudo-classes. (If you define e.g. both `a:focus { outline: 2px dotted blue;}` and `a:focus-visible { outline: 2px dashed green; }`, the rule that is triggered by keyboard focus is whichever of these rules comes after the other in your stylesheet.)

patrickhlauke:
> 1. Did you mean `:hover` instead of `:focus`? If not, does "not set focus ... using the keyboard" refer to script-triggered focus?

No, i meant `:focus`, and by "not set focus ... using the keyboard" I mean both script-triggered focus, but also ... as a result of clicking a control using a mouse/touch/stylus

This is in the context of the *lengthy* discussion in wai-ig discussing how mouse users also benefit from seeing where focus is, and why `:focus-visible` should never be used... (a rather opinionated take, of course)

> 2. In the second sentence, do you mean a single style rule that covers both `:focus-visible` and `:focus` or separate styles for both pseudo-classes. (If you define e.g. both `a:focus { outline: 2px dotted blue;}` and `a:focus-visible { outline: 2px dashed green; }`, the rule that is triggered by keyboard focus is whichever of these rules comes after the other in your stylesheet.)

I leave this up to the reader. but yes, one approach would be to provide a (maybe slightly more subtle) `:focus` style in general, and then a much more visible/stronger `:focus-visible` specifically for keyboard users

techniques/css/C40.html
```diff
@@ -71,6 +71,9 @@ <h2>Related Techniques</h2>
 			<ul>
 				<li><a href="../general/G149">G149</a></li>
 				<li><a href="../general/G195">G195</a></li>
+				<li><a href="C15">C15</a></li>
+				<li><a href="../general/G165">G165</a></li>
+        <li><a href="C40">C45</a></li>
```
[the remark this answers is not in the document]
patrickhlauke: ah, well spotted. copy-pasta...

understanding/20/focus-visible.html
```diff
@@ -18,10 +18,13 @@ <h2>Intent of Focus Visible</h2>
...
+
+      <div class="note">
+         <p>There may be situations where mouse/pointer users could also benefit from having a visible focus indicator, even though they did not set focus to an element using the keyboard. As a best practice, consider still providing an explicit focus indicator for these cases.</p>
+      </div>
       
    	<p>Note that a keyboard focus indicator can take different forms.<span class="wcag22"> This criterion does not specify what the form is, but <a href="focus-appearance-minimum">Focus Appearance</a> does define how visible the indicator should be. Passing <a href="focus-appearance-minimum">Focus Appearance (Minimum)</a> would pass this success criterion.</span></p>
```
nitedog, line 26: PR #3710 removes the word "(Minimum)" from the second link in this paragraph -- it seems to be a left over from changes made during the creation of this SC.

techniques/css/C45.html
```diff
@@ -0,0 +1,98 @@
...
+			</section>      
+		</section>
+		<section id="tests">
+			<h2>Tests</h2>
+			<section class="test-procedure">
+				<h3>Procedure</h3>
+				<p>For each user interface component that can receive keyboard focus:</p>
+				<ol>
+					<li>Set focus to the interface component using the keyboard (generally, navigating to the component using <kbd>Tab</kbd>/<kbd>Shift+Tab</kbd>)</li>
+          <li>Verify that once the component has received focus, there is a visible focus indication on or around the component</li>
```
alastc, line 71:
```suggestion
          <li>Verify that once the component has received focus, a focus indicator is visible</li>
```

alastc: There's nothing normative about where the indicator is, so suggest keeping it short.

patrickhlauke: Love it

bruce-usab:
Discussed on TF call 4/12 but not advanced.  @gundulaniemann might you review?

bruce-usab:
Discussed on call 4/26 and Ready for approval.

gundulaniemann:
<p>There may be situations where mouse/pointer users could also benefit from having a visible focus indication, even though they did not set focus to an element using the keyboard. As a best practice, consider still providing an explicit <code class="language-css">:focus</code> style for these cases.</p>

I suggest to drop the word 'could', as mouse/pointer users definitely benefit from seeing the focus indicator once they set the focus to a specific element.

Is the text supposed to contain placeholders?
The metadata box does not seem to be complete.
The Example is numbered as 'X'.
The links to related techniques point to a location with "patrickhlauke-focus-visible-technique" in it and yield a 404 error

patrickhlauke:
> mouse/pointer users definitely benefit from seeing the focus indicator once they set the focus to a specific element

there are scenarios where forcing a visible focus indicator gets visually confusing, such as controls/buttons that trigger an immediate action. that's the whole reason why we go `:focus-visible` in the first place (things like carousel left/right buttons, where after clicking with a mouse you don't want to have that one button visually highlighted, but you do want to provide a strong highlight when navigating with the keyboard)

understanding/20/focus-visible.html
```diff
@@ -83,52 +87,46 @@ <h2>Techniques for Focus Visible</h2>
...
             </li>
-            
             <li>
-               									         
-               <a href="../Techniques/client-side-script/SCR31" class="script">Using script to change the background color or border of the element with focus</a>
-               								       
+               <a href="../../techniques/css/C45">Using CSS :focus-visible to provide keyboard focus indication</a>
+            </li>
+            <li>
+               <a href="https://www.w3.org/WAI/WCAG21/Techniques/client-side-script/SCR31" class="script">Using script to change the background color or border of the element with focus</a>
```
mbgower, line 126:
```suggestion
               <a href="../Techniques/client-side-script/SCR31" class="script">Using script to change the background color or border of the element with focus</a>
```

PR #3277, opened 2023-07-12
Focus not obscured margin technique

### LINKED ISSUE #3187: Sufficient Technique C43 is not a sufficient technique
This technique doesn't work properly, in that:

- it doesn't work for caret browsing
- it incorrectly assumes that the footer height is known and fixed

A sufficient technique would use scroll-padding and would re-calculate it dynamically using a resize observer. 

See the following for notes I made about this approach, based on Alastair's original proof of concept: https://www.tpgi.com/prevent-focused-elements-from-being-obscured-by-sticky-headers/#an-alternative-approach

fstrr:
This update addresses issue 3187
1. Adds support for dialog element of unknown height.
2. Adds support for caret browsing.
3. Retitled C43 and working example as they now use `scroll-padding` instead of `margin-padding`.
4. Updated Resources links

Closes #3187


fstrr:
[rawgit version, if needed](https://raw.githack.com/w3c/wcag/focus-not-obscured-margin-technique/techniques/css/C43.html)

PR #3278, opened 2023-07-12
Issue 3138 numbering

### LINKED ISSUE #3138: Examples are mis-numbered
In the spec, the examples are numbered across section. 5.2.3 has the first example ("example 1") and the glossary item for Abbreviation has examples 2-4. This continues until the glossary item for "web content" has examples 38-42.

It seems that the ideal situation is:

- Items with individual examples would have "Example"
- items with multiple examples would have "Example 1", "Example 2", etc
- The same logic should apply to Notes. We lost numbering for multiple notes when we did WCAG 2.1

michael-n-cooper:
fixes #3138 

Note this changes the text of note and example headings, but they continue to have respec-generated ids which won't "match up", assuming that's minor.

### DIFF
--- a/script/wcag.js
+++ b/script/wcag.js
@@ -83,10 +83,41 @@ function termTitles() {
 	});	
 }
 
+function numberNotes() {
+	var sectionsWithNotes = new Array();
+	document.querySelectorAll(".note").forEach(function(note) {
+		sectionsWithNotes.push(note.parentNode);
+	});
+	
+	sectionsWithNotes.forEach(function(sec) {
+		if (sec.processed) return;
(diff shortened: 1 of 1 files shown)


michael-n-cooper:
[Preview](https://raw.githack.com/w3c/wcag/issue-3138-numbering/guidelines/index.html)

alastc (approved):
It seems to work, I didn't realise it would be done in JS, but I guess that's easier than with XSL!

awkawk:
Great, thanks.

alastc:
Would this also need merging into WCAG 2.1 for republishing?

PR #3280, opened 2023-07-13
Parsing update for WCAG 2.1 understanding

alastc:
Adjusts the understanding document on the basis that #3152 is merged. 

I had thought to include more of the note from 3152, but in the understanding document that note will be included with the normative text, so it will be right above the intent section.


alastc:
Approved: https://www.w3.org/2023/07/18-ag-minutes#t11

PR #3281, opened 2023-07-13
Issue 1814 on timing adjustable

### LINKED ISSUE #1814: Suggested improvement to Understanding 2.2.1: Timing Adjustable
https://www.w3.org/WAI/WCAG21/Understanding/timing-adjustable.html

The issue that I'm getting asked about is for toast messages, or snackbar messages like what we refer to in the Understanding 4.1.3 Status Messages document.

The core question is "Do toast messages need to do something different in order to not result in a failure of 2.2.1?"

I would say that as long as a toast message is not the only way that a user can be aware of something happening (e.g., an email arriving) or that they have successfully accomplished something (e.g., image uploaded to personal folder in online directory) then the criticality of the message being viewed as a toast is less. Those toast messages would need to support 4.1.3 Status Messages, but that is a separate issue.

Leaving the toast in view until the user dismisses each one is not desirable as in some cases you might have dozens or more individual messages and if they need individual attention then it is making the UI harder to use for everyone.

In 4.1.3 there is a reference to Material Design's snackbar (https://material.io/components/snackbars) which are designed for low-priority messages:
After a user puts a photo in an album in an online photo app, a snackbar displays the message "Saved in 'Wedding' album", which is also read by a screen reader.

I propose that for Understanding 2.2.1 we incorporate the following text into the Intent section right before the notes regarding server time limits:
Not all content that operates on a timer requires that the timing is adjustable. For example, a web application such as an email client provides notification of new email arriving in toast messages in the lower right-hand side of the interface, and the toast messages disappear after 5 seconds. The users are able to verify the arrival of email through other means, such as viewing the Inbox, so the disappearance of the toast message does not set a time limit on the end user's ability to verify if new mail has arrived. If the messages conveyed via toast could not be verified separately, then the message would need to meet this Success Criterion in order to provide the user with sufficient time to access the information.

alastc:
Closes #1814


understanding/20/timing-adjustable.html
```diff
@@ -40,6 +40,8 @@ <h2>Intent of Timing Adjustable</h2>
 
       <p>In some cases, however, it is not possible to change the time limit (for example, for an auction or other real-time event) and exceptions are therefore provided for those cases.</p>
       
+      <p>Not all content that operates on a timer requires that the timing is adjustable. For example, a web application such as an email client provides notification of new email arriving in toast messages in the lower right-hand side of the interface, and the toast messages disappear after 5 seconds. The users are able to verify the arrival of email through other means, such as viewing the Inbox, so the disappearance of the toast message does not set a time limit on the end user's ability to verify if new mail has arrived. If the messages conveyed via toast could not be verified separately, then the message would need to meet this Success Criterion in order to provide the user with sufficient time to access the information.</p>
```
patrickhlauke, line 43: I'd change "end user" to just "user"

scottaohara, line 43: following on from Patrick's suggestion:
The following changes 
- the first instance of "The users" to just "Users"
- the second instance "the end user's" I change to "their" instead of Patrick's suggestion, since the sentence starting talking about users plural, and then switched to a single user.  
- the third instance of "the user" was changed to "users" to maintain parity with the previous plural mention of them.

Additionally changed "If the messages conveyed via toast..." to "If the toast messages..." to be consistent with the first mention of "toast messages", and because "conveyed via toast" is making it seem like "toast" is more than just a design/dev jargon term for a JS powered UI pattern, and an actual mechanism for message delivery.

Finally, changed "...then the message would need...." to "then each message would need..." because the sentences was starting to talk about messages plural and then changed to the singular.

```suggestion
      <p>Not all content that operates on a timer requires that the timing is adjustable. For example, a web application such as an email client provides notification of new email arriving in toast messages in the lower right-hand side of the interface, and the toast messages disappear after 5 seconds. Users are able to verify the arrival of email through other means, such as viewing the Inbox, so the disappearance of the toast message does not set a time limit on the their ability to verify if new mail has arrived. If the toast messages could not be verified separately, then each message would need to meet this Success Criterion in order to provide users with sufficient time to access the information.</p>
```

patrickhlauke: agree with making it sound clearer/simpler/more direct

bruce-usab:
- I agree with proposed edits.
- I agree with making it sound clearer/simpler/more direct &mdash; but am not clear on the suggestion.
- I recommend using "Toast" only as an example of the behavior we allowing for as "decorative".

understanding/20/timing-adjustable.html
alastc, line 43:
```suggestion
      <p>Content that operates on a timer does not need to be time adjustable if there is an alternative that does not rely on a timer. For example, a web application such as an email client provides notification of new email arriving in toast messages in the lower right-hand side of the interface, and the toast messages disappear after 5 seconds. The users are able to verify the arrival of email through other means, such as viewing the Inbox, so the disappearance of the toast message does not set a time limit on the end user's ability to verify if new mail has arrived. If the messages conveyed via toast could not be verified separately, then the message would need to meet this Success Criterion in order to provide the user with sufficient time to access the information.</p>
```

scottaohara (approved):

understanding/20/timing-adjustable.html
```diff
@@ -40,6 +40,8 @@ <h2>Intent of Timing Adjustable</h2>
 
       <p>In some cases, however, it is not possible to change the time limit (for example, for an auction or other real-time event) and exceptions are therefore provided for those cases.</p>
       
+      <p>Content that operates on a timer does not need to be time adjustable if there is an alternative that does not rely on a timer. For example, a web application such as an email client provides notification of new email arriving with a temporary message (such as a 'toast' message) in the lower right-hand side of the interface, and the message disappears after 5 seconds. Users are able to verify the arrival of email through other means, such as viewing the Inbox, so the disappearance of the message does not set a time limit on the their ability to verify if new mail has arrived. If the message could not be verified separately, then each message would need to meet this Success Criterion in order to provide users with sufficient time to access the information.</p>
```
mbgower, line 43:
```suggestion
      <p>Content that operates on a timer does not need to be time adjustable if there is an alternative that does not rely on a timer. For example, a web application such as an email client provides notification of new email arriving with a temporary message (such as a 'toast' message) in the lower right-hand side of the interface, and the message disappears after 5 seconds. Users are able to identify the arrival of email through other means, such as viewing the Inbox, so the disappearance of the message does not set a time limit on the their ability to determine if new mail has arrived. If the user has no other means of discovering the same information (or performing the same function), then each message would need to meet this Success Criterion in order to provide users with sufficient time to access the information.</p>
```

PR #3286, opened 2023-07-17
1.4.1 Use of Color Understanding doc: add note concerning same color case

### LINKED ISSUE #1467: Looking for definitive clarification regarding 1.4.1
Concerning 1.4.1, I've been having multiple discussions lately on how to handle situations where links have had their default underline removed, and color globally changed to match the static body text of a web page.

Consider the following (inline styles used for example purposes):
```html
<p style="color: #000;">
  I'm a paragraph of content with 
  <a href="#" style="color: #000; text-decoration: none;">a link</a>
  to navigate to a new page!
<p>
```
Some argue that this would be an instance of an author purposefully changing the styles of links to instead visually present them via color alone, as the other visual identifier (underline) has been removed.  If these links were a dark blue instead of black, they would definitely be called out in an audit.

However, the other side of the argument is that because the body text and the link text are the same color, there _is no_ failure of use of color here, and thus color is not being used to promote the links as actionable vs the static text.
From my reviewing of github issues and related docs to 1.4.1, I can see this has been mentioned in [1272](https://github.com/w3c/wcag/issues/1272#issuecomment-673343594) as a current gap area.  So I would ask that this issue be used to explicitly mention, at least in the understanding doc for 1.4.1, whether this scenario is or is not a failure.

Thank you

scottaohara:
closes #1467

Per the feedback in the thread, the use case of a hyperlink being styled to look no different than adjacent static text would not be a failure of 1.4.1 since there is no difference in color to compare against.  

This PR adds a note to call that out.

Additionally this PR makes some corrections to the source code.  Namely:
- removes the `</meta>` and `</link>` in the header - these are self-closing tags so these shouldn't be there.
- fixes an instance where a list was nested in a `p` element, which is invalid and was causing the HTML parser to kick it out/created an empty paragraph after the list.

### DIFF
--- a/understanding/20/use-of-color.html
+++ b/understanding/20/use-of-color.html
@@ -1,9 +1,9 @@
 <!DOCTYPE html>
 <html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml">
 <head>
-   <meta charset="UTF-8"></meta>
+   <meta charset="UTF-8" />
    <title>Understanding Use of Color</title>
-   <link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove"></link>
+   <link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove" />
 </head>
 <body>
    <h1>Understanding Use of Color</h1>
@@ -63,6 +63,16 @@
 
       </div>
 
+     <div class="note">
+       
+       <p>This criterion is not applicable when assessing adjacent content that is rendered as the same color. For instance,
+         a hyperlink which has been styled to appear no different than the static text surrounding it. As there would be no color 
+         differentiation between the actionable hyperlink text and the static text adjacent to it, this would be a poor experience 
+         for all sighted users, but not a failure of this Success Criterion.
+       </p>
+       
+     </div>
+
       <div class="note">
 
          <p>This criterion does not directly address the needs of users with assistive technologies.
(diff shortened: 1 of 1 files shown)


understanding/20/use-of-color.html
```diff
@@ -63,6 +63,16 @@ <h2>Intent of Use of Color</h2>
 
       </div>
 
+     <div class="note">
+       
+       <p>This criterion is not applicable when assessing adjacent content that is rendered as the same color. For instance,
```
patrickhlauke, line 68: i'd go even broader to start with: `this criterion does not apply to situations where colour has *not* been used to convey information, indicating an action, prompting a response, or distinguishing a visual element.` and then move into the more specific example/case

scottaohara: pulled in your suggestion.  thanks

understanding/20/use-of-color.html
```diff
@@ -65,10 +65,11 @@ <h2>Intent of Use of Color</h2>
 
      <div class="note">
        
-       <p>This criterion is not applicable when assessing adjacent content that is rendered as the same color. For instance,
-         a hyperlink which has been styled to appear no different than the static text surrounding it. As there would be no color 
-         differentiation between the actionable hyperlink text and the static text adjacent to it, this would be a poor experience 
-         for all sighted users, but not a failure of this Success Criterion.
+       <p>This criterion does not apply to situations where color has not been used to convey information, indicating an action,
```
patrickhlauke, line 68:
```suggestion
       <p>This criterion does not apply to situations where color has <em>not</em> been used to convey information, indicating an action,
```

mbgower:
The change is the note circa lines 65-75. All other changes are results of formatting, and do not represent changes to content.

understanding/20/use-of-color.html
```diff
@@ -63,6 +63,17 @@ <h2>Intent of Use of Color</h2>
 
       </div>
 
+     <div class="note">
+       
+       <p>This criterion does not apply to situations where color has <em>not</em> been used to convey information, indicating an action,
+         prompting a response, or distinguishing a visual element. For instance, a hyperlink which has been styled to appear no
```
dbjorge, line 69: Text looks good to me! Grammar nIt:

```suggestion
       <p>This criterion does not apply to situations where color has <em>not</em> been used to convey information, indicate an action,
         prompt a response, or distinguish a visual element. For instance, a hyperlink which has been styled to appear no
```

ShaneDittmar:
I have a few issues with the note:
 * I think the language should explain that this isn't a failure of this SC but might be a failure of others.
 * I don't think that "all sighted users" is the best way to identify the group of people for whom an un-styled link would be a poor experience.
 * I think it causes more problems than it solves to explicitly call this a "poor experience."
 * Editorially, there is also a sentence fragment.

I propose this paragraph instead:
```
This criterion does not apply to situations where color has <em>not</em>
been used to convey information, indicate an action, prompt a response, or
distinguish a visual element. For instance, a hyperlink which has been
styled to appear no different than the static text surrounding it would not
fail this success criterion, as there would be no color differentiation
between the actionable hyperlink text and the static text adjacent to it.
However, such a design might fail other success criteria or cause other
usability issues for users.
```

P.S. Would add this as a suggestion, but the paragraph is spread across multiple lines and the GitHubb function for multi-line comments isn't accessible.

patrickhlauke:
> I think the language should explain that this isn't a failure of this SC but might be a failure of others

that is the baseline assumption for any statement in the understanding documents, really. nothing is ever fully exempt from any other potential failures...so don't think it'd be necessary to explicitly call it out here.

understanding/20/use-of-color.html
```diff
@@ -63,6 +63,17 @@ <h2>Intent of Use of Color</h2>
 
       </div>
 
+     <div class="note">
+       
+       <p>This criterion does not apply to situations where color has <em>not</em> been used to convey information, indicate an action,
+         prompt a response, or distinguish a visual element. For instance, a hyperlink which has been styled to appear no
+         different than the static text surrounding it. As there would be no color differentiation between the actionable
```
detlevhfischer, line 70: > visual element. For instance

Change to 
> visual element, for instance

...or make the sentence starting with "For instance" a complete one.

detlevhfischer, line 70: The construct 

> As there would be no color differentiation between the actionable hyperlink text and the static text adjacent to it, this would be a poor experience for all sighted users, but not a failure of this Success Criterion.

seems a bit difficult to parse. I suggest a simpler version:

> No perceivable difference between the link and the text around creates a poor experience for all sighted users, but is not a failure of this Success Criterion.

ShaneDittmar:
> > I think the language should explain that this isn't a failure of this SC but might be a failure of others
> 
> that is the baseline assumption for any statement in the understanding documents, really. nothing is ever fully exempt from any other potential failures...so don't think it'd be necessary to explicitly call it out here.

Great point. I'd remove the last sentence of my version entirely then:
```
This criterion does not apply to situations where color has <em>not</em> been used to convey information, indicate an action, prompt a response, or distinguish a visual element. For instance, a hyperlink which has been styled to appear no different than the static text surrounding it would not fail this success criterion, as there would be no color differentiation between the actionable hyperlink text and the static text adjacent to it.
```

understanding/20/use-of-color.html
mbgower: @detlevhfischer I'd add "it" to read "...text around it creates..."

mbgower, line 70:
```suggestion
         different than the static text surrounding it. Having no perceivable difference between the link 
```

mbgower: Actually, I'd suggest rewording a bit more. How about...
> For example, if link text has been styled to appear no different than the text around it. As there would be no color differentiation between the actionable link and static text, this would be a a poor experience for all sighted users, but not a failure of this Success Criterion.

understanding/20/use-of-color.html
```diff
@@ -63,6 +63,17 @@ <h2>Intent of Use of Color</h2>
 
       </div>
 
+     <div class="note">
+       
+       <p>This criterion does not apply to situations where color has <em>not</em> been used to convey information, indicate an action,
+         prompt a response, or distinguish a visual element. For instance, a hyperlink which has been styled to appear no
```
scottaohara, line 69:
```suggestion
         prompt a response or distinguish a visual element, for instance, a hyperlink which has been styled to appear no
```

understanding/20/use-of-color.html
```diff
@@ -63,6 +63,16 @@ <h2>Intent of Use of Color</h2>
 
       </div>
 
+     <div class="note">
+       
+       <p>This criterion does not apply to situations where color has <em>not</em> been used to convey information, indicate an action,
+         prompt a response or distinguish a visual element. For instance, a hyperlink which has been styled to appear no different than neighboring 
+         static text would not fail this success criterion, as there would be no color differentiation between the actionable hyperlink text 
+         and the adjacent static text. Such styling would result for poor usability for many users, regardless of disability.
```
alastc, line 71:
```suggestion
         and the adjacent static text. Such styling would result in poor usability for many users, regardless of disability.
```

understanding/20/use-of-color.html
```diff
@@ -63,6 +63,16 @@ <h2>Intent of Use of Color</h2>
 
       </div>
 
+     <div class="note">
+       
+       <p>This criterion does not apply to situations where color has <em>not</em> been used to convey information, indicate an action,
+         prompt a response or distinguish a visual element. For instance, a hyperlink which has been styled to appear no different than neighboring 
+         static text would not fail this success criterion, as there would be no color differentiation between the actionable hyperlink text 
+         and the adjacent static text. Such styling would result in poor usability for many users, regardless of disability.
```
alastc, line 71:
```suggestion
         and the adjacent static text. Such such lack of styling differentiation could result in poor usability for anyone looking at the interface, regardless of disability.
```

understanding/20/use-of-color.html
```diff
@@ -63,6 +63,16 @@ <h2>Intent of Use of Color</h2>
 
       </div>
 
+     <div class="note">
+       
+       <p>This criterion does not apply to situations where color has <em>not</em> been used to convey information, indicate an action,
+         prompt a response or distinguish a visual element. For instance, a hyperlink which has been styled to appear no different than neighboring 
+         static text would not fail this success criterion, as there would be no color differentiation between the actionable hyperlink text 
+         and the adjacent static text. Such such lack of styling differentiation could result in poor usability for anyone looking at the interface, regardless of disability.
```
mbgower, line 71:
```suggestion
         and the adjacent static text. Such lack of styling differentiation could result in poor usability for anyone looking at the interface, regardless of disability.
```

PR #3295, opened 2023-07-19
Tweak 2.4.6 Headings and Labels understanding

### LINKED ISSUE #3291: 2.4.6 - Describe topic or purpose IN or OUT of context?
Hello!

What is the normative stance on whether 2.4.6 requires labels to describe topic or purpose out of context?

patrickhlauke:
Remove incorrect suggestion that headings/labels should make sense out of context, reformat lists (particularly what seemed to be a definition list, but marked up with `<strong>`? dear me...)

Closes #3291


understanding/20/headings-and-labels.html
```diff
@@ -49,152 +48,82 @@ <h2>Intent of Headings and Labels</h2>
...
-            
-            <p>
-               								       
-               <strong>Consistent headings in different articles</strong>
-               							     
-            </p>
-            
-            <p>A Web site contains papers from a conference. Submissions to the conference are required
+               the headings.</dd>
+         <dt>>Consistent headings in different articles</dt>
```
fstrr, line 78: Looks like there's an extra > here

patrickhlauke: good spot

awkawk (approved):

scottaohara (approved):

alastc:
@patrickhlauke that _looks_ like quite a lot of changes, but is it just lines 58-60?

patrickhlauke:
@alastc core change of content is 58-60 yes. the rest is removing oodles of empty lines, replacing minus with an actual em dash, and reformatting the unordered list using `<strong>` with an actual definition list (i have a separate PR that makes this particular change everywhere else in the understanding)

dbjorge (approved):

melaniephilipp:
CURRENT: "...labels and headings **are meaningful when read out of context,** for example, in a Table of Contents, or when jumping from heading to heading within a page...." to "...labels and headings **are clearer when presented in a different format** – for example, in an automatically generated list of headings, a table of contents, or when jumping from heading to heading within a page."

patrickhlauke:
> the proposed text doesn't clarify what kind of context can be taken into account

because "context" is a red herring that wasn't part of the SC (not in the way that 1.4.4 tried to define it). The SC is much more handwavy about how to determine if headings/labels are appropriate enough to "define topic or purpose" (naturally, as that's such a wooly subjective concept to begin with)

melaniephilipp:
Thanks @patrickhlauke - we are always wrestling with defining when labels (especially regarding multiple buttons with the same accessible name) are appropriate enough to "define topic or purpose". The subjectiveness makes it difficult indeed for consistency in assessments. When I saw the issue come through I was hoping by some miracle the update would define it less subjectively. :-) 

At any rate, removing the incorrect suggestion that headings/labels should make sense out of context is a good move. 

Thanks!

PR #3296, opened 2023-07-20
Reformat definition lists, ordered lists, fix Reading Level understanding

patrickhlauke:
* seems back in the WCAG 2.0 days, definition lists weren't used. This retrospectively fixes it
* `<ol>` used instead of `<ul>` when no order is intended/necessary
* fix the completely broken listing of education levels for Reading Level understanding

### DIFF
--- a/understanding/20/audio-description-or-media-alternative-prerecorded.html
+++ b/understanding/20/audio-description-or-media-alternative-prerecorded.html
@@ -95,69 +95,41 @@
    
    <section id="examples">
       <h2>Examples of Audio Description or Media Alternative (Prerecorded)</h2>
-      
-      
-      <ul>
-         
-         <li>
-            
+
+      <dl>
+         <dt>A movie with audio description</dt>
+         <dd>
             <p>
-               								       
-               <strong>A movie with audio description.</strong>
-               							     
-            </p>
-            
-            <p> 
-               
                <strong>Describer: </strong>A title, "Teaching Evolution Case Studies. Bonnie Chen." A teacher shows photographs
                of birds with long, thin beaks.
             </p>
             
             <p> 
-               
                <strong>Bonnie Chen: </strong>"These photos were all taken at the Everglades."
             </p>
             
             <p> 
-               
                <strong>Describer: </strong>The teacher hands each student two flat, thin wooden sticks.
             </p>
             
             <p> 
-               
                <strong>Bonnie Chen: </strong>"Today you will pretend to be a species of wading bird that has a beak like this."
             </p>
             
             <p> 
-               
                <strong>Describer: </strong>The teacher holds two of the sticks to her mouth making the shape of a beak.
             </p>
             
-            <p>Transcript of audio based on the first few minutes of "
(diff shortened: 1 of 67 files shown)


understanding/20/audio-description-or-media-alternative-prerecorded.html
```diff
@@ -95,69 +95,41 @@ <h2>Benefits of Audio Description or Media Alternative (Prerecorded)</h2>
...
-         <li>
-            
-            <p>
-               								       
-               <strong>An alternative for time-based media for a training video</strong>
-               							     
-            </p>
-            
-            <p>A company purchases a Training video for use by its employees and puts it on the companies
+            <p>Transcript of audio based on the first few minutes of "<a href="http://www.pbs.org/wgbh/evolution/educators/teachstuds/tvideos.html">Teaching Evolution Case Studies, Bonnie Chen</a>" (copyright WGBH and Clear Blue Sky Productions, Inc.)</p>
```
scottaohara, line 123: unrelated to the changes, but would be good to use `https` here instead.

```suggestion
            <p>Transcript of audio based on the first few minutes of "<a href="https://www.pbs.org/wgbh/evolution/educators/teachstuds/tvideos.html">Teaching Evolution Case Studies, Bonnie Chen</a>" (copyright WGBH and Clear Blue Sky Productions, Inc.)</p>
```

understanding/20/captions-prerecorded.html
```diff
@@ -79,9 +78,7 @@ <h2>Examples of Captions (Prerecorded)</h2>
             
             <p>for the likes of sailors, soldiers and woodsmen.." </p>
             
-            <p>From Sample Transcript Formatting by Whit Anderson.</p>
-            
-         </li>
+            <p>From Sample Transcript Formatting by Whit Anderson.</dd>
```
scottaohara, line 81: looks like some errant changes here?  maybe the file didnt' get fully committed?

understanding/20/reading-level.html
```diff
@@ -169,78 +160,43 @@ <h2>Benefits of Reading Level</h2>
    <section id="examples">
       <h2>Examples of Reading Level</h2>
       
-      
-      <ol>
+      <ul>
```
scottaohara, line 163: this was changed to a `ul`, but the items within are `dt` and `dd`s.

patrickhlauke: oops, thanks for spotting this (and the other one). clearly a find/replace across all files that went wrong (caused by the infurtiatingly inconsistent structure of the original files). if you do spot any more, shout, and I'll do a big cleanup commit on this PR

scottaohara (changes requested):
the majority of this looks good. just a few instances of markup errors i found, so probably a good idea to do another pass through to double check that i also didn't miss stuff in the review.

patrickhlauke:
Setting this temporarily to draft while I fix any tiny find/replace snafus

patrickhlauke:
Did another pass to fix anything obviously broken. Should all be good now.

scottaohara (approved):

patrickhlauke:
Hurrah 💯

PR #3299, opened 2023-07-20
Remove the `class="remove"` sections, remove old page break navigation understanding

patrickhlauke:
* the `<section class="remove">` parts in the understanding documents are confusing and redundant, since they get overwritten at publication time with the actual guideline text from the `guidelines` folder
* the page break navigation understanding document was left over, even though the SC was removed


patrickhlauke:
@alastc possibly another one for @iadawn to consider

alastc:
Hi Patrick,

The problem is that if we edit the understanding documents and preview them, there is no SC text included.

It is a pain keeping them synced, but I think we had more confusion from not having them included.

patrickhlauke:
> I think we had more confusion from not having them included

my experience has been the opposite...particularly when things do get out of sync, and you're working directly in the repo

patrickhlauke:
now that 2.2 is published, i'd still strongly push for this cleanup here to take place @iadawn as otherwise we once again run the risk of having outdated scraps of older wordings of SCs hang around in our repo

iadawn:
I am going to merge this in to remove the sections of content that are outdated. The question of how we deal with previews is separate I feel and can be picked up elsewhere

patrickhlauke:
wonderful, thanks @iadawn

Issue #3289, opened 2023-07-18
[design] consider in the Understanding docs if we want to provide a link to /TR/ doc


shawna-slh:
comments in [AG WG telecon 18 July 2023](https://www.w3.org/2023/07/18-ag-minutes.html#t10) starting at:

> laura: side comment, from Understanding, how does one get to spec ?

including Shadi:
> [2.0 structure sent people in circles because people who came from the QuickRef to the Understanding were sent back to the spec rather than to the QuickRef, and they were disoriented]

scottaohara:
i had made this issue during the meeting as well https://github.com/w3c/wcag/issues/3288

patrickhlauke:
Making this the canonical version of the issue, as there's 3 other ones on the same topic.

and yes, @scottaohara was in there first, so credit to him for raising the original issue when others didn't even know/notice that the links used to be there but are now gone:

> here is a version from the [web archive where 1.1.1](http://web.archive.org/web/20190322213913/https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html) has intro text that links back to the SC,
> 
> If you look through each understanding doc from the wayback machine links, you'll find that every single Understanding Doc started with the normative text, where the first bit of text in each one was "Success Criterion [number/name] (Level X)" - and the number/name text linked back to the primary specification.
> 
> that doesn't exist in the newest versions.
> 
> Additionally, I've noticed direct links to SCs, such as 4.1.2 in the SC text of [1.1.1 in the current docs](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html#name-role-value) do not link anywhere anymore, where they used to link to the SC in the official spec.

Further comment/description from @lauracarlson on one of the other duplicates:

> The previous versions of WCAG had a link to the specification [for example 2.0's 1.2.1 ](https://www.w3.org/TR/UNDERSTANDING-WCAG20/media-equiv-av-only-alt.html) and  [2.1's 1.2.2](https://web.archive.org/web/20190713065229/https://www.w3.org/WAI/WCAG21/Understanding/audio-only-and-video-only-prerecorded). The 2.2 understanding docs currently lack a link, [for example 2.2's 1.2.1 ](https://www.w3.org/WAI/WCAG22/Understanding/audio-only-and-video-only-prerecorded).
> 
> Many people will find the understanding doc from search and can't use back button.

patrickhlauke:
[...]

Issue #3288, opened 2023-07-18
Understanding docs no longer link back to the spec


scottaohara:
here is a version from the [web archive where 1.1.1](http://web.archive.org/web/20190322213913/https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html) has intro text that links back to the SC,

If you look through each understanding doc from the wayback machine links, you'll find that every single Understanding Doc started with the normative text, where the first bit of text in each one was "Success Criterion [number/name] (Level X)" - and the number/name text linked back to the primary specification.

that doesn't exist in the newest versions.

Additionally, I've noticed direct links to SCs, such as 4.1.2 in the SC text of [1.1.1 in the current docs](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html#name-role-value) do not link anywhere anymore, where they used to link to the SC in the official spec.

scottaohara:
new related issues:

- https://github.com/w3c/wcag/issues/3293
- https://github.com/w3c/wcag/issues/3292

patrickhlauke:
Making https://github.com/w3c/wcag/issues/3289 the canonical version and closing this as a duplicate.

PR #3310, opened 2023-07-28
Update identify-input-purpose.html

mbgower:
Replaced relative link with FQDN link https://www.w3.org/TR/WCAG21/#input-purposes since the existing one is triggering nothing when it appears at the top of the Understanding document.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/pull/3310.html" title="Last updated on Jul 28, 2023, 2:16 PM UTC (388cad8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/3310/66c5fec...388cad8.html" title="Last updated on Jul 28, 2023, 2:16 PM UTC (388cad8)">Diff</a>

### DIFF
--- a/guidelines/sc/21/identify-input-purpose.html
+++ b/guidelines/sc/21/identify-input-purpose.html
@@ -7,7 +7,7 @@
 	<p>The purpose of each input field collecting information about the user can be <a>programmatically determined</a> when:</p>
   
   <ul>
-    <li>The input field serves a purpose identified in the <a href="#input-purposes">Input Purposes for user interface components section</a>; and</li>
+    <li>The input field serves a purpose identified in the <a href="https://www.w3.org/TR/WCAG21/#input-purposes">Input Purposes for user interface components section</a>; and</li>
     <li>The content is implemented using technologies with support for identifying the expected meaning for form input data.</li>
   </ul>
   


scottaohara:
this is related to the larger issue of the [understanding docs losing their link back to the spec](https://github.com/w3c/wcag/issues/3289) - and that [all links back to the spec from understanding docs are broken](https://github.com/w3c/wcag/issues/3288) - as this PR further identifies.

fixing these should obviously be done, but is a one off PR the way to go, or is there a way to fix all of these broken links at once?

alastc:
I think this is a separate thing. There is an the include for SCs that is pulled into both the Spec and the understanding docs.

For definitions, the relative link works because they are also on both. However, this is to an appendix, the build-template doesn't account for that.

So I think a hard-coded link is the way to go, but we'll need to do different ones for WCAG 2.1 and 2.2.

scottaohara:
i understand your take on this @alastc, i was coming from the mindset that in other specs like HTML AAM a generic URL can be used in the source and through the build process of the spec, the necessary URLs can be populated in.  So, if that's not a feature that can be used over here, then yes.  agreed.  a separate issue than the primary issue that i linked to.  But, i did intend to be calling out these broken links, as well, in the original issue i filed,

TLC82 (changes requested):
Delete

kfranqueiro:
The Eleventy-based build system handles this case correctly for 2.2 now, and will eventually also be used to rebuild 2.1 docs (see #4007). This change should not be necessary (and we should avoid hard-coding links to specific versions of the spec), so I'll close this.

Issue #3314, opened 2023-07-30
WCAG 2.2 understanding documents: notes for SCs are not indicated as notes (fails 1.3.1 Info and Relationships)


patrickhlauke:
Take for instance https://w3c.github.io/wcag/understanding/focus-not-obscured-minimum.html

![screenshot of the top of the understanding document, showing the SC title and a box containing the SC wording](https://github.com/w3c/wcag/assets/895831/4f4f11be-4599-48b4-8f9f-56d536ff4263)

The second and third paragraph in the boxout for the SC wording are *notes*, but this is not indicated in the understanding document.

In https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-minimum.html notes are at least "boxed-out", but this is only a visual style, and the word "Note" is not present. As such, the fact that these are notes is not conveyed (failing 1.3.1 Info and Relationships in the process)

![screenshot of the top of the understanding document, showing the SC title and a box containing the SC wording, with the notes at least visually boxed-out](https://github.com/w3c/wcag/assets/895831/208ce9a4-4d29-4e4e-bc76-41f8615f4036)

Notes are currently identified (in the main spec) using CSS and generated content. This here demonstrates why that may not actually be a good idea.

Suggest that:

* either add the "Note:" text into the actual HTML, or
* make sure the understanding documents have the same CSS that also does the generated content/extra styling

mraccess77:
I would also prefer if we don't solve this with pseudo content as well as sometimes that can't be searched and may be used by people who customize style sheets.

iadawn:
I have just updated a change to the formatting of notes in the Understanding documents. Main changes:

* Included the word 'Note' - as proper content, no pseudo-content here
* removed the box styling - when in SC this box styling visually dominated quite a bit. I can easily put it back.
* removed the role="heading" from the word 'Note' - as present in the Understanding documents, I am not 100% that these are headings. Open to any debate on that one though.

iadawn:
Closing as there was no dissent

PR #3318, opened 2023-08-01
Address #3314

iadawn:
@alastc Worth a quick discussion on use of headings for Notes in Understanding docs


AbdullahWins (changes requested):
review the code again please!

PR #3331, opened 2023-08-07
Updating the WCAG 2.2 reference

alastc:
Just adjusting "from WCAG 2.2" to "as of 2022" for the WCAG 2.1 version.

### DIFF
--- a/guidelines/terms/20/general-flash-and-red-flash-thresholds.html
+++ b/guidelines/terms/20/general-flash-and-red-flash-thresholds.html
@@ -44,7 +44,7 @@
       of two opposing transitions.
    </p>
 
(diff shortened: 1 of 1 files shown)


bruce-usab (approved):
I am okay with  file as-is, but I am not confident that qualifier of "new" is accurate.  (Yes, new to WCAG, but new to the field?

Proposed edit:

`<p class="note">The definition <q><strong>pair of opposing transitions involving a saturated red</strong></q> (as of 2022) is a pair of opposing transitions where...`

Another editorial nit pick is using both quotation marks and bold, but if so, I think the quotation marks (semantic) should be outside the presentational markup.

PR #3347, opened 2023-08-21
Text-spacing note for I18n

### LINKED ISSUE #3338: Text spacing exception unclear
https://www.w3.org/TR/WCAG22/#text-spacing

SC 1.4.12 says:

> In content implemented using markup languages that support the following [text](https://www.w3.org/TR/WCAG22/#dfn-text) [style properties](https://www.w3.org/TR/WCAG22/#dfn-style-properties), no loss of content or functionality occurs by setting all of the following and by changing no other style property:
>*    Line height (line spacing) to at least 1.5 times the font size;
>*    Spacing following paragraphs to at least 2 times the font size;
>*    Letter spacing (tracking) to at least 0.12 times the font size;
>*    Word spacing to at least 0.16 times the font size.
>
>Exception: Human languages and scripts that do not make use of one or more of these text style properties in written text can conform using only the properties that exist for that combination of language and script.

In various discussions, including in #2680 and the [understanding link](https://www.w3.org/WAI/WCAG22/Understanding/text-spacing.html), it's been made clear that these criteria are related to changes that the user makes when overriding the author's spacing.

I18N observes that there are two problems with the criteria. First, the proposed numbers may not be the right recommendation for all writing systems to which they can be meaningfully applied. For example, some writing systems might require more or less line spacing to be fully legible and thus the guidelines adjusted to match those requirements.

The more significant problem is that each of the bullet items does not always apply to a specific writing system's usage. This is what the exception is trying to address.

For example, many writing systems do not use spaces between words, so adjusting word spacing up to 0.16 times the font size probably has no meaning for languages written in those scripts. But in other cases, a concept may have meaning in the writing system, but still be contrary to common usage. For example, Chinese and Japanese do not generally use inter-paragraph spacing different from line-height, but there is a concept of inter-paragraph spacing. It is unclear if the exception applies to these.

Rewording the exception to be clearer about the intentions of the SC and nature of the exception would help here. Perhaps:

> Be aware that users of human languages or writing systems that have different needs with regard to spacing might not adjust these values or might choose to adjust other values (such as paragraph start indent) in order to improve legibility. Writing systems have different needs and this success criterion might not be sufficient for those needs. Authors should allow for values that exceed these if their language requires it or be mindful of additional modifications appropriate to that language.
> 
> Some human languages and writing systems do not have spaces between words or support modification of letter spacing. The display of text in these languages will not be affected by the user changing these settings.
>
> Some writing systems have the style properties listed here (such as paragraph spacing) but do not generally apply these in text. For example, most CJK documents do not have different paragraph spacing versus line spacing. Authors should still allow for adjustments consistent with this success criterion, even though users will probably not use these settings in these languages.

---

Some additional notes:

>    Word spacing to at least 0.16 times the font size.

Not all writing systems have word spacing. Writing systems like Balinese, Batak, Tai Le, Khmer, Lao, Burmese, Thai, Tibetan, Javanese and Chinese/Japanese don't have word spacing. Some of these writing systems use space as a phrase separator, some of them only have syllable separators, and some don't use any space character at all.

Also, in Samaritan and sometimes in Amharic, words are separated by special punctuation instead of a space character.

In our discussion, we also discussed this as wording for an exception:

> These success criteria only apply to human languages written in the Latin script. Human languages and writing systems do not always use one or more of these text style properties. In such cases, the text is conformant as long as it applies properties appropriate for that language and/or writing system.

alastc:
Closes #3338

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/pull/3347.html" title="Last updated on Sep 22, 2023, 2:49 PM UTC (b91b33e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/3347/48b07d4...b91b33e.html" title="Last updated on Sep 22, 2023, 2:49 PM UTC (b91b33e)">Diff</a>


scottaohara (approved):

AbdullahWins (approved):

guidelines/sc/21/text-spacing.html
```diff
@@ -14,5 +14,8 @@ <h4>Text Spacing</h4>
    </ul>
  
  <p>Exception: Human languages and scripts that do not make use of one or more of these text style properties in written text can conform using only the properties that exist for that combination of language and script.</p>
+
+ <p class="note">Content is not required to use the values specified, the requirement is to ensure no-loss of content or functaionality when a user over-rides the authored styles with the values.</p>
```
r12a, line 18: functaionality -> functionality

and, i think, over-rides -> overrides https://dictionary.cambridge.org/dictionary/english/override

r12a: perhaps also: no-loss -> no loss, or 'ensure no-loss' -> avoid loss

understanding/21/text-spacing.html
```diff
@@ -23,7 +23,8 @@ <h2>Intent</h2>
...
...cified metrics, nor does the SC intend to imply that all users will adjust the specified metrics. Rather, it specifies that an author's content has the ability to be set to those metrics without loss of content or functionality. The author requirement is both to not interfere with a user's ability to override the author settings, and to ensure that content thus modified does not break content in the manners shown in figures 1 through 3 in Effects of Not Allowing for Spacing Override. The values in the SC are a baseline. For content written in languages where the specified metrics would help increase readability, authors are encouraged to surpass the values specified, not see them as a ceiling to build to. These metrics are not meant to represent a ceiling for authors to build to, nor as minimum requirements to adhere to, as some metrics may not even be applicable to certain languages. For example, languages such as Japanese do not use spacing following paragraphs, meaning that users won't necessarily force any paragraph spacing metrics in practice. If the user chooses to go beyond the metrics specified any resulting loss of content or functionality is the user's responsibility.</p>
```
r12a, line 26: I have heard that China has some related typographic guidelines, and suspect that that may also be the case for Japanese.  Therefore, i'd suggest adding a sentence or paragraph to recommend that content should be amenable to supporting any local guidelines for adjusting typography in order to improve readability.

alastc: How about "It is benficial for users if authors use any locally available guidance for improving readability in the local language or writing system."

NB: We have to avoid RFC-like words such as "should" for this sort of recommendation, some people see that as adding requirements by the back door.

murata2makoto: It appears that neither the preview button nor the diff button work.

understanding/21/text-spacing.html
```diff
@@ -23,7 +23,9 @@ <h2>Intent</h2>
...
+         <p>This SC does not dictate that authors must set all their content to the specified metrics, nor does the SC intend to imply that all users will adjust the specified metrics. Rather, it specifies that should a user choose to set any of these metrics they can do so without any loss of content or functionality. The author requirement is both to not interfere with a user's ability to override the author settings, and to ensure that content thus modified does not break content in the manners shown in figures 1 through 3 in Effects of Not Allowing for Spacing Override.</p>
+         <p>These metrics are not meant to represent minimum requirements to adhere to, as some metrics may not even be applicable to certain languages. For example, languages such as Japanese do not use spacing following paragraphs, meaning that users are unlikely to make any paragraph spacing changes in practice. If the user chooses to go beyond the metrics specified any resulting loss of content or functionality is the user's responsibility. It is benficial for users if authors use any locally available guidance for improving readability in the local language or writing system.</p>
```
dbjorge, line 27: "These metrics are not meant to represent minimum requirements to adhere to" is just incorrect; that's *exactly* what these metrics are meant to represent, it's just that some of them are inapplicable in some circumstances.

```suggestion
         <p>In some human languages and scripts, some of the metrics specified by the SC are inapplicable. For example, languages such as Japanese do not use spacing following paragraphs, meaning that users are unlikely to make any paragraph spacing changes in practice. The exception in this SC allows authors to ignore text style properties which are inapplicable to the combination of language and script being used.</p>
         <p>If the user chooses to go beyond the metrics specified, any resulting loss of content or functionality is the user's responsibility. It is beneficial for users if authors use any locally available guidance for improving readability in the local language or writing system.</p>
```

guidelines/sc/21/text-spacing.html
mraccess77: I'd personally go further than saying it's not required to use the value to say it's not implying or recommending these values be used as the default..  Nothing in the SC is encouraging their use as the default method at all and many folks would not find it useful - for me personally I would find content harder to read with those values.  The focus as we rightly say when they are applicable for the ability for them to be overwritten up to those values.

understanding/21/text-spacing.html
alastc: @murata2makoto, no they are auto-generated but don't account for how the WCAG spec works. I can make a manual [preview of the understanding document](https://raw.githack.com/w3c/wcag/Issue3338-text-spacing-i18n/understanding/21/text-spacing.html), but not the SC text.

The whole [SC text file](https://github.com/w3c/wcag/pull/3347/files#diff-1db1537d239fb0076187e01dc53b9d8c2a0dedb423e89350a445dbb922c640f8) is only 20 lines, so should be ok to review as code.

alastc: The suggested line has been added.

guidelines/sc/21/text-spacing.html
mraccess77: I just thought it might help the objections if we were even stronger in the language that it wasn't just required but not recommended as the default.

alastc: I'd rather avoid any 'recommend' language if we can, we have to focus on the requiments in the normative language / area.

mraccess77: Sure, we could say something such as this does not imply these be used as the default values.

guidelines/sc/21/text-spacing.html
```diff
@@ -14,5 +14,8 @@ <h4>Text Spacing</h4>
    </ul>
  
  <p>Exception: Human languages and scripts that do not make use of one or more of these text style properties in written text can conform using only the properties that exist for that combination of language and script.</p>
+
+ <p class="note">It is not required that content use the values specified. The requirement is to ensure no loss of content or functionality when a user overrides the authored styles with the values.</p>
+ <p class="note">Some human languages or writing systems have different needs and may adjust other values (such as paragraph start indent) in order to improve legibility. This success criterion might not be sufficient for those needs.</p>
```
aphillips, line 19: In our TPAC meeting, I18N suggests the following text to replace this paragraph, which incorporates some text from the understanding document (slightly modified)

```suggestion
<aside class="note">
<p>Some human languages or writing systems have different needs and may adjust other values (such as paragraph start indent) in order to improve legibility. This success criterion might not be sufficient for those needs. The exception in this SC allows authors to ignore text style properties which are inapplicable to the combination of language and script being used.</p>

<p>It is beneficial for users if authors use any locally available guidance for improving readability in the local language or writing system. If the user chooses to go beyond the metrics specified, any resulting loss of content or functionality is the user's responsibility.</p>
</aside>
```

understanding/21/text-spacing.html
```diff
@@ -23,7 +23,10 @@ <h2>Intent</h2>
...
+         <p>In some human languages and scripts, some of the metrics specified by the SC are inapplicable. For example, languages such as Japanese do not use spacing following paragraphs, meaning that users are unlikely to make any paragraph spacing changes in practice. The exception in this SC allows authors to ignore text style properties which are inapplicable to the combination of language and script being used.</p>
+         <p>If the user chooses to go beyond the metrics specified, any resulting loss of content or functionality is the user's responsibility. It is beneficial for users if authors use any locally available guidance for improving readability in the local language or writing system.</p>
```
aphillips, line 28: In the above suggestion, we swapped the first and second sentences. Would suggest the same here.

aphillips (changes requested):
I18N reviewed this in TPAC as prep for our meeting tomorrow. We suggest the change below.

guidelines/sc/21/text-spacing.html
alastc: Hi @aphillips, it's getting pretty long now, it's about 25% more words than the SC itself. 
The notes should only be to clarify the normative requirement.

Can we remove the sentence starting "The exception" as it replicates the SC text?

Also, the last paragraph is adding a beneficial other thing for authors to do. The place for that is the understanding doc, can we remove that paragraph from the notes?

aphillips: I18N discussed in our WG call today (2023-09-21). We think it is okay to remove the "The exception..." sentence but we think that our second paragraph is important.

Issue #3226, opened 2023-05-31
Minor clarity around a "change initiated by the user"


scottaohara:
In reviewing the 3.2.6 consistent help text, the question came up about what the expectations would be for the location of help / what's even offered as help when a user is interacting with a logged in version of a website vs a logged out version.   e.g., in a logged out version there's one location/form of help.  and in a logged in version that information and potentially even location changes because legit reasons.

To me, this sort of variation could be covered under "a set of web pages" - as while it's not explicitly stated there, I would infer from the e-commerce example that a logged out version of a website would be one set of pages - while a logged in version would constitute another set - so if a variance was found, that'd be ok so long as there was consistency for all logged in versions of the pages, and consistency for all logged out versions of the pages, but not necessarily with each other.  (if i'm wrong on this, please let me know, as while not the primary purpose of this issue, hey, i'd like to be corrected if needed)

But in talking through that, it was mused that logging into a website sure sounds like "a change initiated by a user".  Being an undefined term/phrase, I could see why one could reasonably come to that conclusion.  Being that this is phrase is used twice now in normative text, but its meaning is only expanded on in the understanding documents / as a note for consistent help, should this be defined as a term instead - e.g., so as to at least consistently call out its intent is around a user modifying the web page by use of zooming, orientation changing, possibly due to a browser setting of some other kind?, etc..  - And (assuming) not due to logging into a website, or changing a setting of the website itself - which are still changes initiated by the user, but not in the context of how this phrase has been described in its two uses?

scottaohara:
thanks @alastc appreciate your thoughts here.

>Overall, the intent was for user-agent driven changes, such as zoom...

and that's what i came to realize after reviewing the notes / informative docs for these two SCs, but that's why I was wondering if it didn't make sense to formally define that.  So as to mitigate such random musings of "well, a user interacting with the website could result in / be considered a change initiated by the user"

alastc:
Discussion in the WCAG backlog meeting:

- The second note covers the intent pretty well.
- Mixed opinions on whether a login would intrinsically count as a 'change initiated by the user'. It depends on the results rather than the action of login.
- We could add some content around 'change initiated by the user'. E.g. highlight what the changes might entail.

bruce-usab:
Add some examples to Understanding per backlog call 8-18.

dbjorge:
I'll create a PR that updates the understanding document with a paragraph of explanation + example

scottaohara:
[...]

PR #3349, opened 2023-08-22
3.2.6 Consistent Help Understanding: clarify "change initiated by the user"

dbjorge:
[Rendered version](https://raw.githack.com/dbjorge/wcag/3226-consistent-help-set-of-web-pages/understanding/22/consistent-help.html#limitations) (see mostly-new "Limitations and Exceptions" section)

This PR is intended to address #3226 by adding some additional clarification and examples based on WCAG 2 backlog meeting discussion + these two comments [[1]](https://github.com/w3c/wcag/issues/3226#issuecomment-1572066047) [[2]](https://github.com/w3c/wcag/issues/3226#issuecomment-1594897338).

Summary of the new explanations:

* Different "sets of web pages" *are* allowed to have different help mechanism locations (but it's best to be as consistent as possible)
* The SC text excepting cases where "a change is initiated by a user" is intended to cover cases where a user performs an action with the intent of changing the display or layout of a page, such as changing the zoom level, orientation, or viewport size
* That exception is *not* intended to treat treat every action that a user might take as a "user-initiated change" - specifically, navigating between pages within a set of web pages doesn't count, and neither does logging into a page (unless doing so would present the user with a new set of web pages)

The main "intent" section was getting pretty long with these additions, so I created a new "Limitations and Exceptions" subsection. I moved a few other existing paragraphs that felt relevant to the new section.

### DIFF
--- a/understanding/22/consistent-help.html
+++ b/understanding/22/consistent-help.html
@@ -70,7 +70,7 @@
          <section id="limitations">
             <h3>Limitations and Exceptions</h3>
 
-            <p>It is not the intent of this Success Criterion to require authors to provide help / access to help. The Criterion only requires that <em>when</em> one of the listed forms of help is available across multiple pages that it be in a consistent location. It does not require authors to provide help information on PDFs or other static documents that may be available for viewing/download from the Web pages. PDFs and other static documents are not considered part of the "set of web pages" from which they are downloaded.</p>
+            <p>It is not the intent of this Success Criterion to require authors to provide help or access to help. The Criterion only requires that <em>when</em> one of the listed forms of help is available across multiple pages that it be in a consistent location. It does not require authors to provide help information on PDFs or other static documents that may be available for viewing/download from the Web pages. PDFs and other static documents are not considered part of the "<a>set of web pages</a>" from which they are downloaded.</p>
 
             <p>It is also not the intent of this Success Criterion to require a human be available at all times. Ideally, if the human contact is not available during certain hours or certain days then information would be provided so the user can tell when it will be available. </p>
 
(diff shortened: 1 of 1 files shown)


AbdullahWins (approved):
well done

understanding/22/consistent-help.html
```diff
@@ -62,18 +62,28 @@ <h2>Intent of Consistent Help</h2>
...
 
-         <p>It is not the intent of this Success Criterion to require authors to provide help / access to help. The Criterion only requires that <em>when</em> one of the listed forms of help is available across multiple pages that it be in a consistent location. It does not require authors to provide help information on PDFs or other static documents that may be available for viewing/download from the Web pages. PDFs and other static documents are not considered part of the "set of web pages" from which they are downloaded.</p>
+         <section id="limitations">
+            <h3>Limitations and Exceptions</h3>
+
+            <p>It is not the intent of this Success Criterion to require authors to provide help / access to help. The Criterion only requires that <em>when</em> one of the listed forms of help is available across multiple pages that it be in a consistent location. It does not require authors to provide help information on PDFs or other static documents that may be available for viewing/download from the Web pages. PDFs and other static documents are not considered part of the "set of web pages" from which they are downloaded.</p>
```
scottaohara, line 73: light suggestion while this is being updated, but the lack of this instance of 'set of webpages' being linked, and the use of '/' instead of 'or' always struck me as odd.  but, no need to merge this if people disagree with my nits.

```suggestion
            <p>It is not the intent of this Success Criterion to require authors to provide help or access to help. The Criterion only requires that <em>when</em> one of the listed forms of help is available across multiple pages that it be in a consistent location. It does not require authors to provide help information on PDFs or other static documents that may be available for viewing/download from the Web pages. PDFs and other static documents are not considered part of the "<a>set of web pages</a>" from which they are downloaded.</p>
```

scottaohara (approved):
this update covers exactly what i was hoping to have clarified.  thank you.

bruce-usab (approved):

Issue #2109, opened 2021-10-25
Clarify if SC 2.4.4 requires paragraph to be a P element in HTML


mraccess77:
SC 2.4.4 allows for link purpose within a paragraph.   Paragraph is not defined - so generally we fall to the dictionary definition of paragraph which appears to be:

"a distinct section of a piece of writing, usually dealing with a single theme and indicated by a new line, indentation, or numbering."

In that case a link within a span, section, div, p, etc. could still pass as being in a requirement if it meet the above.

Related to issues #1745

detlevhfischer:
In my book the difinition would apply to `div` the same way as it applies to `p`.
I have included an example for both sentence in `p` and in `div` at the bottom of http://3needs.org/en/testing/code/link-issues.html
In Jaws, _ALT + up arrow_ behaves the same in both cases (reading sentence up to the link focused).

scottaohara:
should reference HTML to make it more clear that paragraph is a `p`, but **doesn't have to be** a `p`.

https://html.spec.whatwg.org/multipage/dom.html#paragraphs

mraccess77:
JAWS also supports the control+numpad 5 command to read the current paragraph for divs and other flow content.

detlevhfischer:
> JAWS also supports the control+numpad 5 command to read the current paragraph for divs and other flow content.

Strange - that was the first command I tried but that didn't work here (just red the link itself)...

PR #3362, opened 2023-08-25
Update mention of "list" to "list item" in programmatically-determine…

### LINKED ISSUE #3361: Clarity for programmatically determined link context term example
The current (and by that I mean since wcag 2.0) [programmatically determined link context](https://w3c.github.io/wcag/guidelines/#dfn-programmatically-determined-link-context) term example text reads:

>In HTML, information that is programmatically determinable from a link in English includes text that is in the same paragraph, list, or table cell as the link or in a table header cell that is associated with the table cell that contains the link.

The issue I see with this is the simple mention of 'list', rather than 'list item' - particularly in relation to the other examples provided.

Without that qualification, it reads that the following would be fine:

```
<ul>
  <li>All my context</li>
  <li>...</li>
  <li><a href=...>learn more</a></li>
</ul>
```

but based on all the issues/discussion that already exist concerning this topic, and the additional examples related to tables in that example text, the following (text and link within the same list **item**) is likely what was actually intended):

```
<ul>
  <li>All my context <a href=...>learn more</a></li>
  <li>...</li>
</ul>
```

Additionally, I think to help resolve issues like https://github.com/w3c/wcag/issues/2109 - since this example text is already mentioning "in html", it might be a good idea to link the word 'paragraph' to the [HTML definition of paragraph](https://html.spec.whatwg.org/multipage/dom.html#paragraphs), so as to help squash the question of "does paragraph mean 'paragraph' or `p` element?" without having to amend that text further to indicate that `<div>.... <a href>...</a></div>` or `<blockquote>.... <a href=>...</a> ...</blockquote>`, etc. are also providing said context.

scottaohara:
closes #3361

updates the term example to mention "list item" instead of "list" to be more accurate to what the intent of that likely was trying to convey.

Additionally, links to the HTML term of 'paragraph', which I would submit could resolve issues such as #2109

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/pull/3362.html" title="Last updated on Aug 25, 2023, 7:14 PM UTC (44450fb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/3362/6b2b211...44450fb.html" title="Last updated on Aug 25, 2023, 7:14 PM UTC (44450fb)">Diff</a>

### DIFF
--- a/guidelines/terms/20/programmatically-determined-link-context.html
+++ b/guidelines/terms/20/programmatically-determined-link-context.html
@@ -5,8 +5,8 @@
    </p>
    
    <aside class="example"><p>In HTML, information that is programmatically determinable from a link in English
-      includes text that is in the same paragraph, list, or table cell as the link or in
-      a table header cell that is associated with the table cell that contains the link.
+      includes text that is in the same <a href="https://html.spec.whatwg.org/multipage/dom.html#paragraph">paragraph</a>, 
+      list item, or table cell as the link or in a table header cell that is associated with the table cell that contains the link.
    </p></aside>
    
    <p class="note">Since screen readers interpret punctuation, they can also provide the context from


bruce-usab (approved):
Thanks for circling back to this detail.

guidelines/terms/20/programmatically-determined-link-context.html
```diff
@@ -5,8 +5,8 @@
    </p>
    
    <aside class="example"><p>In HTML, information that is programmatically determinable from a link in English
-      includes text that is in the same paragraph, list, or table cell as the link or in
-      a table header cell that is associated with the table cell that contains the link.
+      includes text that is in the same <a href="https://html.spec.whatwg.org/multipage/dom.html#paragraph">paragraph</a>, 
+      list item, or table cell as the link or in a table header cell that is associated with the table cell that contains the link.
```
WilcoFiers, line 9: I don't think WCAG should link to an evergreen HTML spec. These links can and do change. We shouldn't put ourselves in a position where we have to republish WCAG because WHATWG updated a definition.

alastc: The link has been stable for [10 years](https://web.archive.org/web/20140901000000*/https://html.spec.whatwg.org/multipage/dom.html#paragraph), and if the paragraph in HTML were tweaked, wouldn't we want to use the updated version?

WilcoFiers: The problem isn't the definition changing. It's the link changing. WHATWG makes no guarantee about the location of their definitions never changing. ACT has had to update rules because HTML definitions were moved. It happens. That's an easy fix when this is in an understanding document. Not so much for a normative document. This isn't the first link to WHATWG in WCAG, but the more we do this, the greater the chance of a broken in the spec at some point in the future.

scottaohara: the html standard is evergreen.  you can point to a snapshot of it, but you'll get a message like this:
>This document contains the contents of the standard as of the [9025fb30a05b2275016909bc782925d2855a9a03 commit](https://github.com/whatwg/html/commit/9025fb30a05b2275016909bc782925d2855a9a03), and should only be used as a historical reference. This commit may not even have been merged into the main branch.

>  
>Do not attempt to implement this version of the standard. Do not reference this version as authoritative in any way. Instead, see https://html.spec.whatwg.org/ for the living standard.

this seems worse to me than linking to the actual specification.  more and more w3c specs are also becoming living standards.  it seems maybe a conversation between the spec editors to make sure links can remain stable is a potential avenue to consider, rather than not linking to the HTML spec.?

WilcoFiers: No it hasn't. I for one intend to object to any proposal that would result in having to update the million copies of WCAG annually. I'm not going to count on that for a solution. I think an agreement with WHATWG that certain links will never be changed could work, but such an agreement doesn't exist.

Probably the easiest solution would be for this to link to a W3C page that redirects to WHATWG. If the URL ever changes, updated where the redirect goes. That can be done without any change to WCAG.

iadawn: @WilcoFiers To fix a link we could do an in-place change. This would not result in a change to the date of the document. There would be a note included in the Status section of the doc and the document title would get an addition to note the change. In terms of Process this is not arduous. Given the low likelihood of the WHATWG URI changing would this be acceptable?

Your idea for having a redirect is technically feasible. I would need to check what Webmaster would say about that since it is allowing for a change in a REC document without the change being documented in the document. Basically, in-place change without the Status note.

WilcoFiers: @iadawn Linking to an evergreen spec always comes with the risk of an undocumented change. I think that's okay for an example. I think a redirect is a better solution here, since this note is going to end up being copied, link and all, into other places. With a redirect you prefer those copies from having issues too.

PR #2614, opened 2022-08-12
add one-time-code to list of autocomplete attrs

fstrr:
Closes #2613

PR #3380, opened 2023-09-08
Update input-purposes for syntax highlighting

fstrr:
In [section 7](https://www.w3.org/TR/WCAG22/#input-purposes) of WCAG 2.1/2.2, there is a list of "Input Purposes for User Interface Components".

This change simply updates the code from using `<strong>` tags to `<code>` tags, including the language attribute so they are styled as HTML.

The related [PR 2614](https://github.com/w3c/wcag/pull/2614/) includes this change, but also added a normative change. This PR is just to add syntax highlighting support.

This would apply to WCAG 2.1 and 2.2.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/pull/3380.html" title="Last updated on Feb 14, 2024, 4:38 PM UTC (a79c0e4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/3380/8467c21...a79c0e4.html" title="Last updated on Feb 14, 2024, 4:38 PM UTC (a79c0e4)">Diff</a>


fstrr:
@alastc per today's backlog meeting, here's Input Purposes without the one-time-password line

scottaohara (approved):

alastc:
@iadawn I think this is editorial (code change not content), but it's in the normative doc. I assume we don't need group approval (due to it being editorial), but won't appear until 2.2 is re-published?

mbgower:
@alastc I have moved this to the Done column, even though it is not merged -- and cannot be until republish, I believe. We don't really have another column to put it. I'm just going to leave you assigned to it for now.

jenstrickland:
Regarding the conversation in AGWG today around the `dfn` element the following is valid, semantic markup in the html validator.

```
<!doctype html>
<html lang="en">
<head>
<title>proof of concept</title>
</head>
<ul>
<li><dfn>term</dfn> description</li>
<li><dfn>term2</dfn> description2</li>
</ul>
</html>
```

Reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dfn

mraccess77:
While it could be valid - my understanding is that the parent p, section, or dd/dt would become the implied  definition per the spec and information in that MDN article.  I think that is why we would need to use a description list if we used dfn otherwise the definition would include too much.

jenstrickland:
I've used the dfn for over 15 years and it is applicable in a list item. Dan Cederholm and Ethan Marcott's books use the pattern quite a lot. 

It's only necessary for the definition to be included in the ancestor, which it is.

patrickhlauke:
is there any actual benefit (for users, right now, not just theoretically) to using `<dfn>` ? or is this just semantics for semantics' sake?

jenstrickland:
For some of us with cognitive considerations it helps. It's also useful programmatically for the non-humans parsing content. I also read that screen reader users can choose settings to leverage the dfn and abbr tags.

patrickhlauke:
> It's also useful programmatically for the non-humans parsing content

that falls mostly under the "theoretical" part, unless there's actual examples out there of such tools.

> I also read that screen reader users can choose settings to leverage the dfn and abbr tags

that's news to me, I have to admit. would be good to know if this is actually happening, rather than again just being theoretical. purely for the `<abbr>` case, it's not looking good https://www.powermapper.com/tests/screen-readers/labelling/acronym-abbr-title/

mbgower:
@jenstrickland 
Do you see the existing changes proposed in this PR  -- replacing the strong element with the code element-- as producing a better or worse outcome than what exists right now, or making no difference? The reason I ask is that if you view code as preferable to strong, I believe the easiest path forward is to adopt the PR, and make a new issue, wherein the relative merits of <def>, <dt> and any other definition pairs can be properly hashed out.
If you view the use of the code element as detrimental to the existing structure, then I think you should vote thumbs down on this PR, which may cause it to be sent back In Progress in the TF, where the approach can be debated.

jenstrickland:
The 'code' element is definitely an improvement from the 'strong'.

'dfn' would be an enhancement. I figured better to do it at the same time. Happy to submit a PR separately. Thank you for your kind response!

bruce-usab:
Discussed on call, `dfn` should be a separate issuee

alastc:
Hi Jen,

> Regarding the conversation in AGWG today around the `dfn` element the following is valid, semantic markup in the html validator.

I think something missing from the conversation is what the `dfn` is intended for, which is where you have multiple instances of a term on the page, and the `dfn` is for the defining version of that term. (Hat tip to @scottaohara for that point.)

That isn't very explicit in the specs, but if you look at the examples it become clearer.

That usage is not what we're doing in WCAG, as each is the only instance of the term on the page.

Given that each item is intended to be an attribute value in HTML code, the `<code>` element makes the most sense to me.

Using a definition list is potentially an alternative, but it would be styled like the other instances (e.g. [1.1.1](https://www.w3.org/TR/WCAG22/#non-text-content)), I don't see any practical benefit from changing it from the current PR.

jenstrickland:
I'm not challenging the `code` element at all. 

I'm saying that it can be wrapped in a `dfn` as it is the defined term within the `li`. The language within the `li` explains the `code`.

The `dfn` does not require the use of a `dl`.

This is a separate issue from changing `strong` to `code`. I'm wishing I hadn't asked the question during the call.

bruce-usab:
@jenstrickland please open a new issue for your recommendation to use `dfn` in this context (or elsewhere in Understanding, if you like).

PR #3412, opened 2023-09-21
Fix #2914

### LINKED ISSUE #2914: Bug with highlight js
The redesign lets us have syntax highlighting via the Highlight JS library. It looks like all the code blocks we have need updating to add a `<code class="language-*">` attribute, which I'm happy to do. In playing around with a live page, though, it looks like there's a bug. Using [H37](https://www.w3.org/WAI/WCAG21/Techniques/html/H37) as an example:

## Current appearance and code
screen shot of html source code for an image element. The source code is displayed in the Courier typeface in black.

```html
<pre xml:space="preserve">&lt;img src="newsletter.gif" alt="Free newsletter. 
 Get free recipes, news, and more. Learn more."&gt;</pre>
```

## Appearance and code with a `code` element and highlight js added via dev tools
the same source code for an image, but the text is in a dark blue monospaced typeface

```html
<pre xml:space="preserve"><code class="language-html hljs">&lt;img src="newsletter.gif" alt="Free newsletter. 
 Get free recipes, news, and more. Learn more."&gt;</code></pre>
```

It's clear that something is happening because the font has changed style and color, and there's a white background to the code. However, the attribute names and values aren't highlighted as they are in the [Highlight JS demo](https://highlightjs.org/static/demo/).

## Console error
screenshot of a Dev Tools error stating Uncaught SyntaxError: Unexpected token '&' (at H37:201:59).

screenshot of the JavaScript referenced in the Dev Tools image.

iadawn:
When XSLT call-template is used and the template includes Javascript (probably anything really) with '=>' it helpfully converts the '>' to the HTML entity '&gt;'. This of course breaks the Javascript when it comes round to running the stuff.

This change uses a different approach to setting up highlight.js that does not need '=>'.

I should have probably worked out why the '>' is being converted to '&gt;' but realistically I know more about Javascript than I do about XSLT and life is short.


alastc:
I removed my attempt to fix this as well, which didn't seem to have worked on live.

PR #3436, opened 2023-09-29
WCAG 2.2 outdated files cleanup

patrickhlauke:
* Remove outdated page-break-navigation

In addition, I was going to try and expunge some of the very outdated rendered pages like

https://w3c.github.io/wcag/understanding/hidden-controls.html
https://w3c.github.io/wcag/understanding/visible-controls.html
https://w3c.github.io/wcag/understanding/page-break-navigation.html
https://w3c.github.io/wcag/understanding/dragging.html
https://w3c.github.io/wcag/understanding/focus-appearance-minimum.html
https://w3c.github.io/wcag/understanding/focus-appearance-enhanced.html
https://w3c.github.io/wcag/understanding/focus-visible-enhanced.html
https://w3c.github.io/wcag/understanding/findable-help.html
https://w3c.github.io/wcag/understanding/fixed-reference-points.html
https://w3c.github.io/wcag/understanding/pointer-target-spacing.html

but it seems these aren't actually in the repository ... I suspect these are somehow magically generated at build stage - @iadawn i also suspect that the directory where these then end up is not "purged" every time things are rebuilt, meaning that these old files that were generated at some point in the past always remain there, even when their original source isn't even there anymore ... worth adding a "delete everything in the dist folder" step to the build process?


patrickhlauke:
(Noting, for posterity, that this PR itself only removed that one outdated file ... the rendered ghost copies of understanding documents listed in the first comment here are still there, and will need to be removed/the publish process fixed more fundamentally, separately)

patrickhlauke:
@iadawn remembering the conversation we had late last year about removal of outdated documents ... not sure if the ones listed here in the opening comment would also qualify for outright deletion, or if they somehow should be preserved in amber for posterity

> https://w3c.github.io/wcag/understanding/hidden-controls.html
> https://w3c.github.io/wcag/understanding/visible-controls.html
> https://w3c.github.io/wcag/understanding/page-break-navigation.html
> https://w3c.github.io/wcag/understanding/dragging.html
> https://w3c.github.io/wcag/understanding/focus-appearance-minimum.html
> https://w3c.github.io/wcag/understanding/focus-appearance-enhanced.html
> https://w3c.github.io/wcag/understanding/focus-visible-enhanced.html
> https://w3c.github.io/wcag/understanding/findable-help.html
> https://w3c.github.io/wcag/understanding/fixed-reference-points.html
> https://w3c.github.io/wcag/understanding/pointer-target-spacing.html

Again, I just have concerns here because the rendered versions listed here have no connection anymore with the actual source files in the repository, so they're orphaned (and if at any point the rendered web version was purged and rebuilt, they'd disappear without a trace anyway).

PR #3443, opened 2023-10-04
Fix #2441

### LINKED ISSUE #2441: Ambiguous language defining Audio Description in Understanding 1.2.5
A common misinterpretation of 1.2.5 that I have noticed among accessibility professionals is that a separate Audio Description Track is required to narrate visual content, when in fact, the Audio Description may be included in the main (and only) soundtrack.

If my understanding is correct, "Audio Description" (AD) is narration that describes relevant information in the video content (information about actions, characters, scene changes, on-screen text, and other visual content.) AD may be part of the main soundtrack, or a separate track. 

Some of the language on Understanding implies that the AD must be a separate track (in addition to the main audio track). For example, the first line of the Note (green box) in the "Intent" section states, "For 1.2.3, 1.2.5, and 1.2.7, if all of the information in the video track is already provided in the audio track, **no audio description is necessary**". To me, the phrasing "no audio description is necessary" seems to be conflating AD with a separate audio track. 

Another example, from the first line of the "audio description" definition in the "Key Terms" section states, "narration added to the soundtrack to describe important visual details that cannot be understood from the main soundtrack alone". This language is ambiguous to me in that it states "...added to the soundtrack..." followed by, "main soundtrack alone". This implies that there are 2 soundtracks (soundtrack and main sound track).

I think it would be very beneficial to very clearly define Audio Description as the narration that may be included in the primary soundtrack or in a separate soundtrack.

mitchellevan:
Clarifications in Understanding docs consistent with the normative definitions, and conveying the spirit of comments in #2441:

- "Audio description" is not necessarily a separate audio track; it can also be insertions in the single audio track of a video. Rephrase anything that implies otherwise, such as "an audio description" referring to an audio track.
- "Audio description" is only something added, not something integral. If the video is sufficiently described by the original audio, it does not need audio description.

Also edited three Techniques for consistency with the Understanding edits, and for consistency with each other.


techniques/general/G173.html
```diff
@@ -1,10 +1,12 @@
...
-      <p>Since it is not helpful to have this new information obscure key audio information in the original sound track (or be obscured by loud sound effects), the new information is added during pauses in dialogue and sound effects. This limits the amount of supplementary information that can be added to program.</p>
-      <p>Providing a second version of the movie that includes audio descriptions as the primary sound track will make this content accessible for blind people who need to hear not only the dialogue, but also the context and other aspects of the video that are not communicated by the characters' dialogue alone.</p>
+      <p>The objective of this technique is to provide audio description in video content, so that it is possible for people who cannot see to understand audio-visual material.</p>
```
nigelmegitt, line 4: This description presupposes a specific solution, rather than describing the goal.

```suggestion
      <p>The objective of this technique is to ensure that it is possible for people who cannot see to understand audio-visual material. This is generally achieved by describing the video content within the audio, either with or without augmenting it with audio description.</p>
```

techniques/general/G173.html
```diff
@@ -1,10 +1,12 @@
...
+      <p>Since most user agents today cannot merge multiple soundtracks, this technique adds audio description information into the single audio track of a movie. The added information focuses on actions, characters, scene changes, and on-screen text (not captions) that are important to understanding the content.</p>
```
nigelmegitt, line 5: This is too prescriptive about what the solution needs to be. It is possible to arrange for multiple soundtracks to be merged, for example by using Web Audio.

It also very specifically excludes audio describing captions, but there are occasions, for example videos intended to demonstrate captions as a feature, where you _would_ want to describe them. (conversely I have had occasion to subtitle/caption the audio description in videos that demonstrate audio description as a concept)

```suggestion
      <p>This technique allows for audio description information to be made available either by creating an alternative audio track with audio description mixed into the audio track of a movie or by providing a separate audio track that is mixed with the main audio track by the player. The added information focuses on actions, characters, scene changes, and on-screen text (not usually captions) that are important to understanding the content. This is not needed if the main audio already describes the video.</p>
```

techniques/general/G173.html
```diff
@@ -1,10 +1,12 @@
...
+      <p>Since it is not helpful to have this new information obscure key audio information in the original soundtrack (or be obscured by loud sound effects), the new information is added during pauses in dialogue and sound effects. This limits the amount of supplementary information that can be added to the program.</p>
```
nigelmegitt, line 6: It's normal in most audio description use cases to dip the volume of the main soundtrack prior to adding in the audio description audio.

```suggestion
      <p>Since it is not helpful to have additional information obscure key audio information in the original soundtrack (or be obscured by loud sound effects), the new information is added during pauses in dialogue and sound effects. This limits the amount of supplementary information that can be added to the program. It is also possible to reduce the audio volume level of the original soundtrack before mixing audio description so that it is easier to hear and understand.</p>
```

techniques/general/G173.html
```diff
@@ -1,10 +1,12 @@
...
+      <p>Providing audio description in the movie's soundtrack will make this content accessible for blind people who need to hear not only the dialogue, but also the context and other aspects of the video that are not communicated by the characters' dialogue alone.</p>
```
nigelmegitt, line 7: This presupposes that the characters' dialogue is the only possible source of other context within the original soundtrack.

```suggestion
      <p>Providing audio description in the movie's soundtrack will make this content accessible for blind people who need to hear not only the dialogue, but also the context and other aspects of the video that are not communicated in the original soundtrack.</p>
```

techniques/general/G78.html
```diff
@@ -1,10 +1,9 @@
...
-      <p>Since it is not helpful to have this new information obscure key audio information in the original sound track (or be obscured by loud sound effects), the new information is added during pauses in dialogue and sound effects. This limits the amount of supplementary information that can be added to the program.</p>
-      <p>The soundtrack with the audio description (of visual information) can either be an alternate sound track that the user can choose, or it can be the standard sound track that everyone hears.</p>
+      <p>Since most user agents today cannot merge multiple soundtracks, this technique adds audio description information into a copy of the original soundtrack and allows users to switch to the other soundtrack. The added information focuses on actions, characters, scene changes, and on-screen text (not captions) that are important to understanding the content.</p>
```
nigelmegitt, line 5:
```suggestion
      <p>This technique allows for audio description information to be made available either by creating an alternative audio track with audio description mixed into the audio track of a movie or by providing a separate audio track that is mixed with the main audio track by the player. The added information focuses on actions, characters, scene changes, and on-screen text (not usually captions) that are important to understanding the content. This is not needed if the main audio already describes the video.</p>
```
```

File: techniques/general/G78.html [hunk 6]
```diff
@@ -1,10 +1,9 @@
...
+      <p>Since it is not helpful to have this new information obscure key audio information in the original soundtrack (or be obscured by loud sound effects), the new information is added during pauses in dialogue and sound effects. This limits the amount of supplementary information that can be added to the program.</p>
```
nigelmegitt, line 6:
```suggestion
      <p>Since it is not helpful to have additional information obscure key audio information in the original soundtrack (or be obscured by loud sound effects), the new information is added during pauses in dialogue and sound effects. This limits the amount of supplementary information that can be added to the program. It is also possible to reduce the audio volume level of the original soundtrack before mixing audio description so that it is easier to hear and understand.</p>
```

techniques/general/G78.html
```diff
@@ -1,10 +1,9 @@
...
-      <p>Since most user agents today cannot merge multiple sound tracks, this technique adds the additional audio information to synchronized media by providing an option which allows users to replace the soundtrack with a new copy of the original soundtrack that has the additional audio description added. This added information focuses on actions, characters, scene changes and on-screen text (not captions) that are important to understanding the content.</p>
-      <p>Since it is not helpful to have this new information obscure key audio information in the original sound track (or be obscured by loud sound effects), the new information is added during pauses in dialogue and sound effects. This limits the amount of supplementary information that can be added to the program.</p>
-      <p>The soundtrack with the audio description (of visual information) can either be an alternate sound track that the user can choose, or it can be the standard sound track that everyone hears.</p>
```
nigelmegitt, line 7: Seems like the text from earlier could be used instead of this deletion:

```
<p>A single movie with audio description in its single audio track can be presented to all users. However, audio description can be distracting for those who do not need it, so versions of the movie with and without audio description can be provided, with a mechanism for users to choose the version they prefer.</p>
```

understanding/20/audio-description-or-media-alternative-prerecorded.html
```diff
@@ -58,23 +61,22 @@ <h2>Intent of Audio Description or Media Alternative (Prerecorded)</h2>
          
          <p>
             For 1.2.3, 1.2.5, and 1.2.7, if all of the information in the video track is already
-            provided in the audio track, no audio description is necessary. 
-            
-            
+            integrated into and provided by the main soundtrack, it is not necessary to add audio
+            description.
```
nigelmegitt, line 65: This is a key point, but it's no good only having it in the "understanding" section. Hence my edits to the normative sections earlier in this PR, which state this explicitly.

nigelmegitt (changes requested):
This is a good start to resolving #2441. I've made a variety of suggestions to bring it into line with common practice, based on experience - my organisation, the BBC, originates a very large amount of audio described content each year, but it is not an appropriate technique in all cases. In all cases though, the goal is for the content to be understandable by people not able to see the video image.

Therefore the most important thing from my perspective is that the normative part of 1.2.5 needs to make clear what was already in the Understanding text, that if the original soundtrack _already_ includes sufficient description of the video image, then there is no need to provide an additional audio described version.

Issue #323, opened 2018-06-08
Glossary: definition of "prerecorded"


jkshapiro:
The glossary term for "prerecorded" in WCAG2.1, https://www.w3.org/TR/WCAG21/#dfn-prerecorded, simply says "information that is not live". However, the definition for "live", https://www.w3.org/TR/WCAG21/#dfn-live, specifically excludes information that is "completely computer generated".

I think there is computer generated information that doesn't meet the usual English definition of "prerecorded". some examples:
* a real-time clock display
* a screensaver
* a pseudorandom, time-sensitive code used for two-factor authentication

Would it make sense to create a third category for computer generated information? Or, alternatively, to specifically call it out as being considered "prerecorded"?

jkshapiro:
There's also some overlap with https://www.w3.org/TR/WCAG21/#dfn-real-time-events

cstrobbe:
Here are all the phrases from success criteria that use the adjective "prerecorded":

* prerecorded audio-only [media] (SC 1.2.1),
* prerecorded video-only media (SC 1.2.1),
* prerecorded audio content in synchronized media (SC 1.2.2, ,
* prerecorded video content (SC 1.2.3),
* prerecorded audio content in synchronized media (SC 1.2.6, 1.2.7).

With regard to the examples:
1. A real-time clock fits neither the current definition of "precorded" nor the definition of live but is more likely to be provided as updating text than in a video format. (If we interpret video as a sequence of images, a real-time clock in video format fails SC 1.4.5: Images of text.)
2. A screensaver will either disappear or be overlayed with a log-in screen when you do something with an input device. Since you don't interact with the screensaver as such, and it is supposed to run when you're not using your device, I don't see an issue with a screensaver not having subtitles or an audio description.
3. A pseudorandom, time-sensitive code used for two-factor authentication: would this be provided as a video or as audio-only content?

Are there any other types of computer-generated information that would be provided in audio or video format and that would require subtitles or audio description?

bruce-usab:
Are there any other types of computer-generated information that would be provided in audio or video format and that would require subtitles or audio description?

Yes, anything interactive like games or immersive environments.  Those settings need not be "live" but are scripted and programmed and not "prerecorded".

patrickhlauke:
need to be careful not to start overlapping with things that would fall under 1.1.1 Non-text Content or 4.1.3 Status Messages, for instance.

GreggVan:
Yea

[...]

PR #3476, opened 2023-10-12
Add paragraph about non-transcribed CAPTCHA audio

### LINKED ISSUE #3198: SC 3.3.7 Accessible authentication and CAPTCHA challenge
A typical CAPTCHA requires you to recognize images and has an audio challenge.  The audio challenge is required for users who cannot see or who have some other visual disability.

We say that object recognition includes audio - but the transcription aspect is not addressed. SC 3.3.7 allows for the object recognition to pass - but the audio version would not pass the SC because it requires transcription.  In effect a typical CAPTCHA that had an image with audio challenge could fail unless the audio challenge was set up differently with multiple choice answers for example - but then again that requires memorization.

Do I understand this correctly that typical CAPTCHA audio challenges would fail at the AA level because of the transcription?

fstrr:
Closes #3198

Note: Issue #323, from 2018, does bring up the issue of what generated on the fly audio is (live? prerecorded?) and it wasn’t resolved. It’s not clear to me whether this fits under 1.2.1 (pre-recorded audio only) or 1.2.9 (audio-only live), but I’m not sure it matters as both require an alternative.

### DIFF
--- a/understanding/22/accessible-authentication-minimum.html
+++ b/understanding/22/accessible-authentication-minimum.html
@@ -72,7 +72,7 @@
             
             <p>Remembering a site-specific password is a <a>cognitive function test</a>. Such tests are known to be problematic for many people with cognitive disabilities. Whether it is remembering random strings of characters, or a pattern gesture to perform on a touch screen, cognitive function tests will exclude some people. When a cognitive function test is used, at least one other authentication method must be available which is not a cognitive function test.</p>
 
-            <p>Some <abbr title="Completely Automated Public Turing test to tell Computers and Humans Apart">CAPTCHA</abbr> systems have an audio alternative of the visible text that the user needs to transcribe. Unless the audio has been been transcribed, it can't be used as an alternative method.</p>            
+            <p>Some <abbr title="Completely Automated Public Turing test to tell Computers and Humans Apart">CAPTCHA</abbr> systems have an audio alternative of the visible text. If the user needs to transcribe this audio, it cannot be used to meet the Alternative exception.</p>            
 
             <p>If there is more than one step in the authentication process, such as with multi-factor authentication, all steps need to comply with this Success Criterion to pass. There needs to be a path through authentication that does not rely on cognitive function tests.</p>
 
--- a/understanding/22/accessible-authentication-minimum.html
+++ b/understanding/22/accessible-authentication-minimum.html
@@ -72,6 +72,7 @@
             
             <p>Remembering a site-specific password is a <a>cognitive function test</a>. Such tests are known to be problematic for many people with cognitive disabilities. Whether it is remembering random strings of characters, or a pattern gesture to perform on a touch screen, cognitive function tests will exclude some people. When a cognitive function test is used, at least one other authentication method must be available which is not a cognitive function test.</p>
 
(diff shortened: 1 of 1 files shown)


understanding/22/accessible-authentication-minimum.html
```diff
@@ -72,6 +72,7 @@ <h3>Cognitive Function Tests</h3>
             
             <p>Remembering a site-specific password is a <a>cognitive function test</a>. Such tests are known to be problematic for many people with cognitive disabilities. Whether it is remembering random strings of characters, or a pattern gesture to perform on a touch screen, cognitive function tests will exclude some people. When a cognitive function test is used, at least one other authentication method must be available which is not a cognitive function test.</p>
 
+            <p>Some <abbr title="Completely Automated Public Turing test to tell Computers and Humans Apart">CAPTCHA</abbr> systems have an audio alternative of the visible text that the user needs to transcribe. Unless the audio has been been transcribed, it can't be used as an alternative method.</p>
```
mbgower, line 75: My concern with this phrasing is that I don't see how a provided transcription of the audio could then be any form of test. I think maybe it could be tweaked as follows:
```suggestion
            <p>Some <abbr title="Completely Automated Public Turing test to tell Computers and Humans Apart">CAPTCHA</abbr> systems have an audio alternative of the visible text. If the user needs to transcribe this audio, it can't be used as an accessible alternative method.</p>            
```

understanding/22/accessible-authentication-minimum.html
```diff
@@ -72,6 +72,7 @@ <h3>Cognitive Function Tests</h3>
             
             <p>Remembering a site-specific password is a <a>cognitive function test</a>. Such tests are known to be problematic for many people with cognitive disabilities. Whether it is remembering random strings of characters, or a pattern gesture to perform on a touch screen, cognitive function tests will exclude some people. When a cognitive function test is used, at least one other authentication method must be available which is not a cognitive function test.</p>
 
+            <p>Some <abbr title="Completely Automated Public Turing test to tell Computers and Humans Apart">CAPTCHA</abbr> systems have an audio alternative of the visible text. If the user needs to transcribe this audio, it cannot be used to meet the Alternative exception.</p>
```
detlevhfischer, line 75: How is CAPTCHA part of authentication? I seem to remember there was a discussion on that on Github or the mailing list. Since CAPTCHA is not used to identify a particular person but just differentiate human vs. automatic capabilities, this seems misplaced here?

patrickhlauke: it's a cognitive test *as part of* an authentication process (rather than a cognitive test in general). but agree, as mentioned before, this is the the weirdness of this particular SC (that really should have been two separate SCs) special-casing things under the mantle of authentication only

mbgower: @detlev, if you're asking this question, maybe we need to do more than just modify this portion of the document. However, my suggestion would be that spawning a new issue asking that question may be a better approach than adding this contextual information -- unless it is your feeling that this changes makes your underlying question _more_ dominant.

detlevhfischer: I'm easy (adding [rendered view ](https://raw.githack.com/w3c/wcag/06f7b71a542627149dd4d8c0b18765409a6553ed/understanding/22/accessible-authentication-minimum.html)for easier review)

PR #3534, opened 2023-11-03
Update C43 to add test case for Focus Not Obscured Enhanced

### LINKED ISSUE #3532: C43 is cited as sufficient for both the AA and AAA versions of Focus Not Obscured, but the test is only sufficient for the AA
[C43: Using CSS scroll-padding to un-obscure content](https://www.w3.org/WAI/WCAG22/Techniques/css/C43) is cited in the AA version of Focus Not Obscured and also in the [AAA](https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-minimum) version.

However, the test as written is only sufficient to meet AA
>Procedure
For each user interface component that can receive keyboard focus:
>1. Check that the user interface component is not entirely hidden when it receives keyboard focus.

>Expected Results
Check 1 is true.

This can be resolved by adding modifying #1 and adding a #2 that is checking to make sure that no part of the UIC is obscured, and then making 1 true for AA and 1 and 2 true for AAA

fstrr:
Closes #3532


techniques/css/C43.html
```diff
@@ -167,12 +167,14 @@ <h3>Procedure</h3>
 				<p>For each user interface component that can receive keyboard focus:</p>
 				<ol>
 					<li>Check that the user interface component is not entirely hidden when it receives keyboard focus.</li>
+          <li>Check that no part of the user interface component is hidden when it receives keyboard focus.</li>
```
mbgower, line 170:
```suggestion
          <li>Check that no part of the user interface component is hidden by author-created content when it receives keyboard focus.</li>
```

bruce-usab:

patrickhlauke:
@detlevhfischer the deploy links need to be manually expanded to include the actual file you're trying to look at (the deploy preview has no root/index). so in this case, the deploy preview is https://deploy-preview-3534--wcag2.netlify.app/, the file changed here is `techniques/css/C43.html`, so you want to manually go to https://deploy-preview-3534--wcag2.netlify.app/techniques/css/c43

(with hat tip to @kfranqueiro for pointing that out to me recently)

It would of course be super helpful if the deploy preview had some form of initial index (even if just a throwaway page that points to `/techniques/` and `/understanding/`

bruce-usab (approved):

Issue #3535, opened 2023-11-06
2.5.8 & 2.5.1 "single-pointer" definition should not include path-based gestures


philljenkins:
The definition of "single-pointer" incorrectly includes "path-based gestures", which it should not since "path-based gestures" by definition are dragging.

https://www.w3.org/WAI/WCAG22/Understanding/dragging-movements.html#dfn-single-pointer
> **single-pointer**: pointer input that operates with one point of contact with the screen, including single taps and clicks, double-taps and clicks, long presses, and path-based gestures 

This fails to make sense when replacing the definition for the term in the SC
2.5.7 ... can be achieved by a [_pointer input that operates with one point of contact with the screen, including single taps and clicks, double-taps and clicks, long presses, and path-based gestures_] without dragging,

also fails where it's using a circular definition in the SC 
2.5.1 ... path-based gestures for operation can be operated with a [_pointer input that operates with one point of contact with the screen, including single taps and clicks, double-taps and clicks, long presses, and **path-based gestures**_] without a **_path-based gesture_**, unless ...

Removing the "**_path-based gestures_**" from the definition of "**_single-pointer_**" would solve the problem.
[Understanding 2.5.1 Pointer Gestures](https://www.w3.org/WAI/WCAG21/Understanding/pointer-gestures.html#dfn-single-pointer) 
[Understanding 2.5.7  Dragging Movement](https://www.w3.org/WAI/WCAG22/Understanding/dragging-movements.html#dfn-single-pointer)

philljenkins:
- Apple defines "[Standard Gestures](https://developer.apple.com/design/human-interface-guidelines/gestures#Standard-gestures)" as: **_Tap, Swipe, Drag, Touch (or pinch) and hold, Double tap, Zoom, and Rotate_**
- In addition to the [standard gestures](https://developer.apple.com/design/human-interface-guidelines/gestures#Standard-gestures) supported in **all platforms**, iOS and iPadOS support a few other gestures that people expect: **_Three-finger swipe, Three-finger pinch, Four-finger swipe (iPadOS only)_**
- visionOS defines two types of gestures: indirect and direct. Direct gestures include: **_Touch, Touch and hold, Touch and drag, Double touch, Swipe, With two hands, pinch and drag together or apart, With two hands, pinch and drag in a circular motion_**.

patrickhlauke:
[...]

Issue #394, opened 2018-06-29
single pointer definition includes path-based gestures?


michael-n-cooper:
_From @becka11y on April 18, 2018 20:54_

This relates to #634 which is closed.  I am confused by the how definition of single pointer is used in the Pointer Gestures SC because the single pointer definition includes path-based gestures?  
 The definition in the 17 April 2018 editor's draft is:

pointer input that operates with one point of contact with the screen, including single taps and clicks, double-taps and clicks, long presses, and path-based gestures

But 2.5.1 Pointer Gestures says:

All functionality that uses multipoint or path-based gestures for operation can be operated with a single pointer without a path-based gesture, unless a multipoint or path-based gesture is essential.

How can this be achieved if the definition of single pointer includes path-based gesture? 

I get the intent, the user must be able to operate using a single point of contact but without moving it along a path.  But the wording of the SC implies that a single pointer definition can be split into two parts - single point of contact with no movement and single point of contact with movement along a path.  Isn't that really two separate definitions?

I suspect I am splitting hairs. 

_Copied from original issue: w3c/wcag21#868_

michael-n-cooper:
_From @detlevhfischer on April 19, 2018 7:39_

I think seen from the context of 2.5.1 Pointer Gestures it looks indeed odd to have path gestures included in the definition of „single pointer“. I don‘t know when that crept in, and why. Having said that, the SC text says „ single pointer without a path-based gesture“ so it can bee seen as referring to the single Pointer definition while at the same time excluding the path-based aspect of it. Is there some reason why „path-based“ must remain in the definition? I would prefer to get rid of it....

michael-n-cooper:
_From @detlevhfischer on April 26, 2018 16:27_

Proposed WG response:
Thank you for your comment.
The Success Criterion 2.5.1 Pointer Gestures consciously refers to a subset of the single pointer definition "All functionality (...) can be operated with a single pointer **without a path-based gesture**" because this is what is targeted here. The wider definition including the path-based aspect is needed in case of SC 2.5.1 Pointer Cancellation, for example, to cover drag-and-drop actions.

michael-n-cooper:
_From @greg-lowney on May 1, 2018 4:52_

Personally, I'd prefer that instead of talking around it, our response would be honest and straightforward enough to admit that the definition is badly written and should not include the reference to path-based gestures, even though it may be too late in the process to remove it from the document.

michael-n-cooper:
[...]

PR #809, opened 2019-06-28
Expand/clarify "single pointer" definition

patrickhlauke:
Closes https://github.com/w3c/wcag/issues/749

Issue #749, opened 2019-05-20
"single pointer" definition not quite accurate?


patrickhlauke:
the (new in 2.1) definition of "single pointer" is, to my mind, not quite accurate.

> pointer input that operates with one point of contact with the screen, including single taps and clicks, double-taps and clicks, long presses, and path-based gestures

https://www.w3.org/TR/WCAG21/#dfn-single-pointer

noting that a pointer (at least in the Pointer Events sense) also includes mice, which don't have a "contact with the screen" (only in a very abstracted way)

a possible clarification here could be

> pointer input that operates with one point of contact with the screen / targets only one specific position on the page, including single taps and clicks, double-taps and clicks, long presses, and path-based gestures. This includes mouse inputs and interactions with a single finger or stylus with a touchscreen.

(or something to that effect)

happy to submit a PR if this sounds like something we'd want to clarify.

as pointer gesture and pointer cancellation SCs (which reference this key term) do apply to mouse already, changing this would likely be an editorial errata?

jake-abma:
related to https://github.com/w3c/wcag/issues/746

Please comment on 2 & 3 about if path-based gestures needs to be here.

1. > The "one point of contact" can be read as: "only one point on the screen" not allowing hundreds of points when moving through a path
3. > The "one point of contact" can be read as: "one point at a time, but this (one) point may change hundreds of times due to moving the pointer

patrickhlauke:
i believe the definition is not so much currently defining "single pointer", but "single pointer gesture" here (as it lists all the various interactions that counts as a "gesture", like taps/clicks, path-based movements, etc.

if just defining a "single pointer" per se, it should then probably just focus/emphasise that it stands in contrast to multi-pointer (really, multi-touch in practice, as most [possibly all] devices don't allow simulatenous use of different pointers like two mice, of a mouse and a single finger or stylus, or even multiple styluses).

so taking it further, if just trying to define "single pointer" (rather than single pointer gestures, which wouldn't be appropriate as the term is referenced from 2.5.2 as well), it probably needs a more fundamental rewrite...

> certain devices allow the use of more than one simultaneous pointer, such as multiple fingers on a touchscreen device. Single pointer interactions are exclusively limited to interactions where only one point of contact with the screen / only one specific position on the page is used. This includes mouse inputs and interactions with a single finger or stylus with a touchscreen.

patrickhlauke:
maybe then expand that definition (rather than creating another one) to clarify/list

[...]

PR #3536, opened 2023-11-06
Update definition for single pointer

patrickhlauke:
The definition for "single pointer" has had issues for a long time, as it mixes the idea of what a pointer *is* with the action(s) *performed* using a pointer.

I originally tried to fix this, but there was no appetite for it once 2.1 was released. However, with 2.2 and the new 2.5.7 Dragging Movement SC, the broken definition is causing actual misunderstandings/illogical non-sequiturs.

See https://github.com/w3c/wcag/issues/749#issuecomment-494146357 and the recent https://github.com/w3c/wcag/issues/3535 where this is once again causing a non-sequitur

Closes https://github.com/w3c/wcag/issues/3535

(this is effectively a follow-up to https://github.com/w3c/wcag/pull/809 which had disambiguated things, but the definition had since been changed further/again to reintroduce the ambiguous wording we have at this point which confuses input with action)

This would be applied to WCAG 2.1 and 2.2, unless there is a decision to only apply it to 2.2.

EDIT: Also closes https://github.com/w3c/wcag/issues/394

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/pull/3536.html" title="Last updated on Mar 8, 2024, 7:30 PM UTC (6c36df1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/3536/afbf9ee...6c36df1.html" title="Last updated on Mar 8, 2024, 7:30 PM UTC (6c36df1)">Diff</a>

### DIFF
--- a/guidelines/terms/21/single-pointer.html
+++ b/guidelines/terms/21/single-pointer.html
@@ -1,7 +1,5 @@
 <dt><dfn>single pointer</dfn></dt>
 <dd>
-   	
-   <p>pointer input that operates with one point of contact with the screen, including single taps and clicks, double-taps and clicks, long presses, and path-based gestures</p>
-  				
+   <p>a single mouse, stylus, or finger on a touch screen, that only targets a single point on the screen (as opposed to multi-pointer/multi-touch scenarios)</p>
 </dd>
 
--- a/guidelines/terms/21/single-pointer.html
+++ b/guidelines/terms/21/single-pointer.html
@@ -1,5 +1,5 @@
 <dt><dfn>single pointer</dfn></dt>
 <dd>
-   <p>a single mouse, stylus, or finger on a touch screen, that only targets a single point on the screen (as opposed to multi-pointer/multi-touch scenarios)</p>
+   <p>an input – such as a mouse, stylus, or finger on a touch screen – that only targets a single point on the screen (as opposed to multi-pointer/multi-touch scenarios)</p>
 </dd>
 
--- a/guidelines/terms/21/single-pointer.html
+++ b/guidelines/terms/21/single-pointer.html
@@ -1,5 +1,6 @@
 <dt><dfn>single pointer</dfn></dt>
 <dd>
-   <p>an input that only targets a single point on the page/screen at a time – such as a mouse, stylus, or single finger on a touch screen – as opposed to multi-pointer/multi-touch scenarios</p>
+   <p>an input that only targets a single point on the page/screen at a time – such as a mouse, stylus, or single finger on a touch screen.</p>
+   <p class="note">In contrast to single pointer inputs, multipoint interactions involve the use of two or more pointers at the same time – such as two finger interactions on a touchscreen, or the simultaneous use of a mouse and stylus.</p>
 </dd>
 
--- a/guidelines/terms/21/single-pointer.html
+++ b/guidelines/terms/21/single-pointer.html
@@ -1,6 +1,6 @@
 <dt><dfn>single pointer</dfn></dt>
 <dd>
-   <p>an input that only targets a single point on the page/screen at a time – such as a mouse, stylus, or single finger on a touch screen.</p>
+   <p>an input that only targets a single point on the page/screen at a time – such as a mouse, single finger on a touch screen, or stylus.</p>
    <p class="note">In contrast to single pointer inputs, multipoint interactions involve the use of two or more pointers at the same time – such as two finger interactions on a touchscreen, or the simultaneous use of a mouse and stylus.</p>
 </dd>
 
--- a/guidelines/terms/21/single-pointer.html
+++ b/guidelines/terms/21/single-pointer.html
@@ -1,5 +1,5 @@
 <dt><dfn>single pointer</dfn></dt>
 <dd>
-   <p>an input – such as a mouse, stylus, or finger on a touch screen – that only targets a single point on the screen (as opposed to multi-pointer/multi-touch scenarios)</p>
+   <p>an input that only targets a single point on the page/screen at a time – such as a mouse, stylus, or single finger on a touch screen – as opposed to multi-pointer/multi-touch scenarios</p>
 </dd>
 


benja11y:
> The definition for "single pointer" has had issues for a long time, as it mixes the idea of what a pointer _is_ with the action(s) _performed_ using a pointer.
> 
> See [#749 (comment)](https://github.com/w3c/wcag/issues/749#issuecomment-494146357) and the recent #3535 where this is once again causing a non-sequitur
> 
> Closes #3535
> 
> [Preview](https://pr-preview.s3.amazonaws.com/w3c/wcag/pull/3536.html) | [Diff](https://pr-preview.s3.amazonaws.com/w3c/wcag/3536/afbf9ee...c69c778.html)

Not sure the new definition catches everything, such as joysticks, light pens or 'simulated' versions of pointer (think Dragon's mousegrid feature) for example. Will try to come up with a catch-all.

guidelines/terms/21/single-pointer.html
```diff
@@ -1,7 +1,5 @@
 <dt><dfn>single pointer</dfn></dt>
 <dd>
-   	
-   <p>pointer input that operates with one point of contact with the screen, including single taps and clicks, double-taps and clicks, long presses, and path-based gestures</p>
-  				
+   <p>a single mouse, stylus, or finger on a touch screen, that only targets a single point on the screen (as opposed to multi-pointer/multi-touch scenarios)</p>
```
patrickhlauke, line 3:
```suggestion
   <p>an input – such as a mouse, stylus, or finger on a touch screen – that only targets a single point on the screen (as opposed to multi-pointer/multi-touch scenarios)</p>
```

patrickhlauke:
@benja11y it doesn't need to be comprehensive ... I rejigged it so it doesn't give the impression that it's an exhaustive list

benja11y:
Changed my - to a +

detlevhfischer (approved):
Looks good to me.

guidelines/terms/21/single-pointer.html
```diff
@@ -1,7 +1,5 @@
 <dt><dfn>single pointer</dfn></dt>
 <dd>
-   	
-   <p>pointer input that operates with one point of contact with the screen, including single taps and clicks, double-taps and clicks, long presses, and path-based gestures</p>
-  				
+   <p>an input – such as a mouse, stylus, or finger on a touch screen – that only targets a single point on the screen (as opposed to multi-pointer/multi-touch scenarios)</p>
```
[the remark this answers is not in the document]
patrickhlauke: but this definition does not try, nor need to, define what a path-based gesture is or isn't. this is the definition purely of what a "single pointer" is. not what a gesture is. which was the whole problem initially where the definition completely messed up the subject of the definition and tried to define interactions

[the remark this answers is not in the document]
patrickhlauke: the problem of "what is a path-based gesture" is one that needs to be disambiguated/explained/patched in 2.5.1 Pointer Gestures and (now that we have it) 2.5.7 Dragging Movements.

and to be clear, I absolutely agree that the whole "intermediate point" thing was an ugly, ugly fudge that was used to retrospectively make 2.5.1 apply to things that we only later realised 2.5.1 wasn't actually covering, even though we thought it did. it was crowbarred into the understanding for 2.5.1 because it was too late once it was realised to go back to the drawing board and change/expand the normative SC wording... and now that we *also* have 2.5.7 there's even more confusion (as potentially now certain types of sliders/interactions would fall under just 2.5.7, or both 2.5.1 and 2.5.7, and it often looks arbitrary which one's which)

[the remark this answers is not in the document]
detlevhfischer: Yes, makes sense to separate the interaction. I think the definition is fine. Will try to tackle the interaction bit separately (even though I fear there is little appetite for changing path-based gesture definition now that we also have 2.5.7).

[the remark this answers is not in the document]
patrickhlauke: >  I fear there is little appetite for changing the path-based gesture definition now that we also have 2.5.7

https://en.wikipedia.org/wiki/Appetite_for_Destruction

but seriously, i think the actual existence (now) of 2.5.7 fives us more leverage to perhaps remove the fudges that were made in 2.5.1

bruce-usab:
On deck for 1/23 call.

guidelines/terms/21/single-pointer.html
alastc, line 3:
```suggestion
   <p>an input – such as a mouse, stylus, or finger on a touch screen – that only targets a single point on the screen (in contrast to multi-pointer/multi-touch scenarios)</p>
```

alastc, line 3:
```suggestion
   <p>an input that only targets a single point on the screen. Such as a mouse, stylus, or finger on a touch screen, in contrast to multi-pointer/multi-touch scenarios.</p>
```

mbgower: could add something like 'or an assistive device using the pointer API' or something

mbgower: "vector" input?

mbgower: patrick suggested "on page" instead of "on screen"

mbgower: or specific coordinate

bruce-usab:
From the call 1/12, in no particular order.
- @gundulaniemann preferred current definition to the changed proposed.
- No one disagreed that the current definition is not very good.
- There seemed to be general consensus about adding "&mdash; such as a mouse" as having explanatory utility.  But probably additional examples (as part of the defined term) are too much.
- @dav-idc noted that "on screen" could be taken too literally (as physical).  Eye gaze is an example of single pointer where nothing is touched.
- @dbjorge noted that "one point at a time" could be understood as only covering the start of an interaction (e.g. mouse-down event).

guidelines/terms/21/single-pointer.html
patrickhlauke: as with the discussion around other alternative input methods, like sip and puff, these methods like "AT firing synthetic pointer events" all boil down to still being mouse/touch/stylus, so in the interest of brevity/readability, i'd suggest not starting to list all other ways in which these events may get fired (this is in relation to the "assistive device..." comment).

as for "vector" input (an input with a magnitude/direction), that's not really relevant here (and risks confusing matters further when we then talk about dragging movements etc)

patrickhlauke:
@alastc @mbgower @bruce-usab et al ... updated the definition to incorporate some of the suggestions/feedback from yesterday's WCAG 2.x backlog meeting

guidelines/terms/21/single-pointer.html
patrickhlauke: problem with "coordinate" is that you can't just say "a coordinate"...you need "a set of coordinates", "a pair of coordinates", or just plural "coordinates", which then introduces ambiguity again about only being a single point. and then also precludes things like VR/AR where the coordinates may be in 3D space

bruce-usab (approved):
Very clean edit, thanks!

guidelines/terms/21/single-pointer.html
```diff
@@ -1,7 +1,5 @@
 <dt><dfn>single pointer</dfn></dt>
 <dd>
-   	
-   <p>pointer input that operates with one point of contact with the screen, including single taps and clicks, double-taps and clicks, long presses, and path-based gestures</p>
-  				
+   <p>an input that only targets a single point on the page/screen at a time – such as a mouse, stylus, or single finger on a touch screen – as opposed to multi-pointer/multi-touch scenarios</p>
```
alastc, line 3:
```suggestion
   <p>an input that only targets a single point on the page/screen at a time. For example a mouse, stylus, or single finger on a touch screen rather than multi-pointer/multi-touch input</p>
```

mbgower:
```suggestion
   <p>an input that only targets a single point on the page/screen at a time – such as a mouse, stylus, or single finger on a touch screen</p>
```

dbjorge, line 3: Combining one of @alastc 's suggestions with a suggested note to disambiguate make it clear whether stuff like double-clicks counts, based on @scottaohara 's observation that this is a recurring point of confusion for readers.

```suggestion 
   <p>an input that only targets a single point on the page/screen at a time</p>

   <p class="example">A mouse, stylus, or single finger on a touch screen. A multi-touch interface requiring the use of multiple fingers simultaneously would not be "single pointer".</p>
   
   <p class="note">Functionality might require using a single pointer multiple times in sequence; as long as the input only targets a single point at a time, this still represents using a single pointer. For example, double-clicking, long-pressing, and dragging a slider are common operations performed using a single pointer.</p>
```

bruce-usab:
We had significant discussion 1/19 call, but wording was not finalized.  My few notes:
- SC mention of "point" can be read to mean a single pixel.
- "multi-touch scenarios" is confusing.
- AT presents as a mouse, so that is not a conflating issue.
- Some SC definitions include examples (block) which might be a good option.
- Preference is to keep definition work as copy/paste, but that can be just the first sentence of defined term.
- SC provide some context which might take some of load off the defined term.  For example, [2.5.1](https://www.w3.org/TR/WCAG22/#pointer-gestures) exception includes (undefined) term "multipoint" so it follows that "multipoint" need not be part of definition for single pointer.

guidelines/terms/21/single-pointer.html
detlevhfischer, line 3: If "multi-point scenarios" is undesirable - how about 

> ...as opposed to input targeting more than one point at the same time (such as a split tap)


...where some people may not know what a split tap is... sigh

patrickhlauke:
As a possible alternative, I've moved the mention of multipoint (rather than "multi-pointer/multi-touch") into a note (following precedent set in other definitions that also include a note and/or example).

guidelines/terms/21/single-pointer.html
patrickhlauke: went with using "multipoint" and then explaining it a bit further

bruce-usab (commented):
Looks fine to me, but is normative change, so will want to highlight that for AG review.

scottaohara (approved):

guidelines/terms/21/single-pointer.html
```diff
@@ -1,7 +1,6 @@
 <dt><dfn>single pointer</dfn></dt>
 <dd>
-   	
-   <p>pointer input that operates with one point of contact with the screen, including single taps and clicks, double-taps and clicks, long presses, and path-based gestures</p>
-  				
+   <p>an input that only targets a single point on the page/screen at a time – such as a mouse, stylus, or single finger on a touch screen.</p>
```
alastc, line 3:
```suggestion
   <p>an input that only targets a single point on the page/screen at a time – such as a mouse, single finger on a touch screen, or stylus.</p>
```

patrickhlauke:
@mraccess77

> is this always the case?

what situations are you thinking of where this is *not* the case?

dav-idc:
> Single point on the screen/page could be interpreted by some as single pixel (x,y) point - is this always the case?

A single point on a screen can be made up of several pixels, such as in the case of a retina display. Hence, it's helpful to not say specifically a pixel, but rather a point.

The only case I can reasonably think of where it's not an x,y point (but not necessarily a single pixel) is in a 3D display system where it might be an x,y,z point. But I think this updated definition still covers current 3d technologies, where an x,y,z point is still delivered through a screen.

alastc:
@iadawn this went through CFC, so needs adding as an errata. (That's been a Staff task so far, I'm not sure where it is edited?) We'll keep it on the project board until that's confirmed.

alastc:
@WilcoFiers added on list:

> This new definition seems to exclude path-based gestures from the single-pointer input. That creates a contradiction with 2.5.1. 2.5.1 explicitly says a path-based single-pointer gesture is allowed if it is essential. If single-pointer input is not path based, how can it be allowed? The second problem with it is it changes the scope of 2.5.2. If path-based gestures are not considered a single-pointer input, then using a path-based gesture on a down event with no way to abort or undo the action is now allowed. It makes this SC 2.5.2 more permissive than it used to be.

Could someone respond? (Here or on list.)

patrickhlauke:
> This new definition seems to exclude path-based gestures from the single-pointer input

I don't see how, to be honest. The new definition disentangles the fact that the term defined was the input (as in the mechanism), not the gesture (tap, swipe, etc) itself, because the previous definition was what was causing the non-sequitur in #3535

a path-based *gesture* can still be reliant on a single-pointer *input* (the mechanism of input, i.e. the finger, the mouse, the stylus), and nothing in the new definition contradicts that - unless you interpreted "input" as being the gesture rather than the mechanism (but under that interpretation, we had that contradiction of #3535). and the pointer events spec itself defines input as the mechanism too.

dbjorge:
Discussed this with @WilcoFiers; the confusion seems to be coming from the use of the term "at a time" in the new definition. Wilco suggested that an extra note/examples clarifying which actions a single pointer can perform would have helped clarified; perhaps we can reconsider adding something like this?

```html
<p class="example">Clicking, double-clicking, long-pressing, and dragging a slider are operations that can be performed using a single pointer.</p>
```

bruce-usab:
Discussed 9/13, Patrick offers to tweak, by 9/17 AG call.

patrickhlauke:
@bruce-usab 

So this PR has already been merged - making tweaks to it would need a separate new PR. Before proceeding with that, and seeing @alastc's comment above https://github.com/w3c/wcag/pull/3536#issuecomment-2346425776 about *not* adding in an extra note giving *examples* of gestures that can be performed with a single pointer, I'm wondering if changing just the start of this from

```
an input that only targets a single point on the page/screen at a time – such as a mouse, single finger on a touch screen, or stylus
```

to

```
an input mechanism that only targets a single point on the page/screen at a time – such as a mouse, single finger on a touch screen, or stylus
```

(adding the word "mechanism" after input, to make it clear that this is explicitly what this is referring to, rather than "input" in the sense of the "gesture") would help clarify things.

patrickhlauke:
Made the small PR just adding "mechanism" here https://github.com/w3c/wcag/pull/4070 @bruce-usab @alastc

mbgower:
Reviewed and finalized on Sep 20 TF call. Returned to the Errata for CFC column.

Issue #3264, opened 2023-07-03
3.3.8/3.3.9: How is "authentication process" defined?


dbjorge:
As we're starting to develop step-by-step guidance for testing 3.3.8 and 3.3.9, we're noticing that different folks are a bit unclear about exactly what constitutes an "authentication process" for the purposes of applicability to these requirements, since WCAG 2.2 doesn't formally define it.

I see that there were some past discussions of this in https://github.com/w3c/wcag/issues/1256 and https://github.com/w3c/wcag3/issues/176

Some examples at the boundary:

1. A website allows anonymous comments, but requires users fill out a CAPTCHA to post one. Is filling out a CAPTCHA on its own with no further identifying information an "authentication process"?
    * The past issues I linked above suggest "no", but I think this is contentious; [IBM documents that they think "yes"](https://www.ibm.com/topics/captcha), and without a normative definition in WCAG, it's easy to find supporting examples of both sides.
2. A website's user registration process includes a step that verifies a user's email address by sending them an email with a one-time confirmation link. Does the act of sending the confirmation link mean that the user registration process as a whole is now an "authentication process", where without that verification step it would only be a later login process that would count?
3. In the w3c/wcag#1256 discussion, folks were generally agreeing that just the act of offering an identity (eg, an email address) without any step to confirm the authenticity of that information was probably not enough to constitute "authentication". What is the boundary of how private information must be to constitute "authentication"? For example, which of the following count as "authentication processes"?
    1. An online bank account sign-up form that requests your social security number
    2. An employment application form that requests just the last 4 digits of your social security number
    3. An airline boarding pass website that asks for a flight confirmation number
    4. An airline boarding pass website that asks for a flight number + last name
    5. A marketing website for an R-rated move that asks for your birthday before being allowed to view the website
    6. An online poker game allows game hosts to set password for their table. The host's friends can use the password to join the table (even if they are not logged in as a specific user)

Ideally, there'd be a normative definition that disambiguates this, but since that's probably not feasible to add at this point, it'd be nice for us to at least define the term in the understanding doc.

patrickhlauke:
authentication, for me, has always read as "logging into an existing account" and along the lines of https://en.wikipedia.org/wiki/Authentication#Authentication_factors

my take would be:

1. no, just solving a CAPTCHA does not verify the *identity* of a user. it just verifies (supposedly) if they're human or not. it doesn't verify *who* they are, but *what* they are
[...]

PR #3538, opened 2023-11-07
Amend process definition to remove registration example

patrickhlauke:
As "process" is now referred to from the new 3.3.8/3.3.9 Authentication SCs, and these SCs are scoped to "authentication" only (rather than "registration"), the current definition causes confusion/a non-sequitur. see https://github.com/w3c/wcag/issues/3264#issuecomment-1798253931

Note that this change does not impact the normative interpretation of what a process is or isn't. it simply tweaks one of the non-exhaustive examples that are included, to not cause an immediate non-sequitur with the new authentication SCs which are about authentication, rather than account creation.

https://deploy-preview-3538--wcag2.netlify.app/guidelines/#dfn-processes


bruce-usab:
I agree it would have been better to have authentication as an example of a process (as compared to account registration).  That said, I am not enthusiastic about including this as an erratum (since it is normative).

patrickhlauke:
it causes real confusion for the already confusing "accessible authentication" concepts. and while yes, it's a change to a normative part, it only changes the non-exhaustive examples within the definition

guidelines/terms/20/process.html
```diff
@@ -3,13 +3,15 @@
...
-      shipping information and provide payment information.
-   </p></aside>
+      shipping information and provide payment information.</p>
+   </aside>
    
-   <aside class="example"><p>An account registration page requires successful completion of a <a href="https://www.w3.org/TR/turingtest/">Turing test</a> before
-      the registration form can be accessed.
-   </p></aside>
+   <aside class="example">
+      <p>After filling in a login/authentication form, the user has to successfully completete a
```
patrickhlauke:
```suggestion
      <p>After filling in a login/authentication form, the user has to successfully complete a
```

awkawk:
Agree with Bruce from Nov 2023 - this is a normative change that doesn't feel like it is necessary. As an example, it isn't a problem to refer to one type of situation and doesn't limit applicability to other situations.

patrickhlauke:
> it isn't a problem

it does cause an apparent non-sequitur - see https://github.com/w3c/wcag/issues/3264#issuecomment-1798253931

awkawk:
> it does cause an apparent non-sequitur - see https://github.com/w3c/wcag/issues/3264#issuecomment-1798253931

It is still a process, and it could be fully in-line with 3.3.8/9 - we have no idea what the Turing test is in the example.

patrickhlauke:
> It is still a process, and it could be fully in-line with 3.3.8/9 - we have no idea what the Turing test is in the example.

sure, it's a process, but 3.3.8/3.3.9 apply to authentication processes, which creation of a new account is not (which is the whole point of that discussion, which gets then muddied by account creation being listed as an example of a process)

so making a tiny tweak to the example provided (while not trying to imply that account creation isn't a process) removes a small stumbling block in understanding 3.3.8/3.3.9

awkawk:
> sure, it's a process, but 3.3.8/3.3.9 apply to authentication processes, which creation of a new account is not (which is the whole point of that discussion, which gets then muddied by account creation being listed as an example of a process)

The process definition is to define "process" not "authentication process". the process definition is linked from ~4 SC and the 5.2.3 Complete processes.

detlevhfischer:
It's worth noting that the text says "authentication [process](https://www.w3.org/TR/WCAG22/#dfn-processes)" but only the second (generic) term is linked, so I would not expect the process definition to be tailored to authentification or even necessarily include authentification.

patrickhlauke:
to me the utility of this comes down to avoiding lengthy explanations for people who haven't grappled with the distinction between "authentication" and other uses of cognitive function tests elsewhere.

it avoid the run-around having to explain to people:

> "these two SCs only apply to authentication. that's basically logging into an existing account."
> "i looked at the definition of process, and it gives an example of creating a new account."
> "ah, but that is only a general example of a process. see, for these SCs, it's only scoped to authentication processes, which that one example in the definition of process is not."

and yes, "process" is referenced by other SCs, but changing the non-exhaustive. informative example in the definition does not impact these other SCs in the slightest.

or, you know, we keep it all as is, and just always have to rerun through the conversation above with folks...because why should WCAG be easy :)

bruce-usab (approved):
I agree that this is a good editorial correction that [does not add new features](https://www.w3.org/policies/process/#correction-classes).

Link to [diff preview](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FTR%2F2024%2FREC-WCAG22-20241212%2F&doc2=https%3A%2F%2Fdeploy-preview-3538--wcag2.netlify.app%2Fguidelines%2F#dfn-processes) at term in glossary.

patrickhlauke:
Discussed in last TF meeting. Agreed to move to the potential errata column for the next CfC

mraccess77:
I think a registration or account creation process is still a valid example to include although I agree the word "registration" is likely to narrow.  I'm not clear why both that and authentication cannot be listed.

kfranqueiro:

PR #3560, opened 2023-11-19
Add anti-aliasing note to 1.4.11

patrickhlauke:
Matches the anti-aliasing notes we recently included for 1.4.3 and 1.4.6 (see https://github.com/w3c/wcag/pull/3020)

### DIFF
--- a/understanding/21/non-text-contrast.html
+++ b/understanding/21/non-text-contrast.html
@@ -29,7 +29,7 @@
 			</div>
 
 			<div class="note">
-				<p>Because authors do not have control over user settings for font smoothing/anti-aliasing, when evaluating this
+				<p>Because authors do not have control over user settings for font smoothing and anti-aliasing, when evaluating this
 					 Success Criterion, refer to the colors obtained from the user agent, or the underlying
 					 markup and stylesheets, rather than the non-text elements as presented on screen.</p>
 				<p>Due to anti-aliasing, particularly thin lines and shapes of non-text elements may be rendered by user agents with
--- a/understanding/21/non-text-contrast.html
+++ b/understanding/21/non-text-contrast.html
@@ -28,6 +28,18 @@
 				<p>The 3:1 contrast ratios referenced in this Success Criterion is intended to be treated as threshold values. When comparing the computed contrast ratio to the Success Criterion ratio, the computed values should not be rounded (e.g. 2.999:1 would not meet the 3:1 threshold).</p>
 			</div>
 
+			<div class="note">
+				<p>Because authors do not have control over user settings for font smoothing/anti-aliasing, when evaluating this
+					 Success Criterion, refer to the colors obtained from the user agent, or the underlying
+					 markup and stylesheets, rather than the non-text elements as presented on screen.</p>
+				<p>Due to anti-aliasing, particularly thin lines and shapes of non-text elements may be rendered by user agents with
+					 a much fainter color than the actual color defined in the underlying CSS. This can lead to situations where
+					 non-text elements have a contrast ratio that nominally passes the Success Criterion, but have a much lower contrast
+					 in practice. In these cases, best practice would be for authors to avoid particularly thin lines and shapes,
+					 or to use a combination of colors that exceeds the normative requirements of this Success Criterion.
+				</p>
+		 </div>
+
 			<section id="user-interface-components">
 				<h3>Active User Interface Components</h3>
 


scottaohara (approved):

bruce-usab (commented):
My own preference would be to address this in Understanding.  Moved to for discussion 1/26.

bruce-usab:
As discussed 1/26, note mirrors note in place for text contrast.

WilcoFiers (changes requested):
This seems wrong to me. I regret not spotting that change on 1.4.3 / 1.4.6, because that's the bigger concern to me. If an author chooses a font that's so small and/or thin that the original color is completely lost in aliasing, that's a problem they created. They shouldn't get a pass for doing that. It's also incorrect to state authors have no control over aliasing. While not standard, `font-smooth` is supported in just about every browser. And for canvas there is `image-rendering`. 

That decision also contradicts the AGWG approved ACT Rule [Text has minimum contrast](https://www.w3.org/WAI/standards-guidelines/act/rules/afw4f7/). It measures text contrast based on the real color values in pixels, not based on properties set with CSS.

On this particular SC its worth noting that CSS doesn't allow borders/outlines thinner than 1px. A 0.2px border will be drawn as a 1px border. So you have to go well out of your way to get a border that's thinner than that. It can be done with something like a box-shadow though, at which point I don't think its reasonable to say that's not something the author has control over.

JAWS-test:
> That decision also contradicts the AGWG approved ACT Rule [Text has minimum contrast](https://www.w3.org/WAI/standards-guidelines/act/rules/afw4f7/). It measures text contrast based on the real color values in pixels, not based on properties set with CSS.

I don't think that's true. All examples in the ACT rule always refer to the color values in the CSS and never to the displayed color values. Only the [definition of foreground color](https://www.w3.org/WAI/standards-guidelines/act/rules/afw4f7/#foreground-colors-of-text) states that this also includes the anti alias pixels, but it does not say that these must be taken into account in the calculation

> The colors of all the pixels of a [visible](https://www.w3.org/WAI/standards-guidelines/act/rules/afw4f7/#visible) character in a [text node](https://dom.spec.whatwg.org/#text) that change color when the CSS color property is changed. This includes anti-aliased pixels.

If the ACT rule definition means that anti-alias plays a role in the calculation, then the ACT rule must be adapted (regardless of the WCAG) because it is inherently contradictory, as all examples always refer to the CSS color values

mbgower:
This suggested text may not actually contradict the existing anti-aliasing note in the definition of contrast, but it at least has potential to cause a bit of confusion.
The contrast ratio definition appears at the bottom of the 'built' Understanding document and reads:
> Note
Because authors do not have control over user settings as to how text is rendered (for example font smoothing or anti-aliasing), the contrast ratio for text can be evaluated with anti-aliasing turned off.

I suspect the majority of testers do not turn off anti-aliasing when checking for colour contrast, in text or otherwise.
I do know that when checking for text colour contrast, it is very common practice to ensure one picks the darkest pixel in a series of letters to ensure one is actually getting the valid colour. In other words, there is an almost intuitive understanding that one doesn't fail text contrast because one has the ability to pick an edge colour that is clearly not at 4.5:1. So, ultimately anti-aliasing IS being disregarded in spot checks for text, to a great degree. Likewise, when one is checking non-text colour, one looks for the 'solid colour in a non-text objective of any mass, not the border pixels. 

So, ultimately I think this spot-check method should lead to the same outcome as turning anti-aliasing off, in the vast majority of situations. I just wouldn't mind this new text doing a better job of building off the existing definition.

For an automated checker, an inspection of the values set in the CSS seems like a fully defendable way of establishing the colour of an object against its background.

WilcoFiers:
@patrickhlauke The ACT rule says to treat aliasing as part of the text, and for each letter to take the highest contrast between the text and the background. If a letter only has aliasing (which rarely happens) you take the color that contrasts most strongly with the background as the color of the text, not the CSS color. That was a deliberate decision, and the idea of using highest contrast was controversial in AGWG. We spent a lot of time on that.

First up, to be clear about where the problem is. On the vast majority of pages there are no texts so thin/small that they will end up as all aliased pixels. The only time this happens is with very small texts, or very thin fonts. So most of the time using CSS values is fine (assuming you avoid the million pitfalls there are with using those).

Only once text goes 13px or lower do we get to a point where the whole letter may end up getting aliased. It varies a bit by font, aliasing settings, OS, whatever, but that's the number I get when testing. That's problematically small text. The other case is for fonts that are designed to be unusually thin. Browser fonts at 16px doesn't go so thin that they become fully aliased, even on font-weight:100 that doesn't happen. This can only happen with fonts that are designed to be unusually thin. Those kinds of fonts aren't designed to be used at 16px, they have to be much larger to compensate for the thin letters.

If a font is so thin users struggle to read it, its precisely because the real color contrast falls below the 4.5:1 requirement, even if the css values do not. **Authors shouldn't be let off the hook for using thin fonts.**

There are other problems with using CSS values too. Most obviously that not everything is CSS. This approach doesn't work on PDF or canvas, or a variety of other places where you can't open an inspector. If testers need to use the color value before aliasing prevents testing in systems where you can't get those values. Surely that's not the intent here. Even if you have access to the `color` value before aliasing, that doesn't tell you the real color. There is opacity and alpha values to consider, there is blend modes, CSS filters, color fonts, font varients. I don't believe there is a single tool on the market that can accurately compute the color values of text (plural, because text isn't necessarily one color).

CSS colors is a nice and neat short hand. It'll work fine 95% of the time, but we shouldn't be stating that authors can ignore those last 5% when that last 5% only happens when the author chooses to do things that they can reasonably expect can result in contrast issues.

patrickhlauke:
> but that's the number I get when testing

and that is the problem in a nutshell, with emphasis on the "I". there are too many factors outside of the author's control, and the tester's control, and variance (as said...just compare somebody testing on a high-dpi screen compared to somebody testing on a regular dpi screen ... they will get different results purely due to the mechanical difference of one screen having more physical pixels that the OS can then take advantage of).

patrickhlauke:
> There are other problems with using CSS values too

for that, I'd propose an amendment then to both the notes in 1.4.3, 1.4.6, and now this note here for 1.4.11, expanding "CSS" to something like "CSS / underlying technology"

dbjorge:
For discussion purposes, here's a version of the note (for 1.4.3, to be adapted for the others) that Wilco and I wordsmithed some, but I'm not sure it's consistent enough with the contrast ratio note as-is...

```html
      <div class="note">
         <p>User agents which render text will often apply anti-aliasing, font smoothing, and other techniques
            which can cause some pixels of text to render in fainter colors than those specified by authors.
            When evaluating this Success Criterion, this is usually not relevant; as long as the applied font,
            size, and weight ensure that some pixel of each glyph reaches the author-specified color, evaluators
            may refer to the foreground and background colors obtained from the user agent, or the underlying
            markup and stylesheets, rather than evaluating individual pixels within the text as presented on screen.
         </p>
         <p>However, particularly thin, small, or unusual fonts may be rendered by user agents with a much fainter
            color than the actual text color defined in the underlying CSS. This can lead to situations where a
            user agent reports that text has a contrast ratio that nominally passes the Success Criterion, but
            the text as-rendered has a much lower contrast in practice and fails the Success Criterion.
         </p>
         <p>In these cases, best practice would be for authors to choose a font with stronger/thicker lines,
            or to aim for a foreground/background color combination that exceeds the normative requirements
            of this Success Criterion.
         </p>
      </div>
```

dbjorge:
I'll follow up with a PR proposing that note text in the 3 understanding docs in question (1.4.3, 1.4.6, 1.4.11)

dbjorge:
Followup PR with stricter proposal at #3671

patrickhlauke:
> For discussion purposes, here's a version of the note (for 1.4.3, to be adapted for the others) that Wilco and I wordsmithed some, but I'm not sure it's consistent enough with the contrast ratio note as-is...
> 
> ```
>       <div class="note">
>          <p>User agents which render text will often apply anti-aliasing, font smoothing, and other techniques
>             which can cause some pixels of text to render in fainter colors than those specified by authors.
>             When evaluating this Success Criterion, this is usually not relevant; as long as the applied font,
>             size, and weight ensure that some pixel of each glyph reaches the author-specified color, evaluators
>             may refer to the foreground and background colors obtained from the user agent, or the underlying
>             markup and stylesheets, rather than evaluating individual pixels within the text as presented on screen.
>          </p>
>          <p>However, particularly thin, small, or unusual fonts may be rendered by user agents with a much fainter
>             color than the actual text color defined in the underlying CSS. This can lead to situations where a
>             user agent reports that text has a contrast ratio that nominally passes the Success Criterion, but
>             the text as-rendered has a much lower contrast in practice and fails the Success Criterion.
>          </p>
>          <p>In these cases, best practice would be for authors to choose a font with stronger/thicker lines,
>             or to aim for a foreground/background color combination that exceeds the normative requirements
>             of this Success Criterion.
>          </p>
>       </div>
> ```

For what it's worth, I prefer the above (possibly with a few minor tweaks) to the much stricter #3671 proposal

understanding/21/non-text-contrast.html
```diff
@@ -28,6 +28,18 @@ <h2>Intent</h2>
 				<p>The 3:1 contrast ratios referenced in this Success Criterion is intended to be treated as threshold values. When comparing the computed contrast ratio to the Success Criterion ratio, the computed values should not be rounded (e.g. 2.999:1 would not meet the 3:1 threshold).</p>
 			</div>
 
+			<div class="note">
+				<p>Because authors do not have control over user settings for font smoothing/anti-aliasing, when evaluating this
```
dav-idc, line 32: Just clarifying the meaning of the `/` slash, for easier reading. 
Is it meant to represent 'and', 'nor', or 'or'? In this case I think it doesn't matter too much, but 'and' is the most accurate, as the sentence implies an inclusive list of things that authors cannot control.

```suggestion
				<p>Because authors do not have control over user settings for font smoothing and anti-aliasing, when evaluating this
```

HaTheo (approved):

understanding/21/non-text-contrast.html
patrickhlauke: it's more that they're different terms for mostly the same thing ... some people with talk about font smoothing, others about anti-aliasing, but in the end it's all the same concept

dbjorge:
@patrickhlauke [wrote](https://github.com/w3c/wcag/pull/3560#issuecomment-1927728657):

> For what it's worth, I prefer the above (possibly with a few minor tweaks) to the much stricter https://github.com/w3c/wcag/pull/3671 proposal

Can you clarify your stance here? The bit you quoted as preferring is *already* basically-identical to what's in #3671 (+ minor tweaks). I just double checked with a diff between the quoted text and the text #3671 uses for contrast (minimum) and the only differences are:

- The quoted text uses the term "glyph"; in #3671 I updated it to use "character" instead
- The quoted text uses the term "underlying CSS"; in #3671 I updated it to use "underlying markup and stylesheets" instead
- The quoted text uses "stronger/thicker"; in #3671 this is updated to "stronger or thicker"

understanding/21/non-text-contrast.html
dav-idc: Ahh I see @patrickhlauke makes sense. I think in that case either 'and' and 'or' still work just fine. The one that won't work in that case is 'nor'.

They're 2 ways to refer to 1 concept that can't be controlled by an author, but generally they are interchangeable.

patrickhlauke:
@dbjorge oh, apologies for the confusion (on my part), i somehow missed the `authors *may* refer` bit in this version, but it somehow jumped out at me in #3671 ... and that's one of the bits in that PR that rubbed me particularly the wrong way as it yet again handwaved things. not sure why I missed it here, so same comments as on https://github.com/w3c/wcag/pull/3671/files#r1478025213 apply here too

understanding/21/non-text-contrast.html
patrickhlauke: to be difficult, it could even be "and/or" as certain OSs/browsers/users may refer to these interchangeably, or actually see them as two different ways to achieve the same thing... which is why i opted for the slash. happy to change it to either "and" or "or" though, if that makes it easier...

dbjorge:
Having now given more consideration to the feedback in #3671, I think I've been convinced by @patrickhlauke 
 and @JAWS-test that in the common case where a color of a given bit of text/icon/UI component are markup-specified as a single color, we should specify that testers should use the markup color and *should not* use pixel-picking to evaluate pass/fail.

@WilcoFiers 's concerns that this will miss some cases where particularly small/thin fonts really do have accessibility issues resonates with me; I agree that this decision will create some edge cases with false negatives. But I think this is outweighed by the need for repeatable tool-based testing and the false-positives that pixel-picking creates with subpixel rendering.

I think the currently-merged 1.4.3 and 1.4.6 text already cover this adequately for those SCs. I'm going to go thumbs-down my own PR with the stricter proposal for them :)

However (for a separate PR, not this one), I note that we do regularly run into cases in practice where it's not possible to extract a markup-specified color (eg, text rendered over an image background), and I think for those cases specifically, it'd be nice for us to decide on a suggested testing methodology and note it in the understanding docs (and incorporate it more formally/normatively in WCAG3). That's going to need to allow for some sort of pixel-wise color determination for those colors that can't be markup-determined.

WilcoFiers:
My main concern with this note is specifically with stating that it is best practice to ensure that thin lines end up rendered with sufficient contrast. If the author chooses to use lines / fonts that end up less than 1 pixel thick, they are responsible for ensuring those end up with sufficient contrast. For anything 1px and over I think it is perfectly reasonable to use the CSS value.

@JAWS-test results were interesting. I had done similar tests myself a while back and I wasn't getting whole letters missing the defined color at 16px. That gave some insight into the issue. That hasn't changed my view of it much though. I don't think WCAG requires literally every glyph in a text to be surrounded by background pixels (to some undefined distance) that have a 4.5:1 contrast to the text. There are a million edge cases, many of them I suspect most people would never fail. WCAG doesn't say every glyph in a text needs to have a 4.5:1 contrast. It just says the "text", I think that has a degree of reasonableness written into it. 

I would consider aliasing causing a single letter to drop below 4.5:1 an allowable edge case, just like if there was a single background pixel that caused that. It's not a big deal, it doesn't notably hinder the legibility of the text. It's quite another thing though to say that text can pass even if no pixels meet the contrast. Roboto thin at 16px for example looks like this. It has a whopping 3 black pixels in it. That doesn't seem like a black text to me:

The quick brown fox jumps of the lazy dog, written with Roboto thin 16px

Lastly, I want to counter this idea that CSS values are easier for tools. I have spent a LOT of time on axe-core's color contrast testing over the last 8 years. It works reasonably well, but there are a lot of places in which it isn't reasonably possible to compute color values from the CSS. We've started augmenting that in our Chrome extension with screenshot testing, which has worked quite well. That's not always definitive either, but it moved things forward significantly. But if real color values aren't deemed "correct" by AGWG, what is? Because its not just aliasing that depends on the system. Lots of things are. Scaling, blurs, shadows, etc. all vary slightly between systems. It's so bad that people can be fingerprinted by how such things are rendered to a canvas element.

patrickhlauke:
@wilco taking your example of the 16px Roboto ... as said, how would you determine normatively if it's pass or fail if the results of checking it "as rendered" differ based on aspects like "is the author/tester using a low-dpi or high-dpi device"? For argument's sake, I've taken a screenshot of https://fonts.google.com/specimen/Roboto on my Windows machine (running a regular / low-dpi display) and then on my Macbook (high-dpi / retina display). which environment is the "correct" one to make a normative pass/fail assessment?



combined screenshot of the google fonts roboto preview for 16px, showing both the low-dpi windows rendering, and the high-dpi macbook rendering - in the latter, which is 4 times the size due to the high-dpi increased number of pixels, the text is crisp even at thin 100, where on the windows version it's badly antialiased

patrickhlauke:
@dbjorge additionally, either as part of this, or as a separate follow-up PR, i'd love to expand the notes (in 1.4.3, 1.4.6, 1.4.11 to - in a similar way to your other PR, but in a more "best practice" way - acknowledge and mention that there will be situations where it's simply not feasible/possible to get to the "underlying values defined in CSS" (as it may not directly be CSS at all, and also to cover cases like images with burnt-in text) and that in *those* cases ... it may indeed have to be done by color-picking/checking the actual pixel values, which may at least mitigate some further edge cases even though it'll still end up handwaving it for *those* cases.

gundulaniemann:
In the text proposed for decision, anti-aliasing was to be disregarded when measuring color contrasts, because the antialiasing at the user's, together with monitor size, monitor resolution etc, cannot be controlled by the author.
Now, the exact opposite is being proposed, after collecting votings.
On one hand, I suggest to delete all votings and to revite.
On the other, it is not feasible for the page author the ensure that for each and every end user there are pixels which comply to the contrast. resolution, color profile, color management, monitor capabilities and other factors related to color display ae unforeseeable and not controllable by the author. Furthermore, users can switch of Anti-Aliasing.
Therefore I do not agree to this recent change.

alastc:
TF discussion:

- Two testers / authors on different setups (e.g. high / low res displays) can get different results.
- Pixel picking can give very odd results.
- Pixel picking is the only way to measure contrast against a variable background or with translucent text.
- Small / thin fonts are still problematic, they may not output any of the color specified. However, there is no way for us to define that, and users have some control over that with text-resize. 

Overall, the programmatic values need to be the primary way of determining pass/fail, but acknowledging that pixel-picking is needed in some situations such as variable background or with translucent text.

Next step: Take the original PR as written, and follow up with separate PR on when to use Pixel picking guidance.

alastc:
The people who thumbed-down have said it is ok to merge, so we'll merge at the end of the review period (13th).

dbjorge:
> The people who thumbed-down have said it is ok to merge, so we'll merge at the end of the review period (13th).

Yes, confirming that @WilcoFiers and I discussed and agreed on the strategy of merging this now and separately handling improvements to specifying when pixel-picking is acceptable/required in another PR.

WilcoFiers:
👍  from me. Addressing the color picker question should be its own PR.

Issue #3524, opened 2023-10-27
Should there be a section in 1.3.1 on the differences between layout tables and data tables


ljoakley:
Although most tables created currently are data tables there are hundreds of layout tables that exist. The axe accessibility library checks and flags when either data tables and layout tables are built incorrectly. Axe checks that a layout table does not use headers, captions or summary.

mbgower:
@ljoakley if you want to take a stab at drafting something, we can look at it.

I'd argue that the use of tables for layout has trended down significantly over the 15 years since WCAG 2.0 was released, and if this has never been considered necessary until now, it is probably not a high priority. So I would definitely advise not spending too much time on it.

bruce-usab:
@ljoakley I think [F46](https://www.w3.org/WAI/WCAG22/Techniques/failures/F46.html) is close to what you are looking for, but maybe tweak it a bit?
> Failure of Success Criterion 1.3.1 due to using TH elements, caption elements, or non-empty summary attributes in layout tables

fstrr:
F46 might need a bit of a tweak as the `summary` attribute isn't part of the HTML living standard. Related to that, it looks as though I'd forgotten to go through those documents and fix links, remove references to XHTML, etc.

mbgower:
@ljoakley do you want to take a stab at drafting this?

ljoakley:
Bruce has added a comment of where the differences between the types of tables are detailed.

bruce-usab:
Closing since
> mbgower moved this from Fully reviewed to Done in WCAG 2.x 2 weeks ago

@ljoakley please draft a PR if you would like to reopen this issue separately from #3566 (and I see that you actively commenting on that one).

PR #3566, opened 2023-11-22
Update F46 to cover tables with structural elements that are marked presentational

### LINKED ISSUE #3565: Failure F46 for markup in layout tables doesn't taken into account role presentation or none
[F46](https://www.w3.org/WAI/WCAG21/Techniques/failures/F46), "Failure of Success Criterion 1.3.1 due to using th elements, caption elements, or non-empty summary attributes in layout tables" would seem to fail tables with markup even if role of presentation or none is present.  Should this technique be updated to cover this situation?

mraccess77:
Resolves #3565


detlevhfischer:
Looks good to me.

techniques/failures/F46.html
```diff
@@ -1,12 +1,12 @@
...
                     caption elements, or non-empty summary attributes in
-                    layout tables</h1><section class="meta"><p class="id">ID: F46</p><p class="technology">Technology: failures</p><p class="type">Type: Failure</p></section><section id="applicability"><h2>When to Use</h2>
+                    layout tables that are not marked presentational</h1><section class="meta"><p class="id">ID: F46</p><p class="technology">Technology: failures</p><p class="type">Type: Failure</p></section><section id="applicability"><h2>When to Use</h2>
       <p>HTML and XHTML</p>
    </section><section id="description"><h2>Description</h2>
       <p>The objective of this technique is to describe a failure that occurs when a
-                        table used only for layout includes either th elements, a
+                        table used only for layout and is not marked presentational is used only for layout includes either th elements, a
```
mbgower, line 9:
```suggestion
                        table that is used only for layout is both 1) not marked as presentational and 2) includes either th elements, a
```

techniques/failures/F46.html
```diff
@@ -1,12 +1,12 @@
...
                     caption elements, or non-empty summary attributes in
-                    layout tables</h1><section class="meta"><p class="id">ID: F46</p><p class="technology">Technology: failures</p><p class="type">Type: Failure</p></section><section id="applicability"><h2>When to Use</h2>
+                    layout tables that are not marked presentational</h1><section class="meta"><p class="id">ID: F46</p><p class="technology">Technology: failures</p><p class="type">Type: Failure</p></section><section id="applicability"><h2>When to Use</h2>
       <p>HTML and XHTML</p>
    </section><section id="description"><h2>Description</h2>
       <p>The objective of this technique is to describe a failure that occurs when a
-                        table used only for layout includes either th elements, a
+                        table that is used only for layout is both 1) not marked as presentational and 2) includes either th elements, a
                             summary attribute, or a caption element. This
                         is a failure because it uses structural (or semantic) markup only for
```
mbgower, line 11:
```suggestion
                        is a failure because it exposes structural (or semantic) markup only for
```

techniques/failures/F46.html
```diff
@@ -1,14 +1,14 @@
...
-                    layout tables</h1><section class="meta"><p class="id">ID: F46</p><p class="technology">Technology: failures</p><p class="type">Type: Failure</p></section><section id="applicability"><h2>When to Use</h2>
+                    layout tables that are not marked presentational</h1><section class="meta"><p class="id">ID: F46</p><p class="technology">Technology: failures</p><p class="type">Type: Failure</p></section><section id="applicability"><h2>When to Use</h2>
       <p>HTML and XHTML</p>
    </section><section id="description"><h2>Description</h2>
       <p>The objective of this technique is to describe a failure that occurs when a
-                        table used only for layout includes either th elements, a
+                        table that is used only for layout is both 1) not marked as presentational and 2) includes either th elements, a
                             summary attribute, or a caption element. This
-                        is a failure because it uses structural (or semantic) markup only for
+                        is a failure because it exposes structural (or semantic) markup only for
```
mbgower, line 11:
```suggestion
                        is a failure because it exposes structural (or semantic) markup in content only intended for
```

techniques/failures/F46.html
```diff
@@ -1,14 +1,14 @@
...
-                    layout tables</h1><section class="meta"><p class="id">ID: F46</p><p class="technology">Technology: failures</p><p class="type">Type: Failure</p></section><section id="applicability"><h2>When to Use</h2>
+                    layout tables that are not marked presentational</h1><section class="meta"><p class="id">ID: F46</p><p class="technology">Technology: failures</p><p class="type">Type: Failure</p></section><section id="applicability"><h2>When to Use</h2>
       <p>HTML and XHTML</p>
    </section><section id="description"><h2>Description</h2>
       <p>The objective of this technique is to describe a failure that occurs when a
-                        table used only for layout includes either th elements, a
+                        table that is used only for layout is both 1) not marked as presentational and 2) includes either th elements, a
                             summary attribute, or a caption element. This
-                        is a failure because it uses structural (or semantic) markup only for
+                        is a failure because it exposes structural (or semantic) markup in content only intended for
```
mbgower, line 11:
```suggestion
                        is a failure because it exposes structural (or semantic) table markup in content only intended for
```

techniques/failures/F46.html
```diff
@@ -1,14 +1,14 @@
...
       <p>HTML and XHTML</p>
    </section><section id="description"><h2>Description</h2>
       <p>The objective of this technique is to describe a failure that occurs when a
-                        table used only for layout includes either th elements, a
+                        table that is used only for layout is both 1) not marked as presentational and 2) includes either th elements, a
                             summary attribute, or a caption element. This
-                        is a failure because it uses structural (or semantic) markup only for
+                        is a failure because it exposes structural (or semantic) table markup in content only intended for
                         presentation. The intent of the HTML and XHTML table elements is to present data. </p>
       <p>Although not commonly used in a layout table, the following structural markup would also be failures of Success Criterion 1.3.1 if used in a layout table: </p>
```
mbgower, line 13:
```suggestion
      <p>Although not commonly used in a layout table, the following structural markup would also be failures of Success Criterion 1.3.1 if exposed in a layout table: </p>
```

techniques/failures/F46.html
```diff
@@ -84,8 +85,9 @@
       </section>
       <section class="results"><h3>Expected Results</h3>
          <ul>
-            <li>If any check above is false, then this failure condition applies
-                                    and the content fails the Success Criterion.</li>
+            <LI>If check one above is true, then this failure condition does not apply</LI>
```
mbgower, line 88:
```suggestion
            <LI>If check #1 above is true, then this failure condition does not apply</LI>
```

techniques/failures/F46.html
```diff
@@ -84,8 +85,9 @@
...
+            <li>If check one above is false, and any subsequent check above is false, then this failure condition applies
```
mbgower, line 89:
```suggestion
            <li>If #1 is false, and any subsequent step is false, then this failure condition applies
```

techniques/failures/F46.html
```diff
@@ -72,6 +72,7 @@
                <p>If the table is used only to visually lay out elements within the
                                     content </p>
                <ol>
+                  <li>Check that the table is marked presentational by methods such as role none or presentation on the table element</li>
```
mbgower, line 75:
```suggestion
                  <li>Check that the table is marked presentational by methods such as role none or presentation on the table element.</li>
```

techniques/failures/F46.html
```diff
@@ -72,6 +72,7 @@
                <p>If the table is used only to visually lay out elements within the
                                     content </p>
```
mbgower, line 73:
```suggestion
                                    content:</p>
```

techniques/failures/F46.html
```diff
@@ -70,8 +70,9 @@
                                     table element</li>
             <li>
                <p>If the table is used only to visually lay out elements within the
-                                    content </p>
+                                    content:</p>
                <ol>
+                  <li>Check that the table is marked presentational by methods such as role none or presentation on the table element.</li>
```
mgifford, line 75: I'm just wondering if we should be clear that we're talking about role="none" and role="presentation".

It seems that the non-technical approach is both more confusing and more lengthy. Is there a reason not to be specific?

techniques/failures/F46.html
```diff
@@ -84,8 +85,9 @@
       </section>
       <section class="results"><h3>Expected Results</h3>
          <ul>
-            <li>If any check above is false, then this failure condition applies
-                                    and the content fails the Success Criterion.</li>
+            <LI>If #1 is true, then this failure condition does not apply.</LI>
```
mgifford, line 88: I like lower case \<li> elements, but that's pretty trivial.  
  
I think there must be an easiier way to phrase this. Whether #1 is true or false, this fails the SC. I think this statement is what it is trying to say.

mbgower: I find the resulting 'double negative' a bit harder to parse in your suggestion.

The phrasing used in the test is the normal way such things are expressed in the failure techniques (although there is variation in the 100+ [failure techniques](https://www.w3.org/WAI/WCAG22/Techniques/#failures)).

techniques/failures/F46.html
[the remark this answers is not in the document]
patrickhlauke: redoing this @mbgower as you can't use markdown here

patrickhlauke, line 75:
```suggestion
                  <li>Check that the table is marked presentational by methods such as <code>role="none"</code> or <code>role="presentation"</code> on the table element.</li>
```

patrickhlauke:
@mbgower made the discussed change of adding `<code>` around `role="none"` and `role="presentation"` (using markup rather than markdown)

scottaohara (changes requested):
looking at this from the wcag 2.x taskforce meeting, i raised a concern that while the procedure that was added to determine if a table is a layout table or not - there are many other factors that could result in a table's exposed role being changed, or not.

it might be a whole lot simpler to instead write a procedure that instructs people to check the exposed role for the table using browser dev tools - and then mention that there are many factors (including browser heuristics) that could result in a table being exposed as such, or not).

techniques/failures/F46.html
mgifford: There's a lot of content in WCAG to maintain...  And ya, double negatives are annoying.   
  
Could it be simplified to:   
  "If #1 is true, then this does not apply."

techniques/failures/F46.html
```diff
@@ -70,8 +70,9 @@
                                     table element</li>
             <li>
                <p>If the table is used only to visually lay out elements within the
```
awkawk, line 72: We've done the setup line in a different way usually - e.g., "For each table used only to visually lay out elements within the page:"

techniques/failures/F46.html
```diff
@@ -84,8 +85,9 @@
       </section>
       <section class="results"><h3>Expected Results</h3>
          <ul>
-            <li>If any check above is false, then this failure condition applies
-                                    and the content fails the Success Criterion.</li>
+            <LI>If #1 is true, then this failure condition does not apply.</LI>
+            <li>If #1 is false, and any subsequent step is false, then this failure condition applies
```
awkawk, line 89: If there are headings in a table that is marked presentational, what do the browsers expose to AT about that? I assume they don't convey table semantics in that case.

mraccess77: I would assume that was well - but I'm not clear what you are saying needs to be updated.

awkawk: I guess I'm finding the construction confusing. Perhaps we just have one expected result which would be:

If 1 and either 2 or 3 are false, then the failure condition applies.

mbgower:
@mraccess77 are you able to take this feedback and produce something? @scottaohara is happy to assist.

mraccess77:
Do we have any example techniques were we have looked at the accessibility tree in the browser? This seems like a new approach from what our technique typically look like and I want to make sure we are consistent.  Can we assume that is the new way to go for all techniques?  For example, should all of our image with alternative text techniques be rewritten to cover the accessible name?

patrickhlauke:
i believe @scottaohara's idea - if i follow it correctly - would be to generalise the wording to something more like:

```
Failure of Success Criterion 1.3.1 due to using th elements, caption elements, or non-empty summary attributes in layout tables that are still exposed as data tables
```

and then adding an explanation that is more generic, such as "unless the semantics of the table have been removed - for instance, using `role="none"` or `role="presentation"` on the `<table>` element - they will still be exposed as data tables to screen readers" or similar...

scottaohara:
linking related issues (layout tables) https://github.com/w3c/wcag/issues/3524

techniques/failures/F46.html
ljoakley: I think that changing the text to reflect how the actual tag appears is the correct thing to do here.

techniques/failures/F46.html
ljoakley: > If there are headings in a table that is marked presentational, what do the browsers expose to AT about that? I assume they don't convey table semantics in that case.

Yes, a screen reader will skip over a header labeled role-presentational. However, if a table has headers, that makes it a data table and now the rules for data tables apply.

ljoakley:
I find it interesting in reading through the definition of tables in both MDN and W3schools they both are using data tables and make no distinct between layout tables and data tables.

scottaohara:
@ljoakley specific things you can link to?

i ask because there are layout tables in the scope of accessibility, and then there are also layout-table CSS properties

bruce-usab:
Discussed 9/13, moved to To Do.

techniques/failures/F46.html
```diff
@@ -1,16 +1,16 @@
...
                     caption elements, or non-empty summary attributes in
-                    layout tables</title><link rel="stylesheet" type="text/css" href="../../css/editors.css" class="remove"></link></head><body><h1>Failure of Success Criterion 1.3.1 due to using th elements,
+                    layout tables that are not marked presentational</title><link rel="stylesheet" type="text/css" href="../../css/editors.css" class="remove"></link></head><body><h1>Failure of Success Criterion 1.3.1 due to using th elements,
                     caption elements, or non-empty summary attributes in
-                    layout tables</h1><section class="meta"><p class="id">ID: F46</p><p class="technology">Technology: failures</p><p class="type">Type: Failure</p></section><section id="applicability"><h2>When to Use</h2>
+                    layout tables that are not marked presentational</h1><section class="meta"><p class="id">ID: F46</p><p class="technology">Technology: failures</p><p class="type">Type: Failure</p></section><section id="applicability"><h2>When to Use</h2>
       <p>HTML and XHTML</p>
```
fstrr, line 6:
```suggestion
      <p>HTML</p>
```

techniques/failures/F46.html
```diff
@@ -1,16 +1,16 @@
...
    </section><section id="description"><h2>Description</h2>
       <p>The objective of this technique is to describe a failure that occurs when a
-                        table used only for layout includes either th elements, a
+                        table that is used only for layout is both 1) not marked as presentational and 2) includes either th elements, a
                             summary attribute, or a caption element. This
-                        is a failure because it uses structural (or semantic) markup only for
+                        is a failure because it exposes structural (or semantic) table markup in content only intended for
                         presentation. The intent of the HTML and XHTML table elements is to present data. </p>
```
fstrr, line 12:
```suggestion
                        presentation. The intent of the HTML table element is to present tabular data.</p>
```

techniques/failures/F46.html
```diff
@@ -1,16 +1,16 @@
...
                     caption elements, or non-empty summary attributes in
-                    layout tables</h1><section class="meta"><p class="id">ID: F46</p><p class="technology">Technology: failures</p><p class="type">Type: Failure</p></section><section id="applicability"><h2>When to Use</h2>
-      <p>HTML and XHTML</p>
+                    layout tables that are not marked presentational</h1><section class="meta"><p class="id">ID: F46</p><p class="technology">Technology: failures</p><p class="type">Type: Failure</p></section><section id="applicability"><h2>When to Use</h2>
+      <p>HTML</p>
    </section><section id="description"><h2>Description</h2>
       <p>The objective of this technique is to describe a failure that occurs when a
-                        table used only for layout includes either th elements, a
+                        table that is used only for layout is both 1) not marked as presentational and 2) includes either th elements, a
                             summary attribute, or a caption element. This
```
fstrr, line 10:
```suggestion
                            , or a caption element. This
```

fstrr, line 10:
```suggestion
                        table that is used only for layout is both 1) not marked as presentational and 2) includes either th elements, or a caption element. This
```

techniques/failures/F46.html
```diff
@@ -1,16 +1,16 @@
 <!DOCTYPE html><html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml"><head><title>Failure of Success Criterion 1.3.1 due to using th elements,
                     caption elements, or non-empty summary attributes in
```
fstrr, line 2:
```suggestion
<!DOCTYPE html><html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml"><head><title>Failure of Success Criterion 1.3.1 due to using th elements or
                    caption elements in
```

techniques/failures/F46.html
```diff
@@ -1,16 +1,16 @@
...
-                    layout tables</title><link rel="stylesheet" type="text/css" href="../../css/editors.css" class="remove"></link></head><body><h1>Failure of Success Criterion 1.3.1 due to using th elements,
+                    layout tables that are not marked presentational</title><link rel="stylesheet" type="text/css" href="../../css/editors.css" class="remove"></link></head><body><h1>Failure of Success Criterion 1.3.1 due to using th elements,
                     caption elements, or non-empty summary attributes in
```
fstrr, line 4:
```suggestion
                    caption elements in
```

fstrr:
[The summary attribute for tables is obsolete](https://html.spec.whatwg.org/multipage/obsolete.html), so would suggest updating the name of the technique to remove references to it (we've removed other `summary`-related content before. This includes re-naming the Technique. I've added suggestions to the PR.

I've also removed references to XHTML.

techniques/failures/F46.html
```diff
@@ -1,16 +1,16 @@
 <!DOCTYPE html><html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml"><head><title>Failure of Success Criterion 1.3.1 due to using th elements,
                     caption elements, or non-empty summary attributes in
-                    layout tables</title><link rel="stylesheet" type="text/css" href="../../css/editors.css" class="remove"></link></head><body><h1>Failure of Success Criterion 1.3.1 due to using th elements,
+                    layout tables that are not marked presentational</title><link rel="stylesheet" type="text/css" href="../../css/editors.css" class="remove"></link></head><body><h1>Failure of Success Criterion 1.3.1 due to using th elements,
```
detlevhfischer:
```suggestion
                    layout tables that are not marked as presentational</title><link rel="stylesheet" type="text/css" href="../../css/editors.css" class="remove"></link></head><body><h1>Failure of Success Criterion 1.3.1 due to using th elements,
```

techniques/failures/F46.html
```diff
@@ -70,8 +70,9 @@
                                     table element</li>
             <li>
                <p>If the table is used only to visually lay out elements within the
-                                    content </p>
+                                    content:</p>
                <ol>
+                  <li>Check that the table is marked presentational by methods such as <code>role="none"</code> or <code>role="presentation"</code> on the table element.</li>
```
detlevhfischer:
```suggestion
                  <li>Check that the table is marked as presentational by methods such as <code>role="none"</code> or <code>role="presentation"</code> on the table element.</li>
```

techniques/failures/F46.html
```diff
@@ -1,22 +1,30 @@
...
-         <li> 
-               <code>headers</code> attributes </li>
-         <li> 
-               <code>scope</code> attributes 
-              </li>
+         <li><code>headers</code> attributes</li>
+         <li><code>scope</code> attributes</li>
       </ul>
       <p>Assistive technologies use the structure of an HTML table to present data to
                         the user in a logical manner. The th element is used to mark
```
patrickhlauke:
```suggestion
                        the user in a logical manner. The <code>th</code> element is used to mark
```

techniques/failures/F46.html
```diff
@@ -28,8 +36,7 @@
                             summary attribute information available to users. The
```
patrickhlauke:
```suggestion
                            <code>summary</code> attribute information available to users. The
```

techniques/failures/F46.html
```diff
@@ -28,8 +36,7 @@
...
                             caption element is part of the table and identifies the
                         table.</p>
-      <p>
-            Although WCAG 2 does not prohibit the use of layout tables, CSS-based layouts are recommended in order to retain the defined semantic meaning of the HTML table elements and to conform to the coding practice of separating presentation from content. 
+      <p>Although WCAG 2 does not prohibit the use of layout tables, CSS-based layouts are recommended in order to retain the defined semantic meaning of the HTML table elements and to conform to the coding practice of separating presentation from content. 
             When a table is used
                         for layout purposes the th element should not be used. Since
```
patrickhlauke:
```suggestion
                        for layout purposes the <code>th</code> element should not be used. Since
```

techniques/failures/F46.html
patrickhlauke:
```suggestion
                        a <code>summary</code> attribute and do not use the <code>summary</code>
```

patrickhlauke:
```suggestion
                        navigating the content via a screen reader. Empty <code>summary</code>
```

patrickhlauke:
```suggestion
                                title as <code>th</code>.</p>
```

techniques/failures/F46.html
```diff
@@ -1,44 +1,51 @@
...
-              </li>
+         <li><code>headers</code> attributes</li>
+         <li><code>scope</code> attributes</li>
       </ul>
       <p>Assistive technologies use the structure of an HTML table to present data to
-                        the user in a logical manner. The th element is used to mark
+                        the user in a logical manner. The <code>th</code> element is used to mark
                         the column and row headers of the table. A screen reader uses the
                         information in th elements to speak the header information that
                         changes as the user navigates the table. The summary attribute on the
```
patrickhlauke:
```suggestion
                        changes as the user navigates the table. The <code>summary</code> attribute on the
```

patrickhlauke:
I started fixing this PR up a bit, but now taking a step back I'm wondering: what is this failure actually trying to achieve? Isn't it already a failure to have a layout table that is still exposed as a data table, in general? i.e. would we not already fail a table that is used for layout but does NOT include something like `role="presentation"` or `role="none"` or similar?

IF a table is exposed as presentational, then it shouldn't make a difference if there are `th` or similar structural elements, as they won't be exposed as such.

so really the failure currently seems to be a weird mitigation - IF you're using a layout table, but you didn't mark it as presentational (you already failed?), make sure you don't use structural elements (which would fail...harder?)

/cc @scottaohara

bruce-usab:
> Isn't it already a failure to have a layout table that is still exposed as a data table, in general?

Yes.

> i.e. would we not already fail a table that is used for layout but does NOT include something like role="presentation" or role="none" or similar?

No.  Historically, a layout table merely means avoiding `TH` or similar.  See WCAG 1.0 (!) [Create tables that transform gracefully](https://www.w3.org/WAI/GL/WCAG10/wcag10.html#gl-table-markup).

patrickhlauke:
@bruce-usab your "Yes" and then "No" are...in contradiction?

a layout table is a table used for layout. every table is exposed as a table by default (though browsers often use heuristics to suppress this themselves, but not sure they're documented officially anywhere?). you only suppress the semantics of data table when you explicitly mark it as such. in wcag 1.0 days the ability to do this probably didn't exist (with `role="presentation"` or similar)

bruce-usab:
> @bruce-usab your "Yes" and then "No" are...in contradiction?

If a table is being used for layout, but using `TH` to emphasize certain cells (and not as a label for row/column/row group) then that fails 1.3.1.  So **yes**, in general, it’s already a failure to have a layout table that is still exposed as a data table.

> you only suppress the semantics of data table when you explicitly mark it as such. in wcag 1.0 days the ability to do this probably didn't exist (with `role="presentation"` or similar)

Each of these data points supports my conclusion that, **no**, one doesn’t need anything as explicit as role="presentation" for layout tables.  It could be written up as a best practice.  And maybe use of role="presentation" _excuses_ using `TH` for emphasis (rather than as a label) and creates a loophole?

I don't think 1.3.1 is generally relevant to layout tables, except for an author misusing code (e.g., `TH` for emphasis, not as a label).  The main concern for layout tables is SC 1.3.2 Meaningful Sequence.

One pernicious pattern I’ve come across (more than once) is where the headers are one table and the data is in a second table immediately following.  Visually it looks like a single data table.  [The pages were being generated from some legacy databases.  (Maybe Oracle, but it could have been COBOL for all I know.)]  The assertion, by the owner, that these were layout tables was obviously disingenuous, but it’s the sort of thing which was easier to fail under WCAG 1.0 (and original 508).  I couldn’t point to 3.3.2 Labels or Instructions because the context wasn’t input, only an output report.  1.3.1 is so overloaded, but I don’t have any great ideas for improving that.

patrickhlauke:
> one doesn’t need anything as explicit as role="presentation" for layout tables

every table is by its very use/nature a data table, unless explicitly marked as presentational, no? i.e. screen readers will announce the presence of a table with X rows and Y columns, for instance.

or is your definition for layout tables more lax?

and again, if the group has the same lax view ("it's ok to use a table structure for layout purposes, without explicitly marking it/conveying it as presentational/neutralising table semantics, as long as no other semantic structures like `th` are used") then that needs to be explicitly documented in a note/exception

bruce-usab:
Yes, my definition for layout table is more lax.  Is there a non-subjective distinction?  I’ll know one when I see one!  Doesn’t the current Example 1 of F46 demonstrate that it’s okay to use a table structure for layout purposes — without explicitly marking it/conveying it as presentational/neutralizing table semantics?

If you want to add to the above:  as long as no other semantic structures like `th` are <ins>ab</ins>used

How is it problematic for screen reading software to announce the presence of a table with X rows and Y columns when encountering a layout table (that passes 1.3.2)?

patrickhlauke:
> Doesn’t the current Example 1 of F46 demonstrate that it’s okay to use a table structure for layout purposes — without explicitly marking it/conveying it as presentational/neutralizing table semantics?

Yes, which I think is problematic

> How is it problematic for screen reading software to announce the presence of a table with X rows and Y columns when encountering a layout table (that passes 1.3.2)?

Imagine a table, with various other nested tables, and further nested tables... are we saying that it's ok, despite SRs then announcing the structural information to users?

Semantically, "The [table](https://html.spec.whatwg.org/multipage/tables.html#the-table-element) element [represents](https://html.spec.whatwg.org/multipage/dom.html#represents) data with more than one dimension, in the form of a [table](https://html.spec.whatwg.org/multipage/tables.html#concept-table)." https://html.spec.whatwg.org/multipage/tables.html#the-table-element - so by definition all tables are data tables. the spec also says "Tables must not be used as layout aids."

mraccess77:
I think a table used for layout that includes THs would fail unless marked with role presentation to neutralize the table headers.  However, I think a layout table without any headers, scope, THs, etc. would pass if used for layout without role presentation. So for me - it's the presence of those data table type markers.

pkra:
@patrickhlauke 

> every table is exposed as a table by default (though browsers often use heuristics to suppress this themselves, but not sure they're documented officially anywhere?). 

Not officially but there's a draft to document them in html-aam: https://github.com/w3c/aria/pull/2576

patrickhlauke:
@mraccess77 

>  I think a layout table without any headers, scope, THs, etc. would pass if used for layout without role presentation

would that also apply to double/triple/deeper nested table constructs?

(I'm angling to document this somewhere as a note, likely in the 1.3.1 understanding itself rather than by inference in a technique)

mraccess77:
> @mraccess77
> 
> > I think a layout table without any headers, scope, THs, etc. would pass if used for layout without role presentation
> 
> would that also apply to double/triple/deeper nested table constructs?

Layout tables nested within layout tables generally aren't an issue - however, what could be more problematic is data tables within data tables and possibly layout tables within data tables.  If a layout table within a cell of a datable table causes issues for assistive technology then I would consider that problematic - I believe this used to be the case - but I'm not sure if it's an issue today or not.  Previously, tables within table header cells and tables within data cells could cause screen readers to loose track of what sell they were in.  So, in some cases it could be a fail if it caused that under accessibility support of 1.3.1.

stevefaulkner:
As defined in the [HTML spec](https://html.spec.whatwg.org/multipage/tables.html#the-table-element):
> The [table](https://html.spec.whatwg.org/multipage/tables.html#the-table-element) element [represents](https://html.spec.whatwg.org/multipage/dom.html#represents) data with more than one dimension, in the form of a [table](https://html.spec.whatwg.org/multipage/tables.html#concept-table).The [table](https://html.spec.whatwg.org/multipage/tables.html#the-table-element) element [represents](https://html.spec.whatwg.org/multipage/dom.html#represents) data with more than one dimension, in the form of a [table](https://html.spec.whatwg.org/multipage/tables.html#concept-table).

If it is not being used to represent multi-dimensional data (i.e. its being used for layout) then the semantics should not be exposed. `role="none"` or `role="presentation"` is a robust method to hide incorrect semantics and therefore conform with 1.3.1 Info and Relationships.

bruce-usab:
>  the spec also says "Tables must not be used as layout aids."

I submit that, from a WCAG perspective, this is new success criterion.  

Apart from concerns for linearization, layout tables per se haven’t been an accessibility barrier.  Does not `role=presentation` post-date first publication of WCAG 2.1?  How can it retroactively now be a requirement for conformance?

Nested tables seems like a whole other issue.

patrickhlauke:
> Does not role=presentation post-date first publication of WCAG 2.1? How can it retroactively now be a requirement for conformance?

the option to use CSS layout rather than misusing tables for layout purposes predates WCAG 2.0 ...

either way, none of this seems to be explicitly mentioned/documented. in fact, the only place in understanding for 1.3.1 where tables are mentioned (other than in the list of techniques), they're mentions of data tables 

> items that share a common characteristic are organized into a table where the relationship of cells sharing the same row or column and the relationship of each cell to its row and/or column header are necessary for understanding

> A bus schedule table where the headers for each cell can be programmatically determined

And in all the techniques, F46 is currently the only one that mentions layout tables...

https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships.html

if the intention was always to "allow" layout tables, then it should have been mentioned more explicitly, not by stealth in F46?

if we DO want to keep this exemption of tables used for layout, it should at least be a note in 1.3.1 understanding, explaining that historically, tables have been used in the past for layout, and they're still acceptable provided they don't have any further structural elements or attributes that only make sense for data tables... (which this PR's change to F46 - with some tweaks/additions of mine - covers a bit more broadly than the current F46)

bruce-usab:
I submit that it would be significant departure for WCAG 2.x AA to always fail layout tables.  Especially, of course, through an informative document.  Yes, of course CSS is the better alternative.

I believe that’s this the tightrope F46 was trying to walk.  Layout tables are bad, and they _might_ fail SC 1.3.1.

Here’s the F46 in TR space:  https://www.w3.org/TR/WCAG20-TECHS/F46.html

_The example code **is** meant to illustrate a failure!_  At the very least, the page should be updated to be explicit about that.  Ironically, by having this as a failure example, it’s the also the evidence you are looking for that (absent other defects) WCAG AA permits layout tables.

stevefaulkner:
Use of `role=presentation` is not a WCAG 2 requirement, its one technique to remove incorrect semantics being conveyed to a subset of users, the other technique is not to use `<table>`'s for layout.

stevefaulkner:
WCAG 1.0 says

> Tables should be used to mark up truly [tabular information](https://www.w3.org/TR/WAI-WEBCONTENT/#tabular-information) ("data tables"). Content developers should avoid using them to lay out pages ("layout tables")

source: https://www.w3.org/TR/WAI-WEBCONTENT/#gl-table-markup

so it is not a new idea....

bruce-usab:
Not, not a new ides.  But citation uses _should_ not _must_ and WCAG2 has not advanced (for better or worse) beyond that.

Also see [F49](https://www.w3.org/WAI/WCAG22/Techniques/failures/49) Failure of Success Criterion 1.3.2 due to using an HTML layout table that does not make sense when linearized.  Which starts out:
> Although WCAG 2 does not prohibit the use of layout tables…

I submit that requiring `role=presentation` would be a [class 4 change](https://www.w3.org/policies/process/#correction-classes).  Ping to @patrickhlauke

yatil:
Layout tables impede programmatically determination of information, structure, and relationships (because they inject irrelevant information into the otherwise relevant data), so I would think they fail info and relationship.

If the relationship visually says it is not a table, but the code says it is, then the "not a table" information that is visually conveyed is not available to assistive technologies.

The absence of structure/information is structure/information itself.

patrickhlauke:
> I submit that requiring role=presentation would be a [class 4 change](https://www.w3.org/policies/process/#correction-classes).

and again, this is NOT about `role=presentation` per se. it's about: are layout tables passing or failing 1.3.1 ... using a markup structure that by definition is intended to mark up data that has a horizontal/vertical relationship for a use that the markup was not intended for.

patrickhlauke:
> Also see [F49](https://www.w3.org/WAI/WCAG22/Techniques/failures/49) Failure of Success Criterion 1.3.2

so IF it was always the intention to let layout tables slide, despite them being a bastardised misuse of markup ... then this should be made clear at least in understanding, not handwaved somewhere in a technique. something like (what i riffed on before in https://github.com/w3c/wcag/pull/3566#issuecomment-4381697624)

"historically, tables have been used in the past for layout. while this is not an appropriate use of this markup structure, WCAG 2 does allow the use of layout tables / layout tables do not fail 1.3.1 Info and Relationships, but only provided that provided that these tables don't have any further structural elements or attributes that only make sense for data tables" (and maybe even mention that nesting layout tables also then starts tipping them into failure?)

stevefaulkner:
@patrickhlauke wrote
> acceptable provided they don't have any further structural elements or attributes that only make sense for data tables.

When does the presence in the acc tree of `role=table`, `role=row` and `role=cell` make sense for layout tables?

awkawk:
> so IF it was always the intention to let layout tables slide, despite them being a bastardised misuse of markup ... then this should be made clear at least in understanding, not handwaved somewhere in a technique. something like (what i riffed on before in https://github.com/w3c/wcag/pull/3566#issuecomment-4381697624)

At the time of course it was not a misuse of markup, or at least it recognized that CSS was not sufficiently well-supported to eliminate the need for layout tables. The WG certainly accepted that layout tables were acceptable, and people also wanted to see CSS support improve to reduce/eliminate the need for their use.

Techniques and Understanding are equal in status. It seems like you are suggesting that Understanding > Techniques here.

> And in all the techniques, F46 is currently the only one that mentions layout tables...

Not exactly:
H39: https://www.w3.org/WAI/WCAG22/Techniques/html/H39
"Although WCAG 2 does not prohibit the use of layout tables, CSS-based layouts are recommended in order to retain the defined semantic meaning of the HTML table elements and to conform to the coding practice of separating presentation from content. If a table is used for layout, the caption element is not used. The purpose of a layout table is simply to control the placement of content; the table itself is "transparent" to the user. A caption would "break" this transparency by calling attention to the table. See [F46: Failure of Success Criterion 1.3.1 due to using th elements, layout tables](https://www.w3.org/WAI/WCAG20/Techniques/failures/F46) for details."

H43: https://www.w3.org/WAI/WCAG22/Techniques/html/H43
"This technique is not recommended for layout tables since its use implies a relationship between cells that is not meaningful when tables are used for layout."

F43: https://www.w3.org/WAI/WCAG22/Techniques/failures/F43
"Note that the use of HTML tables for layout is not an example of this failure as long as the layout table does not include improper structural markup such as <th> or <caption> elements."

F49: https://www.w3.org/WAI/WCAG22/Techniques/failures/F49
"Although WCAG 2 does not prohibit the use of layout tables, CSS-based layouts are recommended in order to retain the defined semantic meaning of the HTML table elements and to conform to the coding practice of separating presentation from content. If a layout table is used, however, it is important that the content make sense when linearized."

patrickhlauke:
> Techniques and Understanding are equal in status. It seems like you are suggesting that Understanding > Techniques here.

from a "discoverability of information" perspective, certainly not equal. or do you expect folks to trawl through all techniques, just in case they miss some important information that was buried in there?

patrickhlauke:
> At the time of course it was not a misuse of markup, or at least it recognized that CSS was not sufficiently well-supported to eliminate the need for layout tables. The WG certainly accepted that layout tables were acceptable, and people also wanted to see CSS support improve to reduce/eliminate the need for their use.

well then, let's put our money where our mouth is, and make an explicit note in understanding (for discoverability, if nothing else)

bruce-usab:
> well then, let's put our money where our mouth is, and make an explicit note in understanding (for discoverability, if nothing else)

I am okay with that, but _which_ Understanding doc, 1.3.1?

From this conversation, I think at least one of the caveats in H43 wrt layout tables should be strengthened.  Maybe:
> This technique is not recommended for layout tables since its use implies a relationship between cells that is not meaningful when tables are used for layout <ins>and may cause a failure of success criterion 1.3.1 Info and relationships</ins>.

stevefaulkner:
> well then, let's put our money where our mouth is, and make an explicit note in understanding (for discoverability, if nothing else)

I am not OK with that. There is no normative exemption for tables in SC 1.3.1, if there is to be that would be the place for it.

yatil:
There is no disagreement that tables for layouts are OK when the layout table is marked as such. [HTML does not allow layout tables at all.](https://html.spec.whatwg.org/multipage/tables.html#the-table-element:~:text=Tables%20must%20not%20be%20used%20as%20layout%20aids.) Just not using structural elements has no impact from an HTML/programmatically deterministic view:

> Tables must not be used as layout aids. Historically, some web authors have misused tables in HTML as a way to control their page layout. This usage is non-conforming, because tools attempting to extract tabular data from such documents would obtain very confusing results. In particular, users of accessibility tools like screen readers are likely to find it very difficult to navigate pages with tables used for layout.

While the spec defines heuristics for determining layout tables, this is a workaround. As a tester, anything but `role=none`/`presentation` depends on the browser heuristics. There is no advantage for testers and implementers to not be unambiguous in this requirement.

(Edit: Typo.)

yatil:
This also opens the door for all other types of heuristics: Big and bold text? Browsers could interpret it as a heading. Emojis in front of lines of text: Browsers could interpret it as a list.

These heuristics make it "programmatically guessable", not "determinable".

PR #3568, opened 2023-11-25
Add clarification that change of color still needs sufficient 3:1 ratio, darken focused star

patrickhlauke:


bruce-usab (approved):
Looks good!

PR #3571, opened 2023-11-28
Update README.md translations 2.1 -> 2.2

shawna-slh:


rdwz:
Unfortunately there is no branch WCAG-2.2
https://github.com/w3c/wcag/branches/all?query=WCAG-2

patrickhlauke:
> Unfortunately there is no branch WCAG-2.2 https://github.com/w3c/wcag/branches/all?query=WCAG-2

indeed. the correct branch here is simply `main`

shawna-slh:
Thanks for the catch! That's what I get for trying to help by doing an item on Kevin's To Do list.

@iadawn could you confirm that indeed we want translators to use main, and update the translations info. cc @remibetin

remibetin:
@iadawn Is the `main` branch always in sync with the version published on the live website, or is the `main` branch sometimes ahead of the live version?

PR #3593, opened 2023-12-06
New CSS technique: "Using CSS z-index to un-obscure focused content"

### LINKED ISSUE #3578: New CSS Technique for Focus Not Obscured: z-index focus modifier
I'm proposing a potentially ill-advised CSS technique to meet the AAA level criterion, '2.4.12 Focus not Obscured (Enhanced)'.

As of the creation of this GitHub issue, there are no techniques exclusive to the AAA level of 'Focus not obscured'. The one technique offered, [technique C43](https://www.w3.org/WAI/WCAG22/Techniques/css/C43), is also applied to the AA level.

## The problem
The AAA level criterion for 'Focus not obscured' has a very high standard for success:
> "When a [user interface component](https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-enhanced.html#dfn-user-interface-component) receives keyboard focus, no part of the component is hidden by author-created content."

This creates a difficult scenario, which doesn't yet appear to have a sufficient technique to solve it. 
There's a chance to fail 2.4.12 any time that all the following are true:
- there's at least **one 'pop-over' element**
- the pop-over element is situated **over top of 'body content'** areas (basically, situated where it may overlap any [flow content](https://html.spec.whatwg.org/multipage/dom.html#flow-content) and not situated over a blank area)
- the pop-over **persists without focus**

When these 3 things are true, there's a chance that focused 'body content' elements become obscured by the pop-over element.

## The solution

Basically: on focus, set the z-index really high for 'body content' elements in the given area.

This is a potentially controversial solution, but I believe it's worth considering due to the current absence of any alternative AAA-level 'focus not obscured' techniques.

### Example CSS

```
.thing:focus {
    position: relative;
    z-index: calc(infinity) !important;
}'
```

## Pros and cons
This works because:
1. it can be applied to a specific section of content (such as a side-navigation)
2. it will guarantee that the focused element is right on top, unobscured
3. it uses simple CSS properties and selectors

This might be controversial because:
1. playing with z-index when it's not necessary can be risky (at least, I am sensing a vibe that [it's taboo to use z-index frivolously](https://www.codegram.com/blog/stop-ab-using-z-index/))
2. there's risks of broken interface if it's applied to a whole page / site
3. it requires careful implementation, since elements need a declared [position property](https://developer.mozilla.org/en-US/docs/Web/CSS/position)

## Additional details

This solution requires all focusable elements which might be obscured by the pop-over element to have:
- a CSS position property (like `position: relative` or `position: absolute`
- a solid background of some type behind the text (like `background: white`)

It also requires careful selection of a z-index value. I tried various index amounts, to try and find a universally high enough value to override pop-over elements like chatbots, cookie banners, menus, etc. The only one that is guaranteed to work is pretty much just infinity. It seems ridiculous, but `calc(infinity)` might be appropriate to guarantee focus is never obscured.

dav-idc:
**Technique title:** Using CSS z-index to un-obscure focused content

First draft of an HTML page for a new technique, described in detail in this GitHub issue:
- https://github.com/w3c/wcag/issues/3578

<del>Right now the technique number is set to 'C44', but that can be changed if needed</del>

[Preview](https://deploy-preview-3593--wcag2.netlify.app/techniques/css/technqiue-using-css-z-index-to-un-obscure-focused-content.html)

(edited to update technique file name and preview link as we re-purposed C44)


fstrr:
[Preview page](https://deploy-preview-3593--wcag2.netlify.app/techniques/css/technqiue-using-css-z-index-to-un-obscure-focused-content.html)

kfranqueiro:
The applicability portion of this should probably be rewritten and partially moved. Most of its content makes more sense to be in the description, and it is currently not phrased in a way that is consistent/compatible with the same section in other SCs. (Note how the first sentence is currently awkward in the preview, due to the text that is prepended automatically.)

scottaohara:
i think we'd definitely need to make mention of `popover` and how css z-index wouldn't do any good for content rendered in the top-layer - since that is beyond what z-index can reach.

i could also see a recommendation to make popups using the popover attribute - so that if they do obscure content, and they are `auto` popovers, then they can be dismissed using the `esc` key, without a user having to advance their keyboard focus.

scottaohara (changes requested):
see my prior comment about needing to mention z-index vs top layer conent

patrickhlauke:
assigning to @scottaohara as he's got the closest read and what could still be added/mentioned

patrickhlauke:
moving to "in progress" for now until @scottaohara had a chance to review/comment again

Issue #3594, opened 2023-12-07
Link in Related Resources not working as expected in Understanding SC 3.3.8: Accessible Authentication (Minimum)


giacomo-petri:
- **URL**: https://www.w3.org/WAI/WCAG22/Understanding/accessible-authentication-minimum.html
- **Issue description**: In the Understanding document of SC 3.3.8: Accessible Authentication (Minimum), the link in Realted Resources [Making Content Usable for People with Cogntive and Learning Disabilities 4.7.1 Provide a Login that Does Not Rely on Memory or Other Cognitive Skills](https://www.w3.org/TR/coga-usable/#provide-a-login-that-does-not-rely-on-memory-or-other-%20%20%20%20%20%20%20%20%20%20%20%20%20cognitive-skills-pattern) is not working as expected.
The anchor is broken and fails to navigate the user to the intended section of the page.
- Together with this I'll also fix a typo in this link ("Cogntive").

Opening a pull request.

PR #3595, opened 2023-12-07
Link in Related Resources not working as expected in Understanding SC 3.3.8: Accessible Authentication (Minimum)

giacomo-petri:
Closes: #3594

- URL: https://www.w3.org/WAI/WCAG22/Understanding/accessible-authentication-minimum.html
- Fixes: 
    - In the Understanding document of SC 3.3.8: Accessible Authentication (Minimum), the link in Realted Resources [Making Content Usable for People with Cogntive and Learning Disabilities 4.7.1 Provide a Login that Does Not Rely on Memory or Other Cognitive Skills](https://www.w3.org/TR/coga-usable/#provide-a-login-that-does-not-rely-on-memory-or-other-%20%20%20%20%20%20%20%20%20%20%20%20%20cognitive-skills-pattern) is not working as expected.
The anchor is broken and fails to navigate the user to the intended section of the page.
    - Together with this I'll also fix a typo in this link ("Cogntive").


mbgower:
Removed the PR from Project board, since the Issue is already there.

Issue #3561, opened 2023-11-20
Broken link/need for updated reference on Understanding: Non-text contrast


matt-t-tinkers:
On the [understanding doc for Non-text contrast](https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast), **Figure 10** has a broken link to Focus Appearnace (https://www.w3.org/WAI/WCAG22/22/focus-appearance-minimum.html). 

Also, as Focus Appearance (minimum) is a checkpoint for achieving AAA, it would be particularly useful for clearer guidance how around having a thicker line can pass this checkpoint at AA.

mbgower:
@matt-t-tinkers, originally there were AA and AAA criteria for Focus Appearance, but these were combined into a single AAA criterion due to some difficulties with consistent testing and interpretation.
That [Understanding](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance-minimum.html) document provides good guidance on improving the visibility of focus indicators, regardless of the level of WCAG you are targetting. We will fix the broken URL.

Does that answer you question?

PR #3605, opened 2023-12-12
Fixed Figure 10 broken link to Focus Appearance on the Understanding Document for Non-text contrast

giacomo-petri:
Closes: #3561

Fixed Figure 10 broken link to Focus Appearance on the Understanding Document for [Non-text contrast](https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast).

I've also updated links pointing to WCAG 2.1, replacing the absolute URLs with a relative one.

### DIFF
--- a/understanding/21/non-text-contrast.html
+++ b/understanding/21/non-text-contrast.html
@@ -78,7 +78,7 @@
 
 				<h4 id="related-color">Relationship with Use of Color</h4>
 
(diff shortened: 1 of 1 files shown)


scottaohara (approved):

alastc:
Unfortunately the local links are different from the published locations, so I updated the "../../techniques" (correct locally) to "../Techniques" (works when published).

alastc (approved):

PR #3620, opened 2024-01-04
Add IDs to avoid rewriting during translations

remibetin:
### Current Issue:

By default:
- ReSpec uses heading's text content to generate IDs. See https://respec.org/docs/#custom-heading-ids
- Sections are automatically numbered and uniquely id's. See https://respec.org/docs/#sections
- IDs for glossary entries (`dfn` elements) are generated based on description term (or `data-lt` value when existing)

Therefore, when main headings, guidelines headings and definitions terms are translated in another language, IDs also change. 

Issues include:
- That does not work properly for non-latin characters. For example, for Simplified Chinese headings, automatically created IDs are `generatedID`, `generatedID-0`, etc.
- Translations checks are more difficult when IDs differ from the original version.
- Restoring the English IDs after exporting to HTML is very tedious.

### Proposed Solution:

Stable IDs are manually added to main headings, guidelines headings and definition terms, so that they are not overwritten by ReSpec during translation.

In this pull request, I have used the currently generated IDs from the live original version.

***
<a href="https://pr-preview.s3.amazonaws.com/remibetin/wcag/pull/3620.html" title="Last updated on Jan 11, 2024, 1:00 PM UTC (81818c5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/3620/cf3d2fb...remibetin:81818c5.html" title="Last updated on Jan 11, 2024, 1:00 PM UTC (81818c5)">Diff</a>

### DIFF
--- a/acknowledgements/funders.html
+++ b/acknowledgements/funders.html
@@ -1,4 +1,4 @@
-<section>
+<section id="enabling-funders">
 	<h2>Enabling funders</h2>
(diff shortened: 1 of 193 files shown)


remibetin:
Note: The preview is not working well. It cannot get the SC and terms. Is it because my branch is from another repo?

PR #3653, opened 2024-01-23
Remove SCR21 and update page that links to it

### LINKED ISSUE #3652: Do we still need Technique SCR21 ("Using functions of the Document Object Model (DOM) to add content to a page")?
SCR21 [Using functions of the Document Object Model (DOM) to add content to a page](https://www.w3.org/WAI/WCAG21/Techniques/client-side-script/SCR21) seems to be less about the title of the document and more about certain scripting functionality that the author thinks should be avoided. For example:

> "…and the innerHTML property is not part of the DOM specification and thus should be avoided."

and the Procedure, which consists of one item:

> "Examine the source code and check that the new content is not created using document.write(), innerHTML, outerHTML, innerText or outerText."

As such:

1. Do we still this document?
2. If we do, then updating it to bring it up to date and re-writing the Procedure seems to be the way to go.

fstrr:
Closes #3652


patrickhlauke (approved):

bruce-usab (approved):

patrickhlauke:
Randomly ... why does https://www.w3.org/WAI/WCAG22/Techniques/client-side-script/SCR21 still exist? Is this the usual "we don't actually remove files?" thing? @iadawn

PR #3656, opened 2024-01-23
Update H69 for accurate 'accessibility supported'

### LINKED ISSUE #1712: Are 'Headings' enough to pass 2.4.1: Bypass Blocks
A sufficient technique for 2.4.1: Bypass Blocks is:

mitchellevan:
~fixes~ does not fix, but moves in the right direction for #1712


awkawk:
Bringing in cost here is problematic.

mitchellevan:
> Bringing in cost here is problematic.

My intention was to highlight normative WCAG, which does use the word "cost." I paraphrased by saying "no cost" instead of same cost, which in the content of browser and browser plug-ins I believe is accurate.

awkawk:
"The technology is supported in a widely-distributed plug-in that is also accessibility supported;"

There's no cost mentioned in this second option under the second condition of accessibility supported. The 4th option mentions cost but the 4th option isn't required. That's why it is problematic.

patrickhlauke:
i think this proposed addition is a bit awkward/heavy handed. https://github.com/w3c/wcag/issues/1712 is still a valid concern, but i think this may need some more tweaking of the technique itself, and couching it in some conditional framing (as was suggested in https://github.com/w3c/wcag/issues/1712#issuecomment-2671083219)

patrickhlauke:
I'm actually wondering as well now if H69 should be made advisory, rather than sufficient, for Bypass Blocks - as otherwise we could say that of course it's sufficient to just mark up headings, since keyboard users (without other AT) that need to navigate by headings will have some plugin or specialised browser that lets them do that... but as that's not a common feature of browsers, and there's only a handful of obscure extensions, it seems a stretch to just allow headings and put onus on users to hunt those tools down

PR #3670, opened 2024-02-02
Update keyboard to remove normative acceptance of "completely custom …

detlevhfischer:
Changes 
> "deviating from these conventions does <em>not</em> fail the normative requirement" 

to 
> "deviating from these conventions does <em>not</em> always fail the normative requirement"

### DIFF
--- a/understanding/20/keyboard.html
+++ b/understanding/20/keyboard.html
@@ -75,19 +75,18 @@
             the platform/user agent conventions it may be difficult to use, as users will need
             to learn different interaction methods. As a <em>best practice</em>, content
             should follow the platform/user agent conventions. However, deviating from these
-            conventions does <em>not</em> fail the normative requirement of this Success Criterion.</p>
+            conventions does <em>not</em> always fail the normative requirement of this Success Criterion.</p>
          <p>
             For instance, buttons that have focus can generally be activated using both the
             <kbd>Enter</kbd> key and the <kbd>Space</kbd> bar. If a custom button control
-            in a web application instead only reacts to <kbd>Enter</kbd>
-            (or even a completely custom key or key combination), this still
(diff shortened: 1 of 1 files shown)


patrickhlauke:
saying "does not always fail" without explicitly explaining when it does or doesn't ... won't really help authors or testers in determining whether they do pass or fail

understanding/20/keyboard.html
```diff
@@ -75,19 +75,18 @@ <h2>Intent of Keyboard</h2>
             the platform/user agent conventions it may be difficult to use, as users will need
             to learn different interaction methods. As a <em>best practice</em>, content
             should follow the platform/user agent conventions. However, deviating from these
-            conventions does <em>not</em> fail the normative requirement of this Success Criterion.</p>
+            conventions does <em>not</em> always fail the normative requirement of this Success Criterion.</p>
```
chaals, line 78: Partially dealing with @patrickhlauke's feedback...

```suggestion
            conventions does <em>not</em> automatically fail the normative requirement of this Success Criterion.</p>
```

(The rest of his comment is addressed, IMHO, by the text that follows).

patrickhlauke: that still leaves doubt though about when it does or doesn't, so authors and testers still are none the wiser

patrickhlauke:
> this is likely to fail AT users relying on the keyboard interface

noting that AT generally (need to check, but pretty sure) have pass-through commands that allow users to send the next key press directly through to the page. cumbersome? sure. but wouldn't fail, in my view.

patrickhlauke:
somehow, with the various rebases/updates since it was originally filed, this PR ended up with 0 file changes. @detlevhfischer suggest reopening a fresh one if you still remember what you were trying to achieve here.

PR #3671, opened 2024-02-02
Anti-aliasing notes for 1.4.3, 1.4.6, 1.4.11 understanding docs (more strict proposal)

dbjorge:
Followup from today's TF discussion (see https://github.com/w3c/wcag/pull/3560#issuecomment-1924263718)

This is a more strict variation of the notes added in #3020 and #3560. I think this addresses @WilcoFiers 's [concerns in this comment](https://github.com/w3c/wcag/pull/3560#pullrequestreview-1856237319).

I think this version is more consistent with [this ACT rule that AGWG approved](https://www.w3.org/WAI/standards-guidelines/act/rules/afw4f7/), but maybe concerning that it seems more strict than what "NOTE 2" in the [contrast ratio definition](https://www.w3.org/TR/WCAG22/#dfn-contrast-ratio) suggests.

### DIFF
--- a/understanding/20/contrast-enhanced.html
+++ b/understanding/20/contrast-enhanced.html
@@ -86,7 +86,7 @@
             Criterion, but the text as-rendered has a much lower contrast in practice and fails the Success Criterion.
          </p>
          <p>In these cases, best practice would be for authors to choose a font with stronger/thicker lines,
-            or to aim for a foreground/background color combination that exceeds the normative requirements
+            or to aim for a foreground and background color combination that exceeds the normative requirements
             of this Success Criterion.
          </p>
       </div>
--- a/understanding/20/contrast-enhanced.html
+++ b/understanding/20/contrast-enhanced.html
@@ -73,13 +73,19 @@
       </div>
 
       <div class="note">
-         <p>Because authors do not have control over user settings for font smoothing/anti-aliasing, when evaluating this
-            Success Criterion, refer to the foreground and background colors obtained from the user agent, or the underlying
-            markup and stylesheets, rather than the text as presented on screen.</p>
-         <p>Due to anti-aliasing, particularly thin or unusual fonts may be rendered by user agents with a much fainter
-            color than the actual text color defined in the underlying CSS. This can lead to situations where text has
-            a contrast ratio that nominally passes the Success Criterion, but has a much lower contrast in practice.
-            In these cases, best practice would be for authors to choose a font with stronger/thicker lines,
+         <p>User agents which render text will often apply anti-aliasing, font smoothing, and other techniques
+            which can cause some pixels of text to render in fainter colors than those specified by authors.
+            When evaluating this Success Criterion, this is usually not relevant; as long as the applied font,
+            size, and weight ensure that some pixel of each character reaches the author-specified color, evaluators
+            may refer to the foreground and background colors obtained from the user agent, or the underlying
+            markup and stylesheets, rather than evaluating individual pixels within the text as presented on screen.
+         </p>
+         <p>However, particularly thin, small, or unusual fonts may be rendered by user agents with a much fainter
+            color than the actual text color defined in the underlying markup and stylesheets. This can lead to
+            situations where a user agent reports that text has a contrast ratio that nominally passes the Success
+            Criterion, but the text as-rendered has a much lower contrast in practice and fails the Success Criterion.
+         </p>
+         <p>In these cases, best practice would be for authors to choose a font with stronger/thicker lines,
             or to aim for a foreground/background color combination that exceeds the normative requirements
             of this Success Criterion.
          </p>
--- a/understanding/20/contrast-minimum.html
+++ b/understanding/20/contrast-minimum.html
@@ -80,9 +80,9 @@
             markup and stylesheets, rather than evaluating individual pixels within the text as presented on screen.
          </p>
          <p>However, particularly thin, small, or unusual fonts may be rendered by user agents with a much fainter
-            color than the actual text color defined in the underlying CSS. This can lead to situations where a
-            user agent reports that text has a contrast ratio that nominally passes the Success Criterion, but
-            the text as-rendered has a much lower contrast in practice and fails the Success Criterion.
+            color than the actual text color defined in the underlying markup and stylesheets. This can lead to
+            situations where a user agent reports that text has a contrast ratio that nominally passes the Success
+            Criterion, but the text as-rendered has a much lower contrast in practice and fails the Success Criterion.
          </p>
          <p>In these cases, best practice would be for authors to choose a font with stronger/thicker lines,
             or to aim for a foreground/background color combination that exceeds the normative requirements
--- a/understanding/20/contrast-minimum.html
+++ b/understanding/20/contrast-minimum.html
@@ -72,13 +72,19 @@
       </div>
       
       <div class="note">
-         <p>Because authors do not have control over user settings for font smoothing/anti-aliasing, when evaluating this
-            Success Criterion, refer to the foreground and background colors obtained from the user agent, or the underlying
-            markup and stylesheets, rather than the text as presented on screen.</p>
-         <p>Due to anti-aliasing, particularly thin or unusual fonts may be rendered by user agents with a much fainter
-            color than the actual text color defined in the underlying CSS. This can lead to situations where text has
-            a contrast ratio that nominally passes the Success Criterion, but has a much lower contrast in practice.
-            In these cases, best practice would be for authors to choose a font with stronger/thicker lines,
+         <p>User agents which render text will often apply anti-aliasing, font smoothing, and other techniques
+            which can cause some pixels of text to render in fainter colors than those specified by authors.
+            When evaluating this Success Criterion, this is usually not relevant; as long as the applied font,
+            size, and weight ensure that some pixel of each character reaches the author-specified color, evaluators
+            may refer to the foreground and background colors obtained from the user agent, or the underlying
+            markup and stylesheets, rather than evaluating individual pixels within the text as presented on screen.
+         </p>
+         <p>However, particularly thin, small, or unusual fonts may be rendered by user agents with a much fainter
+            color than the actual text color defined in the underlying CSS. This can lead to situations where a
+            user agent reports that text has a contrast ratio that nominally passes the Success Criterion, but
+            the text as-rendered has a much lower contrast in practice and fails the Success Criterion.
+         </p>
+         <p>In these cases, best practice would be for authors to choose a font with stronger/thicker lines,
             or to aim for a foreground/background color combination that exceeds the normative requirements
             of this Success Criterion.
          </p>
--- a/understanding/20/contrast-minimum.html
+++ b/understanding/20/contrast-minimum.html
@@ -84,7 +84,7 @@
             situations where a user agent reports that text has a contrast ratio that nominally passes the Success
             Criterion, but the text as-rendered has a much lower contrast in practice and fails the Success Criterion.
          </p>
-         <p>In these cases, best practice would be for authors to choose a font with stronger/thicker lines,
+         <p>In these cases, best practice would be for authors to choose a font with stronger or thicker lines,
             or to aim for a foreground/background color combination that exceeds the normative requirements
             of this Success Criterion.
          </p>
--- a/understanding/21/non-text-contrast.html
+++ b/understanding/21/non-text-contrast.html
@@ -29,6 +29,16 @@
 				<p>The 3:1 contrast ratios referenced in this Success Criterion is intended to be treated as threshold values. When comparing the computed contrast ratio to the Success Criterion ratio, the computed values should not be rounded (e.g. 2.999:1 would not meet the 3:1 threshold).</p>
 			</div>
 
+			<div class="note">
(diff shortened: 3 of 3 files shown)


understanding/21/non-text-contrast.html
```diff
@@ -29,6 +29,16 @@ <h2>Intent</h2>
...
+
+				<p>However, some authoring techniques can result in components or graphics which user agents will render using much fainter colors than the colors defined in underlying markup and stylesheets. This can lead to situations where a user agent reports that content has a contrast ratio that nominally passes the Success Criterion, but the content as-rendered has a much lower contrast in practice and fails the Success Criterion.</p>
+
+				<p>In these cases, best practice would be for authors to choose an technique which guarantees that the user agent will render the content with a contrast ratio that meets the Success Criterion. For example, an author seeking to use a line drawn around a button to meet this Success Criterion might choose to use a CSS <code>border</code> or <code>outline</code> property, rather than a CSS <code>box-shadow</code> property; most user agents will attempt to render fractional pixels of a shadow using fainter colors, but will not use this technique to render fractional pixels of a border or outline.</p>
```
fstrr, line 39: choose a technique

JAWS-test:
I think there are two very different concepts of contrast measurement here and it should be calmly considered which interpretation should apply in the future and what effects this has  
* on accessibility (where it would be better to measure the actual colors)
* on the established test methods (where many testers have the text contrasts tested by automated tools that use the CSS color values and where the search for the greatest contrast distance is a somewhat subjective matter that can lead to different results with different testers).

I would like to emphasize that it is not so rare that the CSS contrasts differ from the actual contrasts - i.e. clarification of this issue would be very important. 

I see advantages and disadvantages of both test methods and advocate neither, but would like to see a clear decision that is properly documented in the Understandings.

understanding/20/contrast-enhanced.html
```diff
@@ -73,13 +73,19 @@ <h2>Intent of Contrast (Enhanced)</h2>
...
-            Success Criterion, refer to the foreground and background colors obtained from the user agent, or the underlying
-            markup and stylesheets, rather than the text as presented on screen.</p>
-         <p>Due to anti-aliasing, particularly thin or unusual fonts may be rendered by user agents with a much fainter
-            color than the actual text color defined in the underlying CSS. This can lead to situations where text has
-            a contrast ratio that nominally passes the Success Criterion, but has a much lower contrast in practice.
-            In these cases, best practice would be for authors to choose a font with stronger/thicker lines,
+         <p>User agents which render text will often apply anti-aliasing, font smoothing, and other techniques
+            which can cause some pixels of text to render in fainter colors than those specified by authors.
+            When evaluating this Success Criterion, this is usually not relevant; as long as the applied font,
+            size, and weight ensure that some pixel of each character reaches the author-specified color, evaluators
```
JAWS-test, line 79: It should be carefully considered what exactly is required here: is just one pixel with sufficient contrast in a letter really enough? Imagine the letter x. Due to the slanted lines, the contrasts in the entire letter are often not sufficient, except at the intersection of the two lines. So I have a dot in the middle, but it doesn't help me to recognize the letter itself. If a sensible contrast requirement is to be established here, I would demand that every relevant line of a letter has at least one dot with sufficient contrast at every point, i.e. in the case of the x, for example, both lines over their entire length, except for the serifs.

understanding/21/non-text-contrast.html
WilcoFiers, line 39:
```suggestion
				<p>In these cases it is necessary for authors to choose an technique which guarantees that the user agent will render the content with a contrast ratio that meets the Success Criterion. For example, an author seeking to use a line drawn around a button to meet this Success Criterion might choose to use a CSS <code>border</code> or <code>outline</code> property, rather than a CSS <code>box-shadow</code> property; most user agents will attempt to render fractional pixels of a shadow using fainter colors, but will not use this technique to render fractional pixels of a border or outline.</p>
```

understanding/20/contrast-enhanced.html
WilcoFiers: I don't know how we can be clear about that. There is no defined method for deciding which pixels to compare to which other pixels when measuring contrast, and then how many and by how much those are allowed to follow below the minimum before the text is considered to be failing.

mraccess77: If I recall correctly - the ACT rule is where the one pixel check came from and that rule clarifies that if the check passes further testing may be needed to determine conformance with WCAG SC 1.4.3 which does not specify.  When the AG group talked about this many years ago it was said that all pixels needed to contrast.   I had thrown out a suggestion previously when we were discussing the ACT rule that we might follow the similar principle in 1.4.11 where we require the icon;/letter be identifiable with only contrasting pixels.  That is - remove the non-contrasting pixels and can you still tell what letter it is.  If not - it would not pass- if you can then it would pass.

JAWS-test: One example: I have chosen the font Arial and font size 16 px (this is neither a thin nor a small font). Colours are black (#000) on grey (#757575). Contrast is 4.6:1. 

I have chosen the letters x, y and z. In the two illustrations, the representation is greatly enlarged. In the second illustration, the pixels with sufficient contrast are marked in yellow:
* for x: 3 pixels in the centre and one pixel at the bottom right
* for the y: no pixel
* for z: 5 pixels in the upper and 6 in the lower bar, but no pixel in the horizontal bar. 

The letters cannot be identified using the pixels with sufficient contrast. 

There is also the problem that the pixels with sufficient contrast cannot be recognised with the eye, i.e. if I want to measure the actual contrasts, I have to measure each pixel of each letter individually. 

xyz
xyz_contrast

JAWS-test: And I think every pixel in every letter really needs to be measured, because even with the same letters the contrasts are different depending on their position in the word. The following two images show the letters "by y". In the second illustration, the pixels with sufficient contrast are marked in yellow:
* for b: 2 pixels at the top and 2 pixels at the bottom, 2 pixels in the horizontal bar (so it looks like the letter c)
* for the first y: 3 pixels at the bottom
* for the second y: no pixel

byy
byy_contrast

JAWS-test: I would like to agree with the opinion of @mraccess77: if we measure the pixels, then it only makes sense if all pixels that have sufficient contrast are sufficient to recognise the letter or graphic object. The problem is that this can lead to different results depending on the anti-aliasing settings, which is why the WCAG uses the CSS values and not the actual pixels to calculate the contrasts. For content that is not based on CSS (e.g. pixel graphics), there is no other option than to measure the pixels. But even with pixel graphics, the question would be how I measure: The displayed graphic on the page (which can be enlarged or reduced in size) or the original graphic?

understanding/20/contrast-enhanced.html
```diff
@@ -73,13 +73,19 @@ <h2>Intent of Contrast (Enhanced)</h2>
...
+            may refer to the foreground and background colors obtained from the user agent, or the underlying
```
patrickhlauke, line 80: "may refer" introduces inconsistency in evaluation. some authors/testers will, others won't. and for those that don't, and instead rely solely on pixel-peeping, will have further variation in their findings, because the actual rendered pixels on *their* display will depend on a myriad of different factors:

* their particular display's physical resolution (actual hardware pixels) and logical resolution (what their OS' resolution is) - compare "retina" and "non-retina" displays (paradoxically, this means a tester who uses two monitors - e.g. their laptop's built-in monitor and a nice 8k external monitor set to run at any resolution lower than pure 8k, like 1920x1080 - will get different results when pixel-peeping on one monitor versus the other monitor
* the antialiasing settings on their machine - for instance on Windows, there are various ClearType settings/modes available, which all subtly change the way antialiasing works
* the specific font present on the specific machine - for instance, there are many variations in the wild for even common fonts like Helvetica ... different foundries have subtly different variations, with different hinting/metrics, that can all end up making a difference in how text is rendered on a specific machine; beyond that, there is no guarantee that a particular author, user, tester has the specified font available, and the browser may start to do font substitutions (following the list of fallback fonts specified, which all have again the same issue); similarly, browsers/OSs may also "fake" certain variations, like bold or italic, if there is no explicit font installed on the system that has been designed explicitly as the bold/italic/etc variant
* pixel-peeping will - depending on the platform, and tool used to grab the pixel values - often be influenced by colour management/profiles on the specific machine, which will introduce subtle variations (which, particularly on colour combinations that are "just" on the verge of passing/failing, may make all the difference in shifting the values one way or another)

in short, this still handwaves (with the "may") how this binary pass/fail criterion is to be evaluated. while relying on the colour values as defined in the underlying technology (e.g. CSS) may end up not reflecting the real-world/rendered version as seen on a user's device, it at least provides a much higher likelihood of inter-tester consistency. yes, the underlying defined colour may not reflect the *actual* reality for users in all cases, but that's not new and should be addressed as a strong best practice recommendation instead (to close the gap between "conforms to the letter of the SC" and "is accessible in practice")

understanding/21/non-text-contrast.html
patrickhlauke: as authors have no absolute control over the myriad of possible reasons why something gets antialiased too heavily, i feel very uneasy about making such hard requirements for binary pass/fail here

understanding/20/contrast-enhanced.html
patrickhlauke, line 79: but the problem is that you as an author can't always 100% "ensure" anything...

patrickhlauke: normatively, there's a world of difference between saying "yes, there are ways in which authors can increase their chance that antialiasing is likely not to result in things rendering with too low a contrast" (such as choosing a particular font that, in their testing, generally results in strong enough shapes that don't antialias too much, or setting `font-smoothing` in CSS, or starting off with colours that don't just ride the edge of conformance), and "if authors don't use these techniques, then they normatively fail"

understanding/20/contrast-enhanced.html
```diff
@@ -73,13 +73,19 @@ <h2>Intent of Contrast (Enhanced)</h2>
...
+            markup and stylesheets, rather than evaluating individual pixels within the text as presented on screen.
+         </p>
+         <p>However, particularly thin, small, or unusual fonts may be rendered by user agents with a much fainter
+            color than the actual text color defined in the underlying markup and stylesheets. This can lead to
+            situations where a user agent reports that text has a contrast ratio that nominally passes the Success
+            Criterion, but the text as-rendered has a much lower contrast in practice and fails the Success Criterion.
+         </p>
+         <p>In these cases, best practice would be for authors to choose a font with stronger/thicker lines,
             or to aim for a foreground/background color combination that exceeds the normative requirements
```
dav-idc, line 89:
```suggestion
            or to aim for a foreground and background color combination that exceeds the normative requirements
```

understanding/20/contrast-minimum.html
```diff
@@ -72,13 +72,19 @@ <h2>Intent of Contrast (Minimum)</h2>
...
+            size, and weight ensure that some pixel of each character reaches the author-specified color, evaluators
+            may refer to the foreground and background colors obtained from the user agent, or the underlying
+            markup and stylesheets, rather than evaluating individual pixels within the text as presented on screen.
+         </p>
+         <p>However, particularly thin, small, or unusual fonts may be rendered by user agents with a much fainter
+            color than the actual text color defined in the underlying markup and stylesheets. This can lead to
+            situations where a user agent reports that text has a contrast ratio that nominally passes the Success
+            Criterion, but the text as-rendered has a much lower contrast in practice and fails the Success Criterion.
+         </p>
+         <p>In these cases, best practice would be for authors to choose a font with stronger/thicker lines,
```
dav-idc, line 87:
```suggestion
         <p>In these cases, best practice would be for authors to choose a font with stronger or thicker lines,
```

understanding/20/contrast-enhanced.html
dbjorge: These images are a really interesting case to discuss because they are demonstrations of cases where there is a common, practical, meaningful difference between the **perceived color as rendered on hardware** and **the color of individual pixels as measured by a color picker**. These cases where the left side of a glyph is tinted red and the right side is tinted blue are demonstrating a technique called [Subpixel rendering](https://en.wikipedia.org/wiki/Subpixel_rendering), whereby the text rendering engine is taking advantage of knowing the specific type of screen displaying the text to essentially render pixels of the text "in between" pixels of the screenshot. The user-perceived color in a case like this is going to be closer to the specified color than what a tester could evaluate using a pixel-wise color picker.

Notably, this is *not* the same as anti-aliasing; the note in the contrast ratio definition that says you can ignore anti-aliasing wouldn't apply to this technique.

I think this is a very strong argument against this PR as-written and in favor of allowing testing against specified colors *even if an assessment based on color picking would result in a failure*.

patrickhlauke: > the note in the contrast ratio definition that says you can ignore anti-aliasing wouldn't specifically apply to subpixel rendering.

when that note was written in 2000ish or so, there was no subpixel rendering, i'd hazard to say...

dbjorge:
Per TF discussion, closing this in favor of #3560 + a new issue to separately address adding more guidance about when to use a pixel-picking approach

understanding/20/contrast-enhanced.html
bruce-usab: Thanks @JAWS-test for those examples!

PR #3686, opened 2024-02-12
Update G195 - removal of last line of test (low contrast OK if focus ring > 2px thick)

### LINKED ISSUE #3682: 2.4.7 / 1.4.11 Conflict between 1.4 .11 Understanding text and G159 (author-supplied focus indicator)
Under "Relationship with 2.4.7 Focus visible, the [Understanding text of 1.4.11](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast) says 
> the visual focus indicator for a component must have sufficient contrast against the adjacent background when the component is focused, except where the appearance of the component is determined by the user agent and not modified by the author.

...and "sufficient contrast" means 3:1. There is no reference to the thickness of the indiactor. 

However, the test procedure of Sufficient Technique [G195: Using an author-supplied, visible focus indicator](https://www.w3.org/WAI/WCAG21/Techniques/general/G195). has this as a last step:
> 5. If the focus indicator does not have 3:1 contrast ratio with its adjacent colors, check that it is at least 2px thick.

So that creates uncertainty whether the quite frequent cases of buttons having an outside focus ring several pixels thick but below 3:1 would conform. This is an example:
button-focus^

The light green thick focus ring has a contrast of 1.9:1 to the light grey background. Figure 9 of the 1.4.11 Understanding text seems to support the view that a focus ring with insufficient contrast to the background would not pass. If this is consensus, the test procedure of G159 needs to be revised to take out the last point in the test procedure. I am happy to create a PR for that.

detlevhfischer:
Deletion of last line of test procedure that indicated that contrast below 3:1 is OK if the focus ring is more than 2px thick

Closes https://github.com/w3c/wcag/issues/3682


patrickhlauke (approved):

bruce-usab:
Discussed on TF call 11/8 and moved to Ready for approval after Patrick offered to double check.

mbgower:
@patrickhlauke and @detlevhfischer please see my comments in the issue. I believe we need to decide whether we are to maintain two versions of G195. Either way, we need more changes in the Test (unless we are intentionally exceeding the Non-text Contrast requirements). Also, if we decide _not_ to apply to 2.4.13 in 2.2, we need to make adjustments to the Understanding document to remove the mapping.

mbgower:
Discussed on the TF admin call.
I will do the following:
- rewrite the G195 test to be specific to meet Non-text Contrast
- investigate other techniques that attempt to meet both AA and AAA (e.g., G17 and G18)
- draft a new technique to meet Focus Appearance
- Consider the title and guidance in G195 and the new technique to ensure they are complementary but also distinctive

alastc:
I agree with having two separate techniques:

- G195 is for WCAG 2.1, and we remove the focus-appearance aspects.
- G2xx for WCAG 2.2, which meets the focus-appearance one. Include a note at the top saying something like "This builds on G195 to cover Focus-Appearance". 

This PR could be just for the first part, and separate the second one.

mbgower:
@alastc as per the comments in the issue, I found that if we can do it, it actually makes more sense to extend the test to meet Focus Appearance. If necessary, we can sever the connections from 2.1, but I have now made the test include a check that it meets 1.4.11 requirements, so even though the technique exceeds 1.4.11, it could still stay as being sufficient.

techniques/general/G195.html
```diff
@@ -19,23 +19,25 @@ <h3>Form Elements</h3>
...
+            <ol>
+               <li>is at least as large as a 2 CSS pixel thick perimeter of the unfocused component or sub-component.</li>
+               <li>has a contrast ratio of at least 3:1 between the same pixels in the focused and unfocused states.</li>
+            </ol>
          </ol>
       </section>
       <section class="results"><h3>Expected Results</h3>
          <ul>
-            <li>Checks #2, #4, and #5 are true.</li>
+            <li>If Check #1 is true, all other checks are also true.</li>
```
mbgower, line 40:
```suggestion
            <li>If Check #1 is true, all other checks must also be true.</li>
```

mbgower:
@mbgower to merge the first to test steps into the preamble
For each user interface component on the page that should receive keyboard focus:</p>
         <ol>
            <li>Check that the focus indicator or its background color has been modifed by the author</li>

bruce-usab (approved):

bruce-usab:
Discussed on TF call 12-6, facilitators to discuss on a coordination call.

techniques/general/G195.html
```diff
@@ -2,7 +2,7 @@
       <p>Generally applicable.</p>
    </section><section id="description"><h2>Description</h2>
       <p>The objective of this technique is enhance the focus indicator in the browser, by creating a visible one in the content. The default focus indicator in some browsers is a thin, dotted, black line. It can be difficult to see the line when it is around a form element which already has an outline, when the focused element is inside a table cell, when the focused element is very small, or when the background of the page is a dark color. Some browsers use a pale blue outline, which can be difficult to see on some backgrounds.</p>
```
patrickhlauke, line 4:
```suggestion
      <p>The objective of this technique is to enhance the focus indicator in the browser, by creating a visible one in the content. The default focus indicator in some browsers is a thin, dotted, black line. It can be difficult to see the line when it is around a form element which already has an outline, when the focused element is inside a table cell, when the focused element is very small, or when the background of the page is a dark color. Some browsers use a pale blue outline, which can be difficult to see on some backgrounds.</p>
```

patrickhlauke (approved):

bruce-usab:
Discussed on backlog call 1/17.  Needs another round of editorial review.  Expected Results might mention relationship to other SC.  Patrick volunteered to take another pass.

PR #3694, opened 2024-02-16
remove dependency on spec-prod service

halkk990:


fstrr:
@iadawn Not sure what this is; figuring you might.

PR #3695, opened 2024-02-16
Initial draft update to reflow

### LINKED ISSUE #2119: Clarification on definition of "scrolling in two dimensions" in 1.4.10
Based on my reading of the 1.4.10 Understanding document and some of the issue requests raised here, I'm unclear on whether "scrolling in two dimensions" means:
1. The page should not require users to scroll in two dimensions **at the same time** to read content. Example: When you zoom to 400%, the page continues to have the same dimensions as it does at 100%, requiring the user to pan around, like they would have to do in a (traditionally) zoomed PDF.
**or**
2. The page should not require **any** kind of horizontal scrolling at 400% zoom, outside of the "required for understanding" exceptions, even if the user **doesn't** have to scroll up/down and left/right at the same time. Example: A menu bar in which the user needs to scroll left and right to perceive some of the items, but not up and down.

### LINKED ISSUE #3378: Reflow - unclear requirement leads to wrong tests
When applying the SC 1.4.10 reflow, we experience a lot of disagreement on how it should be tested and which issues are violations looking at the normative text.
This leads to many discussions with customers and test agencies.

Issues:
A) The Success Criterion is unclear when reading the normative text, the intent, the understanding and the test instructions in the techniques.
Customers and test agencies test via zoom to 400%, even regulations established this method as test for Reflow (like BITV 2.0 test instructions, Germany).
It is not understood that the SC is about restricted dimensions and explicitly does not request a text enlargement by 400%. Also it is not understood, that 400% zoom is used in success techniques for verification, yet never in failure techniques to detect issues.

As a consequence:
- instead of restricting virtually one dimension, two dimensions are restricted at the same time, which leads to many more issues found.
- test is often performed on high-end monitors with a page ratio of 16:10 or 16:9 instead of 4:3, which leads to more and different issues being reported

B) The Success Criterion asks that the content can be presented in a width equivalent to 320 CSS px (in case of vertical scrolling content) without the need to scroll in two dimensions.
This way this AA-criterion exceeds SC 1.4.8 Visual Presentation Point 5, which is AAA. This is imbalanced.
Moreover, 1.4.8 talks about "while reading one line of text", which 1.4.10 asks to avoid scrolling in two dimension for the content, which is understood to be the page as a whole. So, also in this way the level AA SC goes beyond the Level AAA SC. 

Therefore I ask to agree on one of the following changes (in the order of preference):
1) turn SC 1.4.10 to AAA to give balance to the WCAG as a whole, and to review the understanding document to give clarity
2) review the understanding document to make clear it is about restricting one dimension (which one depends on the reading direction) and eliminate all references to 400% zoom in the test instructions AND 
make clear that scrolling while perceiving a piece of information shall be avoided, not scrolling the whole page. For example, a content list can be perceived without horizontal scrolling, while scrolling the page is needed to consume the content. Again, consuming a specific line of text or a label-value pair shall not require scrolling in two dimensions, while scrolling might be needed to turn from one text column to the second, for example.
The motivation clearly states that in fact this is intended, not avoiding scrolling of the whole page.
3) turn the applicability to consumption of pure text, exempting all sets of web pages which require interaction beyond navigation AND review the understanding document to give clarity

alastc:
Closes #3378
Closes #2119

### DIFF
--- a/understanding/21/reflow.html
+++ b/understanding/21/reflow.html
@@ -20,7 +20,7 @@
 		</section>
     <section id="intent">
         <h2>Intent of this Success Criterion</h2>
-        <p>The intent of this Success Criterion is to support people with low vision who need to enlarge text and read it in a single column.  When the browser zoom is used to scale content to 400%, it reflows - i.e., it is presented in one column so that scrolling in more than one direction is not necessary.</p>
+        <p>It is important to avoid the need for users to scroll in the direction of reading in order to reveal lines that are cut off by the viewport. Such scrolling significantly increases the effort required to read. It is also important that the act of resizing content (zooming) does not cause some content to not be hidden off-screen. For example, zooming into a vertically scrolling page should not cause content to be hidden to one side due to the page horizontal scrolling.</p>
       <p>For people with low vision, both enlarging and reflowing text are critical to reading. Enlarging text enables the perception of characters. Reflowing text enables users to track from the end of one line to the beginning of the next line.</p>
       <p>Avoiding the need to scroll in the direction of reading in order to reveal lines that are cut off by the viewport is important, because such scrolling significantly increases the effort required to read. It is also important that content is not hidden off-screen. For example, zooming into a vertically scrolling page should not cause content to be hidden to one side.</p>
     	<section>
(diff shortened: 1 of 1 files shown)


alastc:
Notes from the meeting:

- Really need some illustrations for these concepts.
- When talking about column layout, but there is no relationship between columns (e.g. newspaper layout) so it doesn't benefit from the exception, whereas an identical table structure would (due to the relationships).
- Add a tablist example, where it has a relationship with the content but exceeds the width of the page.
- More working going on in the [scratchpad](https://docs.google.com/document/d/1NhXS0Gfdsgl67pIhK0BYmayTq_v45CJMP0Ve36Uui3Q/edit) to clarify this.

PR #3702, opened 2024-02-20
Replace broken link in G56

hidde-logius:
The page this links no longer exists.

### DIFF
--- a/techniques/general/G56.html
+++ b/techniques/general/G56.html
@@ -68,7 +68,8 @@
          <ul>
             <li>#5 is true</li>
          </ul>
-      </section><section id="resources"><h2>Resources</h2>
+      </section>
+    </section><section id="resources"><h2>Resources</h2>
 
          <ul>
             <li>
--- a/techniques/general/G56.html
+++ b/techniques/general/G56.html
@@ -68,5 +68,12 @@
          <ul>
             <li>#5 is true</li>
          </ul>
-      </section>
+      </section><section id="resources"><h2>Resources</h2>
+
+         <ul>
+            <li>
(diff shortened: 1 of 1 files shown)


scottaohara:
the link is still available via the wayback machine - https://web.archive.org/web/20170716020055/http://www.eramp.com/david/audio_contrast_general_techs.htm

just posing the question of should that be linked to instead, or is deleting the link all together fine in this instance?

hidde-logius:
Fair enough, think the wayback URL is more useful than nothing. When I looked, I did not find a different page with similar content either, so seems like this is the best we can do.

techniques/general/G56.html
```diff
@@ -68,13 +68,12 @@ <h3>Insufficient Audio Contrast between a narrator and background music</h3>
          <ul>
             <li>#5 is true</li>
          </ul>
-      </section>
```
kfranqueiro, line 71: It looks like the inner end section tag (corresponding to Expected Results) was removed in this change?

(As far as I can tell, this should not change; there should still be 2 end section tags here, for Expected Results and Tests)

techniques/general/G56.html
```diff
@@ -68,13 +68,12 @@ <h3>Insufficient Audio Contrast between a narrator and background music</h3>
...
-   </section><section id="resources"><h2>Resources</h2>
-      
+      </section><section id="resources"><h2>Resources</h2>
```
patrickhlauke, line 71:
```suggestion
      </section>
    </section><section id="resources"><h2>Resources</h2>
```

techniques/general/G56.html
patrickhlauke: confirmed by looking at this in validator. made the fix

hidde:
Thanks, both!

PR #3703, opened 2024-02-20
Update broken link to working link in G102

hidde-logius:
Found the page that replaced the page we were linking to and added it instead of the broken link.


bruce-usab (approved):
Discussed on TF call 30 Jan.  Noted that SIL example is not great as page has poor contrast, but decided to leave as link rather than deleting link.  If someone wants to look for better examples, that should be a new PR.

PR #3717, opened 2024-02-26
1.4.1 Use of color: adding examples to understanding text

detlevhfischer:
Adding two examples for using color in focusing elements with the keyboard, one for links that have an additional underline, and one for buttons where the background color changes and focused and unfocused background have a contrast difference equal to or better than 3:1.

Background: The Understanding Text of 1.4.11 Non-Text Contrast has Figure 15 where it says: 

> The change of background within the component is not in scope of non-text contrast. However, this would not pass [Use of color](https://www.w3.org/WAI/WCAG22/Understanding/use-of-color.html).

There is so  far no example how 1.4.1 applies to keyboard focus states so this might be helpful.


understanding/20/use-of-color.html
```diff
@@ -167,8 +167,12 @@ <h2>Examples of Use of Color</h2>
                element. A legend shows the color and number for each type of element. Sighted users who
                cannot perceive all the color differences can still understand the image by relying on
                the numbers.</dd>
+         <dt>Focus on links</dt>
+         <dd>A page has text links in blue that have a contrast of 3:1 to the surrounding text. When these links receive keyboard focus, they are shown in a darker color. In addition, they are underlined, so the change of color is not the only indicator of focus.</dd>
```
patrickhlauke, line 171: probably need to clarify how much darker the darker color is - i.e. if the difference between the non-focused blue and focused blue is 3:1 or greater, then the underline isn't necessary since a difference in 3:1 or higher counts as an additional visual indication (and the extra underline wouldn't be needed)

patrickhlauke (commented):
Would be great to provide a small image with each of these to illustrate the point (but yes, I know this is always a bit of a pain to do)

bruce-usab:
@detlevhfischer could you be on the TF call _next_ Friday 2/8 call?

As has been the pattern, probably alastair or Michael will be screen sharing but we would like you (or someone who has reviewed) to walk the group through the PR.

detlevhfischer:
> but i assume what you had in mind here is that the dark blue is not 3:1 against the non-focused blue, right?

I think for a link in blue that has sufficient contrast to a white background AND 3:1 to the surrounding text (assumed to be black here), it would not even be possible to find a color on focus that both has 3:1  or better to the unfocused state AND maintains 3:1 and better to the surrounding text. Showing that this will not work too that some other differentiator (underline, bold or whatever) is needed would be the point for inline links.

detlevhfischer:
@bruce-usab I'll try to be on the Friday call next week.

Kingshittt (approved):

understanding/20/use-of-color.html
```diff
@@ -154,21 +154,34 @@ <h2>Benefits of Use of Color</h2>
...
                cannot perceive all the color differences can still understand the image by relying on
-               the numbers.</dd>
-      </dl>
-      
+               the numbers.</p>
+     
+         <h3>Focus on links</h3>
+         <p>A block of text has a text link in blue that has a contrast of 3:1 to the surrounding text. When this link receives keyboard focus, it turns into a darker blue. The contrast difference between the link color in the focused and the unfocused state is only 2.0:1. In addition, after receiving focus the contrast difference between the link and the surrounding text is now also well below 3:1. To show the keyboard focus clearly, the link is in addition underlined, so the change of color is not the only indicator of keyboard focus. Without an additional focus indicator the example would fail the Success Criterion 1.4.1 Use of Color.</p>
+     <figure id="figure-text-links">
+          <img src="img/textlink.svg" alt="A blue inline text link in a block of black text, before and after receiving keyboard focus" width="500" />
```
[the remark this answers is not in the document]
fstrr: Have reported this spam account.

understanding/20/use-of-color.html
```diff
@@ -154,21 +154,34 @@ <h2>Benefits of Use of Color</h2>
...
+         <p>Students view an SVG image of a chemical compound and identify the chemical elements
                present based <strong>both</strong> on the colors used, as well as numbers next to each
                element. A legend shows the color and number for each type of element. Sighted users who
                cannot perceive all the color differences can still understand the image by relying on
-               the numbers.</dd>
-      </dl>
-      
+               the numbers.</p>
+     
+         <h3>Focus on links</h3>
```
mbgower, line 171: I find the entire description of Focus on links extremely hard to follow, and I do not find the visual examples really help to guide the user to an understanding of whether this is demonstrating a pass or fail situation. This is made more problematic in that some of the items in this new section are passes and some are fails.
If you look at Non-Text Contrast's understanding document, the figure captions explicitly call out whether the example is a pass or fail (in fact, my recollection is we had "Failure example" right at the start of these at one point). The short clear figure captions in https://www.w3.org/WAI/WCAG21/Understanding/text-spacing.html may help guide a better way to approach these.

understanding/20/use-of-color.html
```diff
@@ -154,21 +154,34 @@ <h2>Benefits of Use of Color</h2>
...
+         <p>A block of text has a text link in blue that has a contrast of 3:1 to the surrounding text. When this link receives keyboard focus, it turns into a darker blue. The contrast difference between the link color in the focused and the unfocused state is only 2.0:1. In addition, after receiving focus the contrast difference between the link and the surrounding text is now also well below 3:1. To show the keyboard focus clearly, the link is in addition underlined, so the change of color is not the only indicator of keyboard focus. Without an additional focus indicator the example would fail the Success Criterion 1.4.1 Use of Color.</p>
```
scottaohara, line 172: just some wording suggestions

```suggestion
         <p>A block of text contains a blue hyperlink that has a contrast of 3:1 to the surrounding text. When receiving keyboard focus, the link's text color changes to a darker blue. The contrast difference between the link color in the focused and the unfocused state is only 2:1. In addition, when focused the contrast difference between the link's text color and the surrounding text falls well below a 3:1 ratio. However, to further emphasize the focused state the link is provided an addition underlined. Due to the additional underline, color is not used as the only visual indicator of the focused hyperlink, so it passes Success Criterion 1.4.1 Use of Color. Without the underline, the hyperlink would have failed this Success Criterion.</p>
```

understanding/20/use-of-color.html
mbgower: Since the links one is the only failure example, maybe we should make them all pass examples? We already have a failure technique with links; maybe we should add a visual to that one, or expand it?

understanding/20/use-of-color.html
scottaohara, line 174:
```suggestion
          <img src="img/textlink.svg" alt="Two instances of the same paragraph of text containing a hyperlink. The first instance shows the unfocused hyperlink with blue colored text with no underline. The second instance shows the same hyperlink in the focused state, now with darker blue text and an underline" width="500" />
```

understanding/20/use-of-color.html
```diff
@@ -154,21 +154,34 @@ <h2>Benefits of Use of Color</h2>
...
+     <figure id="figure-text-links">
+          <img src="img/textlink.svg" alt="A blue inline text link in a block of black text, before and after receiving keyboard focus" width="500" />
+				<figcaption>The blue inline text link has a contrast of 5.5:1 to the white background. The link gets darker when receiving keyboard focus and then has a contrast of 10.8:1 to the background. However, the difference of contrast to the link color in unfocused state is only 2.0:1 (i.e., below 3:1). </figcaption></figure>
```
scottaohara, line 175: adjusted the alternative text and the figcaption so that the image alternative text was more descriptive, and the figcaption text was adjusted to remove image description redundancy, and solely focus on the aspects that may not be undrstood by looking at the image alone

```suggestion
				<figcaption>In the default state, the blue hyperlink has a contrast of 5.5:1 to the white background. When focused the hyperlink's darker blue text color has a contrast of 10.8:1 to the background. However, the difference of contrast to the link color in unfocused state is only 2.0:1 (i.e., below 3:1). </figcaption></figure>
```

understanding/20/use-of-color.html
```diff
@@ -154,21 +154,34 @@ <h2>Benefits of Use of Color</h2>
...
+				</section>
+				<section>
+     
+         <h3>Focus on buttons</h3>
+         <p>A page has a row of buttons with white text on a medium blue background. When one of the buttons receives keyboard focus, its background turns black. The difference of contrast between the unfocused and the focused button background is above 3:1. This difference of contrast means that a change of color is not the only indicator of focus.</p>
+            <figure id="figure-buttons">
+          <img src="img/buttons.svg" alt="Three buttons with white text on a blue background. When a button receives keyboard focus, the background turns black" width="400" />
+				<figcaption>A row buttons that all have white text on a mid-blue background with a contrast of 5:1. Keyboard focus turns the background black. The difference of contrast between the focused and the unfocused state of the button background color is 4.2:1.</figcaption></figure>
```
scottaohara, line 183:
```suggestion
				<figcaption>A row buttons that all have white text on a mid-blue background with a contrast of 5:1. The black button represents a button which has received keyboard focus. The difference of contrast between the focused and the unfocused state of the button's background color is 4.2:1.</figcaption></figure>
```

scottaohara (commented):
some suggested wording changes / alt text updates for the added graphics

understanding/20/use-of-color.html
fstrr: Suggestion to clarify the link's visual design:

Current: "A block of text contains a blue hyperlink that has a contrast of 3:1 to the surrounding text."
Proposed: "A block of text contains a non-underlined blue hyperlink that has a contrast of 3:1 to the surrounding text."

Typo: "provided an _addition underlined_".

fstrr:
I must be missing something with this, as [F73](https://www.w3.org/WAI/WCAG22/Techniques/failures/F73), Failure of Success Criterion 1.4.1 due to creating links that are not visually evident without color vision, seems to contradict it.

Here we're saying that a link with no default underline can pass Use Of Color if it gets an underline on hover and/or focus. F73 seems to contradict that:

> While some links may be visually evident from page design and context, such as navigational links, links within text are often visually understood only from their own display attributes. Removing the underline and leaving only the color difference for such links would be a failure because there would be no other visual indication (besides color) that it is a link.

[Example 2 in F73](https://www.w3.org/WAI/WCAG22/Techniques/failures/F73#example-2-removing-the-underline-from-a-link-in-a-sentence-or-paragraph-without-providing-another-visual-cue-besides-color) seems to back that up: an inline link with no underline, but that gets an underline in its hover state, fails.

Am I missing something obvious?

understanding/20/use-of-color.html
scottaohara: thank you @fstrr.  incorporated into my suggested update

scottaohara:
@fstrr the scenario here says that the non-underlined link is a color that passes the 3:1 ratio against the non-hyperlink text.  so that's why it would be a pass.  

maybe it'd be good to call out that even though this is a pass, it is still generally not recommended.

fstrr:
In a conversation with @alastc today, he thought that F73's Example 2 was there to prevent sighted users from having to move their mouse over a screen to find low-contrast links. I've created [a PR](https://github.com/w3c/wcag/pull/3743) to remove example 2 and improve the remaining example, but it seems like we need a discussion on this as it does look like there are conflicting interpretations.

patrickhlauke:
> it seems like we need a discussion on this as it does look like there are conflicting interpretations

unless i'm misunderstanding, @detlevhfischer's suggested example 2 here does acknowledge that the difference in luminosity/contrast is initially sufficient, but then the *focus* state just making the link text a bit darker still won't be enough and falls back as being a "use of color" failure unless there's also something else, like an underline, when focused. so don't think there's a conflict?

understanding/20/use-of-color.html
```diff
@@ -169,18 +169,18 @@ <h3>An examination</h3>
...
             <figure id="figure-buttons">
-          <img src="img/buttons.svg" alt="Three buttons with white text on a blue background. When the button receives keyboard focus, the background turns black" width="400" />
-				<figcaption>Three buttons with white text on a mid-blue background with a contrast of 5:1. Keyboard focus turns the background black. The difference of contrast between the focused and the unfocused state of the background color is 4.2:1.</figcaption></figure>
+          <img src="img/buttons.svg" alt="Three buttons with white text on a blue background. When a button receives keyboard focus, the background turns black" width="400" />
+				<figcaption>A row buttons that all have white text on a mid-blue background with a contrast of 5:1. Keyboard focus turns the background black. The difference of contrast between the focused and the unfocused state of the button background color is 4.2:1.</figcaption></figure>
```
fstrr: Typo: it should be "A row of buttons".

fstrr:
@patrickhlauke Do you mean this example?:

screen grab of the two examples from the proposed update to Use Of Color

The preceding paragraph says:

> "To show the keyboard focus clearly, the link is in addition underlined, so the change of color is not the only indicator of keyboard focus. Without an additional focus indicator the example would fail the Success Criterion 1.4.1 Use of Color."

I assumed that the focus state passes Use Of Color because the underline is a non-color way to identify the link. But maybe I'm wrong as it also says <q>Without an additional focus indicator the example would fail…</q>. Does that mean an additional focus indicator on top of the underline?

Or does that paragraph not relate to the images? If it doesn't, then the underline in the second example is really confusing and, I think, be removed.

I put that example and the example from my PR to update F73 into a [CodePen](https://codepen.io/fstrr/pen/RwOaXXB), if that's of any use.

understanding/20/use-of-color.html
```diff
@@ -154,21 +154,34 @@ <h2>Benefits of Use of Color</h2>
...
+				</section>
+				<section>
+     
+         <h3>Focus on buttons</h3>
+         <p>A page has a row of buttons with white text on a medium blue background. When one of the buttons receives keyboard focus, its background turns black. The difference of contrast between the unfocused and the focused button background is above 3:1. This difference of contrast means that a change of color is not the only indicator of focus.</p>
+            <figure id="figure-buttons">
+          <img src="img/buttons.svg" alt="Three buttons with white text on a blue background. When a button receives keyboard focus, the background turns black" width="400" />
+				<figcaption>A row buttons that all have white text on a mid-blue background with a contrast of 5:1. The black button represents a button which has received keyboard focus. The difference of contrast between the focused and the unfocused state of the button's background color is 4.2:1.</figcaption></figure>
+
```
u9000-Nine, line 184:
```suggestion
     
     <h3>Link state</h3>
     <p>Links on a page appear differently when a user has visited them before. Unvisited links appear blue with a solid underline, and visited links appear purple with a dashed underline. The difference in underline types means that color is not used as the only visual indicator of links' states.</p>
```

Could we add something about distinguishing link states by color? I find this is often overlooked by designers and auditors, so mentioning it explicitly would be helpful.

If it's too late in the review to add this I completely understand. If it could be added, though, I really think it would be helpful.

u9000-Nine (approved):

understanding/20/use-of-color.html
u9000-Nine: PS: Not sure what the process is for suggesting non-blocking changes to commits, so I approved the whole PR to indicate that. Please let me know if I should have done it a different way.

mraccess77: FWIW In the past this group has not addressed visited and unvisited state color use because of the problems of trying to get 3 colors that all contrast sufficiently with each other.  Historically folks have not flagged issues comparing visited and unvisited states because browsers have some data on that and many sites today override that information.

alastc:
@patrickhlauke to make a suggestion. Also, please add the editors.css link so the preview works much better.

patrickhlauke:
@detlevhfischer discussed this at today's backlog meeting - one of the points that wasn't clear for @fstrr was whether or not that underline in the image was meant to somehow highlight the dark blue, or it was meant to be an actual underline as part of the link styles ... the fact it was black rather than dark blue seemed to be at the core of the confusion. I've tweaked the SVG accordingly, and expanded the figcaption just to make it extra clear (also, as part of that, noted that there's a second reason for the underline ... that the dark blue alone also lacks contrast against the black static text otherwise) https://github.com/w3c/wcag/pull/3747

If you're happy with my patch (didn't want to just brutally merge it in your PR here without letting you know first), we can merge my PR into your PR, before then merging your PR into main /cc @alastc

understanding/20/use-of-color.html
```diff
@@ -154,21 +154,34 @@ <h2>Benefits of Use of Color</h2>
...
+				</section>
+				<section>
+     
+         <h3>Focus on buttons</h3>
+         <p>A page has a row of buttons with white text on a medium blue background. When one of the buttons receives keyboard focus, its background turns black. The difference of contrast between the unfocused and the focused button background is above 3:1. This difference of contrast means that a change of color is not the only indicator of focus.</p>
+            <figure id="figure-buttons">
+          <img src="img/buttons.svg" alt="Three buttons with white text on a blue background. When a button receives keyboard focus, the background turns black" width="400" />
+				<figcaption>A row buttons that all have white text on a mid-blue background with a contrast of 5:1. The black button represents a button which has received keyboard focus. The difference of contrast between the focused and the unfocused state of the button's background color is 4.2:1.</figcaption></figure>
```
patrickhlauke, line 183:
```suggestion
				<figcaption>A row of buttons that all have white text on a mid-blue background with a contrast of 5:1. The black button represents a button which has received keyboard focus. The difference of contrast between the focused and the unfocused state of the button's background color is 4.2:1.</figcaption></figure>
```

patrickhlauke:
Swapped out base.css for editors.css for better previewing

patrickhlauke:
Rendered preview https://raw.githack.com/w3c/wcag/1.4.1-examples-added/understanding/20/use-of-color.html

detlevhfischer:
Nice, thanks for all the good edits! In a separate step, it's perhaps a good idea to still add a paragraph undo "Intent of Use of Color"  so the example chimes with something explained before, as @u9000-Nine has hinted at.

Dubersnack (approved):

fstrr:
What I think would be useful would be to have the Focus On Links image match the ratios in the preceeding paragraph. I found that having two almost identical examples, one in text and one as an image, next to each other a little confusing.

In last week's backlog meeting, I think we discussed adding a note stating that if the example's underline was under 3:1 contrast, that would be a Non-Text Contrast issue. If I'm remembering that correctly, adding that would also be useful.

detlevhfischer:
@fstrr Does my latest change of contrast values in the text preceding the illustration of inline link in default and focused state address your concerns?

fstrr:
@detlevhfischer Thanks for this. A couple of things:

1. Is the initial contrast actually 3.83:1 and not just 3.8:1?
2. Would changing the underline note to say this be clearer?:

> Because the focus state's text color has a contrast of less than 3:1 with the surrounding non-link text, the underline must have a contrast of at least 3:1 otherwise the underline would fail Success Criterion 1.4.11 Non-Text Contrast

And, as it's a note, should it be marked up as a note?

Thanks :)

understanding/20/use-of-color.html
mbgower:
```suggestion
         <p>A block of text has a text link in blue that has a contrast of 3:1 to the surrounding text. When this link receives keyboard focus, it turns into a darker blue. The contrast difference between the link color in the focused and the unfocused state is only 2.0:1. In addition, after receiving focus the contrast difference between the link and the surrounding text is now also well below 3:1. To show the keyboard focus clearly, the link is in addition underlined, so the change of color is not the only indicator of keyboard focus. Without this additional focus indicator the example would fail the Success Criterion 1.4.1 Use of Color.</p>
```

patrickhlauke:
Looking at this again on today's call, the "Focus on links" is problematic...

alastc:
It needs an update:
1. It should note the example fails focus-appearance.
2. It would be useful to include a fail example for link text without a underline. That's the common case, and skimming the document it wasn't as clear as this niche case.

patrickhlauke:
I'm going to work some more on this PR (hope that's ok, @detlevhfischer). My initial thinking is:

* it could benefit from a note/addition to the understanding prose (before the examples), so mention how "Use of Color" can also apply in situations relating to focus and on/off toggle states
* make an initial example that just shows the more common "links without underline that rely on color to be distinguishable from regular text"
* tweak Detlev's example of the focus state for the link to become a follow-up to the previous example - might also see if we can do it so that it doesn't fail focus appearance
* make another example that talks about a series of two/three buttons, one of which is toggled/pressed, but only uses a subtle difference in background colour to visually show that it's toggled/pressed

patrickhlauke:
Resolved merge conflicts. Latest preview https://deploy-preview-3717--wcag2.netlify.app/understanding/use-of-color

patrickhlauke:
@kfranqueiro wondering if we should prevent spellcheck from going into `.svg` files https://github.com/w3c/wcag/actions/runs/20512324433/job/58934824964?pr=3717

PR #3734, opened 2024-03-07
Update ARIA12

### LINKED ISSUE #3733: ARIA 12 needs updating to support the ARIA 1.2 requirement that aria-level must be specified.
The ARIA 1.2 specification for [role="heading"](https://www.w3.org/TR/wai-aria/#heading) requires that an [aria-level](https://www.w3.org/TR/wai-aria-1.3/#aria-level) property is also used, which is a change from the ARIA 1.1 spec, where a missing aria-level is treated as a second-level heading (an `h2`).

The [WCAG 2.2 version of ARIA 12](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA12) doesn't mention this requirement in its examples, so needs updating.

The [draft ARIA 1.3 spec for aria-level](https://www.w3.org/TR/wai-aria-1.3/#aria-level) includes a useful note around aria-level values of 10 and higher, which would be good for WCAG to also mention.

fstrr:
Closes #3733 

- Specify that `aria-level` must be used.
- Update examples to use `aria-level`.
- Remove link to APG as the referenced content isn’t there anymore.
- Update test procedure for `aria-level`.
- Add note about `aria-level` values of 10 and greater.
- Add link to `aria-level` spec.
- Add link to HTML headings spec.


techniques/aria/ARIA12.html
```diff
@@ -35,39 +43,53 @@ <h3>Additional heading levels</h3>
    ...
 &lt;div role="heading" aria-level="7"&gt;Jonagold/div&gt;
```
PhilippeVay, line 44: There's a small typo in this example with the lack of `<` in closing tag (it's been err... _not_ there since WCAG 2.0)

patrickhlauke:
```suggestion
&lt;div role="heading" aria-level="7"&gt;Jonagold&lt;/div&gt;
```

fstrr: I think we made that fix at just about the exact same time. The "commit suggestion" button is grayed out, so I can't apply your fix, so I'm going to resolve it.

patrickhlauke (approved):

techniques/aria/ARIA12.html
```diff
@@ -33,41 +41,55 @@ <h3>Additional heading levels</h3>
...
-         <ol>
-            <li> Examine each element with the attribute <code class="language-html">role="heading"</code>.
-</li>
-            <li> Determine whether the content of the element is appropriate as a heading.
-</li>
-            <li> If the element has an <code class="language-html">aria-level</code> attribute, determine whether the value is the appropriate hierarchical level.
-</li>
-         </ol>
+	<aside class="note">
+		<p>Headings with an <code class="language-html">aria-level</code> of 10 or higher can create difficulties for users. At the time of writing, most combinations of user agents and assistive technologies report such levels as being level 2.</p>
```
mbgower, line 47: That seems really odd to me. Is there a reference?

fstrr: I mentioned it in the issue—I found it in the [ARIA 1.3 draft](https://www.w3.org/TR/wai-aria-1.3/#aria-level) and tweaked it a bit.

alastc, line 47:
```suggestion
		<p>Headings with an <code class="language-html">aria-level</code> of 10 or higher can create difficulties for users. At the time of writing (2024), most combinations of user agents and assistive technologies report such levels as being level 2.</p>
```

bruce-usab:
Discussed on TF call 4/12 and Ready for approval

techniques/aria/ARIA12.html
```diff
@@ -1,29 +1,37 @@
...
-</p>
-<pre xml:space="preserve"><code class="language-html">&lt;div role="heading"&gt;Global News items&lt;/div&gt;
+a piece of content as a heading. Applying <code class="language-html">role="heading"</code> to an element causes an AT (like a screen reader) to treat it as though it were a heading. The <code class="language-html">role="heading"</code> property must be paired with the <code class="language-html">aria-level</code> property to define the heading level.</p>
+      <p>When possible, <a href="https://www.w3.org/TR/using-aria/#rule1">use native heading mark-up</a>. For example, it is preferable to use <code class="language-html">h1</code> rather than using <code class="language-html">&lt;div role="heading" aria-level="1"&gt;</code>. However, the use of the heading role, instead of heading <abbr title="HyperText Markup Language">HTML</abbr>, may be necessary.</p>
```
scottaohara, line 22: change 'mark-up' to match the 'markup' used in the `title` for the `<abbr>HTML</abbr>`.

I realize the last sentence of this was a bit of a rejiggering of what the previous doc was saying. But it struck me a bit odd to end this saying to use the heading role when necessary, but not actually outline when or why it could be necessary.  The fact that the APG content no longer exists emphasizes the ambiguity of when it might be necessary.

So, rather than that, i wrote up a bit to explain why / double down on why it's better to use the native HTML element instead.  If people think this is "too much", then we do not need to merge in this suggestion.  I'd just rather see there be a bit more explanation, again, particularly due to the fact the APG resource is no longer available - so it may be harder for some to determine when it's necessary, or not. 

```suggestion
      <p>When possible, <a href="https://www.w3.org/TR/using-aria/#rule1">use native heading markup</a>. For example, it is preferable to use an <code class="language-html">h1</code> element, rather than using <code class="language-html">&lt;div role="heading" aria-level="1"&gt;</code>. Native <abbr title="HyperText Markup Language">HTML</abbr> elements have implicit styling and features that won't necessarily be replicated with ARIA equivalents. For instance, heading elements will have specific styling adjustments made when viewing a page in a browser's Reader Mode. Whereas a <code class="language-html">&lt;div role="heading" aria-level="1"&gt;</code> will not be styled as a heading, because the underlying element is a <code class="language-html">div</code>. ARIA only modifies the way an element is exposed to assistive technology, it does nothing to change the implicit style or functionality of the element it is applied to.</p>
```

patrickhlauke:
Note the additional change that was merged into this https://github.com/w3c/wcag/pull/3734/commits/4c06779f241ee04db5465700f8eb8284de44ea8b

scottaohara (approved):

PR #3588, opened 2023-12-05
Update general understanding documents

patrickhlauke:
* removes the woefully out-of-date (and likely never going to be maintained) "Change log" from the https://www.w3.org/WAI/WCAG22/Understanding/about page
* where possible, makes WCAG mentions "versionless" ("WCAG 2" instead of "WCAG 2.0" / "WCAG 2.1" / "WCAG 2.2")
* <s>where not possible (e.g. links to techniques document), bumps to 2.2</s> EDIT: made techniques document link versionless as well
* <s>change `success criterion`/`success criteria` to `Success Criterion`/`Success Criteria` (and correct minor instances where the latter was used as if it were singular - "is the Success Criteria" > "are the Success Criteria" or "is the Success Criterion")</s> EDIT: as a result of discussions that came after this PR was made, reverting this change, and changing any straggler references that did use uppercase `Success Criteria` to lowercase `success criteria`
* adds `lang` attributes to all documents
* updates some of the links (most prominent: replace the outdated https://www.w3.org/WAI/GL/WCAG20/TECHS-SUBMIT/ to instead point to the github repo)

Closes #3585

Note that this change is based on existing non-normative information under [Background on WCAG 2](https://www.w3.org/TR/WCAG22/#background-on-wcag-2) in the existing TR, which reads in part:
>Where this document refers to WCAG 2 it is intended to mean any and all versions of WCAG that start with 2.

PR #3481, opened 2023-10-13
Broken link fixes to 2.2 documents

fstrr:
- fixed links to Accessible Auth (Minimum) in Accessible Auth (Enhanced)
- Updated understanding Understanding:
  - fix all in-page links
  - replace references to WCAG 2.1 with WCAG 2.2
- fix links to Accessible Auth criteria in Redundant Entry 
- fix broken links in techniques → about. (side note: that list hasn't been updated in over three years)
- fix links to 2.2 criteria

PR #3738, opened 2024-03-09
Chore: update/correct all links in understanding and techniques docs

patrickhlauke:
Currently the understanding and techniques docs are littered with absolute URLs, many of which still point to WCAG 2.1 or 2.0 versions of things. While I suspect there's probably some XSLT magic that tries to correct these before publication (?), it makes more sense to keep the sources clean here.

As a bonus, this PR corrects a broken link to the definition for "cognitive function test" in the accessible authentication (minimum) understanding document.


patrickhlauke:
Contrary to what I originally said, I did in fact set up the local build environment, and the build processes went through without any errors. However, it seems that even once you build it locally, links between locally built documents still don't work correctly (not just for the links I edited here, but throughout), so it's difficult to verify if these really work out correctly. maybe @iadawn has some idea/hint?

techniques/about.html
```diff
@@ -85,52 +85,52 @@ <h2>Change Log</h2>
 
                         <li>
 
-                            <time datetime="2019-09-12">Sept 12th 2019</time>: Added <a href="./techniques/failures/F105">Failure of Success Criterion 2.5.1 due to providing functionality via a path-based
+                            <time datetime="2019-09-12">Sept 12th 2019</time>: Added <a href="./Techniques/failures/F105">Failure of Success Criterion 2.5.1 due to providing functionality via a path-based
```
dbjorge, line 88: I think these still look off - this containing doc is already within the `/Techniques/` path

dbjorge (changes requested):

techniques/about.html
patrickhlauke: ah, good catch @dbjorge - indeed, I just did a find/replace for the uppercase `T`, but it seems these are broken even further even in the live version at the moment https://www.w3.org/WAI/WCAG22/Techniques/about

will push an update (for the about document only, which i think is the only one that's still broken this way)

patrickhlauke: actually...wondering if we want to kill of this (non-maintained) list of "new and updated techniques" and just reference github's change log instead

patrickhlauke: doh, this smelled familiar ... already did this exact change (removing the non-maintained list) over here https://github.com/w3c/wcag/pull/3588

patrickhlauke: re-applying #3588 here (i hadn't merged in the changes we did there before, done now), this circumvents/closes this problem outright

fstrr:
> However, it seems that even once you build it locally, links between locally built documents still don't work correctly (not just for the links I edited here, but throughout), so it's difficult to verify if these really work out correctly. maybe @iadawn has some idea/hint?

Indeed. I'm working locally, but have a branch where I've merged in my [PR 3481](https://github.com/w3c/wcag/pull/3481) with this branch, and we appear to have two different link URL styles. To take [SVR3](https://www.w3.org/WAI/WCAG22/Techniques/server-side-script/SVR3) as an example:

My code:
```html
<p><a href="/WAI/WCAG22/Understanding/conformance#cc1">Conformance Requirement 1</a> allows non-conforming pages to be included within the scope of conformance as long as they have a "<a href="/TR/WCAG/#dfn-conforming-alternate-versions">conforming alternate version</a>".
```

Patrick's code:
```html
<a href="../../Understanding/conformance#conf-req1">Conformance Requirement 1</a> allows non-conforming pages to be included within the scope of conformance as long as they have a "<a href="https://www.w3.org/TR/WCAG21/#dfn-conforming-alternate-version">conforming alternate version</a>".
```

Editing the URLs for the Conformance Requirement 1 link in dev tools results in both my root-relative link and Patrick's page-relative link going to the same [Understanding Conformance page](https://www.w3.org/WAI/WCAG22/Understanding/conformance).

Is there a preferred way to do this? There are 25 merge conflicts between this PR and mine. I'll happily resolve them if @iadawn can say which is the better style of link.

alastc:
I think the relative links (i.e. `.../folder/`) should work better, the `/WAI/WCAG22/` will break things for WCAG 2.1 if we back-port updates to there.

The local one will work if it finished with `.html`. That isn't needed when live as the server treats `document` and `document.html` the same, but your file system doesn't.

If it doesn't have `.html` I wouldn't bother adding it, but if you want to test links locally then you'll need to.

fstrr:
@patrickhlauke Hey.

Looking at [ARIA20](https://github.com/w3c/wcag/pull/3738/files#diff-52d12ab1cbc55a5a459af65ea3ba623dcd8789131e59b684c1ef2008a0fd6bba), you've changed the link on line 4 to ARIA11 to a relative one, but you've not updated the three Related Techniques links later on in the document. Those are all relative (`a href="ARIA11">ARIA11</a>`) and work, but I can go through and update those for completeness if you want.

alastc:
replaced, not lost.

patrickhlauke:
> replaced, not lost.

so, to make this less cryptic...

"superseded by https://github.com/w3c/wcag/pull/3757"

PR #3739, opened 2024-03-10
Updates test steps in F94

patrickhlauke:
- Removes 1280 pixels wide step
- Removes the unnecessary preamble before the test steps, and tweaks the last line of the test procedure to reference the last step, rather than giving it a number
- Additionally, cleans up formatting of the HTML

Closes https://github.com/w3c/wcag/issues/704


alastc:
I was thinking that it should start the procedure with 

> For each page variation (such as break points in media queries) use any of the following methods to resize text when available:

But, given the context of a failure technique about view-port units, is that needed?

patrickhlauke:
> given the context of a failure technique about view-port units, is that needed?

only IF there is a hint about the site that fails using breakpoints with different styles for font sizing, and only one particular breakpoint switching to viewport units. which there isn't in the way the technique is written, so long way of saying ... no?

bruce-usab (approved):

patrickhlauke:
As discussed last week, filed a follow-up https://github.com/w3c/wcag/pull/3876

techniques/failures/F94.html
```diff
@@ -59,7 +57,7 @@ <h3>Procedure</h3>
 			<section class="test-results">
 				<h3>Expected Results</h3>
 				<ul>
-					<li>If step #5 is false, then this failure condition applies and the content fails Success Criteria 1.4.4, Resize Text.</li>
+					<li>If step #4 is false, then this failure condition applies and the content fails Success Criteria 1.4.4, Resize Text.</li>
```
gundulaniemann, line 60: I suggest "Success Criterion 1.4.4 Resize Text." (singular and without comma)

patrickhlauke:
```suggestion
					<li>If step #4 is false, then this failure condition applies and the content fails Success Criterion 1.4.4 Resize Text.</li>
```

mbgower:
@patrickhlauke the follow-up doesn't meaningfully change this PR, right? Is there any reason we wouldn't just incorporate here?

techniques/failures/F94.html
```diff
@@ -59,7 +57,7 @@ <h3>Procedure</h3>
 			<section class="test-results">
 				<h3>Expected Results</h3>
 				<ul>
-					<li>If step #5 is false, then this failure condition applies and the content fails Success Criteria 1.4.4, Resize Text.</li>
+					<li>If step #4 is false, then this failure condition applies and the content fails Success Criterion 1.4.4 Resize Text.</li>
```
mbgower, line 60:
```suggestion
					<li>If the last step is false, then this failure condition applies and the content fails Success Criterion 1.4.4 Resize Text.</li>
```

techniques/failures/F94.html
```diff
@@ -44,11 +43,10 @@ <h3>Procedure</h3>
 				<p>In a desktop browser with a resizable window:</p>
```
mbgower, line 43:
```suggestion
```

techniques/failures/F94.html
```diff
@@ -44,11 +43,10 @@ <h3>Procedure</h3>
...
 				<ol>
 					<li>Set zoom level to 100%.</li>
```
mbgower, line 45:
```suggestion
```

techniques/failures/F94.html
```diff
@@ -59,7 +55,7 @@ <h3>Procedure</h3>
 			<section class="test-results">
 				<h3>Expected Results</h3>
 				<ul>
-					<li>If step #5 is false, then this failure condition applies and the content fails Success Criteria 1.4.4, Resize Text.</li>
+					<li>If the last step is false, then this failure condition applies and the content fails Success Criterion 1.4.4 Resize Text.</li>
```
detlevhfischer, line 58: Can a step be false? You just take a  step. Would you not rather say, "if the last check is false" or "if the result of the last check is false"? Maybre I am overthinkign it...

detlevhfischer (commented):
See inline comment.

techniques/failures/F94.html
patrickhlauke: while I agree it sounds odd, this wording ("If step X is false" / "If X is false") is used pervasively throughout the techniques (as well as the more correct ("If check X is false"). If we want to tackle this, should probably do it throughout *all* techniques in a separate PR

gundulaniemann: I understand the original wording better.
Nevertheless, if I see correctly none of us 3 is a native speaker.
What do native speakers say?

patrickhlauke: i'd say tackle this in a separate, specific PR that covers *all* techniques, rather than trying to hash it out here in this specific one

patrickhlauke: https://github.com/w3c/wcag/pull/3915

gundulaniemann: agree

patrickhlauke: "die drei deutschsprachigen hier sind sich einig" :)

PR #3741, opened 2024-03-11
Errata from 3536

alastc:
Replicating the change from #3536


alastc:
Arg, merging with wrong branch.

PR #3743, opened 2024-03-11
Update F73 to remove example 2

fstrr:
This PR removes example 2, “Removing the underline from a link in a sentence or paragraph without providing another visual cue besides color”, which contradicts guidance in G183. This issue came up in #3717.


techniques/failures/F73.html
```diff
@@ -4,54 +4,57 @@
...
       <div class="note">
-         <p>Red and Pink are the same color (hue) but they have different lightness (which is not color ).   So red and pink would pass the requirement for "not distinguished by color (hue) alone" since they differ by lightness (which is not color) - as long as the difference in lightness (contrast) is 3:1 or greater. For example, if surrounding text is RED and the link is PINK it would pass.  Similarly a light green and a dark red differ BOTH by color AND by lightness so they would pass if the contrast (lightness) difference is 3:1 or greater before focus or pointing.</p>
+         <p>Red and pink are the same color (hue) but they have different lightness (which is not color). So red and pink would pass the requirement for "not distinguished by color (hue) alone" since they differ by lightness (which is not color) - as long as the difference in lightness (contrast) is 3:1 or greater. For example, if surrounding text is red and the link is pink it would pass. Similarly a light green and a dark red differ BOTH by color AND by lightness so they would pass if the contrast (lightness) difference is 3:1 or greater before focus or pointing.</p>
```
patrickhlauke, line 7:
```suggestion
         <p>Red and pink are the same color (hue) but they have different lightness (which is not color). So red and pink would pass the requirement for "not distinguished by color (hue) alone" since they differ by lightness (which is not color) - as long as the difference in lightness (contrast) is 3:1 or greater. For example, if surrounding text is red and the link is pink it would pass. Similarly a light green and a dark red differ <strong>both</strong> by color <strong>and</strong> by lightness so they would pass if the contrast (lightness) difference is 3:1 or greater before focus or pointing.</p>
```

patrickhlauke (approved):

patrickhlauke (changes requested):
I may have got the wrong end of the stick when I approved this ... will have another look at this

mbgower:
The TF agree that example 2 _should_ remain since it is a failure.

PR #3751, opened 2024-03-19
Update W3C staff contact from Michael to Kevin

fstrr:
Replace Michael Cooper with Kevin White

### DIFF
--- a/xslt/base.xslt
+++ b/xslt/base.xslt
@@ -353,7 +353,7 @@
 	<xsl:template name="wai-site-footer">
 		<footer id="wai-site-footer" class="page-footer default-grid" aria-label="Page">
     	<div class="inner" style="grid-column: 2 / 8">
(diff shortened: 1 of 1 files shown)


patrickhlauke (approved):

patrickhlauke (approved):

mbgower:
@GreggVan stated the following in an email response:
> I don’t think we need to change WCAG itself.  That is meant to be  the status when published (or sent for publication) I would not change this in WCAG — just on the website.
and Michael email can be forwarded to Kevin if it comes in.

patrickhlauke:
> If the primary purpose of this xslt is to alter this citation

it's not though, this xslt doesn't affect the normative spec document itself. only the footer in understanding etc. Which, strangely, already list Kevin there (see for instance the footer on https://www.w3.org/WAI/WCAG22/Techniques/about), so looks like this xslt may be out of sync with reality?

iadawn:
Gah I really need to check PRs before I 'fix' little things. In this case I spotted the discrepancy last week and 'fixed' it.

patrickhlauke:
Confirming https://github.com/w3c/wcag/blob/main/xslt/base.xslt is updated in the repo, so this PR is now unnecessary (annoying that git/github doesn't automagically update itself here, but i get why that is ... would be great if there was a handy "auto-rebase/merge" that would surface these sorts of things).

Closing.

Issue #1139, opened 2020-06-01
Flash1: Graphic showing how to create accessible name in Flash has no accessible name


dd8:
Figure 1 in technique Flash 1 has this code:

```
<figure id="figure-flash-authoring-accessibility-panel">
 <img src="img/acc_name_scrn.jpg">			       
 <figcaption>Figure 1The Accessibility panel in the Flash authoring environment.</figcaption>
</figure>
```

https://www.w3.org/WAI/WCAG21/Techniques/flash/FLASH1.html

Chrome and Firefox a11y inspectors show graphic as having null name. VoiceOver with Safari reads "acc underscore name underscore scrn dot jpg, image"

dd8:
The same is also true of Flash2, Flash21, Flash28, Flash29, Flash32, Flash5, Flash6, PDF1, PDF10, PDF11, PDF12, PDF17, PDF20, PDF4, PDF6, PDF9, SL20, SL28, SL8.

Looks like these all assume figcaption provides an accessible name

alastc:
I think the Flash ones will be deprecated soon, but the PDF ones are likely to stick around.

dd8:
The problem seems worse than I first reported, most of the `img` elements in the WCAG 2.1 techniques don't have `alt` so don't have an accessible name, but the equivalent WCAG 2.0 technique do have `alt` on `img`

For example:

https://www.w3.org/TR/WCAG20-TECHS/SCR32.html (has alt)
https://www.w3.org/WAI/WCAG21/Techniques/client-side-script/SCR32 (no alt)

https://www.w3.org/TR/WCAG20-TECHS/G167.html (has alt)
https://www.w3.org/WAI/WCAG21/Techniques/general/G167 (no alt)

https://www.w3.org/TR/WCAG20-TECHS/F69.html (has alt)
https://www.w3.org/WAI/WCAG21/Techniques/failures/F69 (no alt)

From searching the source on Github it looks like there's no `alt` in the majority of places where `img` is used.

fstrr:
Closing due to the removal of Flash techniques

dd8:
I think this should be re-opened - it's not just about the Flash techniques - it's most/all of the WCAG 2.1 techniques with `<img>` elements

https://github.com/w3c/wcag/issues/1139#issuecomment-663854412

fstrr:
[...]

PR #3762, opened 2024-03-26
Add alt text to images

fstrr:
Eight images in WCAG 2.2 were missing `alt` attributes. This PR adds those in.

Ref: #1139


techniques/general/G167.html
```diff
@@ -11,10 +11,10 @@
          <h3>A search function</h3>
          
             <p>A Web page contains a text field where the user can enter search terms and a button labeled "Search" for performing the search. The button is positioned right after the text field so that it is clear to the user that the text field is where to enter the search term.</p>
-            <figure id="figure-button-label-field">
-            	<img src="img/button-as-label.jpg"/>
-            	<figcaption>A text field with a button positioned to the right demonstrating the use of a button to label a field.</figcaption>
-            </figure>
+            <div>
+            	<img alt="A text field with a button positioned to the right demonstrating the use of a button to label a field." src="img/button-as-label.jpg"/>
+            	<figcaption></figcaption>
```
scottaohara, line 16:
```suggestion
```

scottaohara (approved):
pending the remnant figcaption which is no longer needed, since the parent was changed to a div, this otherwise looks good!

fstrr:
> pending the remnant figcaption which is no longer needed, since the parent was changed to a div, this otherwise looks good!

Good catch. I thought I'd caught all of those. Thanks.

Issue #3755, opened 2024-03-20
Success Criterion 1.4.5 Images of Text - Image of text + visual real text that matches the content of the image


giacomo-petri:
In terms of normative requirements, the 1.4.5 Success Criterion is notably clear, as are its exceptions:
- Customizable: The image of text can be visually adjusted to meet the user's preferences.
- Essential: The specific presentation of text is crucial for conveying the information.

Therefore, if the image of text does not fall under one of these exceptions, it is deemed non-compliant.

A common scenario observed frequently involves an image featuring text within the image (image of text) followed by the same text presented in a textual format. Despite the ability to customize the real text (presented in textual format) to suit user preferences, the image of text remains uncustomizable, thus failing to meet the Success Criterion.

I will illustrate two examples below:

Example 1:

<img width="1209" alt="image of text coexists with real text mathing the content of the image" src="https://github.com/w3c/wcag/assets/106394951/227e128c-c7fb-4a9b-a829-f8923569787c">

This instance depicts a background image showcasing T-shirts with a "30% OFF Promo Message" as image of text on the right side, while the left side features the same message presented as real text along with additional description and a "Shop Now" link.

Example 2:

<img width="751" alt="static image of text coexists with real text mathing the content of the image" src="https://github.com/w3c/wcag/assets/106394951/f2973584-5ded-423b-a29e-3b012e32fe6f">

The second example presents non-interactive content containing solely an image of text, accompanied by the same text provided in a textual format akin to a caption below the image.

My sentiments regarding this issue fluctuate:
- On one hand, comprehension of the information within the image is necessary to recognize its correspondence with the text element, indicating the adequacy of the Success Criterion.
- Conversely, individuals unable to visually perceive the image, such as those reliant on screen readers, must depend on alternative text. Likewise, low vision users who may struggle to discern text within the image should also rely on alternatives provided by the author. This suggests a need for a more permissive SC. This holds particularly true when considering that 1.4.5 is an AA requirement, contrasting with its corresponding AAA requirement (1.4.9).

Therefore, ultimately, I am inclined to approve both examples despite their non-compliance with the standards.

Additionally, the sufficient technique C30 (C30: Using CSS to replace text with images of text and providing user interface controls to switch) enables authors to incorporate a switch button to replace the image of text with actual text. However, I find this approach more cumbersome and intricate compared to providing the information directly in textual form adjacent to the image of text.
[...]

PR #3773, opened 2024-04-04
Added "Duplicated text" as guidance to fulfil the 1.4.5 Images of Text success criterion in the Understanding Document

giacomo-petri:
Closes: #3755 

This is a proposal to include "duplicated text" as a sufficient technique to satisfy the 1.4.5 Success Criterion, addressing users needs to personalize text presentation while also ensuring sustainability for companies (as entirely replacing text images with real text often results in inaction).

Updates:
- Included a section in the understanding document to consider this approach.
- Provided an example.
- Introduced a sufficient technique.


patrickhlauke:
@stevefaulkner asked about clarification if the technique implies the visible alternative text needs to *always* be visible, or if would be good to also show an example where the text is in a disclosure widget or similar. would be good to provide a separate example with this

dbjorge:
I don't think I'm on board with this. I agree that this likely *should* be an exception to 1.4.5, but I don't think it *is* an exception today; I think if we agree that this should be an exception, we should be exploring an erratum, not adding a technique that tries to define an exception where none exists.

patrickhlauke:
Additional text is an accessible alternative, in effect. Don't think it becomes an exception per se, more an example

patrickhlauke:
i.e. this is a case of

> alternatives to part of a page's content are considered part of the page when the alternatives can be obtained directly from the page, e.g., a long description or an alternative presentation of a video.

https://www.w3.org/TR/WCAG22/#cc2

so wonder if that even needs a technique as such, OR if the technique should clearly explain that it's not trying to make a new exception, but that it demonstrates the point there of the "Full pages" note 1

stevefaulkner:
putting links here for info purposes:

[https://www.w3.org/TR/WCAG22/#images-of-text](https://www.w3.org/TR/WCAG22/#images-of-text)

This would be an appropriate technique if the image of text is considered [_essential_](https://www.w3.org/TR/WCAG22/#dfn-essential) 

For example infographics. [WebAIM has an example](https://webaim.org/resources/designers/#infographic) which has a structured text alternative below the graphic

giacomo-petri:
Following our previous discussions regarding lengthy PR requests, if further discussion arises here, we can move the discussion back to https://github.com/w3c/wcag/issues/3755. 
Since I garnered some consensus in the initial issue and received no additional feedback, I submitted a proposed solution as a PR. However, I am open to revisiting the original issue for further refinement. 
Nonetheless, I still believe that clarification is required on this matter.

patrickhlauke:
To put it succinctly, after considering this further, this to me feels less like a technique to pass 1.4.5, but an example of "in isolation, that image with text fails 1.4.5, but because there's an accessible alternative, overall the page/sample passes". Which would feel odd to have as a technique, even though I can see how the info is useful to have. Maybe, instead of a technique, worth adding a new example/note to the 1.4.5 understanding itself, which makes the scenario clear ("sometimes, there's just no way around having images with text, but then you should provide an accessible alternative such as a visible text rendition of the same information")

giacomo-petri:
@patrickhlauke,

Do you believe the current proposal could work by merely eliminating the technique?

patrickhlauke:
@giacomo-petri I think even just the change to understanding may need a subtle re-orienting ... maybe it's semantics, maybe it's just me, but i think it's not so much that

```
Meeting this success criterion is still deemed acceptable by duplicating the meaning of the image of text through textual format
```

but perhaps more along the lines of "if it's not possible to meet this SC directly, you must provide an accessible alternative - such as duplicating the meaning/content of the image in text..." or something

understanding/20/images-of-text.html
```diff
@@ -130,6 +138,9 @@ <h2>Examples of Images of Text</h2>
          <dt>Customizable font settings in images of text</dt>
          <dd>A Web site allows users to specify font settings and all images of text on the site
                are then provided based on those settings.</dd>
+         <dt>Text replicates an image of text in a marketing campaign</dt>
+         <dd>The CMS (content management system) allows content creators to incorporate both an image and a caption. While an image of text
```
bruce-usab, line 142: > ...allows content creators to incorporate both an image and a caption. _While_ an image of text

I think "while" should be "when" here.  Also is automated or policy?

bruce-usab (changes requested):
I would like a rendered version please.

giacomo-petri:
Hi @bruce-usab,

As decided by the group, I'll first remove the technique and adjust the SC wording accordingly. I'll tweak the note slightly and add an example instead of a technique.

giacomo-petri:
@patrickhlauke, is it better now?

understanding/20/images-of-text.html
```diff
@@ -44,6 +44,12 @@ <h2>Intent of Images of Text</h2>
       <p>Images of text can also be used where it is possible for users to customize the image
          of text to match their requirements.
       </p>
+
+      <p>Sometimes, it is not possible to replace an image of text with actual text or to allow users to customize the image. 
+         In these situations, the author must provide an [accessible alternative](https://www.w3.org/TR/WCAG22/#h-note-47) by duplicating the meaning or content of the image in text form.
```
alastc, line 49:
```suggestion

      <p>If it is not possible to replace an image of text with actual text or to allow users to customize the image, the author must provide an [accessible alternative](https://www.w3.org/TR/WCAG22/#h-note-47) by duplicating the meaning or content of the image in text form. 
```

patrickhlauke: should we use relative links?

understanding/20/images-of-text.html
```diff
@@ -44,6 +44,11 @@ <h2>Intent of Images of Text</h2>
       <p>Images of text can also be used where it is possible for users to customize the image
          of text to match their requirements.
       </p>
+
+      <p>If it is not possible to replace an image of text with actual text or to allow users to customize the image, the author must provide an [accessible alternative](https://www.w3.org/TR/WCAG22/#h-note-47) by duplicating the meaning or content of the image in text form.
```
alastc, line 48:
```suggestion
      <p>If it is not possible to replace an image of text with actual text or to allow users to customize the image, the author must provide an [accessible alternative](https://www.w3.org/TR/WCAG22/#cc2) by duplicating the meaning or content of the image in text form. 
```

understanding/20/images-of-text.html
```diff
@@ -44,6 +44,11 @@ <h2>Intent of Images of Text</h2>
       <p>Images of text can also be used where it is possible for users to customize the image
          of text to match their requirements.
       </p>
+
+      <p>If it is not possible to replace an image of text with actual text or to allow users to customize the image, the author must provide an [accessible alternative](https://www.w3.org/TR/WCAG22/#cc2) by duplicating the meaning or content of the image in text form. 
+         If the information within the image permits the author to replicate the text verbatim, this should be the preferred approach. 
+         However, if converting the visual image to text results in a loss of meaning or readability, it is acceptable to convey the same message by adding or altering some words.
```
alastc, line 50:
```suggestion
         The alternative could replicate the text verbatim. If converting the visual image to text results in a loss of meaning or readability, it could add or altering some words to convey the style, emphasis, or whatever meaning was lost.
```

bruce-usab (changes requested):
I prefer Alastair's suggestion.

Phrasing "it could add or altering" reads oddly to me.

understanding/20/images-of-text.html
patrickhlauke: `it could add or altering` > `it could add or alter` or even `it could add or change`

giacomo-petri:
Updated with agreed version and fixing some typo.

mbgower (approved):

alastc:
From the meeting: The additional examples do not actually pass the criterion, the page is passing by an alternative.
It would be useful to separate those examples into a separate list and highlight that they don't pass.

E.g.

> The following examples pass the criterion:
> [current list]
> The following examples do not pass the criterion, but the page would pass because there is an alternative.
> [new list]

stevefaulkner (approved):

understanding/20/images-of-text.html
```diff
@@ -44,6 +44,10 @@ <h2>Intent of Images of Text</h2>
       <p>Images of text can also be used where it is possible for users to customize the image
          of text to match their requirements.
       </p>
+
+      <p>If it is not possible to replace an image of text with actual text or to allow users to customize the image, the author must provide an [accessible alternative](https://www.w3.org/TR/WCAG22/#cc2) by duplicating the meaning or content of the image in text form.
```
mbgower, line 48: It is unclear what you are trying to link to with this URL: https://www.w3.org/TR/WCAG22/#cc2

There is no definition of _accessible alternative_. Are you making the case that the page can conform although it fails 1.4.5, by meeting 1.1.1?

mbgower:
IMO, the wording needs to be clearer in these additions.
I believe there was a agreement that text that appears on the same page (not just an ALT) as the image of text (i.e., a caption that states the same information) can be considered to pass 1.4.5, since all information is presented as text.

It's less clear to me that folks would agree that an exact but undisplayed ALT for an image of text passes.

understanding/20/images-of-text.html
giacomo-petri: > It is unclear what you are trying to link to with this URL: https://www.w3.org/TR/WCAG22/#cc2

I'm referring to NOTE 1, which states:

> For the purpose of determining conformance, alternatives to part of a page's content are considered part of the page when the alternatives can be obtained directly from the page, e.g., a long description or an alternative presentation of a video.

----

> Are you making the case that the page can conform although it fails 1.4.5, by meeting 1.1.1?

Not really. SC 1.4.5 and 1.1.1 address different needs. SC 1.1.1 is primarily intended for people who can rely on text alternatives that don't need to be visible, while SC 1.4.5 requires an option to visually customize the text. Therefore, if the same text is presented close to the image, it addresses the requirement.

giacomo-petri:
@mbgower,

The intent is indeed to require a visible text alternative, not an invisible one like an alt attribute.

Does adding "visible" to the sentence (emphasis on """visible"""):

> If it is not possible to replace an image of text with actual text or to allow users to customize the image, the author must provide an [accessible alternative](https://www.w3.org/TR/WCAG22/#cc2) by duplicating the meaning or content of the image in """**visible**""" text form.

solve the problem?

giacomo-petri:
Note: new examples seem already covering this bit clearly.

mraccess77:
As SC 1.4.5 is aimed at sighted users - alt text which is not displayed would not pass in my opinion.  If it would, then there would be no material point in SC 1.4.5 as 1.1.1 already requires a text equivalent for images of text.

patrickhlauke:
randomly dropping here: as the SC is scoped to situations where images of text are used *rather than* text, then the case where images of text are used *in addition* to text, then the SC doesn't apply anyway since it's not the *rather than*

understanding/20/images-of-text.html
```diff
@@ -44,6 +44,10 @@ <h2>Intent of Images of Text</h2>
       <p>Images of text can also be used where it is possible for users to customize the image
          of text to match their requirements.
       </p>
+
+      <p>If it is not possible to replace an image of text with actual text or to allow users to customize the image, the author must provide an [accessible alternative](https://www.w3.org/TR/WCAG22/#cc2) by duplicating the meaning or content of the image in text form. 
+         The alternative could replicate the text verbatim. If converting the visual image to text results in a loss of meaning or readability, the author could add or alter some words to convey the style, emphasis, or any other meaning that was lost.
+      </p>
```
giacomo-petri, line 50: Based on yesterday's discussion, I have drafted a new proposal with a different approach to avoid any ambiguity or conflict with the normative portion of the SC:

Proposal:
```suggestion
      <p>An image of text accompanied by visible text that replicates its content verbatim is exempt from this success criterion, as the image is not used to convey information but serves a decorative purpose, with the text providing the actual information.
Moreover, the text does not always need to replicate the image of text content exactly. Verbatim replication might result in a loss of meaning or readability. Authors can modify or add words to convey style, emphasis, or other meanings that could be lost in a direct replication.
      </p>
```

patrickhlauke: I would actually suggest dropping this, and the previous change (lines 47-50), and maybe instead just adding something like the following (which I think is getting to the same point as we discussed):

```
<div class="note"><p>The Success Criterion is scoped to situations where <a>images of text</a> are used <em>rather than</em> text. If a page has an image of text, but also contains visible text that duplicates the textual content of the image and conveys the same meaning, then this Success Criterion does not apply.</p></div>
```

mbgower: I'm not sure that it doesn't apply, so much as that it meets the requirement.
If text is used, it's met. I'm wondering if we're able to align on some concept and wording like this:

> <p>Note: The Success Criterion is intended to address situations where <a>images of text</a> are used <em>rather than</em> text.  Where images of text are used <em>in addition to text</em> to convey the same information -- both are presented to the user --  this success criterion is met. This allows author to convey any stylistic information they desire, while also providing the information in text, whose presentation can then be manipulated by users to make it more distinguishable.

mbgower: I'm inclined not to include new text that would open the Pandora's box regarding _how much_ the text may deviate from the same information provided in the image of text before this SC is no longer met via this rationale.

understanding/20/images-of-text.html
```diff
@@ -130,6 +134,14 @@ <h2>Examples of Images of Text</h2>
          <dt>Customizable font settings in images of text</dt>
          <dd>A Web site allows users to specify font settings and all images of text on the site
                are then provided based on those settings.</dd>
+         <dt>The text replicates an image of text</dt>
+         <dd>The CMS (content management system) allows content creators to incorporate both an image and a caption. 
+           While an image of text is utilized, the identical message is also presented as text directly beneath the image.</dd>
+           <dt>The text conveys the same meaning as the image of text</dt>
+         <dd>In a banner, textual information is combined with graphical content. Simply replicating
```
mbgower, line 141: I'm having difficulty picturing this. We could mabye include an example? Otherwise, I'm inclined to not add this last example.

patrickhlauke:
Following more discussions in last week's WCAG 2.x backlog meeting, I filed a clean counter-proposal https://github.com/w3c/wcag/pull/4021

alastc:
Suggest we wait for Michael and Giacomo to get back before resolving this one.

alastc:
Superceded by #4021

Issue #3793, opened 2024-04-22
Video on "Understanding SC 1.4.10" page fails SC 1.4.10


front-endian:
On the [Understanding SC 1.4.10](https://www.w3.org/WAI/WCAG22/Understanding/reflow.html) the video showing the BBC website scaling for smaller screen widths doesn't scale with the width of the page. You to need to do horizontal scrolling to see the rest of the video and video controls on windows smaller than about 2,300px wide (at least on Chrome 123.0.6312.124 on MacOS).

It looks like this can be fixed by giving `width:100%` to the `<video>` element itself. I could put up a PR with that change, but I am not sure if y'all want to do a more wholistic fix as the same issue may also be present on other videos on other W3C/WCAG pages.

patrickhlauke:
duplicate, and supposedly fixed by https://github.com/w3c/wcag/pull/3115 but problem still there. @iadawn any idea what's going on?

iadawn:
Missed your follow up comment on the typo 'examples' rather than 'example'. Will raise PR to fix that.

iadawn:
Ignore my PR. @patrickhlauke you are paying better attention to detail than me. More ☕️ needed

patrickhlauke:
no worries @iadawn :)

patrickhlauke:
just documented/commented on that PR of yours, but not to tell you off, just for context/posterity :)

iadawn:
Posterity will forever judge me harshly ;)

PR #3796, opened 2024-04-23
Fix typo #3793

iadawn:


patrickhlauke:
still not correct - `#examples`, not `.examples` (i played the odds and made my PR apply to both the original `.example` and `#examples` just to be safe / if we ever add `<video>` to individual `.example` blocks)

PR #3818, opened 2024-05-04
Remove failure technique F87

### LINKED ISSUE #1554: F87 and ARIA 24 seem to contradict in the use of CSS content  property

### LINKED ISSUE #3781: 1.3.1 / F87 for input fields with mandatory visual indication via pseudo-element

patrickhlauke:
Rationale: this is not *always* a failure. User agent/screen reader support has improved considerably since the technique was first devised, with most (all?) combinations now happily reading/announcing CSS-generated content.

Supersedes https://github.com/w3c/wcag/pull/2800 (which was having too many merge conflicts due to its age)

Closes #1554, closes #3781


techniques/index.html
```diff
@@ -48,6 +48,7 @@ <h1 id="title" class="title p-name">Techniques for WCAG 2.2</h1>
 			<h2>Change Log</h2>
 			<p>A list of new, removed or significantly updated techniques:</p>
 			<ol>
+				<li><time datetime="2024-05-04">4 May 2024</time>: Removed F87 Failure of Success Criterion 1.3.1 due to inserting non-decorative content by using :before and :after pseudo-elements and the 'content' property in CSS</li>
```
scottaohara, line 51: per discussion last week, should we just remove the changelog all together since it's inconsistently appearing in techniques and is not always updated?

dbjorge (approved):
Did some followup discussion with other Deque folks and we're on board with removing the failure technique. We acknowledge that `:before`/`:after` content involves some usability issues:

- Can't Ctrl+F for it
- Can't select/copy/paste it
- Doesn't show up in some "Reader Mode" UIs (eg, Safari's)

...but we agree with points made during today's TF meetings that even though these may create usability issues, these probably don't constitute 1.3.1 failures (other acceptable 1.3.1 techniques have similar usability limitations).

I think it might be nice to add an understanding doc note or advisory technique or something that warns about the potential for usability issues, but I don't think that blocks this PR.

PR #3821, opened 2024-05-06
Update F91 to remove unrelated techniques

### LINKED ISSUE #3788: Technique F91: Related techniques are wrong
It should only be related to [H43: Using id and headers attributes to associate data cells with header cells in data tables](https://www.w3.org/WAI/WCAG22/Techniques/html/H43)

scottaohara:
closes #3788

removes f46, g115 and g63 as they aren't directly related to not marking up table headers correctly.

The issue said all but H43 should be removed, but H51 demonstrates how to build a table with correctly marked up headers, so i see no reason why this shouldn't still be linked.  As it stands, support for the headers attribute is rather lack luster - so pointing to a generic use of setting up a table with correct usage of the th element is worth not removing.


detlevhfischer:
Linkt to [F91](https://www.w3.org/WAI/WCAG21/Techniques/failures/F91.html)

scottaohara:
@detlevhfischer could you clarify what you are trying to communicate with your list of links?  are you saying there should be more links, that these links are still in the doc, that these links (pages) link to this particular technique?

bruce-usab (approved):

PR #3863, opened 2024-05-14
Update F26.html

ljoakley:
Add a link to ARIA14 in F26.

### DIFF
--- a/techniques/failures/F26.html
+++ b/techniques/failures/F26.html
@@ -30,5 +30,6 @@
       </section>
    </section><section id="related"><h2>Related Techniques</h2><ul>
       <li><a href="../html/H37">H37</a></li>
-      <li><a href="../aria/ARIA24">ARIA24</a></li>
+	  <li><a href="../aria/ARIA14">ARIA14</a></li>
+	  <li><a href="../aria/ARIA24">ARIA24</a></li>
    </ul></section><section id="resources"><h2>Resources</h2></section></body></html>


mgifford:
So adding a link here [https://www.w3.org/WAI/WCAG22/Techniques/failures/F26](https://www.w3.org/WAI/WCAG22/Techniques/failures/F26)#related  
  
to add [https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA14](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA14) to the Related Techniques section.  
  
  
So how is Technique **ARIA14: Using aria-label to provide an invisible label where a visible label cannot be used** related to Technique **F26:Failure of Success Criterion 1.3.3 due to using a graphical symbol alone to convey information**?  
  
It isn't as direct a relationship as either [H37](https://www.w3.org/WAI/WCAG22/Techniques/html/H37) or [ARIA24](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA24) in my reading.

mbgower:
> So how is Technique **ARIA14: Using aria-label to provide an invisible label where a visible label cannot be used** related to Technique **F26:Failure of Success Criterion 1.3.3 due to using a graphical symbol alone to convey information**?

@mgifford, since the failure check for F26 is:
> 1. Examine the page for non-text marks that convey information.
> 2. Check whether there are other means to determine the information conveyed by the non-text marks.

aria-label would be a method of resolving step 2.

In regard to your statement

> It isn't as direct a relationship as either [H37](https://www.w3.org/WAI/WCAG22/Techniques/html/H37) or [ARIA24](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA24) in my reading.

H37 is an existing related technique, which makes sense, since putting an ALT on an image is a way of resolving step 2 as well.

ARIA24, which is _not_ in the existing failure technique nor part of this PR, can be proposed in a new issue if you like, but my feeling is that it does not actually satisfy the failure. It resolves where something isn't identified as an image; that is a different use case than where an image has no alternative (which is the failure situation in F24).

bruce-usab:
Perhaps off-topic, but I had expected the build process to automagically add the name of the technique?
- See: https://www.w3.org/WAI/WCAG22/Techniques/failures/F26#related
- Source code in repo is just `<a href="../aria/ARIA24">ARIA24</a>`
Does the page title need to be entered manually?  (as would be needed for any/all off-site links)

patrickhlauke:
for techniques, it doesn't seem to do this automagic thing (see the old/current F26 https://www.w3.org/WAI/WCAG22/Techniques/failures/F26) - if there is some kind of automation that runs at build time, i suspect it doesn't happen for techniques, just understanding docs cc /@iadawn

Issue #3866, opened 2024-05-17
F55 link name is truncated to "F55: received"


filippo-zorzi:
All the links to the Failure F55 are labelled as "F55: received" instead of "F55: Failure of Success Criteria 2.1.1, 2.4.7, 2.4.13, and 3.2.1 due to using script to remove focus when focus is received".

<img width="921" alt="Screenshot 2024-05-17 at 09 55 42" src="https://github.com/w3c/wcag/assets/165811341/e040da79-c2f8-450a-9bcb-01cd891ebe05">

This problem is present in:
 - 2.1.1
 - 2.4.7
 - 2.4.13
 - 3.2.1

PR #3874, opened 2024-05-24
Tweak F55 to fix broken/partial link text

patrickhlauke:
Closes https://github.com/w3c/wcag/issues/3866

(need to actually test that it does, but the fix here is the only reason I can think of why the link is shown as just "received" at the moment)


alastc:
Just needs someone to build it and check it works.

PR #3876, opened 2024-05-24
Remove irrelevant preamble and first step from F94, future-proof expected results

patrickhlauke:
This is a follow-up to https://github.com/w3c/wcag/pull/3739


mbgower:
This was merged into #3739 to make one consolidated PR.

PR #3879, opened 2024-05-24
Change <br> to <br/>

patrickhlauke:
Due to the XML-based publication pipeline, need to have self-closing `<br/>`. Otherwise when trying to build locally at least, `ant` chokes.

### DIFF
--- a/understanding/21/target-size-enhanced.html
+++ b/understanding/21/target-size-enhanced.html
@@ -56,14 +56,14 @@
   	<section id="examples">
   	  <h2>Examples</h2>
       <ul>      
-        <li><strong>Example 1: Buttons</strong><br> Three buttons are on-screen and the touch target area of each button is 44 by 44 CSS pixels.</li>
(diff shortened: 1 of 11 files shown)


scottaohara (approved):
took a look and approving per the rational as to why this needs to happen.  but otherwise, boo :)

bruce-usab (changes requested):
Should Understanding Target Size Enhanced uses something semantic rather than BR ?

fstrr:
Seems like a prima candidate for a description list.

patrickhlauke:
can/should we do that in a separate PR, rather than this very targeted (pun) PR to just fix some xml?

bruce-usab (approved):

bruce-usab:
From call 6/7, concur with just making this bug fix.

fstrr:
'Tis done - #3901

patrickhlauke:
as @iadawn 's https://github.com/w3c/wcag/pull/3899 already fixes the first change for the target size understanding (likely he bumped into the same issue i saw when trying to build the spec/understanding), and the other instances of naked `<br>` do *not* need to be fixed as they don't go through the XML pipeline...closing this as superseded

PR #3888, opened 2024-05-31
Update labels-or-instructions.html

### LINKED ISSUE #3887: Clarification needed for 3.3.2 Understanding doc example bullet
The wording of "A field for entering a date contains initial text which indicates the correct format for the date." has made some people assume that placeholder text is an appropriate means to convey instructions / data formatting requirements. However the technique that best accompanies this example bullet does not use placeholder text, and rather includes the format in the label.

As discussed with the wcag 2.x taskforce today, we agreed that the doc should not be implying that placeholders are sufficient (for a number of reasons even beyond this SC alone), so the wording of this example should change so as to not imply placeholders as being 'ok'

scottaohara:
closes #3887

clarifies that wcag is not recommending use of placeholder text, and instead recommends accompanying text to provide the information (which could then be part of label or description - but those details are not applicable to this SC so not specified in the update)

This helps better match [technique g89](https://www.w3.org/WAI/WCAG22/Techniques/general/G89.html) which includes the date format expectation as part of the label text

### DIFF
--- a/understanding/20/labels-or-instructions.html
+++ b/understanding/20/labels-or-instructions.html
@@ -92,7 +92,7 @@
             correct abbreviation.
          </li>
          
-         <li>A field for entering a date contains initial text which indicates the correct format
+         <li>A field for entering a date is provided text instructions to indicate the correct format
             for the date.
          </li>
          
--- a/understanding/20/labels-or-instructions.html
+++ b/understanding/20/labels-or-instructions.html
@@ -92,7 +92,7 @@
             correct abbreviation.
          </li>
          
-         <li>A field for entering a date is provided text instructions to indicate the correct format
+         <li>A field for entering a date has text instructions to indicate the correct format
             for the date.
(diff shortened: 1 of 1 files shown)


understanding/20/labels-or-instructions.html
```diff
@@ -92,7 +92,7 @@ <h2>Examples of Labels or Instructions</h2>
             correct abbreviation.
          </li>
          
-         <li>A field for entering a date contains initial text which indicates the correct format
+         <li>A field for entering a date is provided text instructions to indicate the correct format
```
patrickhlauke, line 95:
```suggestion
         <li>A field for entering a date has text instructions to indicate the correct format
```
cleaner/simpler

patrickhlauke (changes requested):

patrickhlauke (approved):

bruce-usab (approved):

PR #3912, opened 2024-06-20
Fix display of “New in WCAG 22” text

yatil:
Changing the class from wcag22 to wcag2.2 should fix the mentions of non-existing WCAG version 22 throughout the WCAG Understanding Documents.

Example: https://www.w3.org/WAI/WCAG22/Understanding/focus-visible.html

The text says: 

> Note that a keyboard focus indicator can take different forms. New in WCAG 22: While Focus Visible does not specify what that form is, 2.4.13 Focus Appearance (Level AAA) provides guidance on creating a consistent, visible indicator.

This fix should change it to “New in WCAG 2.2”.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -1,43 +1,43 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- This is a build file for Apache ant (http://ant.apache.org/). This has been tested with Ant 1.7. -->
 <project name="wcag" basedir="." default="usage">
-	
+
 	<description>Generate WCAG 2.1 and related documents</description>
-	
+
 	<!-- Load overriding properties -->
 	<xmlproperty file="build.properties" keeproot="false" semanticattributes="true"/>
-	
+
 	<!-- The following properties need to be set appropriately before doing a build - they should all be overridden by the above property file, but are kept here to serve as defaults -->
 	<property name="outputdir" location="." description="Directory within which the output folders are created, normally &quot;../YYYY&quot; unless doing TR in which case it is &quot;../../../TR/&lt;YYYY&gt;&quot;"/>
 	<property name="uri.prefix" value="file:///" description="Prefix if any that must be prepended to URIs to make it resolve on the platform"/>
-	<property name="guidelines.version" value="22" description="Which version of WCAG is being built"/>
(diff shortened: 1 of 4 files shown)


patrickhlauke:
how does changing the classname help? is there a script somewhere that naively grabs whatever the classname is and generates the text?

yatil:
@patrickhlauke I think that is what this section of the XSLT does, but I’m no XSLT expert. 

https://github.com/w3c/wcag/blob/ba81e62b839c64fb3789dff714934c514514eec0/xslt/base.xslt#L192-L212

yatil:
Reading it again, it looks like the script inserts `$guidelines.version` which is probably defined elsewhere…

yatil:
Looks from my limited local testing that changing the version in the build.xml also does not break things, so I’ve reopened the PR. I apologize for the back and forth, last time when I tried building it, it just exploded in my face. 😂

patrickhlauke:
i'd caution that the build pipeline may be using that `$guidelines.version` to determine the final URL, i.e. https://www.w3.org/TR/WCAG22/ ... but @iadawn will hopefully know best

yatil:
Yeah, I don’t expect this to be merged as is, the house of cards that is the WCAG build process is too fragile for that. In the same time, I’d like to not add additional variables that might then not be updated accidentally.

kfranqueiro:
Changing the value of `guidelines.version` in `build.xml` is likely to have many ripple effects. Moreover, there is already another variable that does what you want:

https://github.com/w3c/wcag/blob/ba81e62b839c64fb3789dff714934c514514eec0/xslt/base.xslt#L11

Changing this line...

https://github.com/w3c/wcag/blob/ba81e62b839c64fb3789dff714934c514514eec0/xslt/base.xslt#L206

...to reference `guidelines.version.decimal` instead should do what you want, with no other changes necessary.

FWIW, this will also be fixed in the new Eleventy-based build process in #3917.

yatil:
I wasn’t aware of an Eleventy build process being in development, @kfranqueiro. (I hope the Jekyll templates can be reused.) I don’t think it’s worth working on this for now.

PR #3913, opened 2024-06-20
Add link underline color to 1.4.11

fstrr:
Adds an example where link and non-link text are both white on a black page background. The only way to identify the link is via a dull-red underline that has a contrast of less than 3:1.

[Rendered page](https://raw.githack.com/w3c/wcag/add-underline-color-to-1.4.11/understanding/21/non-text-contrast.html) (check the User Interface Component Examples table)

Closes https://github.com/w3c/wcag/issues/3987


understanding/21/non-text-contrast.html
```diff
@@ -200,10 +200,14 @@ <h4>User Interface Component Examples</h4>
 						<th>Examples</th>
 						</tr>
 						<tr>
-							<th>Link Text</th>
+							<th>Default link text</th>
 							<td>Default link text is in the scope of <a href="contrast-minimum">1.4.3 Contrast (Minimum)</a>, and the underline is sufficient to indicate the link.</td>
 							<td><img src="img/link-text-default.png" alt="A browser-default styled link, blue with an underline."/></td>
 						</tr>
+						<th>Colored underline is the only indicator of a link - fail</th>
+						<td>Link and non-link text are both white on a black background. The link's underline (<code>#A92329</code>) is the only way to identify the link. The underline's contrast less than 3:1.</td>
```
fstrr, line 208:
```suggestion
						<td>Link and non-link text are both white on a black background. The link's underline (<code>#A92329</code>) is the only way to identify the link. The underline's contrast is less than 3:1.</td>
```

bruce-usab (approved):
Since it is a failure example, my preference would be to distinguish more from the affirmative examples that it is grouped with.

I agree that since the underline contrast < 3:1 that it fails.

I would also be happy to use a gray, to keep "color" out as factor.  That could allow a pass [example with white on black using gray link text](https://webaim.org/resources/linkcontrastchecker/?fcolor=FFFFFF&bcolor=000000&lcolor=767676).

understanding/21/non-text-contrast.html
```diff
@@ -200,10 +200,14 @@ <h4>User Interface Component Examples</h4>
 						<th>Examples</th>
 						</tr>
 						<tr>
-							<th>Link Text</th>
+							<th>Default link text</th>
 							<td>Default link text is in the scope of <a href="contrast-minimum">1.4.3 Contrast (Minimum)</a>, and the underline is sufficient to indicate the link.</td>
 							<td><img src="img/link-text-default.png" alt="A browser-default styled link, blue with an underline."/></td>
 						</tr>
+						<th>Colored underline is the only indicator of a link - fail</th>
```
patrickhlauke, line 207: missing an opening `<tr>`

patrickhlauke: the table is prefaced with "the following are examples that have sufficient contrast" so this fail example is the odd one out now

fstrr: Oh interesting. I didn't spot that text. The table also has "Checkbox - Fail" and "Checkbox - Subtle focus style - fail", so we could either yoink out that "sufficient contrast" sentence or re-do the existing two fails as well.

understanding/21/non-text-contrast.html
```diff
@@ -192,22 +186,22 @@ <h4>User Interface Component Examples</h4>
...
 						</caption>
 						<tr>
 						<th>Type</th>
 						<th>Description</th>
 						<th>Examples</th>
 						</tr>
 						<tr>
-							<th>Link Text</th>
+							<th>Default link text</th>
 							<td>Default link text is in the scope of <a href="contrast-minimum">1.4.3 Contrast (Minimum)</a>, and the underline is sufficient to indicate the link.</td>
```
mbgower, line 198: @fstrr This addresses the concerns we discussed in an email conversation on June 20
```suggestion
							<td>Default link text is covered by <a href="contrast-minimum">1.4.3 Contrast (Minimum)</a>,  not Non-text Contrast. Regardless of whether one considers the underline part of the text or a non-text indicator, since it is the same color as the text, which must meet 4.5:1 to pass, the default underline will always pass the lower contrast requirements of Non-text Contrast.</td>
```

understanding/21/non-text-contrast.html
```diff
@@ -192,22 +186,22 @@ <h4>User Interface Component Examples</h4>
...
+							Passing User Interface Component Examples
 						</caption>
 						<tr>
 						<th>Type</th>
 						<th>Description</th>
 						<th>Examples</th>
 						</tr>
 						<tr>
-							<th>Link Text</th>
+							<th>Default link text</th>
```
mbgower, line 197:
```suggestion
							<th>Link text</th>
```

understanding/21/non-text-contrast.html
```diff
@@ -268,18 +262,36 @@ <h4>User Interface Component Examples</h4>
...
-							<th>Checkbox - Subtle hover style</th>
-							<td>A black border on a white background indicates the checkbox, when the mouse pointer activates the subtle hover state adds a grey background (#DEDEDE). The black border has a 15:1 contrast ratio with the grey background.</td>
-							<td><img src="img/checkbox-example4.png" alt="Blackbox on a circular grey background next to a text label." width="150"/></td>
+						<th>Type</th>
+						<th>Description</th>
+						<th>Examples</th>
+						</tr>
+						<tr>
+						<th>Colored underline is the only indicator of a link</th>
+						<td>Link and non-link text are both white on an almost-black (<code>#0D0F13</code>) background. The link's underline (<code>#B1262B</code>) is the only way to identify the link. The underline's contrast with the background color is less than 3:1.</td>
```
mbgower, line 284:
```suggestion
						<td>Link and non-link text are both white on an almost-black (<code>#0D0F13</code>) background. The link's custom underline (<code>#B1262B</code>) is the only way to identify the link. The red underline contrasts less than 3:1 with the background color.</td>
```

understanding/21/non-text-contrast.html
fstrr: Thanks, Mike.

What about this, which is a little more succinct?:

```suggestion
							<td>The browser's default link text color is covered by <a href="contrast-minimum">1.4.3 Contrast (Minimum)</a>. Since the underline is the same color as the text, which must meet at least 3:1 to pass, the default underline will always pass the requirements of Non-text Contrast.</td>
```

understanding/21/non-text-contrast.html
```diff
@@ -268,18 +262,36 @@ <h4>User Interface Component Examples</h4>
 							<td><img src="img/checkbox-example2.png" alt="Black square border with a tick inside, and a text label." width="150"/></td>
 						</tr>
 						<tr>
-							<th>Checkbox - Fail</th>
-							<td>The grey border color of the checkbox (#9D9D9D) has a contrast ratio of 2.7:1 with the white background, which is not sufficient for the visual information required to identify the checkbox.</td>
-							<td><img src="img/checkbox-example3.png" alt="Grey box on a white background with a black tick in the middle." width="150"/></td>
+							<th>Checkbox - Subtle hover style</th>
+							<td>A black border on a white background indicates the checkbox, when the mouse pointer activates the subtle hover state adds a grey background (<code>#DEDEDE</code>). The black border has a 15:1 contrast ratio with the grey background.</td>
```
mbgower, line 266:
```suggestion
							<td>A black border on a white background indicates the checkbox, but when the mouse pointer hovers on the checkbox, a subtle grey background is added (<code>#DEDEDE</code>). The black border has a 15:1 contrast ratio with the grey background, so the hover effect clearly does not cause the checkbox to fail contrast. Note that the grey hover effect does not itself need to contrast 3:1 with the page background, since the pointer position is the primary indicator of the hover state.</td>
```

fstrr: This sentence needs some work, I think:

> so the hover effect clearly does not significantly reduce the checkbox indicator's contrast

1. I'd remove the word "clearly" — it's likely not clear to some people
2. "does not significantly reduce the checkbox indicator's contrast":
    a. What does "significantly" mean? To reduce the contrast to less than 3:1?
    b. What's the "indicator"? Is it the border?

What about:

"The black border has a 15:1 contrast ratio with the grey background, which is above the required 3:1 ratio."

mbgower (approved):

Gcamara14:
Hi All,

Is this available somewhere to view?

fstrr:
Sure. You need to be on the WCAG 2.2 version of the site and then it's [the first example](https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast.html#:~:text=Colored%20underline%20is%20the%20only%20indicator%20of%20a%20link) in the "failing user interface component examples" table.

PR #3915, opened 2024-06-25
Harmonise wording for techniques checks/steps

patrickhlauke:
Currently, techniques use a mix of wording `If step X...` and `If check X...`, as well as a few outlier "If test procedure X...". Most prefix the step/check number with `#`, but some don't. And conceptually, I think it makes more sense saying that a *check* is true/false, rather than a *step* being true/false.

This PR tries to harmonise this, settling for `check` and adding missing `#` (though, it has to be said, this use of the octothorpe is a bit US-centric, so maybe we want to harmonise to *not* having it?)


fstrr:
Agree on the US-centric use of the `#` symbol. I've often considered yoinking it out and would be fine with it being removed.

fstrr (approved):

patrickhlauke:
Point to discuss in our next WCAG 2.x meeting then ... if we want to yoink `#` out, I can do a big find/replace and add it as a commit here.

alastc:
I think aligning with the earlier examples is probably best (and the least amount of files changed), which is what you've done. That will also ruffle the least feathers.

techniques/failures/F52.html
mgifford, line 1: I think this is better as check #2 and check #3 but in other places it is checks #2 and #3. Just wanted to highlight this.

techniques/failures/F9.html
mgifford, line 1: This is inconsistent. Should be all checks and no steps, right?

ljoakley (approved):
completed with a few changes

techniques/failures/F24.html
```diff
@@ -90,7 +90,7 @@ <h3>Specifying foreground color of link text with CSS</h3>
          </div>
       </section>
       <section class="results"><h3>Expected Results</h3>
-         <p>If step #2 is true but step #3 is false, OR if step #3 is true but step #2 is false then this
+         <p>If check #2 is true but step #3 is false, OR If check #3 is true but step #2 is false then this
```
patrickhlauke, line 93:
```suggestion
         <p>If check #2 is true but step #3 is false, <em>or</em> if check #3 is true but step #2 is false then this
```

techniques/failures/F38.html
```diff
@@ -22,7 +22,7 @@
       </section>
       <section class="results"><h3>Expected Results</h3>
          <ul>
-            <li>If step #1 is true and if step #2 is true, this failure condition applies and content fails the Success Criterion.</li>
+            <li>If check #1 and #2 is true, this failure condition applies and content fails the Success Criterion.</li>
```
patrickhlauke, line 25:
```suggestion
            <li>If check #1 and #2 are true, this failure condition applies and content fails the Success Criterion.</li>
```

mbgower:
@shawna-slh Is there any style decision on use of the pound symbol (#) in front of ordered list references?

shawna-slh:
Thanks for the ping @mbgower ! We don't currently have that in our Style guide. (I prefer without the # - yet don't have any support for or against that.) From a quick search, I don't find guidance in leading Style Guides -- so it needs more research.

The good news is we have a new WAI staff person coming on the end of the month who will be great at citing sources for a decision.

Are y'all comfortable leaving this issue open for a few weeks until we can get back to it?

(p.s. thanks to @patrickhlauke I learned a new term, "octothorpe".)

patrickhlauke:
> Are y'all comfortable leaving this issue open for a few weeks until we can get back to it?

no rush from my perspective. I'd rather wait and then do all the changes in a oner, once we have a decision on the `#`

mbgower:
@shawna-slh Sorry, I missed those final comments when I brought this into our current review cycle.
If someone gets to this by the start of August, we can just incorporate your change, otherwise we'll just push a second update when you make your recommendation.

shawna-slh:
@tamsinewing555 will probably get to it within a few days :)

Issue #3920, opened 2024-06-25
Some definition links (#dfn-*) point to fully-qualified URLs


kfranqueiro:
Most definition links are in the format `<a>term</a>`, intended to be automatically resolved by the build process as indicated in the main README. These are responsible for populating each page's Key Terms section.

However, there are a handful of pages that contain links already pointing directly to TR URLs. For example, both ARIA22 and F103 have links pointing to `/TR/WCAG/#dfn-status-messages`, when instead they could be expressed as `<a>status message</a>` (either singular or plural), which would automatically expand and populate the term under Key Terms (using `guidelines/terms/21/status-message.html`).

I'm unsure if any of these instances are intentionally this way, but some of these links point to specific TR versions (i.e. we end up with some links from 2.2 informative docs to 2.1 normative content), and some even point to github.io URLs.

A search for `/#dfn` across the `techniques` and `understanding` directories should surface these instances.

patrickhlauke:
how do we handle the few cases where the link text is slightly different from the actual term? for instance, `<a href="/TR/WCAG/#dfn-web-page-s">web pages</a>`?

kfranqueiro:
> how do we handle the few cases where the link text is slightly different from the actual term? for instance, `<a href="/TR/WCAG/#dfn-web-page-s">web pages</a>` or `<a href="/TR/WCAG/#dfn-conforming-alternate-versions">conforming alternate version</a>`?

This is handled via alternate term names in `data-lt` attributes in the individual term files. See [Definitions in the README](https://github.com/w3c/wcag?tab=readme-ov-file#definitions) for explanation. (e.g. the two examples you mentioned are already handled.)

PR #3917, opened 2024-06-25
Implement Eleventy-based build system

kfranqueiro:
My first project for W3C has been to develop a replacement for the Ant/XSLT-based build system, to give us something that is more flexible to modify and hopefully easier to maintain, since XSLT can be very hard to follow and is not something many people are well-versed in.

The primary building blocks of this PR are [Eleventy](https://www.11ty.dev/), as a basis for static site generation and includes, and [Cheerio](https://cheerio.js.org/), for in-place markup modifications to take the place of much of what the XSLT was doing in the old build system. TypeScript is also used, as there is a significant amount of custom logic that warrants typings to be easier to work with.

In addition to being able to test this locally (see below), you can now also see the output deployed at:

- https://kfranqueiro.github.io/wcag/techniques/
- https://kfranqueiro.github.io/wcag/understanding/

(A new github action is on a separate branch for after this PR.)

## Goals

The primary goal of the new build system is to replicate the `techniques` and `understanding` output of the existing process as closely as possible to give us a starting point for further improvements. However, there are a couple of noticeable changes and many fixes, noted below.

An additional goal is to avoid the need to touch any input files at this point, given the large number of existing open PRs.

Expected benefits include easier local previewing, and no longer having to worry about HTML doubling as valid XML in order to build successfully.

## Files in this PR

- `_includes` directory: contains as much markup as is feasible to isolate for reusable pieces of pages
  - Note: most `include` directives currently exist within a Liquid extension class where in-place transformations are performed, in order to avoid modifying source files
- `11ty` directory: contains all customizations and data retrieval for the Eleventy build
- A few new top-level files are added to configure Eleventy, TypeScript, and formatting
- One `*.11tydata.json` file is added to each of the `techniques` and `understanding` directories with common data used for processing files in each directory

## How to test-run this locally

- First, add a remote pointing to the fork, and checkout the branch:
  `git remote add kgf https://github.com/kfranqueiro/wcag.git`
  `git checkout kgf-11ty`

There is a README file in the `11ty` directory that explains setup and available commands. The short version is:

- `npm i`
- `npm start`
- Browse to `localhost:8080/techniques` or `/understanding`

I suggest Node 20 (LTS) in the README, but I've now performed test-runs on Node 18, 20, and 22; any of these should work.

## Significant changes in this PR

I discussed these changes with @alastc, @mbgower, and @bruce-usab during the weeks leading up to this PR.

### About this Technique and Applicability merged

[...]

PR #3930, opened 2024-06-25
Simplify/correct definition links

patrickhlauke:
Closes https://github.com/w3c/wcag/issues/3920


kfranqueiro:
This looks good to me for techniques and understanding. I would suggest removing the changes in this PR for conformance-challenges and guidelines since those aren't processed the same way. Sorry I was probably ambiguous about that in the issue.

patrickhlauke:
Reverted those two non-technique/non-understanding changes

kfranqueiro:
**TL;DR: we should not merge this until we're using the Eleventy build process, where I've fixed an underlying issue.**

This isn't so much a problem with this PR as it is with the XSLT build process. In the process of thoroughly testing, I think I understand why some of these (particularly some under `techniques`) were using full URLs.

The XSLT build process generates its own IDs for definitions links, disregarding the already-existing `id` attributes in the markup in `guidelines/terms/*` files. This is fine for Understanding pages, which cross-links to headings under its own Key Terms section. However, Techniques pages don't do this; they link directly to the TR - which is bad if the IDs don't match.

As it stands, the links to "status messages", "change of context", "conforming alternate version", and "web pages" will not scroll to the correct position in the TR when this PR's changes are processed by the XSLT build process. I have added a commit to the Eleventy build process PR (#3917) to address this for Techniques pages, so this PR will work 100% with that build process.

kfranqueiro (approved):
Tthe Eleventy build process has now been successfully used to push the latest site update, and merged into the main branch, so it's now safe to merge this PR.

stevefaulkner (approved):

techniques/failures/F61.html
```diff
@@ -3,7 +3,7 @@
                     automatic update that the user cannot disable from within the content</h1><section class="meta"><p class="id">ID: F61</p><p class="technology">Technology: failures</p><p class="type">Type: Failure</p></section><section id="applicability"><h2>When to Use</h2>
       <p>General </p>
    </section><section id="description"><h2>Description</h2>
-      <p> This document describes a failure that occurs when the content in the main <a href="/TR/WCAG/#dfn-viewport">viewport</a> viewport is automatically updated, and there is no option for a user to disable this behavior. </p>
+      <p> This document describes a failure that occurs when the content in the main <a>viewport</a> viewport is automatically updated, and there is no option for a user to disable this behavior. </p>
```
mbgower, line 6: I'm assuming this is only supposed to be the word "viewport" once?
```suggestion
      <p> This document describes a failure that occurs when the content in the main <a>viewport</a> is automatically updated, and there is no option for a user to disable this behavior. </p>
```

PR #3946, opened 2024-07-05
Update non-text-contrast.html

mbgower:
Changing absolutely nothing. Just saving the file, which is causing only formatting changes in the code file, and no changes whatsoever in the output. so that the changes in my next commit are simple to view


patrickhlauke:
somehow, just looking over the file, it's unclear what additions/changes you've actually made. in particular, i naively searched for "hover" and couldn't spot "two paragraphs on hover state".

as something seems to have gone wrong here, I'd suggest starting with a fresh branch/PR.

PR #3952, opened 2024-07-08
Clean up gh-pages

kfranqueiro:
The `gh-pages` branch contains many files/directories that haven't been actively updated in many years. This PR removes everything that is not currently relevant to the deploy action on the main branch.

In addition to removing old files, this also updates the following:

- Adds missing style rules to `guidelines.css`
- Redirects `21/guidelines/index.html` to `guidelines/` (like `22/guidelines/index.html` already does to `guidelines/22/`), rather than being a slightly older copy of it
- Updates top-level `index.html` to include link to 2.2 guidelines, and to use newer URL to 2.1 guidelines

Preview visible at https://kfranqueiro.github.io/wcag/

(Note that the URLs in the redirect pages mentioned above are hard-coded to the upstream repo)

This PR intentionally does _not_ clear outdated files from under `techniques` and `understanding` (since it seems the existing deploy process never removes files to clean up what has been removed from the repository). This can be handled via a deploy script update at a later point in time.


kfranqueiro:
cc @alastc @mbgower @bruce-usab FYI, RE discussion earlier today

iadawn (commented):
Looks good to me. Couple of things:

* Do we need a .gitignore on _site? Might just have been because I switched branch and my directory hadn't been cleared out.
* Is the ```requirements``` directory doing anything?

kfranqueiro:
>     * Do we need a .gitignore on _site? Might just have been because I switched branch and my directory hadn't been cleared out.

`_site` was added to `.gitignore` in #3917 (the first Eleventy PR).

>     * Is the `requirements` directory doing anything?

I'm honestly not sure what it does, but the existing github workflow actively pushes to that directory. I preserved everything that the workflow touches.

kfranqueiro:
@iadawn Another thing to flag, since I saw your edit to index.html: The only files this cleanup retains are related to 2.1 or 2.2.

Should I restore the WCAG20 and wcag20/wcag21 directories on this branch? (I'm not sure if anything actively references them or if we have any meaningful way of ever updating them.)

Or should we drop mention of 2.0 in index.html?

iadawn:
> `_site` was added to `.gitignore` in #3917 (the first Eleventy PR).

Ah ha... _site isn't in the gh-pages branch one. I don't think it is needed though since there isn't a 'build' in that branch. I think I just ended up with it because I swapped branches in GH Client.

> I'm honestly not sure what it does, but the existing github workflow actively pushes to that directory. I preserved everything that the workflow touches.

Hmm... I might need to look closer at that. My inclination is to remove it... mostly because I don't like mess!

iadawn:
> Should I restore the WCAG20 and wcag20/wcag21 directories on this branch? (I'm not sure if anything actively references them or if we have any meaningful way of ever updating them.)
> 
> Or should we drop mention of 2.0 in index.html?

There isn't an Editor's draft mentioned in [WCAG2.0](https://www.w3.org/TR/WCAG20/). That would have been my only concern. Can pick this up in our catchup

kfranqueiro:
> Should I restore the WCAG20 and wcag20/wcag21 directories on this branch? (I'm not sure if anything actively references them or if we have any meaningful way of ever updating them.)

Update: restored `wcag21/requirements` because it is referenced from the [recommendation](https://www.w3.org/TR/WCAG21/#requirements-for-wcag-2-1)

> Or should we drop mention of 2.0 in index.html?

Update: discussed further and removed specific mention of 2.0 since we don't currently target it in builds, and nothing in the 2.0 recommendation links to this repo's gh-pages.

PR #3708, opened 2024-02-22
Update base.xslt

mbgower:
Closes #3684 
Fixing typo in email address

PR #3957, opened 2024-07-10
Update base.xslt

fstrr:
Fixing typo in email address in WCAG 2.0 content

### DIFF
--- a/xslt/base.xslt
+++ b/xslt/base.xslt
@@ -339,7 +339,7 @@
 				<h2> Help improve this page </h2>
 			</header>
 			<div class="box-i">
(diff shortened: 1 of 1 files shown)


kfranqueiro:
This is a duplicate of #3708 which was already merged. As a result, the commit on this pull request collapses to an empty commit when cherry-picked against `main`.

fstrr:
Interesting. [The bug is still there on the 2.1 content](https://www.w3.org/WAI/WCAG21/Understanding/captions-live.html), which is why I created a PR.

kfranqueiro:
The 2.1 informative content is based on the `WCAG-2.1` branch, separate from `main`, which hasn't had any updates pushed in a long time. This particular issue can be lumped into the overarching task I'm on right now of investigating backporting applicable content fixes from 2.2 to 2.1.

Issue #3787, opened 2024-04-15
1.4.4: Resize text: acceptability of a "link to a full version" 


dan-tripp-siteimprove:
Over at the ACT rules group I [raised the question](https://github.com/act-rules/act-rules.github.io/issues/2163) of whether and how a "link to a full version" would cause clipped text (which would otherwise fail) to pass SC 1.4.4.  

In [SC 1.4.4's Understanding document](https://www.w3.org/WAI/WCAG22/Understanding/resize-text) I found these cases where it does pass: 
1. "... in Web mail applications the subject column may not be wide enough to accommodate every possible subject header, but activating the subject header takes the user to the full message with the full subject header."
2. "In Web-based spreadsheets, cell content that is too long to be displayed in a column can be truncated, and the full content of the cell is available to the user when the cell receives focus."
3. "The content of a user interface component may also become too wide in user interfaces where the user can resize the column width. In this type of user interface component, line wrapping is not required; truncation is acceptable if the component's full content is available on focus or after user activation...."

Cases 1 and 2 are about user-supplied content which has no maximum length.  Case 3 is about a user-specified column width.  I'm thinking of another case, which doesn't fit into those categories, and which I'll call "Case 4": where the content is static and supplied by the author of the web page.  There are no resizable columns.  There is just some text which, at 200% zoom, becomes clipped, and a "link to a full version" appears.  

Does Case 4 pass SC 1.4.4?  

Whatever the answer turns out to be, I would like to propose a new (Sufficient or Failing) technique for Case 4, to make it clear.

mraccess77:
Based on the related discussions for SC 1.4.10 and examples from 1.4.4 I would generally consider that example to pass the criterion.

dan-tripp-siteimprove:
I just spent a few minutes searching for related discussions for SC 1.4.10, and I couldn't find any definitive ones.  If you happen to have links to some, let me know.  Same for the [examples for 1.4.4](https://www.w3.org/WAI/WCAG22/Understanding/resize-text.html#examples ) - maybe I'm missing something.

dan-tripp-siteimprove:
I found this wording in [Understanding 1.4.10](https://www.w3.org/WAI/WCAG22/Understanding/reflow.html): 
* "The Success Criterion is met as long as all content and functionality are still fully available - either directly, or revealed via accessible controls, or accessible via direct links."
[...]

PR #3960, opened 2024-07-10
1.4.4 Resize Text: when truncation is acceptable

dan-tripp-siteimprove:
This PR is follow-up to #3787.


dan-tripp-siteimprove:
@dav-idc

Issue #3927, opened 2024-06-25
Empty sections in source documents (causing broken ToC links in XSLT output)


kfranqueiro:
Several source documents contain sections that are effectively empty, containing only a heading and possibly a placeholder instructions paragraph or list item.

Both the XSLT and Eleventy build processes clean these up, although in the case of the XSLT process it sometimes results in "phantom" table of contents links that lack a corresponding destination.

If this is something there is interest in cleaning up, I can probably work on getting some verbose output from the Eleventy process to point to what needs cleaning.

patrickhlauke:
I'd be up for a clean-up. It's the sort of wide-reaching, tedious task i quite enjoy sometimes

kfranqueiro:
Here's what I get with some temporary logging added to the Eleventy build process. (Only some of these cause empty ToC sections in the XSLT process)

<details>
  <summary>Expand for log</summary>
<pre>
./understanding/20/abbreviations.html: Empty subsection: advisory
./understanding/20/abbreviations.html: Empty subsection: failure
./understanding/20/audio-control.html: Empty subsection: advisory
./understanding/20/audio-control.html: Empty top-level section: resources
./understanding/20/audio-description-or-media-alternative-prerecorded.html: Empty subsection: failure
./understanding/20/audio-description-prerecorded.html: Empty subsection: failure
./understanding/20/audio-only-live.html: Empty subsection: advisory
./understanding/20/audio-only-live.html: Empty subsection: failure
./understanding/20/captions-live.html: Empty subsection: advisory
./understanding/20/captions-live.html: Empty subsection: failure
./understanding/20/bypass-blocks.html: Empty subsection: failure
./understanding/20/captions-prerecorded.html: Empty subsection: advisory
./understanding/20/compatible.html: Empty top-level section: advisory
./understanding/20/consistent-identification.html: Empty subsection: advisory
./understanding/20/consistent-identification.html: Empty top-level section: resources
./understanding/20/contrast-enhanced.html: Empty top-level section: examples
./understanding/20/distinguishable.html: Empty top-level section: advisory
./understanding/20/enough-time.html: Empty top-level section: advisory
./understanding/20/error-prevention-all.html: Empty top-level section: examples
./understanding/20/error-prevention-all.html: Empty top-level section: resources
./understanding/20/error-identification.html: Empty subsection: failure
./understanding/20/error-identification.html: Empty top-level section: resources
./understanding/20/error-prevention-legal-financial-data.html: Empty subsection: failure
[...]

PR #3961, opened 2024-07-10
Remove parsing related techniques

fstrr:
This PR removes all techniques that were solely to do with [4.1.1 Parsing](https://www.w3.org/WAI/WCAG22/Understanding/parsing.html).


kfranqueiro:
PSA: one of the projects I've begun planning for this week involves adding proper metadata (and support for it) to indicate obsolete techniques, so that we would no longer remove them from this repository, but instead add a message to them and avoid emitting cross-links to them from other techniques/understanding pages. That project would ultimately revert the changes in this PR, which has me wondering if we should hold off. However, I'm unsure what the timeline on that project will be, compared to the likely timeline of this PR being merged.

If we do go ahead with this PR, one comment: it might be worth removing the now-empty examples, resources, and techniques sections in `parsing.html`? (See also #3927)

fstrr:
> it might be worth removing the now-empty examples, resources, and techniques sections in parsing.html?

I might be imagining this, but I think I remember there were build issues if those sections weren't in the document, which is why I left them empty. If that's not the case, then removing them definitely makes sense.

Metadata project sounds very useful. Dealing with the "no deleting pages" policy will be useful.

kfranqueiro:
FYI, progress on obsolete technique support in the Eleventy build process is coming along on a [WIP branch](https://github.com/kfranqueiro/wcag/compare/kfranqueiro:kgf-11ty-21...kfranqueiro:kgf-11ty-obsolete), and this has been a great example to start testing with.

Some progress screenshots (apologies to dark mode users, including myself):

Table of contents with obsolete techniques omitted (e.g. F70 and F77 in this screenshot):
A portion of the Failures section of the Techniques table of contents, which does not list F70 and F77

Parsing page no longer rendering Techniques section, based on it being an obsolete SC:
A portion of the Parsing page, showing Benefits immediately followed by Test Rules, with no Techniques section in between

(When layered on top of other work I have in progress for supporting generating WCAG 2.1 docs, the techniques section will remain intact on the page when generating for 2.1, but will be removed when generating for 2.2.)

Links to obsolete techniques within other obsolete techniques' pages remain intact, just as they would in the current situation where pages removed from this repository remain in place on the WAI site. Additionally, obsolete pages get a note in their About box about the obsolete status:
The main heading and About this Technique box for H94, showing that its cross-links to H74 and H93, which are also obsolete, remain intact. Additionally, the box ends with a new paragraph, stating "Obsolete: This technique pertains to 4.1.1 Parsing, which was removed as of WCAG 2.2."

alastc:
Also, I think it also needs to remove the links to those techniques in the understanding documents... unless the new build system works differently?

kfranqueiro:
> Also, I think it also needs to remove the links to those techniques in the understanding documents... unless the new build system works differently?

In general, yes, and that's something I'm going to be keeping an eye on as I tackle other obsolete techniques. In this particular case, as far as I can tell, the parsing page is the only one that references these techniques, and my solution in this case is to suppress its techniques section in 2.2 docs since parsing is obsolete/removed to begin with.

kfranqueiro:
I have sent the PR for supporting obsolete techniques: #3975

I performed a diff of the output of that branch vs. this one, and the only difference between the two is that #3975 also removes the empty section I identified earlier above.

@fstrr, are you okay with closing this in light of the PR that will support the same behavior without removing the source files? (I'd like to avoid having to restore them in my PR if this gets merged)

mbgower:
Duplicate of #3975

PR #3965, opened 2024-07-15
Update README.md – Link to Translating Instructions

remibetin:
Point to https://www.w3.org/WAI/about/translating/wcag/ to avoid redundant/slightly different instructions for translating WCAG.


remibetin:
cc @iadawn @shawna-slh

PR #3969, opened 2024-07-16
Remove all Silverlight working examples

fstrr:
We don't need these any more.


patrickhlauke (approved):

fstrr:
Whoops! Thanks for spotting that shell script.

kfranqueiro:
Leaving a note: I discussed `working-examples` with Kevin this morning as it might relate to preserving obsolete techniques, and we are fine with removing the examples under the expectation that the techniques in question are already obsolete anyway (and will soon be marked as such - or in this case, already are). So +1 from this end.

bruce-usab:
Discussed on TF call.

Issue #3942, opened 2024-07-03
The glossary term "cognitive function test" should not start with a capital letter


mitchellevan:
### Currently

https://www.w3.org/TR/WCAG22/#dfn-cognitive-function-test

The glossary term appears as "Cognitive function test" with an initial capital letter C.

Most other terms are not capitalized.

### Proposed

The glossary term should appear as "cognitive function test" without an initial capital letter.

mitchellevan:
It's a rather minor issue of course. I noticed it causing "Cognitive function test" to stand apart in the generated table of contents for [WCAG2ICT](https://www.w3.org/TR/wcag2ict-22/).

PR #3978, opened 2024-07-19
Don't force term names to lowercase in definition lists

kfranqueiro:
This fixes a couple of issues in the build system that I noticed while reviewing #3038:

- A few term names (e.g. ASCII art) have uppercase letters, which the build was forcing to lowercase (this was apparently also true of the XSLT build, judging by e.g. https://www.w3.org/WAI/WCAG21/Understanding/non-text-content#dfn-ascii-art)
- Any linked terms containing newlines would not be recognized (none of these currently exist, but #3038 added one, which I worked around in that PR so it's not blocked on this)

I also noticed that the term "cognitive function test" includes a capital C, but doesn't seem to warrant one (typically the only term capitalization is for acronyms or proper nouns).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/pull/3978.html" title="Last updated on Jul 19, 2024, 5:53 PM UTC (acec01c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/3978/0440c75...acec01c.html" title="Last updated on Jul 19, 2024, 5:53 PM UTC (acec01c)">Diff</a>

### DIFF
--- a/11ty/CustomLiquid.ts
+++ b/11ty/CustomLiquid.ts
@@ -393,7 +393,7 @@ export class CustomLiquid extends Liquid {
       // where we have access to global data and the about box's HTML
       const $termLinks = $(termLinkSelector);
       const extractTermName = ($el: Cheerio<Element>) => {
-        const name = $el.text().trim().toLowerCase();
+        const name = $el.text().toLowerCase().trim().replace(/\s*\n+\s*/, " ");
         const term = termsMap[name];
         if (!term) {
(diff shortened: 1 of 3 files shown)


iadawn (approved):
Remove change in case for 'cognitive function test'. This is raised and addressed via #3942

Otherwise, fine.

kfranqueiro:
> Remove change in case for 'cognitive function test'. This is raised and addressed via #3942

Done, thanks for spotting that!

PR #3982, opened 2024-07-21
Update CCA link text/url

patrickhlauke:
This change simply updates/corrects the link target for the Colour Contrast Analyser

### DIFF
--- a/techniques/general/G145.html
+++ b/techniques/general/G145.html
@@ -151,7 +151,7 @@
       
          <ul>
             <li>
-                  <a href="https://www.paciellogroup.com/resources/contrastanalyser/">Contrast Analyser – Application</a>
+                  <a href="https://github.com/ThePacielloGroup/CCAe">TPGi Colour Contrast Analyser (CCA) – Application</a>
                 </li>
             <li>
                   <a href="http://juicystudio.com/services/luminositycontrastratio.php">Contrast Ratio Analyser - online service</a>
--- a/techniques/general/G148.html
+++ b/techniques/general/G148.html
@@ -38,7 +38,7 @@
       
          <ul>
             <li>
-                  <a href="https://www.paciellogroup.com/resources/contrastanalyser/">Contrast Analyser – Application</a>
+                  <a href="https://github.com/ThePacielloGroup/CCAe">TPGi Colour Contrast Analyser (CCA) – Application</a>
                 </li>
             <li>
                   <a href="http://juicystudio.com/services/luminositycontrastratio.php">Contrast Ratio Analyser - online service</a>
--- a/techniques/general/G17.html
+++ b/techniques/general/G17.html
@@ -155,7 +155,7 @@
       
          <ul>
             <li>
-                  <a href="https://www.paciellogroup.com/resources/contrastanalyser/">Contrast Analyser – Application</a>
+                  <a href="https://github.com/ThePacielloGroup/CCAe">TPGi Colour Contrast Analyser (CCA) – Application</a>
                 </li>
             <li>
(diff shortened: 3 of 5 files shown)


fstrr:
Would it be more useful to link to [the releases page](https://github.com/ThePacielloGroup/CCAe/releases) rather than the repo homepage?

patrickhlauke:
> Would it be more useful to link to [the releases page](https://github.com/ThePacielloGroup/CCAe/releases) rather than the repo homepage?

the homepage has the readme that explains what it's all about (but granted, for somebody unfamiliar with how github works, finding the releases may not be immediately obvious either)

bruce-usab (approved):

alastc:
And if anyone asks, the use of "Colour" is because it's the name of the product, so the US-spelling doesn't apply.

ronsmantpg:
@alastc for the same reason, it should be "Analyser". The Colour Contrast Analyser product home page at https://www.tpgi.com/color-contrast-checker/ is in the process of being corrected.

GreggVan:
why is "checker" being changed on the site to analyser?    and if it was analyser  -- then that was the name.   

I suggest that it stick with the name  Checker  and avoid all this. 

Also -- why is it called the Color/Colour contrast checker.       It is checking luminance, not color contrast.    We are not interested in whether the colors contrast  (and what does it mean for colors to contrast?  Just be suffiently different hues?) 

Looks like a tempest in a teapot -- but if we are going to update the name I  would suggest something that is accurate.  and if we can avoid names that have words that are spelled differently in different countries it would make it easier to find.     

best

ronsmantpg:
@GreggVan the name of the tool is not being changed. It has always been Colour Contrast Analyser. This has, over the years, led to confusion, including on TPGi's own product web page (at least _that_ is being corrected). The web address was changed from contrastanalyser to color-contrast-checker to try to remove some confusion (unsuccessfully), and maybe for SEO. Despite it being raised regularly, there are currently no plans to change the name.

patrickhlauke:
@GreggVan not quite sure what your angle is here. the tool is called "Colour Contrast Analyser". and has been called that since its creation. and this PR updates the link and the text that links to it to actually use the tool's name

GreggVan:
@patrickhlauke 
I don't know what you mean by "what's your angle".  Do people have "angles" when they comment?

I know what the tool's name is -- and how long it has had it.   I was the one who commissioned (paid for) the adding of the WCAG analysis to the tool long before TPG took it over and the refactored it etc.

RE my comments about the name

a) the TPG link (not ours)  is https://www.tpgi.com/color-contrast-checker/   So the TPG link is already "checker" not "analyser/analyzer".   Someone suggested we tell TPG to change the name in the link.    I suggested that if they changed the name of the tool to checker - it would match the link and solve the problem with different languages.

b) then I  noted that colour/color was also back and forth.   So to make it also use a word that was region-neutral and more accurate we might suggest they change the name to   Contrast Checker since it doesnt actually measure how much the colors (hue and lightness) contrast with other colors but how much the luminosity (lightness / darkness) of samples differ.   This would solve a lot of misconceptions about the tool and colors and contrast.    But I  have never pushed for it because it didnt seem like something people wanted to do.   So we will just have to keep explaining.   (funny how it is so hard to change things.   [Machiavelli has a great quote on change](https://www.azquotes.com/quote/531785)

RE tempest in teapot -
I  ended comment by saying that I thought it was just fine to leave things as they are -- that all this "we need to change the link" was a lot of airtime over something that was unlikely to change -- and the link was actually easier for people to understand (as was intended) than using a word people might misspell and not get the tool.     

So I am signing off on this thread --to not waste more airtime.   TPG will decide on the name and URL and we will just use what they use.  It is evident from @ronsmantpg 's comments that all this is already things they have thought about -- and decided.   So.   Ciao for now

Issue #2923, opened 2023-01-14
Test methods in Techniques are not aligned with the SC 1.4.4 Resize Text definition.


ArnaudDelafosse:
The SC 1.4.4 Resize Text says "[...] text can be resized [...] **up to** 200 percent [...]" and again in [the SC's Understanding](https://www.w3.org/WAI/WCAG21/Understanding/resize-text.html) "Content satisfies the Success Criterion if it can be scaled **up to** 200% [...]" which, to my understanding, means that the test should _theoretically_ be conducted in every incrementation (percentage) up to and including 200%.
In practice, most of us must probably test the browsers's own graphic zoom and/or text resizing incrementations (e.g. 110%, 120%, 133%, 150%, 170% and 200% in Firefox or 110%, 125%, 150%, 175% and 200% in Chrome).

However in [Technique G178](https://www.w3.org/WAI/WCAG21/Techniques/general/G178) and [Technique F94](https://www.w3.org/WAI/WCAG21/Techniques/failures/F94) (and it might also be elsewhere) it is said "Check that the text size can be increased to 200%" and "Check that the text [...] can be resized to at least 200%", possibly implying that testing **only** at 200% will be sufficient to conform to SC 1.4.4.

Is it just me or does that need to be clarified by using "...**up to** 200%" everytime it is needed?

detlevhfischer:
Since the normative text does not specify that the resizing has to be linear (i.e., it does not rule out that at certain viewport widths the CSS **reduces** font size slightly, leading to a more moderate increase than you would see without CSS changes to font size) the only thing that remains to be tested is if there is **any** setting (usually via Browser zoom) where 200% text size without loss of content or functionality can be achieved. (It has been noted that this may be counterproductive for text that is already large at the default size, but no watertight wording to modify the normative text has come forward so far,)

JAWS-test:
@detlevhfischer I am unsure if it is in accordance with the SC to require only 200% zoom, although the SC says otherwise. When the SC was written, as far as I know, there were no media queries that could respond to font size adjustment in the browser or OS with jumps in scaling, i.e. when SC 1.4.4 was written, it was assumed that resizing was linear. This has changed, but we would have to react to this by adapting SC or at least the understandings and techniques - and discuss beforehand in which form this adaptation should take place

ArnaudDelafosse:
[...]

PR #3986, opened 2024-07-22
Change 1.4.4 related mentions of "at least 200%" to "up to 200%"

patrickhlauke:
Closes https://github.com/w3c/wcag/issues/2923

### DIFF
--- a/techniques/failures/F94.html
+++ b/techniques/failures/F94.html
@@ -49,7 +49,7 @@
 							<li>on-page controls for resizing text.</li>
 						</ul>
 					</li>
-          <li>Check that the text resizes by one of the methods above, and can be resized to at least 200% of the default.</li>
+          <li>Check that the text resizes by one of the methods above, and can be resized up to 200% of the default.</li>
 				</ol>
 			</section>
 			<section class="test-results">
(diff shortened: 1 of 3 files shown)


giacomo-petri (approved):

detlevhfischer:
Hmm - I have my doubt that all readers will parse "up to 200%" correctly. It could be understood as meaning: "resize should have an effect, and you should be able to enlarge text to a maximum value of 200% (but not more)" - implying that resizing to 150% would be sufficient as well. I would just scrap the "up" for clarity.

GreggVan:
I  think "up to"  is perfectly understandable.  I  don't  think anyone in any position of authority to judge would think that something must be "adjustable up to 200%"  would be satisfied by something that stopped at 150%.    And also no such judge would think it should be capped at 200%  unless you said  "up  to a max of 200%" or even "up to a max of 200% and not more".      This is not to say that "all readers" would  say this because it is not possible to say that "all readers" will read anything correctly.  But that should not be our goal.    We should make it as understandable as possible.  So if you have alternate wording that you think would be better -- it is always great to hear it.

patrickhlauke:
> I have my doubt that all readers will parse "up to 200%" correctly.

it's the exact wording used in the normative SC text. not disagreeing, just pointing out that if there's confusion, it already starts with the actual SC itself

detlevhfischer:
@GreggVan I just gave my take as a non-native speaker with a good grasp of English. The guidance is read by all kinds of people (not just native speakers) who implement stuff, often without a deep understanding of accessibility  - so that "position of authority to judge" thing strikes me as odd. Anyway, if others think there is no problem, that's fine. Personally I am more comfortable with "at least 200%".

bruce-usab (approved):

alastc:
From the meeting: It's in the SC, the understanding document is quite consistent on "up to", so we think the test-procedure should reflect the SC text.

PR #3995, opened 2024-07-27
Fix mentions to HTML to Platform Accessibility APIs Implementation Guide

momdo:
The [HTML to Platform Accessibility APIs Implementation Guide](https://www.w3.org/TR/html-aapi/) says "See [HTML Accessibility API Mappings 1.0](https://www.w3.org/TR/html-aam-1.0/) instead".

So, this PR tries to replace to HTML-AAM or remove HTML-AAPI.


techniques/aria/ARIA5.html
```diff
@@ -85,9 +85,6 @@ <h3>A slider</h3>
          <li>
                <a href="https://www.w3.org/TR/wai-aria/#states_and_properties">Accessible Rich Internet Applications (WAI-ARIA), Supported States and Properties</a>
             </li>
-         <li>
```
patrickhlauke, line 88: is it intentional that the link was removed completely?

momdo: @patrickhlauke 

> is it intentional that the link was removed completely?

Yes. There are already references to WAI-ARIA, and I think it is unnecessary to mention HTML-AAM in ARIA5.

patrickhlauke: cool cool, just checking

patrickhlauke (approved):

PR #3975, opened 2024-07-18
Support marking obsolete techniques and restore removed ones

kfranqueiro:
This PR supersedes #3961; it accomplishes the same output, plus resolves a leftover empty section remaining in that PR.

Related issues: #3651, #3758

## Note to WG

I would appreciate eyes on `obsoleteMessage` wording, seen at the end of each technique's About box (or at the top of HTML files in this PR) for restored obsolete techniques:

- Parsing techniques (obsolete in 2.2): [F70](https://deploy-preview-3975--wcag2.netlify.app/techniques/failures/f70), [F77](https://deploy-preview-3975--wcag2.netlify.app/techniques/failures/f77), [G134](https://deploy-preview-3975--wcag2.netlify.app/techniques/general/g134), [G192](https://deploy-preview-3975--wcag2.netlify.app/techniques/general/g192), [H74](https://deploy-preview-3975--wcag2.netlify.app/techniques/html/h74), [H75](https://deploy-preview-3975--wcag2.netlify.app/techniques/html/h75), [H93](https://deploy-preview-3975--wcag2.netlify.app/techniques/html/h93), [H94](https://deploy-preview-3975--wcag2.netlify.app/techniques/html/h94) (all should have the same message)
- Other previously-removed techniques: [H4](https://deploy-preview-3975--wcag2.netlify.app/techniques/html/h4), [H35](https://deploy-preview-3975--wcag2.netlify.app/techniques/html/h35), [H46](https://deploy-preview-3975--wcag2.netlify.app/techniques/html/h46), [H60](https://deploy-preview-3975--wcag2.netlify.app/techniques/html/h60), [H73](https://deploy-preview-3975--wcag2.netlify.app/techniques/html/h73), [H45](https://deploy-preview-3975--wcag2.netlify.app/techniques/html/h45), [H59](https://deploy-preview-3975--wcag2.netlify.app/techniques/html/h59), [H70](https://deploy-preview-3975--wcag2.netlify.app/techniques/html/h70), [SCR37](https://deploy-preview-3975--wcag2.netlify.app/techniques/client-side-script/scr37), [G187](https://deploy-preview-3975--wcag2.netlify.app/techniques/general/g187), [F87](https://deploy-preview-3975--wcag2.netlify.app/techniques/failures/f87)
[...]

PR #4006, opened 2024-08-06
Remove the frameset example in G128

momdo:


patrickhlauke:
in light of https://github.com/w3c/wcag/pull/3975 we probably want to keep it but mark it as obsolete instead /cc @kfranqueiro

patrickhlauke (approved):

bruce-usab (approved):
Thank you @momdo , is straightforward edit.

bruce-usab:
Discussed on backlog call 9/6.  @scottaohara may want to open a new issue for surfacing technics for obsolete technologies.  @kfranqueiro might already have that covered.  Regardless, the techniques page (for obsolete technologies) do have same public-facing URLs as before and there is no plan to delete them.  General consensus is that we NOT to link to them from Understanding articles, but that they still have some utility.

PR #4021, opened 2024-08-16
Add note about the scope of 1.4.5 Images of Text (images used *rather than* text) and example

patrickhlauke:
Closes: https://github.com/w3c/wcag/issues/3755

This is an alternative to https://github.com/w3c/wcag/pull/3773 following discussions in WCAG 2.x backlog meeting from 9 August meeting. Filed here as a separate clean PR, rather than trying to modify @giacomo-petri's PR


bruce-usab:
Reviewed on TF call.

giacomo-petri (approved):

understanding/20/images-of-text.html
```diff
@@ -130,139 +120,54 @@ <h2>Examples of Images of Text</h2>
          <dt>Customizable font settings in images of text</dt>
          <dd>A Web site allows users to specify font settings and all images of text on the site
                are then provided based on those settings.</dd>
+         <dt>The text replicates an image of text</dt>
```
detlevhfischer, line 123: > The text replicates an image of text

I have read "replicates" the same way as in the inverse situation of "image as text" where the image replactes the text. The text below than makes clear the content passes becasue equivalent text is rendered below. The heading does not say this clearly, though.

detlevhfischer (commented):
See inline comment.

bruce-usab:
Discussed on backlog call 9/6, Mike assigned to himself.

understanding/20/images-of-text.html
mbgower, line 123: This addresses @detlevhfischer concern
```suggestion
         <dt>The same information is given in text and an image of text</dt>
```

bruce-usab:
On TF today, with much due deliberation, we changed "The text replicates an image of text" to "The text in an image is reproduced in text" as being better plain language.  Consistent with other recent nuanced discussion, we deliberately decided not to link to the defined term.

bruce-usab (approved):

understanding/20/images-of-text.html
```diff
@@ -130,139 +120,54 @@ <h2>Examples of Images of Text</h2>
          <dt>Customizable font settings in images of text</dt>
          <dd>A Web site allows users to specify font settings and all images of text on the site
                are then provided based on those settings.</dd>
+         <dt>The text in an image is reproduced in text</dt>
```
bruce-usab, line 123: Missing period.  Sorry to say, "reproduced in text" reads awkwardly.

> The text in an image is also provided as text.

patrickhlauke: though this goes slightly against the point made here https://github.com/w3c/wcag/pull/4021#issuecomment-2379684390

patrickhlauke:
```suggestion
         <dt>The text in an image is also provided as text.</dt>
```

PR #4040, opened 2024-08-25
Amend definition of motion animation to *not* exclude blurring

### LINKED ISSUE #3949: WCAG Example suggests "blurring" does not affect animation triggering vestibular motion perception (but it can)
The definition of motion animation excludes "blurring", which appears incorrect to me.
https://www.w3.org/TR/WCAG22/#dfn-motion-animation

> motion animation
> addition of steps between conditions to create the illusion of movement or to give a sense of a smooth transition
> 
> EXAMPLE
> For example, an element which moves into place or changes size while appearing is considered to be animated. An element which appears instantly without transitioning is not using animation. Motion animation does not include changes of color, **_blurring_**, or opacity which do not change the perceived size, shape, or position of the element.

Blurring is commonly uses to trigger a "depth of field" effect (simulating an artifact of the eye's iris or camera aperature)... The resulting illusion is one of animation depth changes (in the z index), which is known to be a common trigger for vestibular motion disorders. 

I would suggest the word "blurring" be removed from this example text, or perhaps changed to "dissolving"??? Though keeping both dissolving and opacity may be redundant, so my preference is just to remove the word "blurring."

patrickhlauke:
Closes #3949

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/pull/4040.html" title="Last updated on Oct 4, 2024, 3:07 PM UTC (abe30f6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/4040/a9e49c0...abe30f6.html" title="Last updated on Oct 4, 2024, 3:07 PM UTC (abe30f6)">Diff</a>


guidelines/terms/21/motion-animation.html
```diff
@@ -1,7 +1,5 @@
 <dt><dfn id="dfn-motion-animation">motion animation</dfn></dt>
 <dd>
-   					
    <p>addition of steps between conditions to create the illusion of movement or to give a sense of a smooth transition</p>
-	<aside class="example"><p>For example, an element which moves into place or changes size while appearing is considered to be animated. An element which appears instantly without transitioning is not using animation. Motion animation does not include changes of color, blurring, or opacity which do not change the perceived size, shape, or position of the element.</p></aside>
-   				
+	<aside class="example"><p>For example, an element which moves into place or changes size while appearing is considered to be animated. An element which appears instantly without transitioning is not using animation. Motion animation does not include other changes – such as changes of color or opacity – which do not alter the perceived size, shape, position, or distance/depth of the element.</p></aside>
```
mbgower, line 4:
```suggestion
	<aside class="example"><p>For example, an element which moves into place or changes size while appearing is considered to be animated. An element which appears instantly without transitioning is not using animation. Motion animation does not include other changes – such as changes of color or opacity – which do cause a motion effect through the perceived size, shape, position, or distance/depth of the element.</p></aside>
```

gundulaniemann, line 4: I suggest the following:
```suggestion
	<aside class="example"><p>For example, an element which moves into place or changes size while appearing is considered to be animated. An element which changes instantly without transition is not using animation. Motion animation does not include other changes even with transition – such as changes of color or opacity – which do not alter the perceived size, shape, position, or distance/depth of the element.</p></aside>
```

patrickhlauke: I'd actually propose a further tweak/change

```suggestion
	<aside class="example"><p>For example, an element which moves into place or changes size while appearing is considered to be animated. An element which appears or changes instantly, without transition/animation steps, is not using animation.</p><p>Motion animation does not include changes – such as changes of color or opacity – which do not alter the perceived size, shape, position, or distance/depth of the element.</p></aside>
```

bruce-usab (approved):

bruce-usab:
Discussed on TF call.  IMHO this is a nice minimal edit for the issue raised.

detlevhfischer:
This is a bit odd in that the blurring from the PR title is nowhere mentioned in the amended text (opacity and color change is). As to blurring, one could also argue that transitions from sharp to strong blur to sharp and back to sharp etc. could be quite distracting if not just an initial effect, the more so the large the image is. Admittedly haven't seen that out there yet.

patrickhlauke:
> This is a bit odd in that the blurring from the PR title is nowhere mentioned in the amended text (opacity and color change is)

before, the list of excluded effects included blurring. this PR removes the mention of blurring, so does what the PR title says?

detlevhfischer:
@patrickhlauke argh... not sure how I could misread this.

patrickhlauke:
> My two cents - blurred images give me a headache and I am not able to look at them. This is a big issue with blurred remote meeting backgrounds but also some styles of photography.

@mraccess77 while that is a wider, more general topic, this proposed change here should be welcome for you then because it does now mean that where blurring is used on interaction, it *can* now **fail** Animation from Interactions (while before it seemed to imply that blurring was exempt, just like changes in colour and opacity)

guidelines/terms/21/motion-animation.html
```diff
@@ -1,7 +1,5 @@
 <dt><dfn id="dfn-motion-animation">motion animation</dfn></dt>
 <dd>
-   					
    <p>addition of steps between conditions to create the illusion of movement or to give a sense of a smooth transition</p>
-	<aside class="example"><p>For example, an element which moves into place or changes size while appearing is considered to be animated. An element which appears instantly without transitioning is not using animation. Motion animation does not include changes of color, blurring, or opacity which do not change the perceived size, shape, or position of the element.</p></aside>
-   				
+	<aside class="example"><p>For example, an element which moves into place or changes size while appearing is considered to be animated. An element which appears or changes instantly, without transition/animation steps, is not using animation.</p><p>Motion animation does not include changes – such as changes of color or opacity – which do not alter the perceived size, shape, position, or distance/depth of the element.</p></aside>
```
mbgower, line 4:
```suggestion
	<aside class="example"><p>For example, an element which moves into place or changes size while appearing is considered to be animated. An element which appears or changes instantly, without transition/animation steps, is not using animation.</p><p>Motion animation does not include changes – such as changes of color or opacity – that do not alter the perceived size, shape, position, or distance/depth of the element.</p></aside>
```

scottaohara (approved):

guidelines/terms/21/motion-animation.html
```diff
@@ -1,7 +1,5 @@
 <dt><dfn id="dfn-motion-animation">motion animation</dfn></dt>
 <dd>
-   					
    <p>addition of steps between conditions to create the illusion of movement or to give a sense of a smooth transition</p>
-	<aside class="example"><p>For example, an element which moves into place or changes size while appearing is considered to be animated. An element which appears instantly without transitioning is not using animation. Motion animation does not include changes of color, blurring, or opacity which do not change the perceived size, shape, or position of the element.</p></aside>
-   				
+	<aside class="example"><p>For example, an element which moves into place or changes size while appearing is considered to be animated. An element which appears or changes instantly, without transition/animation steps, is not using animation.</p><p>Motion animation does not include changes – such as changes of color or opacity – that do not alter the perceived size, shape, position, or distance/depth of the element.</p></aside>
```
cookiecrook, line 4:
```suggestion
	<aside class="example"><p>For example, an element which moves into place or changes size while appearing is considered to be animated. An element which appears or changes instantly, without transition/animation steps, is not using animation.</p><p>Motion animation does not include changes – such as changes of color or opacity – that do not alter the perceived size, shape, position, or distance/depth of the element to the viewer.</p></aside>
```

cookiecrook (changes requested):
Adding in my original proposed text that clarifies the it's not about screen motion in a vacuum, but about the perspective of the primary user. IMO, clarifying this is critical in the context of emerging modalities like XR, in which motion may be perceivable from another observable camera angle, but not the primary.

bruce-usab:

guidelines/terms/21/motion-animation.html
patrickhlauke: i'll admit, that reads a bit stilted to me. isn't it already implicit in the fact that this talks about "perceived..." as it's clearly being perceived by a user/viewer?

nattarnoff: Agree, the use of perceived makes "to the viewer" redundant.

alastc: In the context of cookiecrook's comment: "this is critical in the context of emerging modalities like XR, in which motion may be perceivable from another observable camera angle, but not the primary."
I can see why it isn't as redundant as it appears from just the code change.

bruce-usab:
Discussed on backlog call 10/4.  Added "to the viewer" at end of sentence to address concern raised by @cookiecrook.

patrickhlauke:
@bruce-usab wondering whatever happened to this PR? i seem to remember we agree on it, but I see it's still open and was under "No status" column. did this accidentally fall between the cracks? Moving it back to "Drafted"

bruce-usab:
Discussed on backlog call 1/17.  Moved to Ready to review for edit after 9/13/2024.

mbgower:
@GreggVan stated on today's Working Group call that he believes the last sentence of the example in [this definition](https://www.w3.org/TR/WCAG21/#dfn-motion-animation) is not an example.

The example has existed in its current form since 2.1 was published in 2018. This change does not alter the designation of this paragraph; it is marked up an example (see image below), and as such is clearly non-normative (as I documented in my [prior comment](https://github.com/w3c/wcag/pull/4040#issuecomment-2616702474)).
image


I can see a case to be made that it is a note, rather than an example. A note is _also_ informative, so I don't really see a big deal leaving it as it is. However, if we want to redesignate that, a new issue should be opened for that so it can be tackled separately. But in regard to this PR, it is scope creep; and also orthogonal.

bruce-usab:
Discussed on backlog call 2/7.  While examples are not normative, that is not relevant in this case.  The use of “for example” is not the same as having a list of examples.

Regardless, the proposed change is appropriate.  No change in status.

PR #3939, opened 2024-07-01
Make Benefits consistently a top-level section with h2 in source files

kfranqueiro:
Fixes #3924.

Also addresses the Label in Name instance in #3922.

The diff will be much easier to compare with whitespace ignored (`?w=1`).

This amends all instances of the Benefits section in Understanding pages to be a top-level section with a level-2 heading.

I've tested this with both the XSLT and Eleventy build processes to confirm this doesn't actually change any content, since both build processes already ensure Benefits is a top-level section with level-2 heading. The one exception is the removal of the redundant first Benefits section in Understanding 2.5.3 Label in Name, which only meaningfully affects the XSLT process, as the Eleventy process already worked around that issue.

Note: I don't know what `identify-changes.html` is supposed to be, but I updated it regardless; it seems to only contain template boilerplate and is not linked from anywhere.

PR #4042, opened 2024-08-27
Re-apply #3939 cleanly against the latest content on main

kfranqueiro:
When #3939 was merged, conflicts were not resolved correctly, resulting in loss of #2455's update to 2.1.4 and loss of #3939's original update to 2.3.3.

This PR cleanly re-applies my original changes from #3939, restoring those other lost updates.

Screenshot of 2.1.4 in incorrect state from the bad merge (visible on the [netlify main branch preview](https://wcag2.netlify.app/understanding/character-key-shortcuts) until this fix is merged):

image

Screenshot of 2.1.4 on this branch, with other changes restored (visible on [this PR's preview](https://deploy-preview-4042--wcag2.netlify.app/understanding/character-key-shortcuts)):

image

2.3.3 was not visibly affected, because the build already corrects the problem case that #3939 cleans up.


understanding/21/character-key-shortcuts.html
```diff
@@ -19,27 +19,29 @@ <h2>In brief</h2>
...
-				<li>Keyboard-only users who have dexterity challenges can also be prone to accidentally hitting keys. Those users would be able to avoid problematic single character shortcuts by turning them off or modifying them to include at least one non-character key.</li>
-				<li>Allowing <em>all</em> shortcut keys to be remapped can help users with some cognitive disabilities, since the same shortcuts can be assigned to perform the same actions across different applications.</li>
-			</ul>
-
+      <p>The intent of this Success Criterion is to reduce accidental activation of keyboard shortcuts. Character key shortcuts work well for many keyboard users. However, they can be inappropriate and frustrating for speech input users whose dictation is interpreted as strings of letters, and for keyboard users who are prone to accidentally hit keys. To rectify this issue, authors need to allow users to turn off or reconfigure shortcuts that are made up of only character keys.</p>
```
patrickhlauke, line 22:
```suggestion
      <p>The intent of this Success Criterion is to reduce accidental activation of keyboard shortcuts. Character key shortcuts work well for many keyboard users. However, they can be inappropriate and frustrating for speech input users, whose dictation is interpreted as strings of letters, and for keyboard users who are prone to accidentally hit keys. To rectify this issue, authors need to allow users to turn off or reconfigure shortcuts that are made up of only character keys.</p>
```

understanding/21/character-key-shortcuts.html
```diff
@@ -19,27 +19,29 @@ <h2>In brief</h2>
...
+      <h3>Background on the mechanics of speech input</h3>
+      <p>Speech Input users generally work in a single mode where they can use a mix of dictation and speech commands. This works well because the user knows to pause before and after commands, and commands are usually at least two words long. So, for instance, a user might say a bit of dictation, such as "the small boat", then pause, and say a command to delete that dictation, such as "Delete Line". In contrast, if the user were to say the two phrases together without a pause, the whole phrase would come out as dictation (i.e., "the small boat delete line"). Although speech input programs often include modes that listen only for dictation or only for commands, most speech users use the all-encompassing mode all the time because it is a much more efficient workflow. It could decrease command efficiency significantly if users were to change to command mode and back before and after issuing each command.</p>
```
patrickhlauke, line 31:
```suggestion
      <p>Speech input users generally work in a single mode where they can use a mix of dictation and speech commands. This works well because the user knows to pause before and after commands, and commands are usually at least two words long. So, for instance, a user might say a bit of dictation, such as "the small boat", then pause, and say a command to delete that dictation, such as "Delete Line". In contrast, if the user were to say the two phrases together without a pause, the whole phrase would come out as dictation (i.e., "the small boat delete line"). Although speech input programs often include modes that listen only for dictation or only for commands, most speech users use the all-encompassing mode all the time because it is a much more efficient workflow. It could decrease command efficiency significantly if users were to change to command mode and back before and after issuing each command.</p>
```

patrickhlauke (approved):

Issue #1263, opened 2020-08-07
Title of F42 indicates it applies to SC 4.1.2 - but 4.1.2 is not listed in the technique


mraccess77:
4.1.2 is not mentioned in the "This technique relates to:" section
https://www.w3.org/WAI/WCAG21/Techniques/failures/F42.html

4.1.2 is mentioned at bottom in check section.

PR #4052, opened 2024-09-03
Add F42 as failure for 4.1.2

patrickhlauke:
Closes https://github.com/w3c/wcag/issues/1263

(also cleans up the markup of 4.1.2 understanding, removing all unnecessary/excessive line breaks, fixes the invalid use of `</meta>`, and removes the unnecessary techniques link text which is autopopulated at publish time)

### DIFF
--- a/understanding/20/name-role-value.html
+++ b/understanding/20/name-role-value.html
@@ -1,50 +1,41 @@
 <!DOCTYPE html>
 <html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml">
 <head>
-   <meta charset="UTF-8"></meta>
+   <meta charset="UTF-8"/>
    <title>Understanding Name, Role, Value</title>
    <link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove"/>
 </head>
 <body>
    <h1>Understanding Name, Role, Value</h1>
-   
    <section id="brief">
 		<h2>In brief</h2>
 		<dl>
          <dt>Goal</dt><dd>People using assistive technology understand all components.</dd>
          <dt>What to do</dt><dd>Give components correct names, roles, states, and values.</dd>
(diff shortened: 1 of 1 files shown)


patrickhlauke:
See deploy previews https://deploy-preview-4052--wcag2.netlify.app/techniques/failures/f42 and https://deploy-preview-4052--wcag2.netlify.app/techniques/failures/f42

detlevhfischer:
At to [F42](https://deploy-preview-4052--wcag2.netlify.app/techniques/failures/f42#example-4-scripting-a-div-element), section "Example 4: Scripting a div element": the code example at the end has no `tabindex="0"` but the example intro claims the author has made the pseudo-link keyboard-accessible.
Should this technique have a note that if pseudo-links that are fully keyboard accessible **and** carry a role="link", they would not fail?

I can create a PR to implement that if people think it would be worthwhile doing.

patrickhlauke:
> At to [F42](https://deploy-preview-4052--wcag2.netlify.app/techniques/failures/f42#example-4-scripting-a-div-element), section "Example 4: Scripting a div element": the code example at the end has no tabindex="0" but the example intro claims the author has made the pseudo-link keyboard-accessible.

urgh, good catch. suggest making a small PR for that fix

> Should this technique have a note that if pseudo-links that are fully keyboard accessible and carry a role="link", they would not fail?

the technique already has the note about "It is possible to use the ARIA role attribute to identify an anonymous element as link control for assistive technologies …". the "fully keyboard accessible" part would be relevant for the 2.1.1 aspect, yes. Suggest amending/expanding that note to clarify that part yes (can be in the same PR as the above)

kfranqueiro (approved):
I diffed builds before/after the changes in this PR to confirm that nothing unexpected happened due to the formatting cleanup.

PR #4056, opened 2024-09-06
Update 3.3.2 Labels or Instructions Understanding Doc

### LINKED ISSUE #3984: Restore the need for accuracy/correctness to 3.3.2 Labels or Instructions
A recently published update to the [Understanding document for 3.3.2 Labels or Instructions](https://www.w3.org/WAI/WCAG22/Understanding/labels-or-instructions.html) has made changes that effectively render the success criterion pointless. 

In an attempt to disambiguate the slight overlap between 3.3.2 Labels or Instructions and 2.4.6 Headings and Labels, #1792 removed several references to the need for _"clear and unambiguous"_ labels and instructions. The [justification ](https://github.com/w3c/wcag/issues/3795#issuecomment-2071907768) was that 2.4.6 concerns the quality of headings and labels, whereas 3.3.2 concerns the mere presence of labels or instructions (regardless of how good/comprehensive/accurate they are). This, I feel, was a misstep. 

As others and I commented [elsewhere](https://github.com/w3c/wcag/issues/3795), these changes go against the intent (if not the normative wording) of 3.3.2. 

The normative text - _"Labels or instructions are provided when content requires user input"_ - is unhelpfully terse. 
However, the stated goal that _"Users know what information to enter"_ and the intent _"to have content authors present instructions or labels that identify the controls in a form so that users know what input data is expected"_ is more indicative of the intended purpose of 3.3.2. 

By removing any need for clarity or quality or accuracy in 3.3.2, then what is the benefit of conforming to this SC?

(Also, if any requirement for clarity or descriptiveness is punted squarely to 2.4.6, why list [G131: Providing descriptive labels](https://www.w3.org/WAI/WCAG22/Techniques/general/G131) as a sufficient technique?) 

Picking up from [#3795 (comment)](https://github.com/w3c/wcag/issues/3795#issuecomment-2071907768), 3.3.2 now effectively says that anything goes, as long as it looks like a label or instruction. How does that help users?

Without any requirement for accuracy or correctness...
- a nonsense label would pass 3.3.2: `<label for="tel">XWPSWTLNMZ</label>`
- an incorrect label would pass 3.3.2: `<label for="tel">Name</label>`
- an ambiguous label would pass 3.3.2: `<label for="consent">I agree</label>`

Now, you could argue that these examples impact everyone and therefore can be considered "bad UX" and beyond the scope of WCAG. But if that's the case, why have this success criterion at all? Who does the _mere provision of labels or instructions_ actually benefit? 

It would be like saying that 1.1.1 Non-text Content should only be about the presence of a text alternative, not about whether the text alternative accurately describes the content. Admittedly, 1.1.1 has the additional clause of _"a text alternative that serves the equivalent purpose"_ and 1.1.1 also hasn't been split over two closely related success criteria, but that's where we are. In my opinion, 3.3.2 would greatly benefit from a similar qualifying clause, if not in the normative text then in the Understanding document.  

At the very least, accuracy should be an explicit expectation of 3.3.2 (in that, the label or instruction _must_ identify/explain what the control does). Whether it does that clearly or sufficiently descriptively is then a matter for 2.4.6 (although I accept that an accurate/correct label will likely be sufficiently descriptive for 2.4.6 in most cases).

In terms of specific actions, I wouldn't necessarily reverse the changes in #1792, but instead make an explicit expectation of accuracy. So, for example:  

- **What to do:** Provide labels or instructions for inputs **[that accurately describes what data is expected]**.
- "The intent of this Success Criterion is to have content authors present instructions or labels that **[accurately]** identify the controls in a form so that users know what input data is expected."
- "While this Success Criterion requires that controls and inputs have **[accurate]** labels or instructions, whether or not labels (if used) are sufficiently clear or descriptive is covered separately by 2.4.6: Headings and Labels."

And perhaps even an additional sentence to emphasize the need for accuracy: 
- "Although WCAG 3.3.2 focuses primarily on ensuring that labels or instructions are provided, it is still essential that these labels or instructions are accurate."

davidofyork:
An attempt at restoring the need for accuracy/correctness to 3.3.2 Labels or Instructions, as discussed in #3984.

Inserts multiple references to _accurate_ labels and instructions, expands upon one of the examples, and addresses the overlap with 2.4.6. 

resolves #3984

### DIFF
--- a/understanding/20/labels-or-instructions.html
+++ b/understanding/20/labels-or-instructions.html
@@ -12,7 +12,7 @@
 		<h2>In brief</h2>
 		<dl>
          <dt>Goal</dt><dd>Users know what information to enter.</dd>
-         <dt>What to do</dt><dd>Provide labels or instructions for inputs.</dd>
+         <dt>What to do</dt><dd>Provide accurate and descriptive labels or instructions for inputs.</dd>
          <dt>Why it's important</dt><dd>Everyone, especially those with cognitive disabilities, will know how to respond.</dd> 
       </dl>
 
@@ -21,12 +21,12 @@
    <section id="intent">
       <h2>Intent of Labels or Instructions</h2>
       
-      <p>The intent of this Success Criterion is to have content authors present instructions
-         or labels that identify the controls in a form so that users know what input data
+      <p>The intent of this Success Criterion is to have content authors present accurate instructions
+         or labels that precisely and clearly identify the controls in a form so that users know what input data
          is expected. In the case of radio buttons, checkboxes, comboboxes, or similar controls
-         that provide users with options, each option must have an appropriate label so that
+         that provide users with options, each option must have an accurate and descriptive label so that
          users know what they are actually selecting. 
-         Instructions or labels may also specify data formats for data entry fields, especially
+         Instructions or labels may also accurately specify data formats for data entry fields, especially
          if they are out of the customary formats or if there are specific rules for correct
          input. Content authors may also choose to make such instructions available to users
          only when the individual control has focus especially when instructions are long and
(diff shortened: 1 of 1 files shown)


davidofyork:
Judging by the response to this PR, both here and in the ongoing discussion in #3984, I would ignore my attempts to pepper the document with references to "clarity" and "accuracy", and instead focus on my suggestion to provide some sort of acknowledgement/disclaimer that accuracy is implied. It doesn't even have to reference 2.4.6 at all and could just state: 

_“While this Success Criterion requires that controls and inputs have labels or instructions, it is crucial that these labels or instructions are not only present but also accurate.”_

This wouldn’t enforce accuracy, it shouldn't impact the attempt to address the overlap with 2.4.6, but it would at least acknowledge and restate the goal and intent of 3.3.2.

bruce-usab:
This PR (and related Issues) discussed on TF call today. Consensus on call is that this issue is best addressed by updating Understanding 2.4.6. @mbgower volunteered to review and propose something.

alastc:
Given that we are unlikely to get agreement on this PR, and Mike will look at an understanding doc update, I think we can close this PR.

PR #4060, opened 2024-09-07
Example icon font img role

patrickhlauke:
This branch never had a pull request made. Suggest we either take it, or delete the PR and branch.


patrickhlauke:
actually, this seems to already exist https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA24

Closing (and deleting the branch)

PR #4061, opened 2024-09-07
Create aria24.html

patrickhlauke:
This branch never had a pull request made. Suggest we either take it, or delete the PR and branch.

If we do think it's a valuable new technique, it'll need to be renumbered, as there is already an ARIA24

### DIFF
--- a/techniques/aria/aria24.html
+++ b/techniques/aria/aria24.html
@@ -1,0 +1,194 @@
+<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml" lang="en">
+	<head>
+		<title>Using ARIA to match the accessible name to the visible label</title>
+		<link rel="stylesheet" type="text/css" href="../../css/editors.css">
+	</head>
+	<body>
+		<h1>Using ARIA to match the accessible name to the visible label</h1>
+		<section id="meta">
+			<h2>Metadata</h2>
+			<p id="id"></p>
+			<p id="technology"></p>
+			<p id="type"></p>
+		</section>
+		<section id="applicability">
+			<h2>When to Use</h2>
+			<p class="instructions">This technique applies to all web technologies that include interactive controls (such as links or form inputs).</p>
+		</section>
+		<section id="description">
+			<h2>Description</h2>
+			<p class="instructions">The objective of this technique is to ensure that speech input users can operate web content reliably while not adversely affecting other users of assistive technology.</p>
+			<p>When speech input users interact with a web page, they usually speak a command followed by the reference to some visible label (such as text beside an input field or inside a button or link). For example, they may say "click search" to activate a button labelled Search. When speech recognition software processes speech input and looks for matches, it uses the <a href="https://www.w3.org/TR/accname/"> accessible name</a> of controls. Where there is a mismatch between the text in the label and the text in the accessible name, it can cause issues for the user.</p>
+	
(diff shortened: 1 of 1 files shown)


alastc:
@mbgower There are some examples or links to your C drive, and links to WCAG 2.1, could you update that? Assuming this technique is still worthwhile?

techniques/aria/aria25.html
```diff
@@ -0,0 +1,194 @@
...
+						<li>The computation for accessible name and description ensures that the value of <code>title</code> is always available programmatically, either as the accessible name or description. Thus screen readers will always have access to this information. </li>
+					</ol>
+				<p></p>
+				<p>There are a couple of disadvantages to this construction:
+					</p><ol>
+						<li>Screen reader users will hear "Work Phone:" announced twice – as the group name and as the label of the first field. (This is an unfortunate outcome of the new 2.5.3 requirement and will occur for at least one input in a set of inputs with a group label in all implementations.)</li>
+						<li>The keyboard user cannot see a label for the second and third inputs since most browsers do not provide a mechanism to display <code>title</code> via keyboard. (Unless the telephone inputs are redesigned to provide persistently visible labels, this will remain a challenge.)</li>
+					</ol>
+				<p></p>
+				<p class="working-example"><a href="file:///C:/Users/IBM_ADMIN/Documents/GitHub/wcag/working-examples/label-in-name-general/example3.html" target="_blank">Working example of Work Phone input set</a></p>
```
mbgower, line 68: Removing reference to working example, since I know longer have it locally. Easy enough to add later
```suggestion
```

techniques/aria/aria25.html
```diff
@@ -0,0 +1,193 @@
...
+			&lt;td&gt;&lt;input type="radio" name="sizes" id="SD6" value="SD6"&gt;&lt;/td&gt;<br>
+			&lt;td&gt;&lt;input type="radio" name="sizes" id="VD6" value="VD6"&gt;&lt;/td&gt;<br>
+		&lt;/tr&gt;<br>
+	&lt;/tr&gt;<br>
+&lt;/tbody&gt;<br>
+&lt;/table&gt;<br>
+
+</code>
+</p>
+				<p class="working-example"><a href="file:///C:/Users/IBM_ADMIN/Documents/GitHub/wcag/working-examples/label-in-name-general/example5.html" target="_blank">Working example of radio buttons in a matrix</a></p>
```
mbgower, line 134: Removing reference to working example, since I no longer have this file locally
```suggestion
```

mbgower: Remove non-existent working example

mbgower:
> @mbgower There are some examples or links to your C drive, and links to WCAG 2.1, could you update that? Assuming this technique is still worthwhile?

I couldn't find the working examples, so removed the references. I think this technique is still useful, so am creating a PR and adding it to draft.

mbgower:
Hmm, getting some check failures. I'll have to return to this...

bruce-usab:
Discussed on backlog call 11/1.  @stevefaulkner volunteered to review.

kfranqueiro (changes requested):
- Double-check `<p class="instructions">` elements - some of them are likely meant to be kept but should no longer have the class; others should be entirely removed.
- Make sure code examples are indented for readability and have proper class applied for syntax highlighting.
- Make sure any related SCs link to this from their Understanding page, so that the About this Technique box fills in correctly.
- Filename should be uppercase; I will fix this since it is easier for me to do so on my case-sensitive filesystem.

mbgower:
I spent some time on this on Friday, including creating a working example for Example 2 (radio buttons laid out in a matrix). I was about to start on the same for example 3 (inputs laid out in a matrix) However, the reality is that aria is not for the most part used to resolve these examples. Instead, the most effective approach appears to be  a combination of aria-labelledby and title. So I question whether it makes sense to have this as an aria technique.

PR #4070, opened 2024-09-16
Add "modality" to single pointer definition with examples of single pointer interactions

patrickhlauke:
Follow-up to https://github.com/w3c/wcag/pull/3536#issuecomment-2352336750 to potentially address concerns by @WilcoFiers @dbjorge

Tries to more clearly differentiate "input" as in the *modality* from the gestures/interactions that are the *result* of using a particular input *modality*

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/pull/4070.html" title="Last updated on Sep 20, 2024, 8:10 PM UTC (a79bb49)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/4070/4213bf6...a79bb49.html" title="Last updated on Sep 20, 2024, 8:10 PM UTC (a79bb49)">Diff</a>

### DIFF
--- a/guidelines/terms/21/single-pointer.html
+++ b/guidelines/terms/21/single-pointer.html
@@ -1,6 +1,6 @@
 <dt><dfn id="dfn-single-pointer">single pointer</dfn></dt>
 <dd>
-   <p>an input mechanism that only targets a single point on the page/screen at a time – such as a mouse, single finger on a touch screen, or stylus.</p>
+   <p>an input modality that only targets a single point on the page/screen at a time – such as a mouse, single finger on a touch screen, or stylus.</p>
    <p class="note">Single pointer interactions include clicks, taps, dragging motions, and single-finger swipe gestures. In contrast, multipoint interactions involve the use of two or more pointers at the same time, such as two finger interactions on a touchscreen, or the simultaneous use of a mouse and stylus.</p>
 </dd>
 
--- a/guidelines/terms/21/single-pointer.html
+++ b/guidelines/terms/21/single-pointer.html
@@ -1,6 +1,6 @@
 <dt><dfn id="dfn-single-pointer">single pointer</dfn></dt>
 <dd>
-   <p>an input that only targets a single point on the page/screen at a time – such as a mouse, single finger on a touch screen, or stylus.</p>
+   <p>an input mechanism that only targets a single point on the page/screen at a time – such as a mouse, single finger on a touch screen, or stylus.</p>
    <p class="note">In contrast to single pointer inputs, multipoint interactions involve the use of two or more pointers at the same time – such as two finger interactions on a touchscreen, or the simultaneous use of a mouse and stylus.</p>
 </dd>
 
--- a/guidelines/terms/21/single-pointer.html
+++ b/guidelines/terms/21/single-pointer.html
@@ -1,6 +1,6 @@
 <dt><dfn id="dfn-single-pointer">single pointer</dfn></dt>
 <dd>
    <p>an input mechanism that only targets a single point on the page/screen at a time – such as a mouse, single finger on a touch screen, or stylus.</p>
-   <p class="note">In contrast to single pointer inputs, multipoint interactions involve the use of two or more pointers at the same time – such as two finger interactions on a touchscreen, or the simultaneous use of a mouse and stylus.</p>
+   <p class="note">In contrast to single pointer interactions, multipoint interactions involve the use of two or more pointers at the same time – such as two finger interactions on a touchscreen, or the simultaneous use of a mouse and stylus.</p>
 </dd>
 
--- a/guidelines/terms/21/single-pointer.html
+++ b/guidelines/terms/21/single-pointer.html
@@ -1,6 +1,6 @@
 <dt><dfn id="dfn-single-pointer">single pointer</dfn></dt>
 <dd>
    <p>an input mechanism that only targets a single point on the page/screen at a time – such as a mouse, single finger on a touch screen, or stylus.</p>
-   <p class="note">In contrast to single pointer interactions, multipoint interactions involve the use of two or more pointers at the same time – such as two finger interactions on a touchscreen, or the simultaneous use of a mouse and stylus.</p>
+   <p class="note">Single pointer interactions include clicks, taps, dragging motions, and single-finger swipe gestures. In contrast, multipoint interactions involve the use of two or more pointers at the same time, such as two finger interactions on a touchscreen, or the simultaneous use of a mouse and stylus.</p>
 </dd>
 
--- a/guidelines/terms/21/single-pointer.html
+++ b/guidelines/terms/21/single-pointer.html
@@ -1,6 +1,6 @@
 <dt><dfn id="dfn-single-pointer">single pointer</dfn></dt>
 <dd>
    <p>an input modality that only targets a single point on the page/screen at a time – such as a mouse, single finger on a touch screen, or stylus.</p>
-   <p class="note">Single pointer interactions include clicks, double clicks, taps, dragging motions, and single-finger swipe gestures. In contrast, multipoint interactions involve the use of two or more pointers at the same time, such as two finger interactions on a touchscreen, or the simultaneous use of a mouse and stylus.</p>
+   <p class="note">Single pointer interactions include clicks, double clicks, taps, dragging motions, and single-finger swipe gestures. In contrast, multipoint interactions involve the use of two or more pointers at the same time, such as two-finger interactions on a touchscreen, or the simultaneous use of a mouse and stylus.</p>
 </dd>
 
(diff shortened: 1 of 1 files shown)


Hvychvy383:
)

Hvychvy383:
On Mon, Sep 16, 2024, 4:04 AM Stephen Craig Wilson ***@***.***>
wrote:

> (
> On Mon, Sep 16, 2024, 3:53 AM KRUEGER *1*@***.***>

wrote:

bruce-usab (approved):

guidelines/terms/21/single-pointer.html
```diff
@@ -1,6 +1,6 @@
 <dt><dfn id="dfn-single-pointer">single pointer</dfn></dt>
 <dd>
-   <p>an input that only targets a single point on the page/screen at a time – such as a mouse, single finger on a touch screen, or stylus.</p>
-   <p class="note">In contrast to single pointer inputs, multipoint interactions involve the use of two or more pointers at the same time – such as two finger interactions on a touchscreen, or the simultaneous use of a mouse and stylus.</p>
+   <p>an input mechanism that only targets a single point on the page/screen at a time – such as a mouse, single finger on a touch screen, or stylus.</p>
```
WilcoFiers, line 3: I think using "mechanism" makes this more confusing, not less so. WCAG has a definition of "mechanism", which doesn't seem to be the same thing that you're using the word for here.

The phrase that confused me was "at a time". The way I think of path-based gestures is that it is multiple inputs coming in at multiple times. So since it doesn't happen "at a time" but "over time", I figured the definition doesn't apply. After talking to @dbjorge I learned that what you meant here is that it can be over time, just not at the same time. It's that part that I think needs to be clarified.

I think that removing "single taps and clicks, double-taps and clicks, long presses, and path-based gestures" from this definition has made it more confusing. I think these need to be put back either in the definition text or in a note.

WilcoFiers (changes requested):

guidelines/terms/21/single-pointer.html
patrickhlauke: but again, the pointer is not the gesture. the single pointer is the one finger on the touchscreen, the mouse, the stylus. you use pointers to then perform gestures. i can propose a further tweak on friday expanding the note, but this fundamental distinction needs to be finally put to bed (as it causes non-sequiturs like the one the original change addresses, which also caused confusion)

patrickhlauke:
to be clearer: a tap, a swipe, a path-based gesture *is not* a "single pointer". they are examples of gestures *performed* using a single pointerOn 18 Sep 2024, at 16:16, Wilco Fiers ***@***.***> wrote:﻿
@WilcoFiers requested changes on this pull request.



In guidelines/terms/21/single-pointer.html:
> @@ -1,6 +1,6 @@
 <dt><dfn id="dfn-single-pointer">single pointer</dfn></dt>
 <dd>
-   <p>an input that only targets a single point on the page/screen at a time – such as a mouse, single finger on a touch screen, or stylus.</p>
-   <p class="note">In contrast to single pointer inputs, multipoint interactions involve the use of two or more pointers at the same time – such as two finger interactions on a touchscreen, or the simultaneous use of a mouse and stylus.</p>
+   <p>an input mechanism that only targets a single point on the page/screen at a time – such as a mouse, single finger on a touch screen, or stylus.</p>

I think using "mechanism" makes this more confusing, not less so. WCAG has a definition of "mechanism", which doesn't seem to be the same thing that you're using the word for here.
The phrase that confused me was "at a time". The way I think of path-based gestures is that it is multiple inputs coming in at multiple times. So since it doesn't happen "at a time" but "over time", I figured the definition doesn't apply. After talking to @dbjorge I learned that what you meant here is that it can be over time, just not at the same time. It's that part that I think needs to be clarified.
I think that removing "single taps and clicks, double-taps and clicks, long presses, and path-based gestures" from this definition has made it more confusing. I think these need to be put back either in the definition text or in a note.

guidelines/terms/21/single-pointer.html
dbjorge: I agree with Wilco that I think adding "mechanism" makes it more confusing, I'd prefer omitting it.

dbjorge (changes requested):
I agree with Wilco that I don't think this PR addresses the concern that Wilco and I mentioned, so I'd prefer to just close it rather than proceed with this direction if our concerns were the only thing motivating the PR.

Personally, I still think it would be easier to understand overall with a note or example (probably not a normative part of the definition) that clarifies examples of operations that can be performed using a single pointer, but I recognize [@alastc 's comment here](https://github.com/w3c/wcag/pull/3536#issuecomment-2346425776) and that my position doesn't seem to be the consensus one; I can live with the already-merged change given the paragraph @alastc mentioned in the understanding document, but I think it would be easier to understand if that paragraph were less buried.

> Single pointer operations include taps and clicks, double-taps and double-clicks, long presses, swiping, dragging, and path-based gestures. Gestures such as "pinch to zoom" or two-finger swipes are multipoint gestures, as they require two or more pointer inputs - in this case, two fingers on a touchscreen.

patrickhlauke:
if, contrary to the original consensus position, we *were* to expand the informative note, maybe something like this could strike the correct balance:

```
<p class="note">In contrast to single pointer interactions – including clicks, taps, dragging motions, and single-finger swipe gestures  – multipoint interactions involve the use of two or more pointers at the same time – such as two finger interactions on a touchscreen, or the simultaneous use of a mouse and stylus.</p>
```

guidelines/terms/21/single-pointer.html
```diff
@@ -1,6 +1,6 @@
...
+   <p class="note">In contrast to single pointer interactions, multipoint interactions involve the use of two or more pointers at the same time – such as two finger interactions on a touchscreen, or the simultaneous use of a mouse and stylus.</p>
```
patrickhlauke, line 4:
```suggestion
   <p class="note">Single pointer interactions include clicks, taps, dragging motions, and single-finger swipe gestures. In contrast, multipoint interactions involve the use of two or more pointers at the same time, such as two finger interactions on a touchscreen, or the simultaneous use of a mouse and stylus.</p>
```

guidelines/terms/21/single-pointer.html
patrickhlauke:
```suggestion
   <p>an input modality that only targets a single point on the page/screen at a time – such as a mouse, single finger on a touch screen, or stylus.</p>
```

guidelines/terms/21/single-pointer.html
```diff
@@ -1,6 +1,6 @@
 <dt><dfn id="dfn-single-pointer">single pointer</dfn></dt>
 <dd>
-   <p>an input that only targets a single point on the page/screen at a time – such as a mouse, single finger on a touch screen, or stylus.</p>
-   <p class="note">In contrast to single pointer inputs, multipoint interactions involve the use of two or more pointers at the same time – such as two finger interactions on a touchscreen, or the simultaneous use of a mouse and stylus.</p>
+   <p>an input modality that only targets a single point on the page/screen at a time – such as a mouse, single finger on a touch screen, or stylus.</p>
+   <p class="note">Single pointer interactions include clicks, taps, dragging motions, and single-finger swipe gestures. In contrast, multipoint interactions involve the use of two or more pointers at the same time, such as two finger interactions on a touchscreen, or the simultaneous use of a mouse and stylus.</p>
```
mbgower, line 4:
```suggestion
   <p class="note">Single pointer interactions include clicks, double clicks, taps, dragging motions, and single-finger swipe gestures. In contrast, multipoint interactions involve the use of two or more pointers at the same time, such as two finger interactions on a touchscreen, or the simultaneous use of a mouse and stylus.</p>
```

guidelines/terms/21/single-pointer.html
```diff
@@ -1,6 +1,6 @@
 <dt><dfn id="dfn-single-pointer">single pointer</dfn></dt>
 <dd>
-   <p>an input that only targets a single point on the page/screen at a time – such as a mouse, single finger on a touch screen, or stylus.</p>
-   <p class="note">In contrast to single pointer inputs, multipoint interactions involve the use of two or more pointers at the same time – such as two finger interactions on a touchscreen, or the simultaneous use of a mouse and stylus.</p>
+   <p>an input modality that only targets a single point on the page/screen at a time – such as a mouse, single finger on a touch screen, or stylus.</p>
+   <p class="note">Single pointer interactions include clicks, double clicks, taps, dragging motions, and single-finger swipe gestures. In contrast, multipoint interactions involve the use of two or more pointers at the same time, such as two finger interactions on a touchscreen, or the simultaneous use of a mouse and stylus.</p>
```
patrickhlauke, line 4:
```suggestion
   <p class="note">Single pointer interactions include clicks, double clicks, taps, dragging motions, and single-finger swipe gestures. In contrast, multipoint interactions involve the use of two or more pointers at the same time, such as two-finger interactions on a touchscreen, or the simultaneous use of a mouse and stylus.</p>
```

WilcoFiers (approved):
❤️

dbjorge (approved):

bruce-usab:
Reviewed and finalized on Sep 20 TF call.  Returned to the Errata for CFC column.

PR #4093, opened 2024-10-03
Add title and styling to examples

### LINKED ISSUE #2958: Examples not highlighted in the Understanding documents
The examples are no longer at all different from the running text (class=example)

In the main WCAG document they are at least marked in yellow and indented, which is not perfect, but better than nothing (there is already another [issue](https://github.com/w3c/wcag/issues/634) for that)

Example at https://www.w3.org/WAI/WCAG21/Understanding/conformance#conformance-requirements

> A Web page incorporates a new interactive graphic technology called "ZAP". Although ZAP is not accessibility-supported, ...

iadawn:
Fixes #2958

Change styling for examples within the Key Terms section to include text to indicate it is an example and a background change to make it stand out.

example of styling for key terms example

### DIFF
--- a/11ty/CustomLiquid.ts
+++ b/11ty/CustomLiquid.ts
@@ -506,7 +506,7 @@ export class CustomLiquid extends Liquid {
 			</div>`);
     });
     
-    // Add header to example sections in Key Terms
+    // Add header to example sections in Key Terms (aside) and Conformance (div)
     $("aside.example, div.example").each((_, el) => {
       const $el = $(el);
       $el.prepend(`<p class="example-title marker">Example</p>`);
--- a/11ty/CustomLiquid.ts
+++ b/11ty/CustomLiquid.ts
@@ -505,6 +505,12 @@ export class CustomLiquid extends Liquid {
 				<p>${$el.html()}</p>
 			</div>`);
     });
+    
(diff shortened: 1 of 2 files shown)


11ty/CustomLiquid.ts
```diff
@@ -505,6 +505,12 @@ export class CustomLiquid extends Liquid {
 				<p>${$el.html()}</p>
 			</div>`);
     });
+    
+    // Add header to example sections in Key Terms
```
kfranqueiro, line 509:
```suggestion
    // Add header to example sections in Key Terms (aside) and Conformance (div)
```

css/base.css
```diff
@@ -155,6 +155,19 @@ dt div {
   margin-top: 0;
 }
 
+:is(aside, div).example .example-title {
```
kfranqueiro, line 158: This seems to be our first time using `:is` for these docs; flagging to make sure we are okay with its [level of browser support](https://caniuse.com/css-matches-pseudo)

kfranqueiro (approved):
Approved with one suggested comment update and one note

css/base.css
iadawn: `:is` is being used in TR space so I am comfortable with this

Issue #4095, opened 2024-10-04
2.4.6 Headings and Labels appears to prohibit the use of icons as labels


TestPartners:
The normative text says "Headings and labels describe topic or purpose". I don't see how an icon can describe anything, and the Understanding page makes no mention of the use of icons as labels.

It has been stated that icons can be used as labels in discussions regarding SC 3.3.2: Labels or Instructions, which does not require that labels are accurate or descriptive. See [https://github.com/w3c/wcag/issues/3276](https://github.com/w3c/wcag/issues/3276)

If the use of icons as labels conforms with SC 3.3.2 but does not conform with SC 2.4.6, I think this should be made explicit in the Understanding pages for both success criteria.

patrickhlauke:
starting with the premise that "describe" is a fluffy/vague/subjective term anyway, i'm sure though that we can agree that:

* for icons/images, there are two aspects: the visual part, and the (per 1.1.1) alternative text part
* you can make an argument for each of those two parts about whether something is correct/descriptive/appropriate or not ... if a search field is preceded by a poop emoji, it's probably not "descriptive" visually, compared to the same field preceded by a magnifying glass icon; likewise, if the text alternative for that icon/image isn't appropriate in conveying what the form field's use/purpose is

TestPartners:
I agree. For the purposes of this discussion I want to ignore the alternative text because that can be changed without changing the visual appearance. I am sure there have been plenty of discussions about adequate text alternatives.

It's a bit of a stretch to suggest that a magnifying glass icon describes the adjacent textbox (or indeed the button itself) because magnifying glass icons are commonly used for other purposes, such as buttons that enable magnification.

In Adobe Reader's Preflight tool, a magnifying glass icon is used to label the button that opens the Analyze tab. Would that fail this SC on the basis that the magnifying glass icon is reserved for search functionality? If an icon doesn't have a specific meaning, how can it be descriptive?

And I don't think anyone could argue that the three horizontal lines on a burger button describe its purpose. Its purpose is widely, but not universally, recognised now but it wasn't when it first started to be used.

patrickhlauke:
> And I don't think anyone could argue that the three horizontal lines on a burger button describe its purpose. Its purpose is widely, but not universally, recognised now but it wasn't when it first started to be used.

[...]

PR #4102, opened 2024-10-09
Update 3.3.2 understanding doc and g167 technique

### LINKED ISSUE #4088: Clarification on 3.3.2: Does every input field need a persistent visual label?
For SC 3.3.2: Does every input field need a persistent visual label?
What about cases where a input field is used in a particular context, which might not be a form?
For example the WYSIWYG Text Editor in Chat / Messenger applications? It is technically a input, putting a visible label on the page is required by this SC - in a lot of cases it is weird in terms of UI / UX (the context of the messenger makes it clear that the input is for writing a message to someone).

Any guidance on this?

scottaohara:
closes #4088

includes new icon-button example in G167 to showcase how a button represented by an icon alone can be used in certain instances for labeling form fields

also updates the understanding doc to provide two new written examples to the examples section list.

One to not make someone have to know to look to g167 to understand that "icons" can also serve as labels
And the second to provide an example of how two websites could have the same sort of input, but one needs instructions (due to extra requirements) while another doesn't (because they are super chill).


techniques/general/G167.html
```diff
@@ -10,16 +10,33 @@
...
+            <p>A web page contains a text field where the user can enter search terms and a button labeled "Search" for performing the search. The button is positioned right after the text field so that it is clear to the user that the text field is where to enter the search term.</p>
             <div>
             	<img alt="A text field with a button positioned to the right demonstrating the use of a button to label a field." src="img/button-as-label.jpg"/>
             </div>
          
       </section>
+
+      <section class="example">
+         <h3>A "reply" or "send message" field</h3>
+         <p>A chat or email web application will often present users a single or multi-line text field used for composing a message.  The field does not have a visible text label, but the purpose of the field - to compose a message to send - is indicated by it being paired with a button represented by a "send message" icon.</p>
```
gundulaniemann, line 22: Similar as with the text button for 'Search' above, the icon here serves as a label. I suggest to clearly state that.

<p>A chat or email web application will often present users a single or multi-line text field used for composing a message. It is often paired with an icon button labeled with a 'send message' icon (usually a paper plane). The icon also serves as a label for the message field.</p>

techniques/general/G167.html
```diff
@@ -10,16 +10,33 @@
...
+        <div>
+          <img src="img/button-send-message.jpg alt="text field containing the text let's go. The field has a button visually overlaying the right side of it. The button is circular and contains an arrow icon that also resembles a paper airplane - which has increasingly gained popularity as a 'send message' identifier. ">
+        </div>
+        
+         
+            <p>A web page contains a text field where the user can enter search terms and a button labeled "Search" for performing the search. The button is positioned right after the text field so that it is clear to the user that the text field is where to enter the search term.</p>
+            <div>
+            	<img alt="A text field with a button positioned to the right demonstrating the use of a button to label a field." src="img/button-as-label.jpg"/>
+            </div>
+
```
giacomo-petri, line 33:
```suggestion
         
```

This example is repeated above, so I suggest removing it from here and keeping it only in the section above.

bruce-usab (approved):

techniques/general/G167.html
```diff
@@ -10,16 +10,27 @@
...
             </div>
          
       </section>
+
+      <section class="example">
+         <h3>A "reply" or "send message" field</h3>
+         <p>A chat or email web application will often present users a single or multi-line text field used for composing a message. The field does not have a visible text label, but the purpose of the field - to compose a message to send - is indicated by its pairing with a button represented by a "send message" icon. The icon serving the dual purpose of visually labelling the button, as well as the text field.</p>
+
+        <div>
+          <img src="img/button-send-message.jpg alt="text field containing the text let's go. The field has a button visually overlaying the right side of it. The button is circular and contains an arrow icon that also resembles a paper airplane - which has increasingly gained popularity as a 'send message' identifier. ">
```
giacomo-petri, line 25:
```suggestion
          <img src="img/button-send-message.jpg" alt="text field containing the text let's go. The field has a button visually overlaying the right side of it. The button is circular and contains an arrow icon that also resembles a paper airplane - which has increasingly gained popularity as a 'send message' identifier.">
```

giacomo-petri: Just correcting a closing double quotes. Approving already.

giacomo-petri (approved):

techniques/general/G167.html
```diff
@@ -10,16 +10,27 @@
...
             </div>
          
       </section>
+
+      <section class="example">
+         <h3>A "reply" or "send message" field</h3>
+         <p>A chat or email web application will often present users a single or multi-line text field used for composing a message. The field does not have a visible text label, but the purpose of the field - to compose a message to send - is indicated by its pairing with a button represented by a "send message" icon. The icon serving the dual purpose of visually labelling the button, as well as the text field.</p>
+
+        <div>
+          <img src="img/button-send-message.jpg" alt="text field containing the text let's go. The field has a button visually overlaying the right side of it. The button is circular and contains an arrow icon that also resembles a paper airplane - which has increasingly gained popularity as a 'send message' identifier.">
```
mbgower, line 25:
```suggestion
          <img src="img/button-send-message.jpg" alt="text field containing the text 'let's go'. The field has a button visually overlaying the right side of it. The button is circular and contains an arrow icon that also resembles a paper airplane - which has increasingly gained popularity as a 'send message' identifier.">
```

bruce-usab:
Discussed on backlog call 11/1.

detlevhfischer:
I think it would be beneficial to have the far more common case of a search icon (loupe) for the search case, or add that as an icon only variant. I think the commonality of that icon to stand for search is undisputed. That is then the core case on which to hang similar cases like the chat message send icon.

TestPartners:
Actually, the commonality of the loupe icon to stand for search was disputed in a discussion here only a few weeks ago. See [https://github.com/w3c/wcag/issues/4095](https://github.com/w3c/wcag/issues/4095) where I mentioned three different ways in which that icon is used in different applications. That's not a problem for SC 3.3.2 because it does not require that labels are accurate or meaningful, but SC 2.4.6 does. I would argue that icons are almost never unambiguously meaningful with the possible exception of icons that occur in the user agent and are used in the same way in the website.

detlevhfischer:
@TestPartners I agree there are different interpretations of the loupe, but I think this is **not** an issue when you take into account the context in which it is used (after a search field, or on a map). There can hardly be a more ubiquitous pattern on the web than textfield followed by a loup icon?

TestPartners:
You say the context is that the icon is "after a search field", but users don't know it's a search field unless they understand what the icon means. I'm really not happy saying "everyone know what this icon means". Even if we agree that the loupe is universally recognisable, many (most?) other icons are not. I would like to see G167 clearly state that the use of icons as labels is conformant with SC 3.3.2 but that it is not conformant with 2.4.6.

detlevhfischer:
> You say the context is that the icon is "after a search field", but users don't know it's a search field unless they understand what the icon means

I guess it is clear that the field itself would need an accessible name ,so for non-visual use this would not be an issue?

> I'm really not happy saying "everyone know what this icon means"

That's for sure - when could you ever be able to make sucxh a sweeping claim? Exactly, never. There are many places where implict assumptions and conventionality play a role in interpreting and understanding, and you are managing tradeoffs here all the time. You can say the same thing about the play and pause icon - sure you will find some people who don't know them. How far will you get when you say "every visual element needs a text description next to it" or "you must not use the loupe icon, the hamburger icon, the play icon etc. because they fail 2.4.6"?

patrickhlauke:
> I would like to see G167 clearly state that the use of icons as labels is conformant with SC 3.3.2 but that it is not conformant with 2.4.6

I'm not sure everybody agrees with that last bit, I'm afraid. Which is the rub.

TestPartners:
In my view, the last bit is uncontroversial. There are countless examples of icons that have multiple interpretations or no meaning at all. SC 2.4.6 explicitly says it is non-conformant "if the label is not sufficiently clear or descriptive". On that basis, very few icons would be conformant.

patrickhlauke:
"Very few" does not mean "none"...

TestPartners:
I encounter so many icons that even I don't understand, that I would not object to it being an automatic non-conformance even if a small number are understandable. However, I realise that's not going to happen, so the next best thing is a requirement "that authors should carefully consider how (subjectively) understandable they are". That makes it clear that we can report a non-conformance if we think an icon is not sufficiently understandable.

It's worth making the point that text labels can also be subjectively understandable, so it's not just icons.

techniques/general/G167.html
```diff
@@ -10,16 +10,27 @@
...
             </div>
          
       </section>
+
+      <section class="example">
+         <h3>A "reply" or "send message" field</h3>
+         <p>A chat or email web application will often present users a single or multi-line text field used for composing a message. The field does not have a visible text label, but the purpose of the field - to compose a message to send - is indicated by its pairing with a button represented by a "send message" icon. The icon serving the dual purpose of visually labelling the button, as well as the text field.</p>
+
+        <div>
+          <img src="img/button-send-message.jpg" alt="text field containing the text 'let's go'. The field has a button visually overlaying the right side of it. The button is circular and contains an arrow icon that also resembles a paper airplane - which has increasingly gained popularity as a 'send message' identifier.">
```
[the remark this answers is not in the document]
TestPartners: I would say "is increasingly widely used" rather than "has increasingly gained popularity". I think it's a lousy icon that works very badly (for me) because I always look right past it without recognising what it is. Can't we use a better example?

techniques/general/G167.html
```diff
@@ -10,16 +10,27 @@
       <section class="example">
          <h3>A search function</h3>
          
-            <p>A Web page contains a text field where the user can enter search terms and a button labeled "Search" for performing the search. The button is positioned right after the text field so that it is clear to the user that the text field is where to enter the search term.</p>
+            <p>A web page contains a text field where the user can enter search terms and a button labeled "Search" for performing the search. The button is positioned right after the text field so that it is clear to the user that the text field is where to enter the search term.</p>
             <div>
             	<img alt="A text field with a button positioned to the right demonstrating the use of a button to label a field." src="img/button-as-label.jpg"/>
             </div>
```
scottaohara, line 17: <p>Alternatively, a button visually labeled with a "loupe" or "magnify glass" icon, with the alternative text of "search", could be used instead of a button with a text label, as variants of this icon are used to identify search inputs across not just websites, but user interfaces of non-web software as well.</p>

<div>
  A common visual design and labeling pattern for search fields: a text field with a button, visually represented with a common magnify glass icon, positioned at the start (left) of the text field. the icon serves to both label the button and the text field.
</div>

scottaohara, line 17:
```suggestion
         <p>Alternatively, a button visually labeled with a "loupe" or "magnify glass" icon, with the alternative text of "search", could be used instead of a button with a text label, as variants of this icon are used to identify search inputs across not just websites, but user interfaces of non-web software as well.</p>

<div>
  <img alt="A common visual design and labeling pattern for search fields: a text field with a button, visually represented with a common magnify glass icon, positioned at the start (left) of the text field. the icon serves to both label the button and the text field." src="img/search-button-icon-label.png">
</div>
```

techniques/general/G167.html
```diff
@@ -10,16 +10,31 @@
...
+         <p>Alternatively, a button visually labeled with a "loupe" or "magnify glass" icon, with the alternative text of "search", could be used instead of a button with a text label, as variants of this icon are used to identify search inputs across not just websites, but user interfaces of non-web software as well.</p>
```
scottaohara, line 17:
```suggestion
         <p>Alternatively, a button visually labeled with a "loupe" or "magnify glass" icon, with the alternative text of "search", could be used instead of a button with a text label. Variants of this icon are used to identify search inputs across not just websites, but user interfaces of non-web software as well.</p>
```

bruce-usab:
Discussed on TF call 11/22.  It is not controversial that icons (with alt) can be used as labels.  Concurrence with PR.

mraccess77:
In the past the group chose to not require on-screen text for icons.  Now requiring them would mean a significant change which would in my opinion be on the level of an errrata.

scottaohara:
@mraccess77 this pr makes no such claim, nor does it plan to. So no errata necessary

mbgower:
IMO, Detlev's [comment requesting](https://github.com/w3c/wcag/pull/4102#issuecomment-2490625933) an examination of the current wording of the test procedure is valid
> Should the test of the technique also check that the icon (in icon-only cases) has a descriptive programmatically determinable name (or label)? With the old school examples this was more or less a given, not so for icon based buttons?

In the existing version of this technique the context of the button having a text label is explicit in the pre-amble, and the existing examples support this:
>When a button invokes a function on an input field, has a clear text label, and is rendered adjacent to the input field, the button also acts as a label for the input field.

We have introduced icon-only buttons into this technique, but not addressed the preamble, nor the tests, which currently read:
> 1. Check that the field and button are adjacent to one another in the programmatically determined reading sequence.
> 2. Check that the field and button are visually rendered adjacent to one another.

There is no check here that the icon button has an ALT (although that is inferred  in changes to the technique content, specifically "with the alternative text of 'search'"). While it is true that an icon-only button _without_ a name would already fail WCAG (4.1.2, 1.1.1, etc), for the purposes of this technique it seems reasonable to include an additional check, since obviously a icon button without a name cannot be said to be labelled.

I suggest :
3. Where the button does not have a visible text label, confirm it has an accessible name

---
https://www.w3.org/WAI/WCAG21/Techniques/general/G131
Admittedly, the existing technique _also_ makes no me

techniques/general/G167.html
```diff
@@ -1,7 +1,7 @@
...
       <p>All technologies that support forms</p>
    </section><section id="description"><h2>Description</h2>
-      <p>When a button invokes a function on an input field, has a clear text label, and is rendered adjacent to the input field, the button also acts as a label for the input field. This label helps users understand the purpose of the field without introducing repetitive text on the Web page. Buttons that label single text fields typically follow the input field.</p>
+      <p>When a button invokes a function on an input field, has a clear text label, and is rendered adjacent to the input field, the button also acts as a label for the input field. This label helps users understand the purpose of the field without introducing repetitive text on the web page. Buttons that label single text fields typically follow the input field.</p>
```
mbgower, line 4:
```suggestion
      <p>When a button invokes a function on an input field, has a clear text label or name, and is rendered adjacent to the input field, the button also acts as a label for the input field. This label helps users understand the purpose of the field without introducing repetitive text on the web page. Buttons that label single text fields typically follow the input field.</p>
```

mbgower:
Note that I accidentally directed [committed](https://github.com/w3c/wcag/pull/4102/commits/826ea7b438f9f46d9150b2e02b43d7ff59548787) my final suggestion:
> Where the button does not have a visible text label, check that it has an accessible name

mbgower:
Merging this PR, since it has sufficient support, and comments have been addressed.

PR #4104, opened 2024-10-11
Suggested additional note for focus-not-obscured

### LINKED ISSUE #3529: Clarification for Understanding 2.4.11 Focus Not Obscured (Minimum) to ensure consistent testing
In a meeting between accessibility experts from several Scandinavian organisations, the following questions for 2.4.11 Focus Not Obscured were brought up.

The question is not so much what the suggested best practice for this is, but what the minimum requirements are in a legal/monitoring context.

Having written several ACT Rules for WCAG in the past, I also anticipate that further clarification will be needed before it is possible to agree on ACT Rules for this SC.

We suggest that Understanding SC 2.4.11 (and possibly techniques as well) is updated with information to help clarify these points.

**What is the threshold for "not entirely hidden"?**
Should enough of the component be unobscured, that the user can identify what the component is and what it does?
Or should it just be possible to identify that there is a component?
Or is it enough that 1 pixel of the component is not hidden?

**Which part (component, focus indication or both) has to be unobscured?**
The Intent of the Understanding document explains: "knowing the current point of focus is critical. The component with focus signals the interaction point on the page"

However, the SC states that this is about "the component is not entirely hidden". This means that the focus indication could be entirely hidden and this SC still pass, even though the "current point of focus" is not clear in this case. If the focus indication is positioned on one side of the element, e.g. as an underline, an arrow before etc. this scenario is not hard to imagine. 

On the other hand, it is also not clear from the understanding document whether the focus indicator should be considered a part of the element, and having a portion of the focus indicator unobscured would be enough to pass this SC.

So is it enough that either the component or the focus indicator is unobscured? 
Or does both a part of the component AND something that indicates that it is currently focused have to be visible to pass this SC?

alastc:
Closes #3529


understanding/22/focus-not-obscured-minimum.html
```diff
@@ -31,6 +31,8 @@ <h2>Intent of Focus Not Obscured (Minimum)</h2>
 
       <p>Another form of obscuring can occur where light boxes or other semi-opaque effects overlap the item with focus. While less than 100 percent opacity is not causing the component to be <q>entirely obscured</q>, such semi-opaque overlaps may cause a failure of <a href="../understanding/non-text-contrast.html">1.4.11 Non-text Contrast</a>. When a focus indicator can be covered by a semi-opaque component, the ability of the focus indicator to pass 1.4.11 should be evaluated (and pass) while the focus indicator is under the semi-opaque component. The intention in both situations is that the component receiving focus should never be obscured to the point a user cannot tell which item has focus.</p>
 
+      <p class="note">The component does not include the focus indicator when considering "the component is not entirely hidden". Although both are needed by users when tracking the focus, it is a clearer metric to test when not including the indicator.</p>
```
patrickhlauke, line 34: might be worth introducing the clarification that the SC refers to the focused *component*, rather than the focus *indication*

patrickhlauke: might also need further clarification about focus indicator that is around/outside of the component

patrickhlauke:
```suggestion
      <p class="note">Note that this criterion evaluates the focused <em>component</em>, rather than the focus <em>indicator</em>. The component itself does not include the focus indicator when checking that "the component is not entirely hidden" - unless the focus indicator is inside the component, or focus is indicated by a change to the component itself. Although users benefit from both the component and the focus indicator (if external to the component) not being obscured when tracking the focus, for the purposes of this criterion only checking the component provides a clearer metric.</p>
```

patrickhlauke: added a suggested rewording, which introduces an explicit clarification that the SC is concerned with the component, rather than the indicator. also added cross-reference to 2.4.7 (check that the link works)

@alastc suggest that it would be good to have a similar note for the AAA

patrickhlauke:
Please review my suggested tweak/addition above

bruce-usab:
Reviewed and discussed on Backlog call 1/17.   Kept in Ready for approval after edit.  @patrickhlauke volunteered to propose PR for related AAA SC.

understanding/22/focus-not-obscured-minimum.html
```diff
@@ -31,6 +31,8 @@ <h2>Intent of Focus Not Obscured (Minimum)</h2>
...
 
+      <p class="note">This criterion evaluates the focused <em>component</em>, rather than the focus <em>indicator</em>. The component itself does not include the focus indicator when checking that "the component is not entirely hidden" - unless the focus indicator is inside the component, or focus is indicated by a change to the component itself. Although users benefit from both the component and the focus indicator (if external to the component) not being obscured when tracking the focus, for the purposes of this criterion only checking the component provides a clearer metric. However, if the focus indicator is obscured, it will already fail <a href="focus-visible">2.4.7 Focus Visible</a>.</p>
```
kfranqueiro, line 34: Suggested by Giacomo:

```suggestion
      <p class="note">This criterion evaluates the focused <em>component</em>, rather than the focus <em>indicator</em>. The component itself does not include the focus indicator when checking that "the component is not entirely hidden" - unless the focus indicator is inside the component, or focus is indicated by a change to the component itself. Although users benefit from both the component and the focus indicator (if external to the component) not being obscured when tracking the focus, for the purposes of this criterion only checking the component provides a clearer metric. However, if the focus indicator is obscured, it would likely fail <a href="focus-visible">2.4.7 Focus Visible</a>.</p>
```

PR #4124, opened 2024-10-29
Add build support for linking to bibliographical references

kfranqueiro:
Refs #2535

This adds support to the Eleventy build system for resolving bibliographical references within paragraphs in the format `[[ref]]`, and converting them into links within single brackets instead.

This resolves against both specref.org and the local bibliography configured in `biblio.js` in this repo.

Note that there are some references that have no resolution in either of these sources, which is why I'm not marking this as fully resolving the related issue.

Details at https://github.com/w3c/wcag/issues/2535#issuecomment-2445246302

Here's a breakdown of what is successfully replaced: (search for these in single brackets on the page in question on the PR preview; some may require expanding Key Terms)

- 1.4.3 Contrast (Minimum): SRGB
- 1.4.6 Contrast (Enhanced): SRGB
- 1.4.11 Non-text Contrast: SRGB
- 1.4.13 Content on Hover or Focus: HTML
- 2.2.2 Pause, Stop, Hide: ISO_9241-391, SRGB
- 2.3.1 Three Flashes or Below Threshold: ISO_9241-391, SRGB
- 2.3.2 Three Flashes: ISO_9241-391
- 2.4.13 Focus Appearance: SRGB
- 2.5.5 Target Size (Enhanced): pointerevents
- 2.5.8 Target Size (Minimum): pointerevents
- 3.1.5 Reading Level: UNESCO (3 instances)
- 3.2.1 On Focus: UAAG10
- 3.2.2 On Input: UAAG10
- 3.2.5 Change on Request: UAAG10
- 4.1.3 Status Messages: UAAG10

### DIFF
--- a/11ty/biblio.ts
+++ b/11ty/biblio.ts
@@ -3,7 +3,7 @@ import { readFile } from "fs/promises";
 import { glob } from "glob";
 import uniq from "lodash-es/uniq";
 
-export const biblioPattern = /\[\[([\w-]+)\]\]/g;
+export const biblioPattern = /\[\[\??([\w-]+)\]\]/g;
 
 /** Compiles URLs from local biblio + specref for linking in Understanding documents. */
 export async function getBiblio() {
--- a/11ty/biblio.ts
+++ b/11ty/biblio.ts
@@ -1,0 +1,35 @@
+import axios from "axios";
+import { readFile } from "fs/promises";
+import { glob } from "glob";
+import uniq from "lodash-es/uniq";
+
+export const biblioPattern = /\[\[([\w-]+)\]\]/g;
+
+/** Compiles URLs from local biblio + specref for linking in Understanding documents. */
+export async function getBiblio() {
+  const localBiblio = eval(
+    (await readFile("biblio.js", "utf8"))
+      .replace(/^respecConfig\.localBiblio\s*=\s*/, "(")
+      .replace("};", "})")
+  );
+  
+  const refs: string[] = [];
+  for (const path of await glob(["guidelines/**/*.html", "understanding/*/*.html"])) {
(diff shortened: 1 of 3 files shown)


patrickhlauke:
There seem to also be some double-bracketed references here and there (e.g. https://deploy-preview-4124--wcag2.netlify.app/understanding/contrast-minimum#rationale-for-the-ratios-chosen)

11ty/biblio.ts
```diff
@@ -0,0 +1,35 @@
+import axios from "axios";
+import { readFile } from "fs/promises";
+import { glob } from "glob";
+import uniq from "lodash-es/uniq";
+
+export const biblioPattern = /\[\[([\w-]+)\]\]/g;
```
iadawn, line 6: Informative references can be expressed as `[[?lorem]]`. These aren't used often but might need to be aware of the possibility

iadawn:
@kfranqueiro Code looks fine to me. There is one comment relating to informative references which I think would be worth putting in.

Next step would be listing those references that do not resolve to anything. Will need to take those to the TF to come up with the correct references then we can add them into the `biblio.js`

kfranqueiro:
> Next step would be listing those references that do not resolve to anything. Will need to take those to the TF to come up with the correct references then we can add them into the `biblio.js`

Fortunately, this code already lists the references that do not resolve:

`Unresolved biblio refs: LAALS, HEARING-AID-INT, ISO-9241-3, ANSI-HFES-100-1988, ARDITI-FAYE, ARDITI-KNOBLAUCH-1994, ARDITI-KNOBLAUCH-1996, ARDITI, GITTINGS-FOZARD`

iadawn (approved):

PR #4126, opened 2024-10-30
Update reflow.html

mbgower:
An attempt to introduce a primary consideration in isolation in the Understanding document.

### DIFF
--- a/understanding/21/reflow.html
+++ b/understanding/21/reflow.html
@@ -1,136 +1,227 @@
 <!DOCTYPE html>
 <html xmlns="http://www.w3.org/1999/xhtml" lang="en" xml:lang="en">
+
 <head>
-		<meta charset="UTF-8" />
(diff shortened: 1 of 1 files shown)


mbgower:
@scottaohara Another attempt at rewording I wrote some time ago, which I thought I'd just shove into a PR to ensure there is no novel or better-crafted text than what you have created (reviewing that is next on my list after reviewing all these branches I've created that are still lingering).
If there is no value to this, LMK and I'll close it. Otherwise, happy to try to incorporate into the work done on Reflow to date.

scottaohara:
yes this is covered.  not in this exact wording though (which makes sense being different authors n' all :) )

i'll make a note in my draft PR to make sure this text / a version of this text is given a proper highlight

scottaohara:
closing per above comment / adding this as a comment to address in my reflow PR

Issue #157, opened 2016-02-06
Clarification over link/button states


awkawk:
Name: Jim Osborn
Email: (removed)
Affiliation:
Document: UW
Item Number: Understanding Success Criterion 1.4.3
Comment (Including rationale for any proposed change):
There is an implication that contrast requirements cover all states of links, buttons and inputs (hover, focus etc) but this is currently unclear and it would be good to have an explicit note to state this.

Likewise, if the above is an incorrect assumption then a note could clarify the requirements.

Proposed Change:
Add a new note (similar to note 3) to clarify if contrast success criterion for 1.4.3 applies to all states for interactive elements such as links, buttons and inputs.

For example:
"Note 4: Different interactive states (e.g. the focus state of a button) must also provide sufficient contrast to satisfy the minimum contrast success criterion (1.4.3)."

jnurthen:
We should probably exclude transient states from this. For example a button goes into the :active state when it is being pressed. IMO it is not necessary that its text content need be read when it is in this state as it is only like this for a short amount of time.

joshueoconnor:
I'm inclined to agree with @jnurthen on this. Do we need to change this SC so contrast requirements cover all states?

alastc:
I'm not sure about the transient states, for example, I observed a low-vision user in usability testing who was using a magnifier. He was zoomed in to around 800%, and when he moused-over a link, it became pink on white. Quite clear to me, but I found out later his vision struggled with colours like pink & orange on white. From his point of view it disappeared, which was quite disconcerting.
It wasn't a complete dead end, after a few goes he did continue, so I wouldn't object to excluding transient states, but I would ask if anyone else has seen more difficult examples.

jnurthen:
By transient states I didn't mean the focus/hover state. I agree that those need to be readable (except ironically on touch where generally no one can read those states as the finger is over the control).

I was talking about states which happen for only a really short amount of time like those when actually pressing a button.

gclowney:
@jnurthen, I don’t agree with your parenthetical exception: I rely heavily on focus/hover feedback when using mobile devices in order to reduce the number of wrong buttons pushed, especially when they’re closely spaced, and for realizing that the wrong button was pressed without having to switch my gaze to see the results. While my finger may obscure the glyph on an on-screen key I'm pressing, enough of its face is visible to make its color change obvious. I’m not convinced it’s less necessary for those with additional visual impairments.

[...]

PR #4127, opened 2024-11-01
Contrast minimum: Address the active (pressed) state

mfairchild365:
From #157: the active state (the typically split-second state when a button or control is being clicked or pressed) is not explicitly required to meet the contrast requirement.

This question comes up from time to time. The answer to this is provided in #157 but you have to dig to find it. Hopefully this change will make it easier to find the answer.

### DIFF
--- a/understanding/20/contrast-minimum.html
+++ b/understanding/20/contrast-minimum.html
@@ -119,7 +119,9 @@
       <p>This Success Criterion applies to text in the page, including
          placeholder text and text that is shown when a pointer is hovering over an object
          or when an object has keyboard focus. If any of these are used in a page, the text
-         needs to provide sufficient contrast.
+         needs to provide sufficient contrast. The active state (the typically split-second
+         state when a button or control is being clicked or pressed) is not explicitly
+         required to meet the contrast requirement.
       </p>
       
       <p>Although this Success Criterion only applies to text, similar issues occur for content presented


patrickhlauke:
At the risk of reopening old wounds, I'm not sure the decision to somehow special-case/exempt "temporary" states isn't a slippery slope. And frankly I'm not sure why you'd want to have low-contrast active states to begin with. Might be worth bringing this up for discussion again, as I'm really not sure there's a defensible rationale (other than "back in 2014 the group said it's ok, so we'll honour that decision")

stevefaulkner:
> (other than "back in 2014 the group said it's ok, so we'll honour that decision")

not a good enough reason

GreggVan:
What is the rationale for requiring the contrast to pass during the feedback that a button has been pressed. 

It makes sense for it to maintain contrast for hover or focus.
And I can see a slight advantage to having it maintain focus while depressed with a mouse (you can still slide off it and not click it to cancel).  But that is pretty borderline, does not exist for a keyboard activation    But I would think it was more important to have good change in contrast between focus and activated state for feedback.

I don’t think that it is an accessibility problem if the contrast drops during the moment of activation.  The person has all the time in the world to view and read the text on the button before they activate it.  So it is accessible. 

Gregg

patrickhlauke:
what is the need for the exemption? what is this trying to solve/allow? what sites would unnecessarily fail without this strangely specific exemption?

mfairchild365:
I’m all for opening old wounds! 🤣

Honestly, I just created this pull request because in the linked issue it sounded like a settled question. Personally, I think it should fail normatively, but with a minimal impact / severity (of which WCAG has no concept of).

Im totally fine if we want to discuss this further rather then merge.

patrickhlauke:
If the idea is that contrast ratios don't apply to things in general that happen very quickly/for a short period of time - like a quick transition for instance - then it's worth defining this in broad terms, rather than just mentioning the ultra-specific case of `:active` state of something.

mfairchild365:
I don’t think this should apply to things in general that happen very quickly. Sometimes important information will flash on the screen (which may fail a different SC), but that content should still have sufficient contrast.

I think the argument here is that the element itself is not transitory. Instead it’s just a state that is transitory and attached to a permanent element that contains the same text / information that displays in the transitory state. Additionally, the state is triggered by the user and is displayed for only a brief moment before the element is removed. It’s only the moment in time when an element is being activated. 

Again, normatively, I think this should fail. Should I update the PR to clarify that it fails?

mraccess77:
It seems like updating the documentation on active state may require some sort of review by the wider group.

patrickhlauke:
In the first instance, we'll discuss it in the WCAG 2.x backlog meeting/subgroup, to then take it to the wider AGWG

PR #4128, opened 2024-11-01
Update F2.html

### LINKED ISSUE #4109: Maybe remove Example 3 of F2
While trying to fill in all the blanks I have in the understanding docs, I was a bit surprised to find [example 3](https://www.w3.org/WAI/WCAG22/Techniques/failures/F2#example-3-using-css-to-visually-emphasize-a-phrase-or-word-without-conveying-that-emphasis-semantically) of F2 "Failure of Success Criterion 1.3.1 due to using changes in text presentation to convey information without using the appropriate markup or text" (both in WCAG 2.1 and 2.2). This is exactly the kind of accessibility dogoodery I wasted a lot of time on, 20 years ago. 😑

A few objections that come to mind:

1) I couldn't name AT that supports this type of text-level semantics, either at all, or with default configuration.
2) NVDA can be configured to announce text attributes; that works fine for the failure example as well.
3) Unlike the previous two examples, no remediation guidance is offered. Is the &lt;b>-element sufficient?
4) Well actually 🤭 I would fail usage of &lt;strong> or &lt;em> for a task that for any reason depends on conveying emphasis to AT for not being accessibility supported. Am I wrong?
5) Rhetorical complaint: Doesn't the failure example contain an inline quote that isn't conveyed to AT, either? 🤪

This just seems to reinforce what would I would call typical beginner mistakes. But I am also surprised. Am I missing something?

mbgower:
- ~~rephrased the introduction to include mention of assistive technology~~
- removed example 3
- updated outcome logic of checks 1 and 2
- ~~added a third condition to check 2~~

Resolves #4109


techniques/failures/F2.html
```diff
@@ -65,16 +47,17 @@ <h3>Using CSS to visually emphasize a phrase or word without conveying that emph
...
+                  <li>Check that supported assistive technologies surface the change in text styling.</li>
                </ol>
             </li>
          </ol>
       </section>
       <section class="results"><h3>Expected Results</h3>
          <ul>
-            <li>If check #1.1 is true, then #1.2 is true.</li>
-            <li>If check #2.1 is true, then #2.2 is true.</li>
+            <li>If check 1.1 is true and 1.2 is false, this failure condition applies and the content fails this success criterion. If either 1.2 is true, the failure condition does not apply.</li>
```
mbgower, line 59:
```suggestion
            <li>If check 1.1 is true and 1.2 is false, this failure condition applies and the content fails this success criterion. If 1.2 is true, the failure condition does not apply.</li>
```

techniques/failures/F2.html
```diff
@@ -65,16 +47,17 @@ <h3>Using CSS to visually emphasize a phrase or word without conveying that emph
             <li>
                <p>For styled text that conveys information:</p>
                <ol>
-                  <li>Check if there is any styled text that conveys structural information.</li>
-                  <li>Check that in addition to styling, the proper semantic structure is used with the text to convey the information.</li>
+                  <li>Check if there is any styled text whose change in appearance conveys meaning.</li>
+                  <li>Check that in addition to styling, the proper semantic structure is used with the text to convey the visual information.</li>
+                  <li>Check that supported assistive technologies surface the change in text styling.</li>
```
mbgower, line 52:
```suggestion
                  <li>Check that supported assistive technologies surface the change in meaning in text styling.</li>
```

techniques/failures/F2.html
```diff
@@ -1,7 +1,7 @@
...
       <p>All technologies that support images or presentation markup.</p>
    </section><section id="description"><h2>Description</h2>
-      <p>This document describes a failure that occurs when a change in the appearance of text conveys meaning without using appropriate semantic markup. This failure also applies to images of text that are not enclosed in the appropriate semantic markup.</p>
+      <p>This document describes a failure that occurs when a change in the appearance of text conveys meaning without using appropriate semantic markup, or without using text styling that is conveyed by assistive technology. This failure also applies to images of text that are not enclosed in the appropriate semantic markup.</p>
```
dbjorge, line 4: (per TF discussion) It feels off to add this clause when the rest of the PR is specifically *removing* the only example where it would apply. In the headings examples, it wouldn't be sufficient for ATs to convey a change in text styling; that wouldn't be sufficient to convey the semantic heading structure.

I would lean more towards updating the technique to more specifically talk about not using non-semantic visual techniques to present stuff that looks semantically relevant (the 2 remaining examples with headings are good, could also add something like "using a left-margin and images instead of a semantic list to indicate a list" if we want a non-heading example)

dbjorge (changes requested):

bruce-usab:
Discussed on backlog call 11-1.  Concern on call with only having heading examples, since CSS can semantically be used to convey emphasis.

techniques/failures/F2.html
mbgower: okay, thanks @dbjorge.

I will pull back on additive material for now.

mbgower:
Created a new issue to track unincorporated changes https://github.com/w3c/wcag/issues/4132

techniques/failures/F2.html
mbgower: @dbjorge please re-review so this can be merged if/when supported by WG.

techniques/failures/F2.html
```diff
@@ -65,16 +47,15 @@ <h3>Using CSS to visually emphasize a phrase or word without conveying that emph
...
+                  <li>Check that in addition to styling, the proper semantic structure is used with the text to convey the visual information.</li>
                </ol>
             </li>
          </ol>
       </section>
       <section class="results"><h3>Expected Results</h3>
          <ul>
-            <li>If check #1.1 is true, then #1.2 is true.</li>
-            <li>If check #2.1 is true, then #2.2 is true.</li>
+            <li>If, in either situatinoa, check 1 is true and check 2 is false, this failure condition applies and the content fails this success criterion. If condition 2 is true, the failure condition does not apply.</li>
```
mbgower, line 58:
```suggestion
            <li>If, in either situation, check 1 is true and check 2 is false, this failure condition applies and the content fails this success criterion. If condition 2 is true, the failure condition does not apply.</li>
```

techniques/failures/F2.html
```diff
@@ -65,16 +47,15 @@ <h3>Using CSS to visually emphasize a phrase or word without conveying that emph
             <li>
                <p>For styled text that conveys information:</p>
                <ol>
-                  <li>Check if there is any styled text that conveys structural information.</li>
-                  <li>Check that in addition to styling, the proper semantic structure is used with the text to convey the information.</li>
+                  <li>Check if there is any styled text whose change in appearance conveys meaning.</li>
```
mbgower, line 50:
```suggestion
                  <li>Check if there is any styled text that conveys structural information.</li>
```

mbgower:
Once the other non-structural changes have been added in the other Issue, the wording of the first check can be adjusted to something like the a prior commit

dbjorge (approved):

detlevhfischer:
Test: The sentence 

> If condition 2 is true, the failure condition does not apply 

seems superfluous and therefore a bit odd (it seems implied in the first part).

mbgower:
> Test: The sentence
> 
> > If condition 2 is true, the failure condition does not apply
> 
> seems superfluous and therefore a bit odd (it seems implied in the first part).

The existing language does not actually say what results produce a failure. I agree that the second sentence (which declares what does not fail) is restating the other side of a logic loop. If people feel its removal is unlikely to result in erroneous interpretation, I'm fine to remove. But I don't see it causes a problem being there.

> But I have more general concerns about this Failure Technique. People will use it to fail content that use things like `b` and `i` instead of `strong` and `em` so they will let content fail for stuff that is usually not semantically available to screen reader users (unless you explicitly check for styling / emphasis). And there are cases where the author might rightfully claim that `b` or `i` were used for stylistic reasons that do not map onto the HTML semantics. Should there be a not limiting of the Failure to cases like headings where the practical issue is clear (or a warning that often inline semantic markup is not honored by AT)?

As you may see from the discussion, there is general concern with the scope of this failure technique. We are removing with this change example 3 as a first attempt to reduce the problems. There is also another issue open to further this work.
That said, can you provide an example of a situation where bold and italic are used that are not meant to convey meaning that could not be conveyed semantically?

detlevhfischer:
> That said, can you provide an example of a situation where bold and italic are used that are not meant to convey meaning that could not be conveyed semantically? 

I think it could be cases where you don't want to put emphasis on a bit of text but differentiate: technical terms, taxanomic terms, etc. Or a title that is not put into quotation marks:

In Edgar Allan Poe's short story <i>A Descent into the Maelström</i> a man recounts how he survived a shipwreck and a whirlpool.

See also https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i

I admit there is a blurry line between `i` and `em`...

mbgower:
> I think it could be cases where you don't want to put emphasis on a bit of text but differentiate: technical terms, taxanomic terms, etc. Or a title that is not put into quotation marks:
> 
> In Edgar Allan Poe's short story _A Descent into the Maelström_ a man recounts how he survived a shipwreck and a whirlpool.

But that <em>has</em> a precise meaning. Italics used in such a case are meant to identify the title of a book. The appropriate element to use would be the `<cite>` element, which is normally shown as italics by browsers. 

> See also https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i

That page also identifies some of these other semantic meanings some of which are also rendered as italicized text:
>Be sure the text in question is not actually more appropriately marked up with another element.
> - Use [`<em>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/em) to indicate stress emphasis.
> - Use [`<strong>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strong) to indicate importance, seriousness, or urgency.
> - Use [`<mark>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/mark) to indicate relevance.
> - Use [`<cite>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/cite) to mark up the name of a work, such as a book, play, or song.
> - Use [`<dfn>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dfn) to mark up the defining instance of a term.

techniques/failures/F2.html
```diff
@@ -66,15 +48,14 @@ <h3>Using CSS to visually emphasize a phrase or word without conveying that emph
                <p>For styled text that conveys information:</p>
                <ol>
                   <li>Check if there is any styled text that conveys structural information.</li>
-                  <li>Check that in addition to styling, the proper semantic structure is used with the text to convey the information.</li>
+                  <li>Check that in addition to styling, the proper semantic structure is used with the text to convey the visual information.</li>
```
mbgower, line 51:
```suggestion
                  <li>Check that in addition to styling, the proper semantic structure is used with the text to convey the information.</li>
```

mbgower: Based on discussion with Detlev and prior discussions, pulling back the addition of "visual" as that may be unnecessarily confining what kind of information is conveyed.

mraccess77:
> > I think it could be cases where you don't want to put emphasis on a bit of text but differentiate: technical terms, taxanomic terms, etc. Or a title that is not put into quotation marks:
> > In Edgar Allan Poe's short story _A Descent into the Maelström_ a man recounts how he survived a shipwreck and a whirlpool.
> 
> But that _has_ a precise meaning. Italics used in such a case are meant to identify the title of a book or work. The appropriate element to use would be the `<cite>` element, which is normally shown as italics by browsers.

It seems like a citation in a paragraph that didn't use cite could be a failure if there was no indication of what the italics implies, however, I wouldn't consider not using cite it was on a references page that was marked References?

techniques/failures/F2.html
patrickhlauke: information/meaning

mbgower:
Retaining in discussion since there are no thumbs up on this topic.

bruce-usab (approved):

PR #4131, opened 2024-11-01
Dotted underline link technique and working example

hata6502:
Resolves https://github.com/w3c/wcag/issues/4133
[Preview](https://deploy-preview-4131--wcag2.netlify.app/techniques/css/dotted-underlines.html)

New technique and working example for [success criterion 1.4.1: Use of Color](https://www.w3.org/WAI/WCAG21/Understanding/use-of-color).

This technique applies to pages there are a large number of links in the block of text.

This technique is an alternative to use solid underlines.
Solid underlines might be visually distracting for pages there are a large number of links in the block of text.
Links are commonly used to indicate words or phrases that are links within a paragraph or other block of text.
Even using underlines on hover or keyboard focus only, they might not be used for users with touch devices.

Please leave a comment 🙌


hata6502:
I'm not a W3C Working Group member, I'll make a non-member patent licensing commitment if it technique can be acceptable. 
https://github.com/w3c/wcag/blob/main/CONTRIBUTING.md
https://github.com/w3c/csswg-drafts/issues/6866#issuecomment-997867056

patrickhlauke:
Reviewed the technique, and while it's an interesting idea, I don't personally think it merits inclusion as a technique/example. How authors decide to style their links - particularly if there are many of them - is up to them really. If they choose to do this, more power to them, but techniques aren't meant to be exhaustive/cover all possible ways to solve something, and in this case I'm not sure there's value in highlighting this approach to authors.

patrickhlauke:
Discussed this on the 21/02 backlog meeting. Working group agrees that, while interesting, this technique does not warrant inclusion in the understanding/techniques.

Thank you for your suggestion, though.

hata6502:
Thank you!

PR #4123, opened 2024-10-29
Update changelog in guidelines/index.html

mbgower:
Updated changelog to only record changes made that were preserved through WCAG 2.2's Oct 2023 publication.
Added normative errata to the change log incorporated into the republication
**NOTE** @iadawn, I took a guess on the actual republication date of Nov 15, which will need to be updated to match the real date.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/pull/4123.html" title="Last updated on Nov 1, 2024, 5:23 PM UTC (bc23f43)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag/4123/b99ada8...bc23f43.html" title="Last updated on Nov 1, 2024, 5:23 PM UTC (bc23f43)">Diff</a>

PR #2220, opened 2022-02-09
Create update-pr.yml

michael-n-cooper:
[Preview Changes](https://raw.githack.com/w3c/wcag/pr-preview/guidelines/index.html)

Add pull request preview

PR #4141, opened 2024-11-13
Generate guidelines during Netlify PR preview builds

kfranqueiro:
This effectively replaces what the pr-preview bot currently does (and what #2220 sought to do within the repo). The problem with the pr-preview bot is its previews don't fully work (e.g. as seen in #4123).

## Changes

- Remove `.pr-preview.json` to stop pr-preview bot from running
- Add logic to `eleventy.config.ts` to copy `guidelines` folder assets and run through spec-generator (the same thing we use for GH Pages builds; also the same thing used by the pr-preview bot, IIUC) when running a standard build through netlify
- Conditionally add guidelines list item to top-level `index.html` (which is only used for local dev and PR builds)
- Tangentially, rename our github-pages push workflow to be more distinguishable in the Actions tab

### DIFF
--- a/.github/workflows/11ty-publish.yaml
+++ b/.github/workflows/11ty-publish.yaml
@@ -1,4 +1,4 @@
-name: CI
+name: Push to gh-pages branch
 
 # Reference documentation: https://docs.github.com/en/actions/reference
 
--- a/.pr-preview.json
+++ b/.pr-preview.json
@@ -1,4 +1,0 @@
-{
-    "src_file": "guidelines/index.html",
-    "type": "respec"
-}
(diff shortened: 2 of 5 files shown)


iadawn:
Can we get something in the index file to note that this is a preview?

kfranqueiro:
> Can we get something in the index file to note that this is a preview?

I assume you mean for the top-level index page, since the guidelines page already has a big yellow preview banner at the bottom (I saw this in what the pr-preview bot produces and wanted to reproduce it, and figuring out how to do so took me longer than any other part of the PR lol)

image

Anyway, I've also added a paragraph with a link to the PR in the top-level index now:

image

iadawn:
Perfect!

iadawn (approved):

PR #1481, opened 2020-10-17
Change markers: add New visual style and remove unnecessary "text semantics"

patrickhlauke:
In WCAG 2.1 there's one change marker for the third note in "Full pages" https://www.w3.org/TR/WCAG21/#cc2 - this currently looks odd/just like an awkward start of the sentence.

> WCAG 2.1 third note in 'Full pages', with 'New' at the start of the paragraph looking like it's just part of the sentence

In WCAG 2.2, there's a few more change markers, and these use JavaScript to add "text semantics", wrapping them in square brackets. but these don't - to my mind at least - add anything useful ... they're not read out by default by AT unless you go character by character, and as they just live in their own little paragraph, they don't really need to delimit start/end of anything and are just as understandable without the square brackets.

> ![one of WCAG 2.2's '[New]' change markers](https://user-images.githubusercontent.com/895831/96336427-9ca2ce00-1077-11eb-8ca2-f3db0f24bf22.png)

**This PR adds a more visually explicit style, and removes the square brackets**

WCAG 2.1 third note in 'Full pages', with 'New' at the start of the paragraph, now styled as bold white text on a red box with slightly rounded corners, like a tag/badge

one of WCAG 2.2's 'New' change markers, styled as bold white text on a red box with slightly rounded corners, and without the square brackets around the word

PR #4145, opened 2024-11-20
Change "NEW" marker to use W3C blue background

patrickhlauke:
Instead of the original red - see https://github.com/w3c/wcag/pull/1481#issuecomment-2447584448


kfranqueiro (approved):

bruce-usab:
Discussed on TF call 11/22, concurrence that this is a bug fix.  Might be mentioned on AG call 12/3.  Ken to move to Done when merged.

PR #4147, opened 2024-11-21
2.4.6-headings-and-labels-descriptive-icons

### LINKED ISSUE #4095: 2.4.6 Headings and Labels appears to prohibit the use of icons as labels

detlevhfischer:
Closes #4095 
A proposed addition to the 2.4.6 Understanding text that states that conventional icons can in some cases serve as descriptive label of form controls. Included is an example of a conventional icon (loupe for search).

- Addresses the isue raised in https://github.com/w3c/wcag/issues/4095
- Adresses the point regarding the descriptiveness of icon labels brought up in https://github.com/w3c/wcag/pull/4102

See [rendered version of the revised 2.4.6 Understanding text (updated 26 Nov 2024)](https://raw.githack.com/w3c/wcag/a9e2ef7668d136a631fc7467530796a8ce9debdb/understanding/20/headings-and-labels.html)


understanding/20/headings-and-labels.html
```diff
@@ -93,6 +96,9 @@ <h2>Examples of Headings and Labels</h2>
          <dt>A form asking for the name of the user</dt>
          <dd>A form asks for the name of the user. It consists of two input fields to ask for the first
                and last name. The first field is labeled <q>First name</q>, the second is labeled <q>Last name</q>.</dd>
+         <dt>A search field labeled by a loupe icon</dt>
```
scottaohara, line 99: probably also referring to this as a "magnifying glass"

not everyone is familiar with the "loupe" term

i'll similarly add the "loupe" term to the 3.3.2 PR

detlevhfischer: good point, will change to magnifying lass

understanding/20/headings-and-labels.html
```diff
@@ -30,6 +30,9 @@ <h2>Intent of Headings and Labels</h2>
       <p>Labels and headings do not need to be lengthy. A word, or even a single character,
          may suffice if it provides an appropriate cue to finding and navigating content.
       </p>
+       <p>Labels of form controls are usually text-based. In some cases, images can serve as descriptive label without additional text. In these cases, authors should ensure that the image and its use as label in a particular context can be considered conventional and widely understood. 
+         Note that the same icon can be interpreted differently in different contexts and still be considered descriptive in that context. For example, placed after a search field, a loupe icon is commonly interpreted as triggering a search. 
+         Placed on an image, a loupe icon is commonly interpreted as opening a magnified version of the image (or zooming into that image).</p>
```
scottaohara: take or leave these suggestions.  just trying to expand on the idea here.
```suggestion
         Placed on or near another image, a loupe or magnify glass icon is commonly interpreted as a means to view a magnified version of the image (for instance, acting as a mechanism to zoom into the image, or opening a full sized image in a new window).</p>
```

understanding/20/headings-and-labels.html
scottaohara: thanks!   ended up doing a larger review (that i kept having to pickup / put down over the last hour) so some additional suggestions for you to consider.

understanding/20/headings-and-labels.html
scottaohara: outdated due to inflight conflict so redid my suggestion

understanding/20/headings-and-labels.html
```diff
@@ -30,6 +30,9 @@ <h2>Intent of Headings and Labels</h2>
       <p>Labels and headings do not need to be lengthy. A word, or even a single character,
          may suffice if it provides an appropriate cue to finding and navigating content.
       </p>
+       <p>Labels of form controls are usually text-based. In some cases, images can serve as descriptive label without additional text. In these cases, authors should ensure that the image and its use as label in a particular context can be considered conventional and widely understood. 
+         Note that the same icon can be interpreted differently in different contexts and still be considered descriptive in that context. For example, placed after a search field, a magnifying glass icon is commonly interpreted as triggering a search.
```
scottaohara, line 34: Some suggested text for you to consider:
- swapped out the first instance of "icon" with "image", as previously "image" was the term being used, and prior to introducing the example, changing to "icon" seemed awkward.
- Split the first sentence into two, and carried over the "conventional and widely understood" caveat from the previous paragraph to explain why it can be still descriptive in different contexts.
- changed "when placed after" to "accompanying a" since the former was narrowly scoped and not representative of how all instances of search fields might be designed.
- similarly changed 'triggering' to the changed text because sometimes a search icon is merely a label, and sometimes it also acts as a button, but it always is a means to identifying the search widget.

```suggestion
         Note that the same image can be interpreted differently in different contexts. However, it can still be considered descriptive if its use is commonly understood in each context. For example, when accompanying a text field, a loupe or magnifying glass icon is commonly interpreted as indicating the field is for entering and submitting a search query. 
```

understanding/20/headings-and-labels.html
```diff
@@ -30,6 +30,9 @@ <h2>Intent of Headings and Labels</h2>
       <p>Labels and headings do not need to be lengthy. A word, or even a single character,
          may suffice if it provides an appropriate cue to finding and navigating content.
       </p>
+       <p>Labels of form controls are usually text-based. In some cases, images can serve as descriptive label without additional text. In these cases, authors should ensure that the image and its use as label in a particular context can be considered conventional and widely understood.
```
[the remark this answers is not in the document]
scottaohara: re-added my suggestion due to the inflight conflict of the text being updated, while i was still in process of finishing my review suggetions

understanding/20/headings-and-labels.html
```diff
@@ -30,6 +30,9 @@ <h2>Intent of Headings and Labels</h2>
       <p>Labels and headings do not need to be lengthy. A word, or even a single character,
          may suffice if it provides an appropriate cue to finding and navigating content.
       </p>
+       <p>Labels of form controls are usually text-based. In some cases, images can serve as descriptive label without additional text. In these cases, authors should ensure that the image and its use as label in a particular context can be considered conventional and widely understood. 
+         Note that the same icon can be interpreted differently in different contexts and still be considered descriptive in that context. For example, placed after a search field, a loupe icon is commonly interpreted as triggering a search.
```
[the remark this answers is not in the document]
scottaohara: re-added my suggestion due to the inflight conflict of the text being updated, while i was still in process of finishing my review suggestions

detlevhfischer:
Sounds very reasonable. Do you want to create a PR on my PR or do you want me to implement those changes?

scottaohara:
@detlevhfischer i provided them as suggestions so that if you agree with them, you can just commit them into your existing PR via GitHub's "commit suggestion" button that follows each of my proposed changes.

I would rather you be the one to commit these, as personally am against the idea of someone editing/committing to someone else's PR without explicit permission :)

detlevhfischer:
@scottaohara 
> as personally am against the idea of someone editing/committing to someone else's PR without explicit permission :)

I gave you permission, I suggested it  :) - but no problem, I think your suggestions are good and I will work them in myself.

patrickhlauke:
@detlevhfischer

>  I will work them in myself

just to reiterate, if it's not obvious: all you need to do is press the "Commit suggestion" for each here in github, no need to manually try to rework your PR

A11yTea:
Hey I was just reading about this and I was thinking you might want to add a reference to 1.1.1 Non-text Content to make sure they have an equivalent alternative?

understanding/20/headings-and-labels.html
```diff
@@ -30,6 +30,9 @@ <h2>Intent of Headings and Labels</h2>
       <p>Labels and headings do not need to be lengthy. A word, or even a single character,
          may suffice if it provides an appropriate cue to finding and navigating content.
       </p>
+       <p>Labels of form controls are usually text-based. In some cases, images can serve as descriptive labels without additional text. In these cases, authors would need to ensure that the image and its use as label in a particular context can be considered conventional and widely understood. 
+         Note that the same image can be interpreted differently in different contexts. However, it can still be considered descriptive if its use is commonly understood in each context. For example, when accompanying a text field, a loupe or magnifying glass icon is commonly interpreted as indicating the field is for entering and submitting a search query.
```
mbgower, line 34:
```suggestion
         Note that the same image can be interpreted differently in different contexts. However, it can still be considered descriptive if its use is commonly understood in each context. For example, when accompanying a text field, a loupe or magnifying glass icon with an `alt` of "Search" is commonly interpreted as indicating the field is for entering and submitting a search query. 
```

understanding/20/headings-and-labels.html
```diff
@@ -93,6 +96,9 @@ <h2>Examples of Headings and Labels</h2>
          <dt>A form asking for the name of the user</dt>
          <dd>A form asks for the name of the user. It consists of two input fields to ask for the first
                and last name. The first field is labeled <q>First name</q>, the second is labeled <q>Last name</q>.</dd>
+         <dt>A search field labeled by a magnifying glass icon</dt>
+         <dd>A search text input is followed by a button containing a magnifying glass icon that activates the search function. 
+               The icon has the string "search" as programmatically determinable label.</dd>
```
gundulaniemann, line 101: A visible text alternative is also needed. Can we please add an according sentence?

patrickhlauke: isn't this the whole point of this PR? clarifying that if an icon is "descriptive" enough, it *doesn't* need text?

mbgower: Agreed: "In some cases, images can serve as descriptive labels without additional text."

So it needs a text alternative, but the alternative does not need to be visible, just programmatically available (which can be exposed to users in a number of ways, including visibly, through user agents and AT)

understanding/20/headings-and-labels.html
```diff
@@ -30,6 +30,9 @@ <h2>Intent of Headings and Labels</h2>
       <p>Labels and headings do not need to be lengthy. A word, or even a single character,
          may suffice if it provides an appropriate cue to finding and navigating content.
       </p>
+       <p>Labels of form controls are usually text-based. In some cases, images can serve as descriptive labels without additional text. In these cases, authors would need to ensure that the image and its use as label in a particular context can be considered conventional and widely understood.
```
mbgower, line 33: a suggestion to cite 1.1.1, as one commenter requested:
```suggestion
       <p>Labels of form controls are usually text-based. In some cases, images can serve as descriptive labels without additional text. In these cases, authors would need to ensure that the image and its use as label in a particular context can be considered conventional and widely understood (and that the image meets 1.1.1 Non-text Content by providing a text alternative). 
```

understanding/20/headings-and-labels.html
mbgower: @patrickhlauke does this work?

scottaohara: This addition would make it seem like all instances of images directly need to have alt text, when in reality the image could have no alt text because the form field’s accname is clear enough.  Eg having a search icon with alt=search next to a text field with an accname of search is not what anyone should think they need to do in this situation. 

To put it bluntly, this just seems to make this more complicated, and to provide the appropriate nuance, it’d submit more verbiage is necessary.

understanding/20/headings-and-labels.html
scottaohara: Per my above comment, I don’t think this addition is nuanced enough since this would imply every image itself needs alt text, but that isn’t always true

understanding/20/headings-and-labels.html
patrickhlauke: it's the *text input* that needs an accessible name, not the icon/image used as visual label, basically.

understanding/20/headings-and-labels.html
alastc, line 33:
```suggestion
       <p>Labels of form controls are usually text-based. In some cases, images can serve as descriptive labels without additional text. In these cases, authors should ideally ensure that the image and its use as a label (in context) are widely understood. 
```

alastc: I don't think "descriptive" from the normative text means that it has to be considered conventional or widely understood, so I'd frame that as a useful thing to do but not required.

understanding/20/headings-and-labels.html
```diff
@@ -30,6 +30,9 @@ <h2>Intent of Headings and Labels</h2>
       <p>Labels and headings do not need to be lengthy. A word, or even a single character,
          may suffice if it provides an appropriate cue to finding and navigating content.
       </p>
+       <p>Labels of form controls are usually text-based. In some cases, images can serve as descriptive labels without additional text. In these cases, authors would need to ensure that the image and its use as label in a particular context can be considered conventional and widely understood. 
+         Note that the same image can be interpreted differently in different contexts. However, it can still be considered descriptive if its use is commonly understood in each context. For example, when accompanying a text field, a loupe or magnifying glass icon with text alternative of "Search" is commonly interpreted as indicating the field is for entering and submitting a search query.
```
alastc, line 34:
```suggestion
         <p>Note that the same image can be interpreted differently in different contexts. However, it can still be considered descriptive if its use is commonly understood in each context. For example, when accompanying a text field, a loupe or magnifying glass icon with text alternative of "Search" is commonly interpreted as indicating the field is for entering and submitting a search query.</p>
```

understanding/20/headings-and-labels.html
```diff
@@ -30,6 +30,9 @@ <h2>Intent of Headings and Labels</h2>
       <p>Labels and headings do not need to be lengthy. A word, or even a single character,
          may suffice if it provides an appropriate cue to finding and navigating content.
       </p>
+       <p>Labels of form controls are usually text-based. In some cases, images can serve as descriptive labels without additional text. In these cases, authors should ensure that the image and its use as a label (in context) are widely understood. 
+         <p>Note that the same image can be interpreted differently in different contexts. However, it can still be considered descriptive if its use is commonly understood in each context. For example, when accompanying a text field, a loupe or magnifying glass icon with text alternative of "Search" is commonly interpreted as indicating the field is for entering and submitting a search query.</p>
+         Placed on or near another image, a loupe or magnifying glass icon is commonly interpreted as a means to view a magnified version of the image (for instance, acting as a mechanism to zoom into the image, or opening a full-sized image in a new window).</p>
```
alastc, line 35:
```suggestion
        <p>Placed on or near another image, a loupe or magnifying glass icon is commonly interpreted as a means to view a magnified version of the image (for instance, acting as a mechanism to zoom into the image, or opening a full-sized image in a new window).</p>
```

understanding/20/headings-and-labels.html
```diff
@@ -30,6 +30,9 @@ <h2>Intent of Headings and Labels</h2>
       <p>Labels and headings do not need to be lengthy. A word, or even a single character,
          may suffice if it provides an appropriate cue to finding and navigating content.
       </p>
+       <p>Labels of form controls are usually text-based. In some cases, images can serve as descriptive labels without additional text. In these cases, authors should ensure that the image and its use as a label (in context) are widely understood.
```
mbgower, line 33:
```suggestion
       <p>Labels of form controls are usually text-based. In some cases, images can serve as descriptive labels without visible text. In these cases, authors should ensure that the image and its use as a label (in context) are widely understood. 
```

alastc, line 33:
```suggestion
       <p>Labels of form controls are usually text-based. In some cases, images can serve as descriptive labels without additional text. In these cases, authors should ensure that the image and its use as a label (in context) are widely understood.</p>
```

patrickhlauke:
Mentioned this in the call, but for posterity/on the record: I do appreciate that first extra sentence

```
In these cases, authors should ensure that the image and its use as a label (in context) are widely understood.
```

as this to me still leaves the door open for auditors to fail egregiously bad uses of icons without text. If an auditor makes a  subjective judgement call that they believe the icon is not sufficiently understood/clear to users, they can still fail these icon-only labels (though, as this is subjective, there *may* be differing opinions ... but nothing that we don't already have in other cases).

Yes, conceptually, icon-only labels *can* be problematic for certain users - those who are unfamiliar with a particular icon, or users with cognitive disabilities. But I don't believe the original intent of the SC ever forbade the use of icon-only labels, and insisting that icons should always be accompanied by visibile text (next to the icon, or as a hover/tooltip, or by having a setting where a user can choose between "icon only" / "icon with text label" like some software does) would introduce a novel normative requirement by stealth here, I'd say.

TestPartners:
It's ironic that the sentence "In these cases, authors should ensure that the image and its use as a label (in context) are widely understood" is followed by a button whose purpose I do not know because it has an unfamiliar (to me) icon and no text label or tooltip.

understanding/20/headings-and-labels.html
A11yTea: Good point, sorry I missed that. In that case, its basically only 4.1.2 that is needed not 1.1.1.

A11yTea:
> as this to me still leaves the door open for auditors to fail egregiously bad uses of icons without text. If an auditor makes a subjective judgement call that they believe the icon is not sufficiently understood/clear to users, they can still fail these icon-only labels (though, as this is subjective, there _may_ be differing opinions ... but nothing that we don't already have in other cases).

Yes if you leave that phrase in, we will start to fail icons that we don't find intuitive. I can confirm that as an auditor.

PR #4173, opened 2024-12-12
2024 publication fixes

iadawn:


iadawn:
@kfranqueiro, not sure if there is anything in the deploy fails that I need to worry about

kfranqueiro:
The failure is because the WCAG-2.1 branch doesn't have the whole eleventy build system and package.json, and netlify is attempting to build against that for this PR. So that's a red herring.

kfranqueiro (approved):

PR #4183, opened 2024-12-20
Add 2.1 build script that reuses process from main branch

kfranqueiro:
These changes are part 2 of 2 to enable preview builds against the `WCAG-2.1` branch. This part applies to the `WCAG-2.1` branch; See #4182 for part 1, which applies to `main`.

This adds a shell script which downloads a tarball of the latest `main` branch, extracts it to a subfolder, then runs the Eleventy build within it, instructing it to build against 2.1 but using the local `guidelines` folder from the active branch rather than the latest published version.

I have temporarily stood up an example site using a version of this branch modified to run against the branch on my fork used for #4182:

https://kgf-wcag21-preview-test.netlify.app/

I have verified that the only diffs between the output of this script and directly running against the latest 2.1 publication from `main` are date stamps in page footers.

Here's what the deploy options look like for the netlify site:

Netlify deploy configuration for kgf-wcag21-preview-test, including build command set to ./build-21.sh, and publish directory set to wcag-main/_site

### DIFF
--- a/build-21.sh
+++ b/build-21.sh
@@ -1,0 +1,10 @@
+#! /bin/sh
+
+# Download main branch
+curl -Lo wcag.tar.gz https://github.com/w3c/wcag/archive/refs/heads/main.tar.gz
+tar xf wcag.tar.gz
+cd wcag-main
+npm i
+
+# Run build using local 2.1 guidelines from active branch
+WCAG_VERSION=21 WCAG_FORCE_LOCAL_GUIDELINES=../guidelines/index.html npm run build


iadawn (approved):
Presumably this will be attached to a workflow in the 2.1 branch? Or am I missing something?

kfranqueiro:
> Presumably this will be attached to a workflow in the 2.1 branch? Or am I missing something?

It will be used as the build command for a separate netlify site specifically for WCAG 2.1, to be used for PR builds against `WCAG-2.1`.

Currently we only deploy informative docs for 2.2; we would only use this for a github workflow if we were to also add 2.1 informative docs at separate paths within gh-pages.

Issue #3019, opened 2023-02-12
Technique H28 should not be sufficient to meet 3.1.4: Abbreviations


jialiang:
[Technique H28:Providing definitions for abbreviations by using the abbr element](https://www.w3.org/WAI/WCAG21/Techniques/html/H28)

Technically the `<abbr>` tag is used to indicate and define abbreviations. E.g. `<abbr title="pound">lb.</abbr>`.

But in reality, to quote the WHATWG (https://html.spec.whatwg.org/multipage/dom.html#the-title-attribute):

> Relying on the [title](https://html.spec.whatwg.org/multipage/dom.html#attr-title) attribute is currently discouraged as many user agents do not expose the attribute in an accessible manner as required by this specification (e.g., requiring a pointing device such as a mouse to cause a tooltip to appear, which excludes keyboard-only users and touch-only users, such as anyone with a modern phone or tablet).

True enough, in major browsers like Chrome and Firefox, you can't get the text in the title attribute to show up in any way if you're using touch or keyboard.

patrickhlauke:
I would be in favour of removing this technique, or demoting it from sufficient to advisory, exactly because `<abbr>` is not keyboard-supported - so following this technique would lead to a 2.1.1 failure.

PR #4195, opened 2025-01-12
[WCAG AAA] - Move H28 from sufficient to advisory in 3.1.4 Abbreviations SC

giacomo-petri:
Closes: #3019 

[H28](https://www.w3.org/WAI/WCAG21/Techniques/html/H28) is not a sufficient technique, rather than an advisory one, since it would likely fail 2.1.1

Actions: moved H28 from sufficient to advisory.

### DIFF
--- a/techniques/html/H28.html
+++ b/techniques/html/H28.html
@@ -18,7 +18,7 @@
       <h2>Description</h2>
       <p>The objective of this technique is to provide expansions or definitions for
             abbreviations by using the <code class="language-html">abbr</code> element. It is always appropriate to use the <code class="language-html">abbr</code> element for any abbreviation, including acronyms and initialisms.</p>
-            <p>This technique has been altered from a sufficient technique to an advisory technique because it uses the <code>title</code> attribute of <code>abbr</code> to provide the expansion of the abbreviation needed to acheive a sufficient technique when used in combination with <a href="../general/G102">G102</a>. As noted in the <a href="https://html.spec.whatwg.org/multipage/dom.html#the-title-attribute"> specification</a>:
+            <p>This technique has been altered from a sufficient technique to an advisory technique because it uses the <code>title</code> attribute of <code>abbr</code> to provide the expansion of the abbreviation needed to achieve a sufficient technique when used in combination with <a href="../general/G102">G102</a>. As noted in the <a href="https://html.spec.whatwg.org/multipage/dom.html#the-title-attribute"> specification</a>:
 
                <q>Relying on the title attribute is currently discouraged as many user agents do not expose the attribute in an accessible manner as required by this specification (e.g., requiring a pointing device such as a mouse to cause a tooltip to appear, which excludes keyboard-only users and touch-only users, such as anyone with a modern phone or tablet).</q></p>
    </section>
--- a/techniques/html/H28.html
+++ b/techniques/html/H28.html
@@ -18,7 +18,7 @@
       <h2>Description</h2>
       <p>The objective of this technique is to provide expansions or definitions for
             abbreviations by using the <code class="language-html">abbr</code> element. It is always appropriate to use the <code class="language-html">abbr</code> element for any abbreviation, including acronyms and initialisms.</p>
-            <p>This technique has been altered from a sufficient technique to an advisory technique for G102 because it uses the <code>title</code> attribute of <code>abbr</code> to provide the expansion of the abbreviation. As noted in the <a href="https://html.spec.whatwg.org/multipage/dom.html#the-title-attribute"> specification</a>:
+            <p>This technique has been altered from a sufficient technique to an advisory technique because it uses the <code>title</code> attribute of <code>abbr</code> to provide the expansion of the abbreviation needed to acheive a sufficient technique when used in combination with <a href="../general/G102">G102</a>. As noted in the <a href="https://html.spec.whatwg.org/multipage/dom.html#the-title-attribute"> specification</a>:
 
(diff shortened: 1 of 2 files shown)


bruce-usab:
Discussed on Backlog call 1/17.  Moved to Ready for approval.

mbgower:
IMO, there has to be some language inside the technique, and maybe inside the Understanding document, explaining the matter raised in the issue.

The SC language reads:
> A [mechanism](https://www.w3.org/WAI/WCAG21/Understanding/abbreviations.html#dfn-mechanism) for identifying the expanded form or meaning of [abbreviations](https://www.w3.org/WAI/WCAG21/Understanding/abbreviations.html#dfn-abbreviation) is available.

The use of `<abbr>` with an appropriate value for the `title` attribute clearly provides a mechanism for identifying the meaning of the abbreviation. So we have to be clear that the reason this is not sufficient is because of the note in the definition of [mechanism](https://www.w3.org/WAI/WCAG21/Understanding/abbreviations.html#dfn-mechanism)
> The mechanism needs to meet all success criteria for the conformance level claimed.

bruce-usab (changes requested):
I am of the opinion that something should be added to H28 explaining why it has been down graded from sufficient to advisory.

Keyboard support for `abbr` has been weak, but historically we didn't assert that the default behavior of the element fails 2.1.1.  But maybe enough time has passed that we should?

understanding/20/abbreviations.html
```diff
@@ -295,8 +277,9 @@ <h4>Situation B: If the abbreviation means different things within the same Web
       
       <section id="advisory">
          <h3>Additional Techniques (Advisory) for Abbreviations</h3>
-         
-         
+         <ul>
+           <li><a href="../Techniques/html/H28" class="html">Providing definitions for abbreviations by using the abbr and acronym elements</a></li>
```
bruce-usab, line 281: I believe the `acronym` element is deprecated, so maybe not include that one?

bruce-usab:
Discussed on backlog call 2/21.  Keeping in For Discussion as Mike is still has this in his queue.

mbgower:
I have added an explanation to the top of the H28 techniques which I think resolves this.

techniques/html/H28.html
```diff
@@ -18,6 +18,9 @@ <h2>When to Use</h2>
       <h2>Description</h2>
       <p>The objective of this technique is to provide expansions or definitions for
             abbreviations by using the <code class="language-html">abbr</code> element. It is always appropriate to use the <code class="language-html">abbr</code> element for any abbreviation, including acronyms and initialisms.</p>
+            <p>This technique has been altered from a sufficient technique to an advisory technique because it uses the <code>title</code> attribute of <code>abbr</code> to provide the expansion of the abbreviation needed to acheive a sufficient technique when used in combination with <a href="../general/G102">G102</a>. As noted in the <a href="https://html.spec.whatwg.org/multipage/dom.html#the-title-attribute"> specification</a>:
```
patrickhlauke, line 21:
```suggestion
            <p>This technique has been altered from a sufficient technique to an advisory technique because it uses the <code>title</code> attribute of <code>abbr</code> to provide the expansion of the abbreviation needed to achieve a sufficient technique when used in combination with <a href="../general/G102">G102</a>. As noted in the <a href="https://html.spec.whatwg.org/multipage/dom.html#the-title-attribute"> specification</a>:
```

techniques/html/H28.html
```diff
@@ -18,6 +18,9 @@ <h2>When to Use</h2>
       <h2>Description</h2>
       <p>The objective of this technique is to provide expansions or definitions for
             abbreviations by using the <code class="language-html">abbr</code> element. It is always appropriate to use the <code class="language-html">abbr</code> element for any abbreviation, including acronyms and initialisms.</p>
+            <p>This technique has been altered from a sufficient technique to an advisory technique because it uses the <code>title</code> attribute of <code>abbr</code> to provide the expansion of the abbreviation needed to achieve a sufficient technique when used in combination with <a href="../general/G102">G102</a>. As noted in the <a href="https://html.spec.whatwg.org/multipage/dom.html#the-title-attribute"> specification</a>:
```
patrickhlauke, line 21:
```suggestion
            <p>This technique has been altered from a sufficient technique to an advisory technique because it uses the <code>title</code> attribute of <code>abbr</code> to provide the expansion of the abbreviation needed to achieve a sufficient technique when used in combination with <a href="../general/G102">G102</a>. As noted in the <a href="https://html.spec.whatwg.org/multipage/dom.html#the-title-attribute"> specification</a>:</p>
```

techniques/html/H28.html
```diff
@@ -18,6 +18,9 @@ <h2>When to Use</h2>
...
+
+               <q>Relying on the title attribute is currently discouraged as many user agents do not expose the attribute in an accessible manner as required by this specification (e.g., requiring a pointing device such as a mouse to cause a tooltip to appear, which excludes keyboard-only users and touch-only users, such as anyone with a modern phone or tablet).</q></p>
```
patrickhlauke, line 23:
```suggestion
               <blockquote><p>Relying on the title attribute is currently discouraged as many user agents do not expose the attribute in an accessible manner as required by this specification (e.g., requiring a pointing device such as a mouse to cause a tooltip to appear, which excludes keyboard-only users and touch-only users, such as anyone with a modern phone or tablet).</p></blockquote>
```

bruce-usab (approved):
Thanks for addressing my ask!

PR #4196, opened 2025-01-12
Updated 2.5.8 Target Size (Minimum) Understanding Doc to add more clarity to the User agent control bit in Exceptions

### LINKED ISSUE #3998: Minimum Target Size, custom elements and user agent exception
In the user agent exception example it should be made clear that it always applies as long as the size still is default.

The exception in the normative text says:
"User agent control: The size of the target is determined by the user agent and is not modified by the author;"

So it talks about not manipulating the size. It does not talk about any other attribute a control might have, like color.

Yet the example still might be misunderstood as "if any attribute is altered, the exception does not apply".
Therefore is should say:
"User agent control: Browsers have default renderings of some controls, such as the days of the month calendar in an <input type="date">. As long as the author has not modified the user agent default size, the target size for a User agent control is excepted."

(The word 'size' was added after 'user agent default'.)

Some background for scrollbars:
The default size for scrollbars usually is 'auto'. This way any setting a browser or an extension might offer can be applied. This remains unchanged if for example the colors of the scrollbars have been changed by the author.
Only the size needs to remain untouched.

giacomo-petri:
Closes #3998 

Added the word "size" within the "Exceptions" section to enhance clarity.

Note: this is also clearly stated further down, but adding "size" also in the exceptions section won't cause any arm:
> This Success Criterion has an exception for the size of targets determined by a user agent and not modified by the author.


bruce-usab:
Discussed on Backlog call 1/17.  Moved to Ready for approval.

bruce-usab (approved):

PR #4245, opened 2025-02-25
Fix links to Understanding WCAG 2.1 to 2.2

momdo:
The links in the Key Terms section of the Understanding WCAG 2.2 Docs incorrectly reference Understanding WCAG 2.1.
e.g.: <https://www.w3.org/WAI/WCAG22/Understanding/non-text-content#dfn-text-alternative>

This PR might fix not only the Key Terms sections in Understanding documents but also WCAG 2.2 §6. Glossary.


patrickhlauke:
This has since been fixed (indeed, updating the branch now shows that this PR has 0 changes). The fix (https://github.com/w3c/wcag/commit/50ba7ae8506434713ae39401b99864c6f2109787) was applied in the editor's draft and will be live next time the spec is republished.

Issue #4246, opened 2025-02-25
Use of “session” in “In Brief” in Understanding SC 3.3.7: Redundant Entry (Level A)


yatil:
In the [Understanding for 3.3.7](https://www.w3.org/WAI/WCAG22/Understanding/redundant-entry.html), the “In Brief” section postulates:

> What to do: Don't ask for the same information twice in the same session.

The word session is not mentioned in the success criterion. It is part of the Understanding, but only in the context that information[ does not need to be preserved between sessions](https://www.w3.org/WAI/WCAG22/Understanding/redundant-entry.html#:~:text=This%20Success%20Criterion%20does%20not%20add%20a%20requirement%20to%20store%20information%20between%20sessions).

When you have multiple processes within a session, the success criterion permits re-entry, even within the same session.

Example:

A user places an order and enters a different shipping address. If they place another order right after that, the changed shipping address does not need to be preserved despite the user being in the same session.

Introducing the word session in “In Brief” is unnecessary and I would recommend using “process” instead.

mbgower:
The use of "session" made it intact through all the TF and WG reviews, but I agree it is not the best option for the reasons you state.

I'm going to suggest "procedure", simply because "process" already exists in the normative text, and one of the tactics we used was using synonyms rather than restating the normative wording, with the intention of broadening people's context.

> Don't ask for the same information twice in the same procedure.

patrickhlauke:
ah, i commented on the suggested change before seeing your comment here @mbgower - i'd actually say that introducing "procedure" here would make it less understandable, not more, and would opt to keep the term "process"

yatil:
Why would we want to introduce synonyms when there are normative defined terms for things, especially if it’s a commonly used word? The in-brief should be an on ramp, and I see people misinterpreting in-brief as pseudo-normative and trying to find definitions for words in them. Just use the words we have. It’s OK.

mbgower:
> Why would we want to introduce synonyms when there are normative defined terms for things, especially if it’s a commonly used word? The in-brief should be an on ramp, and I see people misinterpreting in-brief as pseudo-normative and trying to find definitions for words in them. Just use the words we have. It’s OK.

I already explained the tactical usage of synonyms in the formation of the In Brief materials. It is intentional and was supported by EO. The use of synonyms is endemic.

In this particular case, we have replaced the word "session" as you requested. The word activity is in usage in the Understanding document, specifically in context of the term "process"
[...]

PR #4247, opened 2025-02-25
Change Session to Process in “In Brief”

yatil:
- Related to w3c/wcag#4246

### DIFF
--- a/understanding/22/redundant-entry.html
+++ b/understanding/22/redundant-entry.html
@@ -12,7 +12,7 @@
         <h2>In brief</h2>
         <dl>
             <dt>Goal</dt><dd>Make it easier for users to complete multi-step processes.</dd>
-            <dt>What to do</dt><dd>Don't ask for the same information twice in the same process.</dd>
+            <dt>What to do</dt><dd>Don't ask for the same information twice in the same activity.</dd>
             <dt>Why it's important</dt><dd>Some people with cognitive disabilities have difficulty remembering what they entered before.</dd>
         </dl>
     </section>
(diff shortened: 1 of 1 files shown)


understanding/22/redundant-entry.html
```diff
@@ -12,7 +12,7 @@ <h1>Understanding Redundant Entry</h1>
         <h2>In brief</h2>
         <dl>
             <dt>Goal</dt><dd>Make it easier for users to complete multi-step processes.</dd>
-            <dt>What to do</dt><dd>Don't ask for the same information twice in the same session.</dd>
+            <dt>What to do</dt><dd>Don't ask for the same information twice in the same process.</dd>
```
mbgower, line 15:
```suggestion
            <dt>What to do</dt><dd>Don't ask for the same information twice in the same procedure.</dd>
```

patrickhlauke: personally, i find "process" more immediately understandable than "procedure".

also, process is a defined term https://www.w3.org/TR/WCAG22/#dfn-processes whereas "procedure" doesn't seem to feature in WCAG at this point

patrickhlauke (approved):

understanding/22/redundant-entry.html
patrickhlauke: "procedure" just makes my mind go straight to "medical procedure" for some reason...

patrickhlauke: if we do want to avoid the word "process", alternatives perhaps: "activity" or "user flow"? still not sure it would help, or whether that would just throw up more confusion for readers when it's not defined and unambiguous terms

scottaohara (approved):

yatil:
I do not think activity is better than session. It’s process. Please don’t merge this with a reference to my pull request.

yatil (commented):
I disagree with this change.

alastc:
I'll close this, Eric isn't happy with it, we can discuss in #4282

PR #4251, opened 2025-02-26
Remove redundant word "Rules"

momdo:
https://www.w3.org/WAI/WCAG22/Understanding/understanding-act-rules#test-rules-rules-are-informative


patrickhlauke:
LGTM

PR #4257, opened 2025-03-01
Replaces the previous name of the working group in Understanding

momdo:
Replace "WCAG" and "Web Content Accessibility Guidelines" with "Accessibility Guidelines".


patrickhlauke:
This one is likely a bit more controversial/not as straightforward, as *at the time* that WCAG came out, I think the WG was still called that. I think it's possibly safer to leave as is, *or* to neutralise it by just referring to "the Working Group" to make it applicable to both the old and new names.

mbgower:
A search on "WCAG working group" revealed 36 results over 19 files in the repo -- more than the 6 files changed here.
I'm going to review each of the other occurrences to confirm if they should likewise be altered.

Note that I also confirmed many occurrences of "Accessibility Guidelines Working Group", so the proposed term is certainly used in the index file and elsewhere.

@alastc note that I did discover inconsistency between the use of "AG WG" (53 results in 20 files) and "AGWG" (5 results in 3 files) in usage. I am going to correct the far less prevalent occurrences of AGWG.

mbgower:
I've determined that other occurrences seem to occur either in old wcag 2 areas not to be updated or in the specification, not relevant non-normative documents, so I will create a separate PR for those the spec changes.

IMO this change looks solid.

mbgower (approved):

mbgower:
@momdo I cannot find your PR, so I could not make additional changes in it.
One thing I noticed: you missed a few other corrections where the case was not correct. Would you please do a search for 
"WCAG Working group" and "WCAG working group" and ensure both are changed to "AG Working Group"

bruce-usab:
Discussed during Backlog call today and Mike is on the case!

momdo:
@mbgower 
I'm sorry for the inconvenience...
I found "WCAG Working group" in `understanding/conformance.html`, so I replaced it.
Could you please check if this commit is correct?

understanding/conformance.html
```diff
@@ -642,7 +642,7 @@ <h3>Level of Assistive Technology Support Needed for "Accessibility Support"</h3
          
          <p>This topic raises the question of how many or which assistive technologies must support
             a Web technology in order for that Web technology to be considered "accessibility
-            supported". The WCAG Working group and the W3C do not specify which or how many assistive
+            supported". The AG Working Group and the W3C do not specify which or how many assistive
```
kfranqueiro, line 645: Would we rather expand this to Accessibility Guidelines in all cases? (Other instances have been expanded from WCAG to Accessibility Guidelines in this PR, but this one is still abbreviated as AG.)

kfranqueiro:
> @kfranqueiro this ID is not in the group. What's the easiest way to address?

I've marked this as non-substantive (which I think is fair for this PR?) in order to pass the check.

understanding/conformance.html
patrickhlauke: I'd be in favour of expanding in all instances, yes

patrickhlauke:
```suggestion
            supported". The Accessibility Guidelines Working Group and the W3C do not specify which or how many assistive
```

PR #4258, opened 2025-03-03
Update links in Understanding SC 4.1.3

momdo:
- Update Material Design's link
- Add link to SC 1.1.1

### DIFF
--- a/understanding/21/status-messages.html
+++ b/understanding/21/status-messages.html
@@ -53,7 +53,7 @@
 					<li>An application displays a progressbar to indicate the status of an upgrade. The element is assigned a suitable role. The screen reader provides intermittent announcements of the progress. </li>
 					<li>After a user submits a form, text is added to the existing form which reads, "Your form was successfully submitted." The screen reader announces the same message.</li> 
 					<li>After a user unsuccessfully fills in a form because some of the data is in the incorrect format, text is added to the existing form which reads "5 errors on page". The screen reader announces the same message.</li>
-					<li>After a user puts a photo in an album in an online photo app, a <a href="https://material.io/design/components/snackbars.html#">snackbar </a> displays the message "Saved in 'Wedding' album", which is also read by a screen reader.</li>
+					<li>After a user puts a photo in an album in an online photo app, a <a href="https://m3.material.io/components/snackbar/overview">snackbar</a> displays the message "Saved in 'Wedding' album", which is also read by a screen reader.</li>
 				</ul>
 			</section>
 			<section class="example">
@@ -82,7 +82,7 @@
 				</section>
 				<section>
 		<h4>Non-textual status content</h4>
-		<p>Changes in content are not restricted to text changes. Where an icon or sound indicates a status message, this information will be surfaced by the screen reader through a combination of two things: 1) existing WCAG requirements governing text alternatives (under SC 1.1.1 Non-Text Content), and 2) the requirement of this current Success Criterion to supply an appropriate role.</p>
+		<p>Changes in content are not restricted to text changes. Where an icon or sound indicates a status message, this information will be surfaced by the screen reader through a combination of two things: 1) existing WCAG requirements governing text alternatives (under <a href="non-text-content" class="sc">SC 1.1.1 Non-Text Content</a>), and 2) the requirement of this current Success Criterion to supply an appropriate role.</p>
 				</section>
 			</section>
 		


understanding/21/status-messages.html
```diff
@@ -82,7 +82,7 @@ <h4>Removal of status text</h4>
 				</section>
 				<section>
 		<h4>Non-textual status content</h4>
-		<p>Changes in content are not restricted to text changes. Where an icon or sound indicates a status message, this information will be surfaced by the screen reader through a combination of two things: 1) existing WCAG requirements governing text alternatives (under SC 1.1.1 Non-Text Content), and 2) the requirement of this current success criterion to supply an appropriate role.</p>
+		<p>Changes in content are not restricted to text changes. Where an icon or sound indicates a status message, this information will be surfaced by the screen reader through a combination of two things: 1) existing WCAG requirements governing text alternatives (under <a href="non-text-content" class="sc">SC 1.1.1 Non-Text Content</a>), and 2) the requirement of this current Success Criterion to supply an appropriate role.</p>
```
bruce-usab:
```suggestion
		<p>Changes in content are not restricted to text changes. Where an icon or sound indicates a status message, this information will be surfaced by the screen reader through a combination of two things: 1) existing WCAG requirements governing text alternatives (under <a href="non-text-content">SC 1.1.1 Non-Text Content</a>), and 2) the requirement of this current success criterion to supply an appropriate role.</p>
```

understanding/21/status-messages.html
adampage:
```suggestion
			<p>This success criterion was intentionally worded to apply primarily when visible text is added to (or becomes visible on) the page. The reason for this is that where new text is displayed, it is intended to be visible to all users. By providing a programmatic means of ensuring the text is also surfaced through assistive technologies, the success criterion provides the same information to users who cannot or may not see it. However, not all changes to content involve the addition of text to the screen. The following are all considerations relevant to this success criterion:</p>
```

detlevhfischer:
I was under the impression that we always capitalize "Success Criterion" / "Success Criteria" elsewhere,  but I am happy either way - just pointing it out for consideration.

patrickhlauke:
@detlevhfischer 

> I was under the impression that we always capitalize "Success Criterion" / "Success Criteria" elsewhere, but I am happy either way - just pointing it out for consideration.

we had clarification a while ago on what the preferred casing should be https://github.com/w3c/wcag/wiki/WCAG-2-style-guide#s (I don't necessarily agree with it myself, but ... at least it's official)

understanding/21/status-messages.html
```diff
@@ -53,12 +53,12 @@ <h3>Status Message Examples</h3>
 					<li>An application displays a progressbar to indicate the status of an upgrade. The element is assigned a suitable role. The screen reader provides intermittent announcements of the progress. </li>
 					<li>After a user submits a form, text is added to the existing form which reads, "Your form was successfully submitted." The screen reader announces the same message.</li> 
 					<li>After a user unsuccessfully fills in a form because some of the data is in the incorrect format, text is added to the existing form which reads "5 errors on page". The screen reader announces the same message.</li>
-					<li>After a user puts a photo in an album in an online photo app, a <a href="https://material.io/design/components/snackbars.html#">snackbar </a> displays the message "Saved in 'Wedding' album", which is also read by a screen reader.</li>
+					<li>After a user puts a photo in an album in an online photo app, a <a href="https://m3.material.io/components/snackbar/overview">snackbar</a> displays the message "Saved in 'Wedding' album", which is also read by a screen reader.</li>
```
jorycunningham: I wonder about the value, and potential peril, of using "snackbar" as a particular example. My concern is that is could be construed as an implicit endorsement of that pattern, and Material UI as a whole.

At the least it seems like an unnecessary external dependency. How often do we check that this link still resolves, or that the component hasn't evolved in a way that fails 4.1.3 or other SC?

Would the impact and value of this sentence be lessened by just saying:

`<li>After a user puts a photo in an album in an online photo app, a "Saved in 'Wedding' album" appears, which is also read by a screen reader.</li>`

scottaohara: per the above comment / mitigating the need to have to keep on top of other potential URL changes in the future, how about:
```suggestion
					<li>After a user puts a photo in an album in an online photo app, the app dynamically displays a popup message of "Saved in 'Wedding' album", which is also read by a screen reader.</li>
```

patrickhlauke: oh apologies, i already merged this before i realised this was in "sent to AGWG" ... normally i'd keep it "stable" until after the 2 week discussion period. sorry if this causes confusion to folks who reviewed it before and are now wondering why this changed

PR #4261, opened 2025-03-04
Remove links to too old Editor's Draft in Understanding SC 4.1.2

momdo:
Following related resources are too old:

- https://www.w3.org/WAI/PF/roadmap/
- https://www.w3.org/WAI/PF/GUI/

EDIT (PHL): substituted with links to

- https://www.w3.org/TR/wai-aria/
- https://www.w3.org/TR/html-aria/
- https://www.w3.org/WAI/ARIA/apg/


patrickhlauke:
Agree those resources are very outdated. Perhaps worth taking the opportunity to link to some more relevant materials, like

* https://www.w3.org/TR/html-aria/
* https://www.w3.org/TR/accname-1.2/

understanding/20/name-role-value.html
```diff
@@ -70,12 +70,6 @@ <h2>Examples of Name, Role, Value</h2>
...
       <ul>
-         <li>
-            <a href="https://www.w3.org/WAI/PF/roadmap/">Dynamic Accessible Web Content Roadmap</a>
-         </li>
-         <li>
-            <a href="https://www.w3.org/WAI/PF/GUI/">Role Taxonomy for Accessible Adaptable Applications</a>
-         </li>
          <li>
             <a href="https://www.w3.org/TR/wai-aria/">Web Accessibility Initiative - Accessible Rich Internet Applications (ARIA)</a>
          </li>
```
patrickhlauke, line 75:
```suggestion
         </li>
         <li>
             <a href="https://www.w3.org/TR/html-aria/">ARIA in HTML</a>
         </li>
         <li>
             <a href="https://www.w3.org/TR/accname-1.2/">Accessible Name and Description Computation 1.2</a>
         </li>
```

@momdo are you happy to add these?

momdo: @patrickhlauke 
Looks good to me

patrickhlauke: great stuff, thanks :)

mbgower:
TF discussed adding https://www.w3.org/WAI/ARIA/apg/

Also discussed whether to link to 1.1 or 1.2 for https://www.w3.org/TR/accname-1.2/

kfranqueiro:
FWIW I checked with Daniel and he said we're better off linking to 1.2 in informative docs, even though it's a WD.

Separately from this PR, we may want to update other references, since we link to just `TR/accname/` in many places, which still resolves to 1.1.

understanding/20/name-role-value.html
```diff
@@ -71,13 +71,13 @@ <h2>Examples of Name, Role, Value</h2>
...
+            <a href="https://www.w3.org/TR/wai-aria/">Web Accessibility Initiative - Accessible Rich Internet Applications (ARIA)</a>
          </li>
          <li>
-            <a href="https://www.w3.org/WAI/PF/GUI/">Role Taxonomy for Accessible Adaptable Applications</a>
+             <a href="https://www.w3.org/TR/html-aria/">ARIA in HTML</a>
          </li>
          <li>
-            <a href="https://www.w3.org/TR/wai-aria/">Web Accessibility Initiative - Accessible Rich Internet Applications (ARIA)</a>
+             <a href="https://www.w3.org/TR/accname-1.2/">Accessible Name and Description Computation 1.2</a>
          </li>
```
patrickhlauke, line 81:
```suggestion
         </li>
         <li>
             <a href="https://www.w3.org/WAI/ARIA/apg/">ARIA Authoring Practices Guide</a>
         </li>
```

patrickhlauke:
> TF discussed adding https://www.w3.org/WAI/ARIA/apg/

Added link to APG as well, as discussed. Think this is now ready to be sent to WG

patrickhlauke:
@kfranqueiro 

> Separately from this PR, we may want to update other references, since we link to just `TR/accname/` in many places, which still resolves to 1.1.

Just checking: is that because 1.2 is still WD? Could it perhaps be set up that there's a redirect from `/accname/` to always go to the latest?

kfranqueiro:
> Just checking: is that because 1.2 is still WD? Could it perhaps be set up that there's a redirect from `/accname/` to always go to the latest?

I checked with Daniel again and yes, 1.2 still being WD is the main reason, so this may resolve itself when it reaches CR (which they're working towards).

PR #4262, opened 2025-03-04
Alternatives match language

### LINKED ISSUE #3873: 1.2.3 / 1.2.5 Audio description - right language required?
As far as I can see, neither the normative text of 1.2.3 and 1.2.5 nor the definition of "audio description" has anything on the 'right' language for such description (maybe due to an english-centric bias where AD in different languages rarely surfaces). In a current audit of a German site, there are videos with a voice-over comment in English describing what happens.

Looking strictly at the normative text we'd have to pass this provided the AD covers the visuals sufficiently - even if the AD language were Chinese (and we'd be able to tell whether it covers the visuals, which we aren't) - right? 

Since there is no markup for spoken language, 3.1.2 Language of parts would not really help here.

Or am I missing something?

mbgower:
Closes https://github.com/w3c/wcag/issues/3873 by adding a concept of matching language to understanding documents and techniques for 1.1 and 1.2

Also for the 1.2 Time-based Media page, added some previously missing content.


understanding/20/time-based-media.html
```diff
@@ -87,6 +88,12 @@ <h2>Intent of Time-Based Media</h2>
       </p>
       
    </section>
+
+   <section><h3>Text alternatives and equivalents match the language of the content</h3>
+      <p>Text alternatives and equivalents should match the human language of the original content (normally the default human language of the page). <a href="https://www.w3.org/TR/WCAG22/#conformance-reqs">5.2 Conformance Requirements<a>, and the defined terms used in the section, state that success criteria be met through accessibility-supported ways (5.2.4), where the technology is used “in the human language of the content.” Where an alternative version is used (5.2.1), it is defined as something that “provides all of the same information and functionality in the same human language.”</p>
```
mbgower, line 93:
```suggestion
      <p>Text alternatives and equivalents should match the human language of the original content (normally the default human language of the page). The <a href="https://www.w3.org/TR/WCAG22/#conformance-reqs">5.2 Conformance Requirements<a> section, through the defined terms used there, states that success criteria be met through accessibility-supported ways (5.2.4), where the technology is used “in the human language of the content.” Where an alternative version is used (5.2.1), it is defined as something that “provides all of the same information and functionality in the same human language.”</p>
```

understanding/20/non-text-content.html
```diff
@@ -76,6 +76,11 @@ <h3>
          </ul>
          
       </section>
+
+      <section>
+         <h3>Note on alternatives matching the language of content</h3>
+         <p>Text alternatives and equivalents should match the human language of the original content (normally the default human language of the page). The <a href="https://www.w3.org/TR/WCAG22/#conformance-reqs">5.2 Conformance Requirements<a> section, and the defined terms used there, states that success criteria be met through accessibility-supported ways (5.2.4), where the technology is used “in the human language of the content.” Where an alternative version is used (5.2.1), it is defined as something that “provides all of the same information and functionality in the same human language.”</p>
```
mbgower, line 82:
```suggestion
         <p>Text alternatives and equivalents should match the human language of the original content (normally the default human language of the page). The <a href="https://www.w3.org/TR/WCAG22/#conformance-reqs">5.2 Conformance Requirements<a> section, through the defined terms used there, states that success criteria be met through accessibility-supported ways (5.2.4), where the technology is used “in the human language of the content.” Where an alternative version is used (5.2.1), it is defined as something that “provides all of the same information and functionality in the same human language.”</p>
```

understanding/20/time-based-media.html
```diff
@@ -87,6 +88,12 @@ <h2>Intent of Time-Based Media</h2>
       </p>
       
    </section>
+
+   <section><h3>Text alternatives and equivalents match the language of the content</h3>
+      <p>Text alternatives and equivalents should match the human language of the original content (normally the default human language of the page). The <a href="https://www.w3.org/TR/WCAG22/#conformance-reqs">5.2 Conformance Requirements<a> section, through the defined terms used there, states that success criteria be met through accessibility-supported ways (5.2.4), where the technology is used “in the human language of the content.” Where an alternative version is used (5.2.1), it is defined as something that “provides all of the same information and functionality in the same human language.”</p>
+      <p>For time-based media, that normally means that where the spoken language of the media is Spanish, the language used in captions, audio descriptions, and media alternatives will also be Spanish.</p>
```
mbgower, line 94:
```suggestion
      <p>For time-based media, that normally means that where the spoken language of the media is, for example, Spanish, the language used in captions, audio descriptions, and media alternatives will also be Spanish.</p>
```

mbgower:
@patrickhlauke on today's call I was momentarily confused into thinking I'd adjusted a definition by this line
` <a class="glossary">audio description</a>. This is done by providing an alternative for time-based media in the same <a>human language</a> as the video or page on which it appears.`

However, as is likely apparent to you, this is not the glossary definition within the Understanding, but a link in the Understanding doc to the definition, followed by more text. So it works fine as is.

PR #4265, opened 2025-03-06
Typo: Change all instances of dismissable to dismissible

yatil:
In Understanding Content on Hover or Focus

- https://www.w3.org/WAI/WCAG22/Understanding/content-on-hover-or-focus.html
- https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html


patrickhlauke:
Adding some context: this is for consistency with the normative wording of the SC, which uses "dismissible". LGTM

patrickhlauke:
There's also one instance in `wcag.json` ... not sure if we want to touch that too? https://github.com/w3c/wcag/blob/946397c0dc458690390c4199b7226d16a3db2fbd/guidelines/wcag.json#L3460

patrickhlauke (approved):
need to decide what to do with `wcag.json`

yatil:
I saw the wcag.json, but I don’t know if it is actually used anywhere. I also don’t know if it is generated from somewhere else and would auto-update.

yatil:
I now also changed wcag.json

kfranqueiro (changes requested):
This branch seems to have more changes than just fixing "dismissible" - I see e.g. the In Brief section being removed, a note being added, and a `ul` being changed to an `ol`.

https://github.com/w3c/wcag/pull/4265/files?w=1

Perhaps a sync with main had conflicts that were not correctly resolved?

Should I try re-syncing this?

understanding/21/content-on-hover-or-focus.html
```diff
@@ -1,145 +1,141 @@
...
-				<ul>
-					<li>perceive the additional content AND</li>
-					<li>dismiss it without disrupting their page experience.</li>
-				</ul>
+  <body>
+    <h1>Understanding Content on Hover or Focus</h1>
+    <section id="intent">
+      <h2>Intent</h2>
+      <p>Additional content that appears and disappears in coordination with keyboard focus or pointer hover often leads to accessibility issues. Reasons for such issues include:</p>
+      <ol>
```
patrickhlauke, line 16: this change to numbered list looks odd ... not sure what other things may have happened here, but that's the one that jumps out at me /cc @kfranqueiro - maybe worth doing a clean re-do of this PR that literally just finds/replaces the words and nothing else (leaving the reformatting for a separate commit or even leaving it as it currently is)

patrickhlauke (changes requested):
Unless I'm misreading it, this makes a lot more fundamental changes to the understanding docs than just the find/replace of the word, so blocking this for now until we can have a proper look

patrickhlauke:
> Perhaps a sync with main had conflicts that were not correctly resolved?
> 
> Should I try re-syncing this?

Try, but if that doesn't solve it I can make a fresh PR with JUST the word changes

patrickhlauke:
As I was already messing with things, I did go ahead and do a clean redo of this https://github.com/w3c/wcag/pull/4292 (tried re-merging main into this one, but it didn't fix the problem)

PR #4268, opened 2025-03-06
Improve wording of first example for path-based gesture

bruce-usab:
Under Intent, we currently have:

> A **path-based gesture** involves an interaction where not just the endpoints matter, but how the pointer moves between these points.
>  
> If the gesture is only recognised if the user moves in a (mostly) straight line from the start point to the end point, it is an example of a path-based gesture.

That second sentence is confusing.  It is an example of a common _characteristic_ of some path-based gestures.  I propose replacing it with:

> Swiping is an example of a **path-based gesture** which is only recognized when the user moves in a (mostly) straight line from the start point to the end point.


bruce-usab:
FWIW, this came up in the context of WCAG 3 complex pointer inputs drafting.  We were looking for a definition of simple pointer inputs.  We used [single-pointer](https://www.w3.org/TR/WCAG2/#dfn-single-pointer) as part of our draft glossary when we noticed [single pointer input](https://w3c.github.io/wcag3/guidelines/#dfn-single-pointer-input) is currently just a place-holder with the WCAG 3 editors draft.

patrickhlauke (approved):
This looks good to me

patrickhlauke:
> We used [single-pointer](https://www.w3.org/TR/WCAG2/#dfn-single-pointer) as part of our draft glossary when we noticed [single pointer input](https://w3c.github.io/wcag3/guidelines/#dfn-single-pointer-input) is currently just a place-holder with the WCAG 3 editors draft.

incidentally, I'm glad to see that WCAG 3 is defining the proper "single pointer input" rather than just "single pointer" ... this had been a blunder in WCAG 2 at the time (which is what caused the various contortions to address this/clarify this somehow - https://github.com/w3c/wcag/pull/3536 and https://github.com/w3c/wcag/pull/4070)

bruce-usab:
Discussed briefly on Backlog call 3/7.  Agreement that there is some grammar error, but maybe could be improved.

PR #4269, opened 2025-03-07
Working group name

mbgower:
Matches the changes in https://github.com/w3c/wcag/pull/4257 for the normative document.


iadawn:
We wouldn't look to update these two documents.

Issue #4241, opened 2025-02-21
Revisit figcaption styles


kfranqueiro:
Figcaptions are currently displayed in bold text, due to styles defined WAI-website-wide in `/WAI/assets/css/style.css`.

This can perhaps be overwhelming for captions that run particularly long, and also limits options to emphasize content within those captions, as pointed out in https://github.com/w3c/wcag/pull/4055#issuecomment-2674874580.

On the other hand, the existing bold styles do make it easy to distinguish figcaptions from ordinary paragraphs of text within the body.

Some thoughts that came up in discussion on the backlog TF call on 2025-02-21:

- Make only the leading "Figure N" part of the caption bold
  - Currently these are added by the build system to Understanding docs as separate paragraphs before the content, but not Techniques
  - One technique (G211) has them written in-line within the content
  - If we want to consistently settle on the former, we should remove the few inline instances and consistently apply this across Techniques and Understanding
- Consider options to ensure long figcaptions are still distinguishable from subsequent content:
  - Increase contrast of the divider line below the figure
  - Use a leading-side border or margin to group the figure and its caption
  - Add styles to `figure` + `figcaption` to make the caption match the width of the image? (This may be possible with CSS `inline-grid` + figcaption width trickery)
    - We would probably want to set a `min-width`, since some figures are not very wide (e.g. Figure 4 in [Non-text Contrast](https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast#intent))

[WIP exploring options on my fork](https://github.com/w3c/wcag/compare/main...kfranqueiro:wcag:kgf-figcaptions)

Example before and (proposed) after:

![Screenshot of Figure 4 from Understanding Non-text Contrast as it exists today, with fully-bolded figcaption text, full width, and extremely low-contrast bottom border](https://github.com/user-attachments/assets/da6d987d-1967-4bdb-a439-4d838f36f0d9)

![Example of updated styles, including "Figure N" in a preceding paragraph within the figcaption, non-bolded figcaption content, limiting the caption width, and choosing a slightly darker WAI theme color for the bottom border](https://github.com/user-attachments/assets/acef0050-f4bc-410f-b136-86cfe333164f)

alastc:
I liked Patrick's suggestion of having a line/block to the left of both the image and the caption. So long as it isn't confused with blockquotes? If there's a page of WAI components somewhere it would be useful to see how it fits in with other styles.

patrickhlauke:
[...]

PR #4276, opened 2025-03-13
Restyle figures and captions

patrickhlauke:
* Change 11ty/liquid script to wrap the auto-generated "Figure X" in a `<span>`
* Add CSS overrides
* Add vertical line at margin that spans the indent of the image and caption

[preview](https://deploy-preview-4276--wcag2.netlify.app/understanding/non-text-contrast)

Closes https://github.com/w3c/wcag/issues/4241


patrickhlauke:
Good example: https://deploy-preview-4276--wcag2.netlify.app/understanding/non-text-contrast

iadawn:
My preference would be for this to be a bit more aligned with the figure caption styling in TR space. Separating the word 'figure' and the figure number from the caption text seems to be peculiar. I am also not overly keen on adding another border - not sure that this sort of attention is necessary.

I was thinking something more like the following example:

alternative presentation for figure captions

11ty/CustomLiquid.ts
```diff
@@ -243,7 +243,7 @@ export class CustomLiquid extends Liquid {
           $("figcaption").each((i, el) => {
             const $el = $(el);
             if (!$el.find("p").length) $el.wrapInner("<p></p>");
-            $el.prepend(`Figure ${i + 1}`);
+            $el.prepend(`<span>Figure ${i + 1}</span>`);
```
kfranqueiro, line 246: If we want to try @iadawn's suggestion of inlining the "Figure N", it would go something like this, in order to insert it into the first paragraph within the figcaption:

```suggestion
            $el.find("p").first().prepend(`Figure ${i + 1} &mdash; `);
```

(To be clear though, RE @iadawn's comment, "Figure N" was already separated like this prior to this PR)

patrickhlauke:
I personally prefer having an extra indicator like a border, I'll admit ... it visually breaks up long pages with lots of images/captions from being just a see of prose with images dotted around. Happy to go for the "Figure..." inline.

Dropping the border and left padding, and inlining the figure, this current PR would then look like this

one of the figures and figure captions, without border of the left

Personally still prefer it with...

same, inlined 'Figured' text, but with border on left

shawna-slh:
A left border would visually highlight it. I think you want the opposite, for the figure caption to be quiet.

patrickhlauke:
To discuss at today's meeting. Happy to mod/change the PR once we decided.

mbgower:
> A left border would visually highlight it and draw attention to it. I think you want the opposite, for the figure caption to be quiet.

@shawna-slh the problem we're encoutering, particularly for large figcaptions, is that if the styling is "quiet" people read it as if it is the main content, and get confused. And by "people" I mean almost everyone who has been working on it :)

If you look at the editor's draft version of the Reflow document, it makes its own case for this problem, IMO.

So, counterintuitively, a 'louder' figcaption styling makes it is easier to distinguish and skip when one is just reading the main narrative. I think there are many examples of this in the wild (for example, with justified main content, center-aligned figcaptions set in italics at a larger point size).

kfranqueiro:
After further discussion and clarification with Shawn:

- Instead of "Figure N" as separate paragraph, inline "Figure N."
  - Ending with period, not dash; regular weight (_not_ bold)
  - This follows multiple common style guides, e.g. Chicago, MLA
- Remove leading border
- Restore bottom border; make darker
  - e.g. `var(--cloudy)` instead of `var(--line-grey)`, as demonstrated in the screenshots and branch shown in #4241
- Ensure there is more space after the bottom border than before it
  - i.e., increase bottom margin of `figcaption` to e.g. `2em`

Examples from Reflow:

Figure 5, where the figure is immediately followed by another paragraph

Figure 6, where the figure is immediately followed by another image/figure


Figure 13, where the figure caption spans 2 paragraphs and is followed by a heading

patrickhlauke:
See https://deploy-preview-4276--wcag2.netlify.app/understanding/non-text-contrast for the proposed changes

shawna-slh:
Thanks Ken and Patrick, it's helpful to see options on a full page (which I think Ken was avoiding so as not to steps on the toes of this PR :).

Currently the text size is smaller for the figcaption, which I think is good as one differentiation from the main text.

Perhaps along with that, using spacing and proximity would be enough.
- Remove the space between the image and the figcaption
- Leave lots of space after the figcaption

I also wonder what are best practices and potential issues with limiting the width of the caption text?

kfranqueiro:
> I also wonder what are best practices and potential issues with limiting the width of the caption text?

@shawna-slh that idea came up when this topic was first brought up, and I originally experimented with that in https://github.com/w3c/wcag/issues/4241 (there are a couple of image examples in there). It ends up creating a lot of potential for problematic edge cases, e.g.:

- It would make already-long captions run even longer vertically
- We need to be careful not to limit it directly based on image width because there are some very narrow images.

alastc:
Could we not associate the figure and caption better? e.g. 
same figure of a button and caption, but with a blue line on the left.

So just something like:
`border-left: #005a9c 3px solid;
  padding-left: 0.5em;`

Not concerned about the colour (I just picked the blue from the header), but this associates the image with the text much better than the alternatives I've seen, at least to me.

patrickhlauke:
@alastc pushed changes trying your suggestion. i could go for that, yeah

https://deploy-preview-4276--wcag2.netlify.app/understanding/non-text-contrast

mbgower:
@shawna-slh we are moving this to ready for review

patrickhlauke:
There was discussion during our Friday meeting wondering whether or not the border on the left should be darker to achieve `3:1` contrast ratio. While normatively not required (as the indent of the content compared to regular prose also visually indicates the separation, even if you can't perceive the border), I'm happy to make this change here before it's merged. from the predefined colours, maybe `--grey` / `#686868`, but it might be a bit too dark/in your face

screenshot from the last deploy, but with the border made dark / #686868

mbgower:
> from the predefined colours, maybe --grey / #686868, but it might be a bit too dark/in your face

Agreed. I think the weight in the [last preview](https://deploy-preview-4276--wcag2.netlify.app/understanding/non-text-contrast) is sufficient 

image
**Figure showing existing preview**

That contrast is similar to examples in the wild, including the vertical rule right here for in github for quoting a reply. 

Bringing it up to 3:1 overemphasizes the line, which I know Shawn was concerned with. 
image
**Figure showing vertical line at 3:1**

I prefer leaving that level of weight to borders that are actionable or are the sole indicator of a component.

PR #4282, opened 2025-03-17
Update redundant-entry.html

### LINKED ISSUE #4246: Use of “session” in “In Brief” in Understanding SC 3.3.7: Redundant Entry (Level A)

alastc:
In-brief update for redundant-entry, replacing Eric's in #4247.

Closes #4246


bruce-usab (approved):

detlevhfischer:
> in the same activity

Since we are talking processes above, is there any reason not to say "in the same process" or "during the same process"?

mbgower:
@detlevhfischer 

> > in the same activity
> 
> Since we are talking processes above, is there any reason not to say "in the same process" or "during the same process"?


The original PR used "process"; however, there are a number of reasons for using a synonym. Please  see discussion in https://github.com/w3c/wcag/issues/4246

scottaohara (approved):

detlevhfischer:
@mbgower 
OK. I hadn't followed that. I see the rationale while still thinking process would be more straightforward (because I equate "activity" to a single sub-part of a process, such as a single activation). No objection here though.

PR #4292, opened 2025-03-21
Typo: Change all instances of dismissable to dismissible

patrickhlauke:
This is a clean redo of https://github.com/w3c/wcag/pull/4265 (which seems to have some unforeseen additional changes)

### DIFF
--- a/guidelines/wcag.json
+++ b/guidelines/wcag.json
@@ -3457,7 +3457,7 @@
 		"TECH:future-content-on-hover-or-focus-1"
 		,
 		"title": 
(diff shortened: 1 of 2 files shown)


patrickhlauke:
Setting this already to "Ready for approval", as we agreed to the original #4265 (before we spotted that that PR wasn't clean)

mbgower:
@kfranqueiro can we remove the json file from this now?

kfranqueiro:
> @kfranqueiro can we remove the json file from this now?

Thanks for flagging, @mbgower, I've just synced with main and yes that means wcag.json is already removed. The next time we perform a site build, the update made to the Understanding document should propagate to the generated JSON.

PR #4295, opened 2025-03-21
Update audio-control.html

### LINKED ISSUE #4230: 1.4.2 Audio Control - Do two-way voice communications features and/or video conference features need to meet this req?
Curious question (because we are testing a web site that has two-way voice communication functionality AND video conference/meeting functionality):

Must a video meeting function (assume the video meeting function is inside a web page) need to meet: 

1.4.2 Audio Control - If any audio on a Web page plays automatically for more than 3 seconds, either a [mechanism](https://www.w3.org/TR/WCAG21/#dfn-mechanism) is available to pause or stop the audio, or a mechanism is available to control audio volume independently from the overall system volume level.

Is the audio in a web based video or VOIP call considered "automatically playing"?  Or, do we need to add a note to clarify that this requirement does not apply to web based VOIP or web based virtual meetings?

(Then I got curious...to see if browsers might give me a way to turn off the incoming audio.  I'm working from a Mac...and here is what I experienced: 

In one video mtg platform...I just tested to see if there was a way for me to mute the incoming sound.  

- In Chrome, I could do it using browser based feature to mute a tab:  https://www.wikihow.com/Mute-and-Unmute-Google-Chrome-Tabs
- In Safari, I found a similar feature...but when I tried to mute the tab..it didn't actually work when I was INSIDE the web based video mtg (but my ability to mute audio on a tab in safari DID WORK on a site like YouTube when I had a video playing).
)

mbgower:
Closes #4230

Added clarity to the existing language for "direct result of user's intentional activation" and added text specific to a conference call.

### DIFF
--- a/understanding/20/audio-control.html
+++ b/understanding/20/audio-control.html
@@ -41,7 +41,8 @@
          </p>
       </div>
       <p>See also <a href="low-or-no-background-audio">1.4.7 Low or No Background Audio</a>.</p>
-      <p>In the context of this Success Criterion, "plays automatically" broadly refers to audio that is not started/played as a direct result of a user's intentional activation. For example, not selecting a link or button.</p>
+      <p>In the context of this success criterion, "plays automatically" broadly refers to audio that is not started/played as a direct result of a user's intentional activation. For example, selecting a link or button with clear labelling or context that it will start an experience where audio will play is an example of starting in response to a user's intended action. This criterion is also not intended to apply to a conference call or other interaction where two-way voice communication may take place; the potential for any participant to speak during a conference call is not equivalent to audio that "plays automatically for more than 3 seconds." However, a mechanism to control the volume of conference call output independently from the overall system volume would be a best practice. </p>
+      <p></p>
    </section>
    <section id="benefits">
       <h2>Benefits of Audio Control</h2>


scottaohara (approved):

patrickhlauke (approved):

understanding/20/audio-control.html
```diff
@@ -41,7 +41,7 @@ <h2>Intent of Audio Control</h2>
          </p>
       </div>
       <p>See also <a href="low-or-no-background-audio">1.4.7 Low or No Background Audio</a>.</p>
-      <p>In the context of this Success Criterion, "plays automatically" broadly refers to audio that is not started/played as a direct result of a user's intentional activation. For example, not selecting a link or button.</p>
+      <p>In the context of this success criterion, "plays automatically" broadly refers to audio that is not started/played as a direct result of a user's intentional activation. For example, selecting a link or button with clear labelling or context that it will start an experience where audio will play is an example of starting in response to a user's intended action. This criterion is also not intended to apply to a conference call or other interaction where two-way voice communication may take place; the potential for any participant to speak during a conference call is not equivalent to audio that "plays automatically for more than 3 seconds." However, a mechanism to control the volume of conference call output independently from the overall system volume would be a best practice. </p>
```
patrickhlauke, line 44:
```suggestion
      <p>In the context of this success criterion, "plays automatically" broadly refers to audio that is not started/played as a direct result of a user's intentional activation. For example, selecting a link or button with clear labelling or context that it will start an experience where audio will play is an example of starting in response to a user's intended action. This criterion is also not intended to apply to a conference call or other interaction where two-way voice communication may take place; the potential for any participant to speak during a conference call is not equivalent to audio that "plays automatically for more than 3 seconds." However, a mechanism to control the volume of conference call output independently from the overall system volume would be a best practice.</p>
```

PR #4301, opened 2025-03-26
Re-implement wcag.json output

kfranqueiro:
This repo contains a `wcag.json` file under the `guidelines` folder, generated by `xml-to-json.xslt`. However, the XSLT is out of date with the formats that some of its consumers use, especially [Quickref](https://www.w3.org/WAI/WCAG22/quickref/), and it has not been fully regenerated for a while (easily over a year).

This PR:

- Re-implements `wcag.json` generation using TypeScript (reusing code already involved in the Eleventy build process where possible)
  - Always generates based on published guidelines (not ED); also relies on built Understanding docs (i.e. it needs to run after Eleventy)
  - Can generate for both 2.1 and 2.2 (since multiple consumers rely on parts of both)
- Adds a `WCAG_JSON` environment variable recognized by the Eleventy build process to also build `wcag.json` at the end
- Updates the `w3c-publish` scripts to include the new environment variable
- Outputs `wcag.json` under `_site` (the build output folder), since it is an output of the build process; there may be some discussion as to the precise behavior of this
- Updates behavior of some fields in the JSON to be easier to work with by their respective consumers, e.g.:
  - Any fields containing HTML have links modified to point to absolute URLs
  - `techniques` is replaced by `techniquesHtml`, as Quickref's version of the former had been manually modified to the point where it would be unrealistic to automate, while the latter can be output as-is
- Removes old XML build targets and XSLT code that this supersedes
- Removes wcag.json from the guidelines folder, as it is now part of build output intended to be published

**Note:** https://github.com/w3c/wcag/pull/4286 also needs to be merged to prevent some data loss when regenerating `wcag.json` (particularly for Quickref)

I have respective branches for each known repository that consumes `wcag.json`, to handle any changes to the format, and add a script that can pull down updates. However, those will be dependent on publishing `wcag.json`, so will have to wait until after this is merged and deployed. In order to test those, I have temporarily pushed JSON for WCAG 2.1 and 2.2 to my fork's gh-pages:

- [2.1](https://kfranqueiro.github.io/wcag/21/wcag.json)
- [2.2](https://kfranqueiro.github.io/wcag/22/wcag.json)


iadawn (approved):
Looks good, just one comment on the README

PR #4055, opened 2024-09-05
Updated Reflow understanding doc

scottaohara:
This initial commit takes the current draft from the google doc that had been worked on for quite some time now, and makes it into a PR for further editing and review.

Not all feedback from the google doc was directly taken/addressed, but there have been additional changes made that attempted to consider a good portion of the unresolved comments.  

Marking this PR as a draft, as there is still work to do (and I also need to upload all the new graphics for the examples - and a few examples still need to be created, which are currently marked as comments in the HTML file).

- [x] identify all the reflow related issues this PR will close or address in some form (see: [this comment](https://github.com/w3c/wcag/pull/4055#issuecomment-2349139912)) - there are others, but this lists at least references all the ones that were considered while working on this PR.
- [x] add all the new graphics to this PR / render properly (further revisions can be made if people would like different/more consistent examples - but as of now at least all examples for this pr have been created)
- [x] follow on issue for [creating more examples/techniques](https://github.com/w3c/wcag/issues/4236)

[The latest updates will be visible via this preview link of the understanding doc](https://deploy-preview-4055--wcag2.netlify.app/understanding/reflow)

Closes #3121

PR #4304, opened 2025-03-27
Migrate Reflow videos to WAI Understanding WCAG YouTube account

kfranqueiro:
This is a proposal; I'm opening this PR to generate a preview to allow for further evaluation/discussion.

This replaces the mp4 files that were merged in #4055 with YouTube video embeds, using unlisted videos posted to a WAI account. It is coded using a pattern that avoids loading until direct interaction. The pattern is centralized in an include file, intended to make instances as easy as possible to adapt from YouTube's iframe embed code snippets.

The goal of this is to establish a pattern to avoid adding binary files of significant size to the repository, which would ultimately have the effect of irreversibly growing the repository size, impacting everyone who needs to work with it locally.

I tried this approach first, since it is feasible for the WAI team to manage without additional dependencies. This particular approach does have downsides:

- YouTube is not accessible in all countries
- YouTube blocks embed playback from VPNs (at least when not signed in)
- JavaScript is required for playback
- The header and control overlays initially mask the top and bottom of the video for a few seconds (there used to be parameters to minimize this but they have been removed)
- Autoplay seems to require `&mute=1` in webkit browsers, even with `allow="autoplay"` included; this is fine for these videos which have no audio, but worth noting for future. (This PR avoids loading the embeds on page load, so autoplay is only taking effect after an initial user interaction; without it, playing would require 2 clicks)

@netlify /understanding/reflow


TestPartners:
I hope you're not planning to use the YouTube embedded media player. The accessibility is very poor and there's nothing you can do about it. Able Player is a much better choice and it can play videos hosted on YouTube, although a few of the accessibility issues remain. The best solution is Able Player with the videos hosted somewhere else.

kfranqueiro:
@TestPartners thanks for the additional feedback. Fortunately, after more conversations with the team, we should have a more direct option available that won't require reliance on YouTube embeds. I'm going to close this.

PR #4323, opened 2025-04-07
Update audio-description.html

mbgower:
Change to informative content (a note) in normative content (definition):

Added the word "important" to note 3 to match the wording of the definition of [audio description](https://www.w3.org/TR/WCAG22/#dfn-audio-descriptions) (which reads "..to describe important visual details...").


bruce-usab:
All of the information >> than important information.  I prefer the important qualifier. 

But what is “video information”?  Is it the same as information conveyed by the video?

I think maybe the sentence could be:

Where all of the important information is already provided in existing <a>audio</a>, no additional audio description is necessary.

bruce-usab (approved):
Important qualifier is important and good.  But please consider deleting “video”.

> Where all of the important information is already provided in existing audio, no additional audio description is necessary.

patrickhlauke:
I'd actually suggest changing the "video information" (which is existing wording, not something this PR adds) to "visual information", which is what this is getting at, I think

guidelines/terms/20/audio-description.html
```diff
@@ -13,7 +13,7 @@
       (See also <a>extended audio description</a>.)
    </p>
    
-   <p class="note">Where all of the <a>video</a> information is already provided in existing <a>audio</a>, no additional audio description is necessary.
+   <p class="note">Where all of the important <a>video</a> information is already provided in existing <a>audio</a>, no additional audio description is necessary.
```
patrickhlauke, line 16:
```suggestion
   <p class="note">Where all of the important visual information is already provided in existing <a>audio</a>, no additional audio description is necessary.
```

bruce-usab (approved):
Discussed on backlog call 4/15.  As Patrick notes, “video information” features in Note 3 of [AD definition](https://www.w3.org/TR/WCAG22/#dfn-audio-descriptions) so that must be okay!

guidelines/terms/20/audio-description.html
mbgower: As discussed on TF call, my inclination is to leave the video>visual change out for now, as tempting as it is. This is just in the interest of tweaking as little as possible. I believe @patrickhlauke will be opening a new issue to consider making this change a bit more globally, and when that's done, we'll mark this suggestion as resolved for now

patrickhlauke:
I had more thorough look, and unfortunately I think we may have to stick with `<a>video</a>` here ... as that is what the normative SCs and various other terms use - confusingly, to refer to the visuals in synchronised media in general (where i'd argue in more common parlance when somebody talks about "video" they refer to the whole thing, a video file like an mp4, which has visuals and audio), and we'd need to start fundamentally changing quite a few normative bits.

wondering if instead we need some sort of note somewhere (maybe even just added to the term definition for "video" https://www.w3.org/TR/WCAG22/#dfn-video) that explains something along the lines of "when we talk about video, we don't necessarily mean the technology/container format ... a CSS/SVG/Canvas animation that provides visual information can be classed as 'video' - the term refers to any visual aspect of synchronised media")

mbgower:
Thanks for having a look, @patrickhlauke. 

There are so many things it would be nice to clean up in Time-based media. At one point, some of us talked about releasing an update to WCAG 2 that attempted to address a bunch of the normative language. If that were ever to happen, there is no shortage of material here!

PR #4324, opened 2025-04-07
Audio description Understanding covers "important" visual content

mbgower:
Two changes to the first note in the informative Understanding document:

- add "additional" to match the first change to the definition note.
- swap out the word "provided" in preference for the synonym "conveyed"

This change is made in the belief that is more accurate. For example, if a scene shows something blowing up, and there is the sound of an explosion that matches, it is more accurate to say the sound conveys the visual explosion than to say the sound "provides" the visual explosion.

### DIFF
--- a/guidelines/terms/20/audio-description.html
+++ b/guidelines/terms/20/audio-description.html
@@ -13,7 +13,7 @@
       (See also <a>extended audio description</a>.)
    </p>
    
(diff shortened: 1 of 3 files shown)


bruce-usab (approved):

bruce-usab:
Discussed on backlog call 4/11.

hidde-logius (approved):

Issue #1768, opened 2021-04-24
How to add audio descriptions to videos?


guyhickling:
There has been a long discussion on the WebAIM forum today on this subject. Most of us came to the conclusion that compliance with SC1.2.5 requires that audio description be provided for all video content that is not adequately explained in the audio track, even if there are not sufficient spaces in the dialog to use for that purpose.

The SC quite clearly states "Audio description is provided for all pre-recorded video content ...". It specifically says "all", without any caveats. That is the normative requirement.

Unfortunately the Understanding document then tries to alter that by claiming that this audio provision is to be done "During existing pauses in dialogue....". The informative document effectively tries to restrict the application of the normative SC, in all cases where videos have large amounts of unexplained visual content with no space in the audio for it.

Not only does that contradict or soften what the SC itself says, but it also blandly ignores the fact that about half of all videos ever made have hardly any spaces in the audio for adding extra. It also ignores that many videos - particularly educational and technical ones - often have detailed diagrams or illustrations that require very long spaces to insert a description. It has also led to widespread confusion in the web dev world.

Whether website and video owners choose to comply with the SC for already existing videos, or instead claim undue burden, is up to them. But the SC itself should be clear, without having an  Understanding doc appearing to say something different. And it should be a basic requirement of the WCAG at Level AA that all video content be fully accessible to blind people in the audio form without having huge chunks of many videos missing or unexplained or not described to them. That should never have been relegated to AAA in the first place; it's such a basic right!

So to solve this contradiction (not to mention to help ensure that blind people have full access to all video content), could I suggest that a green Note be added to the Understanding doc for 1.2.5. It need not contradict what is already said in the Understanding document. I recommend a wording something like:

> Where there is not sufficient space in the existing audio track to insert a particular needed item of audio description, space should be created to allow the insertion. This can be done by freezing the currently displayed video frame for the duration of the audio insertion, or by inserting additional video frames if preferred.

NB: This will, of course, make the AAA SC1.2.7 almost redundant since it says a similar thing. But we should just accept that. Better a redundancy at AAA level (which hardly any website follows anyway), rather than retain this confusing contradiction of 1.2.5.

patrickhlauke:
The quick fix to understanding here is, I think, to remove "During existing pauses in dialogue", so the sentence there starts with "Audio description provides information about actions..."

[...]

PR #4325, opened 2025-04-07
Removed 1.2.7 reference from note 1

mbgower:
This change aligns the guidance in note 1 of the Understanding documents with the SC text for [1.2.7](https://www.w3.org/WAI/WCAG22/Understanding/extended-audio-description-prerecorded.html).

The preamble condition for 1.2.7 reads "Where pauses in foreground audio are insufficient to allow audio descriptions to convey the send of the video...". Based on that preamble, it makes no sense for note 1 to reference 1.2.7, since if all important information is "already conveyed in the audio track", such a video will never meet the preamble condition.

This logic is confirmed by the fact that note 1 does not exist in the 1.2.7 Understanding document.

### DIFF
--- a/understanding/20/audio-description-or-media-alternative-prerecorded.html
+++ b/understanding/20/audio-description-or-media-alternative-prerecorded.html
@@ -57,7 +57,7 @@
       <div class="note">
          
          <p>
-            For 1.2.3, 1.2.5, and 1.2.7, if all of the important information in the video track is already
+            For 1.2.3 and 1.2.5, if all of the important information in the video track is already
             conveyed in the audio track, no additional audio description is necessary. 
             
             
--- a/understanding/20/audio-description-prerecorded.html
+++ b/understanding/20/audio-description-prerecorded.html
@@ -34,7 +34,7 @@
       <div class="note">
          
          <p>
-            For 1.2.3, 1.2.5, and 1.2.7, if all of the important information in the video track is already
+            For 1.2.3 and 1.2.5, if all of the important information in the video track is already
             conveyed in the audio track, no additional audio description is necessary. 
             
             


dbjorge:
> The preamble condition for 1.2.7 reads "Where pauses in foreground audio are insufficient to allow audio descriptions to convey the send of the video...". Based on that preamble, it makes no sense for note 1 to reference 1.2.7, since if all important information is "already conveyed in the audio track", such a video will never meet the preamble condition.

I'm not understanding why "such a video will never meet the preamble condition" would cause the inclusion in the note to "make no sense" - I think it just acts as confirmation that the note *does* make sense in the context of 1.2.7. My preference would be to continue to include it.

> This logic is confirmed by the fact that note 1 does not exist in the 1.2.7 Understanding document.

I disagree; I think this just means the 1.2.7 understanding doc should get an update to add the note, too.

mbgower:
> I'm not understanding why "such a video will never meet the preamble condition" would cause the inclusion in the note to "make no sense" - I think it just acts as confirmation that the note _does_ make sense in the context of 1.2.7. My preference would be to continue to include it.

We can't continue to include something that isn't there; the 1.2.7 Understanding doc, and with good reason.

> > This logic is confirmed by the fact that note 1 does not exist in the 1.2.7 Understanding document.
> 
> I disagree; I think this just means the 1.2.7 understanding doc should get an update to add the note, too.

Can you please elaborate, @dbjorge?
1.2.7 _only_ exists to add extended audio descriptions "where pauses in foreground audio are insufficient to allow audio descriptions to convey the sense of the video".

In other words, in order to even be relevant to 1.2.7, there _must_ be information in the audio track not covered by audio. So in what was can the note apply?

JAWS-test:
> So in what way can a note about "all of the information in the video track is already provided" be relevant?

As I understand it, the video track is the visual part of the video. The audio track is the acoustic part. Then it is self-explanatory for me that I do not need an additional audio description for 1.2.3, 1.2.5, and 1.2.7 if the audio track sufficiently reproduces the visual content. The only difference between 1.2.5 and 1.2.7 is that I have the excuse of missing pauses in the audio track with 1.2.5, but not with 1.2.7. For me, it is clear that the note in the Understanding should be added to 1.2.7:

> For 1.2.3, 1.2.5, and 1.2.7, if all of the information in the video track is already provided in the audio track, no audio description is necessary.

mbgower:
> The only difference between 1.2.5 and 1.2.7

The primary difference between 1.2.5 and 1.2.7 is that 1.2.7 has a "where" preamble which qualifies when it can apply. And that preamble, crucially, disqualifies it from ever needing that first note. 1.2.5 and 1.2.3 lack such a note, and so need an explanation that a sufficiently descriptive audio track can meet the requirement _without_ using any of the listed techniques.

dbjorge:
@mbgower "disqualifies it from ever needing that first note" isn't the point. *None* of the understanding docs *need* the note; if the note were adding new information, it'd be normative and not a note in an understanding. My point isn't that I think 1.2.7 *needs the note to be accurate*, my point is that I think the note is just as true for 1.2.7 as it is for 1.2.3 and 1.2.5. I think it's a helpful bit of clarification for all 3 of the cases.

JAWS-test:
If we are precise, we can remove the note in 1.2.3, 1.2.5 and 1.2.7, as it is already present in the “Key Term” section as note 3 in the definition of Audio Description: https://www.w3.org/WAI/WCAG21/Understanding/audio-description-prerecorded#dfn-audio-description

mbgower:
> @mbgower "disqualifies it from ever needing that first note" isn't the point. None of the understanding docs need the note; if the note were adding new information, it'd be normative and not a note in an understanding.

I strongly disagree with this in regard to the 1.2 criteria. I understand the principle, obviously, but the reality is that time-based media SCs _rely_ on notes. As @JAWS-test writes, this first note is reinforcing the third note of the definition, which reads:
 
> Note 3: Where all of the [video](https://www.w3.org/TR/2008/REC-WCAG20-20081211/#videodef) information is already provided in existing [audio](https://www.w3.org/TR/2008/REC-WCAG20-20081211/#audiodef), no additional audio description is necessary.

I would prefer to _remove_ note 1 from the 1.2.3 and 1.2.5 Understanding documents and rely on people to apply Note 3 in the definition, than I would to add the note into 1.2.7.

But I think we need to do better. The reason it is so important to work towards something more clear and consistent is due to the _very_ inconsistent, and not always great normative language in the SCs and in the definition of "audio description", which without the notes are wide-open to various interpretations.

The 1.2 SCs are the least implemented/supported of any A/AA criteria. Time-based media has become a much larger part of web content. As technology makes it more feasible for the 1.2 SCs to be implemented, more scrutiny is occurring, and confusion amongst even experienced a11y people is rampant.

 If you have not done so, I encourage you to review the series of PRs I've created, in the order I indicated in my email, (which I also cite in [one of the many](https://github.com/w3c/wcag/issues/1768#issuecomment-2784911795) video issues that talk about the confusion of interpretations of 1.2 criteria).

mbgower:
@JAWS-test 
> If we are precise, we can remove the note in 1.2.3, 1.2.5 and 1.2.7, as it is already present in the “Key Term” section as note 3 in the definition of Audio Description

The note does not exist in the 1.2.7 Understanding.
I'll re-emphasize that the 1.2.7 _normative text_ contains a qualifying preamble that renders note 1 _entirely_ superfluous.

However, 1.2.3 and 1.2.5 do not have such qualifying preambles. For those who do not go and look at the normative definitions and chase all the nuances, I think retaining an Understanding doc note 1 is valuable for 1.2.3 and 1.2.5. All this PR does is remove "1.2.7" from that note to make it consistent with reality.

If you look through the other PRs I've created, you'll notice that I am doing small incremental changes to align the notes between the Understanding docs and definitions and normative material. My intent doing it this way was to be able to consider each change in isolation, but I'm beginning to wonder if I've created a different problem doing so.

bruce-usab (approved):
Good catch!

bruce-usab:
Discussed on backlog call 4/11.  We didn't get consensus for the change.

mbgower:
Based on a lack of indicated support, dropping this PR.

PR #4335, opened 2025-04-11
Some editorial updates to Understanding SC 2.5.3

momdo:
- Update HTML Standard link and quotation
- Update and add Understanding success criterion links
- Remove mismatching parenthesis


understanding/21/label-in-name.html
```diff
@@ -37,13 +37,13 @@ <h3>Identifying label text for components</h3>
...
+			<p>Note that placeholder text within an input field is not considered an appropriate means of providing a label. The <a href="https://html.spec.whatwg.org/multipage/input.html#the-placeholder-attribute">HTML Standard</a> states <q>The placeholder attribute should not be used as an alternative to a <code>label</code>.</q> However, it is worth noting that "label" in that HTML5 statement is in code brackets and links to the <code>label</code> element. For the purposes of this Label in Name Success Criterion, "label" is not used in such a programmatic sense but is simply referring to a text string in close visual proximity to a component. As such, in the absence of any other nearby text string (as described in the preceding list), if an input contains placeholder text, such text may be a candidate for Label in Name. This is supported both through the accessible name calculation (discussed later) and from the practical sense that where a visible label is not otherwise provided, it is likely that a speech-input user may attempt to use the placeholder text value as a means of interacting with the input.</p>
```
kfranqueiro, line 40: Maybe also re-word "in that HTML5 statement" since the link no longer contains "HTML5". Maybe "in the previous statement"?

mbgower, line 40:
```suggestion
			<p>Note that placeholder text within an input field is not considered an appropriate means of providing a label. The <a href="https://html.spec.whatwg.org/multipage/input.html#the-placeholder-attribute">HTML Standard</a> states <q>The placeholder attribute should not be used as an alternative to a <code>label</code>.</q> However, it is worth noting that "label" in that statement is in code brackets and links to the <code>label</code> element. For the purposes of this Label in Name Success Criterion, "label" is not used in such a programmatic sense but is simply referring to a text string in close visual proximity to a component. As such, in the absence of any other nearby text string (as described in the preceding list), if an input contains placeholder text, such text may be a candidate for Label in Name. This is supported both through the accessible name calculation (discussed later) and from the practical sense that where a visible label is not otherwise provided, it is likely that a speech-input user may attempt to use the placeholder text value as a means of interacting with the input.</p>
```

momdo: Thanks for suggestion

patrickhlauke (approved):

patrickhlauke:
Preview here https://deploy-preview-4335--wcag2.netlify.app/understanding/label-in-name

All looks good to me, thanks @momdo

scottaohara (approved):

Issue #4346, opened 2025-04-17
Error in Success Criterion 1.3.4 Orientation Note


ryukochan:
I believe the Note to [Success Criterion 1.3.4 Orientation](https://www.w3.org/TR/WCAG21/#orientation) is made incorrect by inclusion of the word 'not.' 
Due to the essential orientation of the provided examples, the note should read:

`Examples where a particular display orientation may be essential are a bank check, a piano application, slides for a projector or television, or virtual reality content where content is necessarily restricted to landscape or portrait display orientation.`

![Image](https://github.com/user-attachments/assets/ec4f8388-9bef-491a-afa2-0d77f4bc607b)

The word 'not' was apparently inserted via an erratum in PR #734 in May of 2019.

patrickhlauke:
agree the "not" there is not correct. not sure #734 is to blame. it seems https://github.com/w3c/wcag/commit/f9a2821c2f54bf07e9146b2f30936dd4f3c58989 included a rewrite/simplification of original wording 

```
where binary display orientation is not applicable
```

into simpler language, and in the process the meaning was accidentally flipped on its head

```
where content is not necessarily restricted to landscape or portrait display orientation
```

/cc @alastc

patrickhlauke:
Filed https://github.com/w3c/wcag/pull/4348

PR #4348, opened 2025-04-18
Remove incorrect "not" from 1.3.4 Orientation note

patrickhlauke:
The "not" was introduced as part of https://github.com/w3c/wcag/commit/f9a2821c2f54bf07e9146b2f30936dd4f3c58989 which rewrote the example from an early draft into simpler language, and in doing so accidentally flipped the meaning from `where binary display orientation is not applicable` to `where content is NOT necessarily restricted to landscape or portrait orientation`

Closes https://github.com/w3c/wcag/issues/4346

### DIFF
--- a/guidelines/sc/21/orientation.html
+++ b/guidelines/sc/21/orientation.html
@@ -6,6 +6,6 @@
    					
   <p>Content does not restrict its view and operation to a single display orientation, such as portrait or landscape, unless a specific display orientation is <a>essential</a>.</p>
    				
-   <p class="note">Examples where a particular display orientation may be essential are a bank check, a piano application, slides for a projector or television, or virtual reality content where content is not necessarily restricted to landscape or portrait display orientation.</p>
+   <p class="note">Examples where a particular display orientation may be essential are a bank check, a piano application, slides for a projector or television, or virtual reality content where content is necessarily restricted to landscape or portrait display orientation.</p>
    					
 </section>


giacomo-petri (approved):

mbgower:
TF discussion shows this is confusing. Rather than simply remove "not", we need more explanation of the behaviour in a virtual reality context.

patrickhlauke:
oh, was the binary display part only related to virtual reality? I originally read that last part as being related to all examples.

patrickhlauke:
i'd actually say let me close this PR, as I clearly misunderstood that the `where binary display orientation is not applicable` is only related to the VR case, and do a new PR. wondering if the intention for the VR case was that the idea is content on the VR screens potentially always remains in a particular orientation (like, say, a HUD) relative to the headset itself, regardless of the orientation of the headset (e.g. your life bar etc remain fixed relative to the headset/your eyes, so you always see it in the same orientation, even though the VR world itself changes orientation in response to your head movement)

giacomo-petri:
> oh, was the binary display part only related to virtual reality? I originally read that last part as being related to all examples.

I read it the same way initially...

giacomo-petri:
I recommend in the new PR to distinguish between what's essential and what's not applicable/not binary.

patrickhlauke:
Superseded by https://github.com/w3c/wcag/pull/4349

PR #4351, opened 2025-04-21
Add link to NIST SP 800-63B 2nd public draft to 3.3.8 / 3.3.9 understanding

patrickhlauke:
Incorporate outside references about allowing pasting into password fields.

This change offsets stances about preventing pasting passwords.  Even NIST plans to recommend allowing it:

> 3.1.1.2 Password verifiers
> 
> [...]
> 
> Verifiers SHALL allow the use of password managers. **Verifiers SHOULD permit claimants to use the “paste” functionality when entering a password to facilitate their use**. Password managers have been shown to increase the likelihood that users will choose stronger passwords, particularly if the password managers include password generators

> 8.2.1 Passwords
> 
> [...]
> 
> Usability considerations for typical usage without a password manager include:
> 
> [...]
> * User experience during entry of the password
>   * **Support copy and paste functionality in fields for entering passwords, including passphrases.**

> 8.2.3 Out-of-band
> 
> [...]
> 
> Usability considerations for typical usage include:
> 
> [...]
> 
> **Consider offering features that do not require text entry on mobile devices (e.g., a copy-paste feature)**, which are particularly helpful when the primary and secondary channels are on the same device. For example, it is difficult for users to transfer the authentication secret manually using a smartphone because they must switch back and forth — potentially multiple times — between the out-of-band application and the primary channel.


fstrr:
Should we link to a revision of the draft of the NIST standard now or is it worth waiting until the final standard is published?

patrickhlauke:
my gut feeling is: looking at the current state of US politics, who knows if this ever actually gets published as a stable document... so I'd rather push it now and update at a later point

Issue #4214, opened 2025-01-31
Consider clarifying what is meant by "where user input is required" in 3.3.2.


mfairchild365:
Please consider updating the Understanding document for 3.3.2 to clarify what is meant by "where user input is required."

I frequently encounter confusion from designers, developers, and testers about this wording. Many incorrectly think that it means that only required fields need to have labels - and that optional fields or fields with default values do NOT require labels because these fields do not REQUIRE input to proceed.

Perhaps adding this line to understanding document would help clarify this:

> This Success Criteria applies to all fields, including optional fields and those that have default values.

patrickhlauke:
x-ref https://github.com/w3c/wcag/issues/2357#issuecomment-1118538464

Issue #2357, opened 2022-05-04
Clarification on 3.3.2: Labels or Instructions with regard to required fields and Sufficient Techniques


joe-watkins:
Hi folks! Can the Working Group comment on the following scenario with regard to [3.3.2: Labels or Instructions](https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html) and required fields in a web form?

**Scenario**: Imagine a web form where fields that own descriptive labels are required for entry but they are not marked visually nor programmatically as required. Upon submission of the form, error validation kicks in and text appears indicating the field/s is required. The form does not have any instructional indicating that there are required fields.

It's my understanding that to satisfy `3.3.2: Labels or Instructions` you can use [G131 Providing descriptive labels](https://www.w3.org/WAI/WCAG21/Techniques/general/G131) **AND** [G83 Providing text descriptions to identify required fields that were not completed](https://www.w3.org/WAI/WCAG21/Techniques/general/G83).

Is my interpretation correct that WCAG does not require authors to indicate that fields are required as part of the label or accessible description as long as the label is descriptive (G131) **and** error validation using text descriptions is present upon form submission (G83)? 

ps. I know this isn't good UX or advised - just strictly parsing WCAG with respect to satisfying the Success Criterion.

patrickhlauke:
some related discussion here https://github.com/w3c/wcag/issues/1698

bruce-usab:
Hello Joe Watkins.

> Is my interpretation correct that WCAG does not require authors to indicate that fields are required as part of the label or accessible description...

That is correct, since there is no SC which says that.  In general, poor UX which effects everyone (and not only PWD), won't be reflected in the normative phrasing of Success Criteria.  It is also the case that the AG WG is reluctant to includes examples in Understanding (et al.) that are just-barely-passing-but-terrible.

JAWS-test:
> Is my interpretation correct that WCAG does not require authors to indicate that fields are required as part of the label or accessible description as long as the label is descriptive (G131) and error validation using text descriptions is present upon form submission (G83)?

My understanding of WCAG is that this is not correct. Required fields must be marked (according to 3.3.2). Regardless, labels must be meaningful.

patrickhlauke:
@JAWS-test it *is* contextual to an extent. If you have a login form where of course both username and password are required, do you fail a page if it doesn't explicitly say "Username (Required)" / "Password (Required)"?

bruce-usab:
[...]

PR #4367, opened 2025-04-27
Add note to 3.3.2 understanding to explain "require" in normative wording

patrickhlauke:
Closes https://github.com/w3c/wcag/issues/4214

See also https://github.com/w3c/wcag/issues/2357#issuecomment-1118538464


scottaohara (approved):

scottaohara:
This LGTM.  There was a question on the call today if this note was sufficient in just the understanding doc, or if it was a note that should be added to the spec as well.

I could go either way on that.  I personally think that the spec text is already clear that this is beyond just form controls in the required state - as if that was the intent it would have said that - rather than "when **content** requires user input" - which I read as "when any type of interactive element needs to be engaged with by a user"

bruce-usab (approved):

PR #4368, opened 2025-04-28
Change "alternate" to "alternative"

patrickhlauke:
Excludes "conforming alternate version" as it's a deeply embedded term


patrickhlauke:
Just to see what this would look like, did a fairly naive find/replace, *but* kept "conforming alternate version" as is (as there was some concern about changing *that*)

bruce-usab:
Discussed briefly on backlog call 5/16.  Agreement that alternative better alternate, but concern that there might still be too many replacements with conforming _alternate_ version being the canonical term.

mbgower:
TF discussed and although this is technically correct, we believe that since the spec has existed with this wording for 17 years, it can continue without negative effect.

PR #4371, opened 2025-04-28
Update synchronized-media.html

mbgower:
Adds a note to the synchronized media definition clarifying that the definition applies to any form of content that is intended to be experienced as part of an experience that combines more than one time-based medium.

### DIFF
--- a/guidelines/terms/20/synchronized-media.html
+++ b/guidelines/terms/20/synchronized-media.html
@@ -5,4 +5,5 @@
       interactive components, unless the media is a <a>media alternative for text</a> that is clearly labeled as such
    </p>
    
+   <p class="note">"Synchronized" refers to the fact the audio and video are intended to be played according to the same time base. It is not intended to imply that images and videos that are out of synch with each other (intentionally or unintentionally) in the same experience are excluded by the definition. Asynchronous sound, for example, is still being experienced in the context of time-based, synchronized media.</p>
 </dd>
--- a/guidelines/terms/20/synchronized-media.html
+++ b/guidelines/terms/20/synchronized-media.html
@@ -5,5 +5,5 @@
       interactive components, unless the media is a <a>media alternative for text</a> that is clearly labeled as such
    </p>
    
-   <p class="note">"Synchronized" refers to the fact the audio and video are intended to be played according to the same time base. It is not intended to imply that images and videos that are out of synch with each other (intentionally or unintentionally) in the same experience are excluded by the definition. Asynchronous sound, for example, is still being experienced in the context of time-based, synchronized media.</p>
+   <p class="note">"Synchronized" refers to all audio and video content intended to be played according to the same time base (i.e., in the same video file or multimedia experience). Images and sounds that are out of synch with each other (intentionally or unintentionally) are included by the definition. Asynchronous sound, for example, is still being experienced in the context of time-based, synchronized media.</p>
 </dd>


guidelines/terms/20/synchronized-media.html
```diff
@@ -5,4 +5,5 @@
       interactive components, unless the media is a <a>media alternative for text</a> that is clearly labeled as such
    </p>
    
+   <p class="note">"Synchronized" refers to all audio and video content intended to be played according to the same time base (i.e., in the same video file or multimedia experience). Images and sounds that are out of synch with each other (intentionally or unintentionally) are included by the definition. Asynchronous sound, for example, is still being experienced in the context of time-based, synchronized media.</p>
```
patrickhlauke, line 8: "according to the same time base" sounds overly convoluted/impenetrable. perhaps something simpler, like "at the same time", "in parallel", or similar?

I think we've been trying to avoid "e.g.", "i.e.", and so on for simpler language. maybe replace here with "that is" or "for instance"?

maybe expand a bit what's mean by "Asynchronous sound", again using simpler language? Maybe a bit wordy, but something like "For instance, even if a video only includes 'incidental' music that does not explicitly synchronise with any of the visuals, ..."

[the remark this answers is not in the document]
TestPartners: That works for me.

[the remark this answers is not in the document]
bruce-usab: Can "played in the same time base" be "played at the same time"?

guidelines/terms/20/synchronized-media.html
```diff
@@ -5,4 +5,5 @@
       interactive components, unless the media is a <a>media alternative for text</a> that is clearly labeled as such
    </p>
    
+   <p class="note">"Synchronized" refers to all audio and video content presented in the same multimedia experience (whether combined in a single file or played in the same time base). Images and sounds that are out of synch with each other (intentionally or unintentionally) are included by the definition. Asynchronous sound, for example, is still being experienced in the context of time-based, synchronized media.</p>
```
mbgower, line 8:
```suggestion
   <p class="note">"Synchronized" refers to all audio and video content presented in the same multimedia experience (whether combined in a single file or played at the same time). Images and sounds that are out of synch with each other (intentionally or unintentionally) are included by the definition. Asynchronous sound, for example, is still being experienced in the context of time-based, synchronized media.</p>
```

alastc, line 8:
```suggestion
   <p class="note">“"Synchronized" refers to all audio and video content presented in the same multimedia experience (whether combined in a single file or played in the same time base). As well as multimedia that is obviously syncronized (for example, seeing people speaking), a presentation with a background track would also considered sycronized. If the multimedia is not "presenting information", it would not require audio description.</p>
```

patrickhlauke: related https://github.com/w3c/wcag/issues/4398

bruce-usab:
My suggestion on call:
> Some multimedia experiences may have artifacts where audio and video are not played exactly at the same time.

We also discussed adding examples.

kfranqueiro:
RE the alternative of adding an example, @dbjorge suggested the following on the May 16 call:

Example: A website presents a slideshow with some background music. If the background music is presenting information, then it is synchronized media. If the background music is synchronized with the slideshow using a time-based interactive component - for example, a pause/play button that affects both the slideshow and the music - then it is also synchronized media. Otherwise, it is not synchronized media.

mbgower:
>  If the background music is presenting information, then it is synchronized media...  Otherwise, it is not synchronized media.

I don't understand what that first sentence means. The _value_ of the background music is irrelevant. The fact there is an audio track that is timed with the video is what makes it possible to add audio descriptions. That is the only consideration we have in regard to our definition of "synchronized" in the context of audio-video.

@alastc I similarly don't understand the additions you have incorporated into your suggestion 'Both the visual and audio need to be "presenting information".' Do you mean that audio isn't just dead air?

I recommend making the first sentence: "If background music is part of the video, it is synchronized media."

> If the background music is synchronized with the slideshow using a time-based interactive component - for example, a pause/play button that affects both the slideshow and the music - then it is also synchronized media.

I also find we are getting into the weeds discussing time-based interactive components. If someone wants to create a _separate_ example for an experience that accepts timed user actions, go ahead. But the specific issue here is confusion of what means "synchronized" in regard to audio-video.

All we seem to be doing here is more comprehensively confusing things.

---

I had a discussion this weekend with someone who does weekly AV broadcasts. They confirmed my prior opinion that anyone involved in this field will immediately grasp a concept of a "time base" in regard to synchronized media. A _lot_ of their work is spent time-correcting cameras, monitor playback, etc. 

I quickly found an [existing definition](https://www.sweetwater.com/insync/time-base/#:~:text=Literally%20means%20timing%20reference.,to%20which%20things%20are%20synchronizing.) that shows this term is in use:

> Time base
Literally means timing reference. For example, on a typical watch the time base is in seconds. In music/video/film production it pertains to [synchronization](https://www.sweetwater.com/insync/synchronization/) and has roughly the same meaning, though the units are often frames. The time base is the timing reference to which things are synchronizing.

Given the confusion with our current definition is arising from a technical point, I think we have to incorporate some consideration of a time base into our update, without needing to burrow into technical concepts.

guidelines/terms/20/synchronized-media.html
mbgower: "this also hinges of whether or not there's actual important content/information being conveyed" How is importance relevant? Importance matters in what gets audio described, but seem irrelevant to a discussion on audio-video synchronization.

mbgower:
> The "presenting information" terminology I used here is a verbatim quote

IMO, the phrase "presents information" in the definition you quote is intended generically. It implies no qualitative measure; however, the way it's used in the suggested example for synchronized media that Ken captured, it implies quality.

It suggests there exists some kind of music that contains no information, and that somehow we're supposed to be able to tell what that is.

guidelines/terms/20/synchronized-media.html
mbgower: Muzak is still information. There is _no_ normative language here about "important". It's just 'information'. I think folks are cross-circuiting the 'important visual' concept in audio descriptions with the notion of synchronized, which is just an amalgam of two time-based media.

TestPartners:
I don't agree that muzak is information. That's like saying the background colour of a page is information. The muzak that accompanies videos could usually be replaced by any other tune or none at all, just as the background colour of a page can be any colour. The muzak and background colour clearly do not convey any information if they can be replaced without changing the meaning of the content. That is not the case when music is specifically chosen to convey a feeling such as danger or suspense and different music would convey a different feeling. I don't think it's difficult to tell the difference.

dbjorge:
> IMO, the phrase "presenting information" in the definition you quote is intended generically. It implies no qualitative measure

I agree that the wording is more ambiguous than I'd like for normative language, but I don't think it's reasonable to pick a meaning that assumes those words were included for no reason.

> It suggests there exists some kind of music that contains no information, and that somehow we're supposed to be able to tell what that is.

This is no different from other parts of WCAG, where we expect evaluators to make exactly this determination to, for example, decide whether an image counts as "pure decoration".

mbgower:
> This is no different from other parts of WCAG, where we expect evaluators to make exactly this determination to, for example, decide whether an image counts as "pure decoration".

What I don't understand about this position is that it makes synchronized media a qualitative assessment. I see zero grounds for that.

### Mechanically it makes no sense
For the purposes of audio descriptions, all we care about is that the audio and video are combined and synchronized, which 1) allows there to _be_ an audio track (i.e., not video-only), and 2) allows someone to align the timing of the audio descriptions in the audio track with the appearance of the relevant visuals.

### Audio descriptions are about visuals
Audio descriptions are not about providing an equivalent for sound, so what do we care what the nature of the existing audio is? The only way the soundtrack even enters into considerations for audio descriptions is in regard to identifying "pauses". We also have a potential consideration of a proposed sufficient technique (not yet approved) to use audio ducking. But those are both orthogonal to the synchronized media definition.

### Focus on audio-video
Finally, the reason I've been pushing back about attempts to incorporate a time-based interaction factor into this discussion is that it has no direct association with audio descriptions. Time-based interactions apply to a certain kind of synchronized media, but it's a consideration for which I've never seen any issues. And if there _were_ any issues flagged, I cannot see how audio quality would fit into the synchronized media consideration either.

Perhaps the solution here is to put in a note in the audio descriptions definition that specifically talks about synchronized media in this context, instead of putting it in the synchronized media definition?

mbgower:
For example:

Note 5:
In the context of audio descriptions, "synchronized" refers to all audio and video content aligned to the same time base, which allows an author to align the timing of the audio descriptions in the audio track with the appearance of the relevant visuals in the video.  Images and sounds that are out of synch with each other (intentionally or unintentionally) are included by the definition. Asynchronous sound, for example, is still being experienced in the context of time-based, synchronized media.</p>

mbgower:
It might help to look at the definition of “synchronized media” with its definitions spelled out:
 
>Synchronized media:
the technology of sound reproduction or the technology of moving or sequenced pictures or images synchronized with another format for presenting information and/or with time-based interactive components
 
You’ll note all that’s being talked about is technologies.
 
I feel like folks are trying to use the flimsy, undefined word “information” here for a qualitative argument, instead of focusing on the much more obvious idea that accessibility solutions of certain kinds only occur when audio and video are presented together.
 
> It is incidental to the video information.
The argument has been advanced that because muzak is meaningless, a video with audio track wholly composed of muzak is not synchronized media.  But that would mean no one is required to provide audio descriptions for the important visual information that is occurring at the same time. Is that really what people want?
 
Or, in another case, because the only thing being shown visually is a meaningless black screen, I do not need to be providing captions of the audio playing over top of the black screen.
 
Why would we even want that as a defence?
 
Think of it this way: because there is synchronized media, we can have a caption (on video), or an audio description (on audio).  What is in the av production and what is in either of these alternatives is immaterial to the technical ability to achieve the alternatives.
 
The author decides (and is judged by) what is a sufficient alternative at the SC level, but we must have the technical reality to even supply the alternative.
 
Mike

bruce-usab (approved):
Discussed on backlog call 5/30.

guidelines/terms/20/synchronized-media.html
mbgower: I'm going to make this an additional note to Audio Description, as discussed, leaving the Synchronization definition alone for now

bruce-usab:
Discussed on backlog call 6/6.

mbgower:
@TestPartners 
There is disagreement on the task force about whether this updated PR addresses your [question](https://github.com/w3c/wcag/pull/1790#issuecomment-2820697325), and adds clarity. Could you please review and comment? BTW, this is now proposed as a note for Audio Descriptions, not for the Synchronized Media definition.

guidelines/terms/20/audio-description.html
```diff
@@ -18,4 +18,5 @@
    
    <p class="note">Also called "video description" and "descriptive narration."</p>
    
+   <p class="note">In the context of audio descriptions, <a>synchronized media</a> primarily refers to all audio and video content intended to be played to the same time base, which allows an author to align the timing of the audio descriptions in the audio track with the appearance of the relevant visuals in the video. Intentionally asynchronous sound can take place in the context of time-based, synchronized media. A video that is accompanied only by background music is still "synchronized media".</p>
```
mbgower: My attempt to remove points of confusion and terminology from the update
```suggestion
   <p class="note">In the context of audio descriptions, <a>synchronized media</a> predominantly describes audio and video technologies that are combined and played in tandem, which allows an author to align the timing of the audio descriptions in the audio track with the appearance of the relevant visuals in the video. In pre-recorded content, there can be intentionally asynchronous sound or images; however, they will always occur at the same time on each playback of synchronized media.</p>
```

mbgower:
@TestPartners if you can take a moment to review the updated draft, I'd like to get this back in front of the task force.

awkawk:
Seems like a fine note for the understanding document.

mbgower:
@mbgower to follow up with Nigel on his advice/feedback on this wording and concept.

mbgower:
@nigelmegitt Can you please review this and give some input on this proposed change?

awkawk:
To clarify my comment, this is a good example of something that will be non-normative as a note on the definition so would get published far faster as an addition to the Understanding document (for 1.2.3/1.2.5) rather than as a change to the 2.x spec.

mbgower:
as per @awkawk's suggestion, I am going to take the latest version of this text and alter it from being a proposed note in a definition to an additional note in the Understanding document. It reads:
`<p class="note">In the context of audio descriptions, <a>synchronized media</a> predominantly describes audio and video technologies that are combined and played in tandem, which allows an author to align the timing of the audio descriptions in the audio track with the appearance of the relevant visuals in the video. In pre-recorded content, there can be intentionally asynchronous sound or images; however, they will always occur at the same time on each playback of synchronized media.</p>`

nigelmegitt:
> ### Audio descriptions are about visuals
> 
> Audio descriptions are not about providing an equivalent for sound, so what do we care what the nature of the existing audio is? The only way the soundtrack even enters into considerations for audio descriptions is in regard to identifying "pauses". We also have a potential consideration of a proposed sufficient technique (not yet approved) to use audio ducking. But those are both orthogonal to the synchronized media definition.

Just want to note an edge case: where the video contains a burned in translation subtitle (i.e. translating dialogue in the audio), the audio description should include those subtitles being read out - in that case the audio description is providing a _translated_ equivalent to speech in the main soundtrack.

Issue #4185, opened 2024-12-24
SC 2.1.2 Tab looping


Que-tin:
There is a problem with the following example that leads to major differences between the browser and open source implementations for the dialogs:

> When the dialog has been opened, focus is trapped within the dialog; tabbing from the last control in the dialog takes focus to the first control in the dialog. The dialog is dismissed by activating the Cancel button or the OK button.

imo there is too much room for interpretation and clearer guidance is needed that defines a "Best Practice" for this specific scenario. Otherwise it leads to further inconsitency between implementations.

The native <dialog> doesn't support tab looping at all while most open source libraries are on the other extrema and don't provide an option to disable it.

With clearer guidance one or the other would need to be adjusted to fulfil the Level A requirements.

fstrr:
It looks like this relates to [No Keyboard Trap](https://www.w3.org/WAI/WCAG22/Understanding/no-keyboard-trap.html).

You say "The native `<dialog>` doesn't support tab looping at all", which isn't the case—see this [quick demo](https://cdpn.io/pen/debug/xbKLpZm/456e62345818b7780d3096b2c86a53a0) where, when the `dialog` is open, it's not possible to navigate to the elements on the underlying page. It does allow the user to navigate to the browser's "chrome", which is fine, and then back into the open modal dialog.

It looks like [the example](https://www.w3.org/WAI/WCAG22/Understanding/no-keyboard-trap.html#:~:text=the%20applet%20itself.-,A%20modal%20dialog%20box,-A%20Web%20application) hasn't been updated since the native `dialog` element became fully supported. It could be updated to clarify that both trapping inside a modal dialog and allowing the user to access the browser's chrome is fine, as long as there's an accessible method to close the modal dialog. Does that work for you?

tay1orjones:
This was been discussed at length in https://github.com/whatwg/html/issues/8339, particularly https://github.com/whatwg/html/issues/8339#issuecomment-1822591131 where @matatk @niklasegger mentioned facilitating updates to APG but it appears that maybe no one followed up on this.

> It could be updated to clarify that both trapping inside a modal dialog and allowing the user to access the browser's chrome is fine, as long as there's an accessible method to close the modal dialog. Does that work for you?

@fstrr Yes. Also the [2.4.3 focus order understanding doc](https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html) needs updated as well. I'd encourage the title of this issue to be updated to include that and mention of `<dialog>`. 

[...]

PR #4381, opened 2025-05-05
Tweak language around modal dialogs

patrickhlauke:
Closes https://github.com/w3c/wcag/issues/4185

Note that this only softens/expands some of the example language in the understanding documents. The original issue #4185 seems to conflate non-normative examples in a non-normative document with requirements of how modal dialogs should be implemented ... they are not. They provide examples of appropriate focus order and absence of keyboard traps. They are not the only ways in which focus order or absence of keyboard traps can be achieved. Examples are non-exhaustive, and there is no need - in my view - to expand these further to specifically address variations that result from different behaviours related to `<dialog>`

### DIFF
--- a/understanding/20/focus-order.html
+++ b/understanding/20/focus-order.html
@@ -134,7 +134,8 @@
          
          <li>A web page implements modal dialogs via scripting. When the trigger button is activated,
             a dialog opens and focus is set within the dialog. As
-            long as the dialog is open, focus is limited to the elements of the dialog. When the
+            long as the dialog is open, focus is generally limited to the elements of the dialog (though,
+            depending on the implementation, focus may also cycle through the user agent controls). When the
             dialog is dismissed, focus returns to the button or the element following the button.
          </li>
          
--- a/understanding/20/no-keyboard-trap.html
+++ b/understanding/20/no-keyboard-trap.html
@@ -64,7 +64,7 @@
                prior to the applet as well as within the applet itself.</dd>
          <dt>A modal dialog box</dt>
          <dd>A web application brings up a dialog box. At the bottom of the dialog are two buttons,
-               Cancel and OK. When the dialog has been opened, focus is trapped within the dialog;
+               Cancel and OK. When the dialog has been opened, focus is generally trapped within the dialog:
                tabbing from the last control in the dialog takes focus to the first control in the
                dialog. The dialog is dismissed by activating the Cancel button or the OK button.</dd>
       </dl>


scottaohara (approved):

mbgower:
I'm a bit concerned about these changes. Pretty consistently in our examples, we are listing best behaviour/practices. As you note here (and in the issue), someone misconstruing an example with a normative edict shouldn't necessarily lead us to made the examples less clear or confusing (which I personally find these to be).
I'd prefer to leave as is, or lessen the change.

giacomo-petri (approved):

understanding/20/focus-order.html
```diff
@@ -134,7 +134,8 @@ <h2>Examples of Focus Order</h2>
          
          <li>A web page implements modal dialogs via scripting. When the trigger button is activated,
```
adampage:
```suggestion
         <li>A web page implements modal dialogs. When the trigger button is activated,
```
Now that we’ve got the declarative [`command`](https://html.spec.whatwg.org/multipage/form-elements.html#attr-button-command) / [`commandfor`](https://html.spec.whatwg.org/multipage/form-elements.html#attr-button-commandfor) approach, “_via scripting_” is no longer a given?

understanding/20/focus-order.html
```diff
@@ -134,7 +134,8 @@ <h2>Examples of Focus Order</h2>
...
             a dialog opens and focus is set within the dialog. As
-            long as the dialog is open, focus is limited to the elements of the dialog. When the
+            long as the dialog is open, focus is generally limited to the elements of the dialog (though,
+            depending on the implementation, focus may also cycle through the user agent controls). When the
```
adampage:
```suggestion
            long as the dialog is open, all web page content outside the dialog becomes 
            inert and cannot receive focus. When the
```

How about sidestepping user agent controls and instead scoping to “_all web page content outside the dialog_” and then emphasizing the inertness of that content rather than trapping the user inside the dialog?

understanding/20/no-keyboard-trap.html
```diff
@@ -64,7 +64,7 @@ <h2>Examples of No Keyboard Trap</h2>
                prior to the applet as well as within the applet itself.</dd>
          <dt>A modal dialog box</dt>
          <dd>A web application brings up a dialog box. At the bottom of the dialog are two buttons,
-               Cancel and OK. When the dialog has been opened, focus is trapped within the dialog;
+               Cancel and OK. When the dialog has been opened, focus is generally trapped within the dialog:
                tabbing from the last control in the dialog takes focus to the first control in the
                dialog. The dialog is dismissed by activating the Cancel button or the OK button.</dd>
```
adampage:
```suggestion
               Cancel and OK. When the dialog has been opened, all web page content outside the dialog becomes 
               inert and cannot receive focus.
               The dialog is dismissed by activating the Cancel button or the OK button.</dd>
```

adampage: Ditto.

adampage (commented):
Heya, @patrickhlauke / @mbgower, how about these tweaks?

understanding/20/focus-order.html
patrickhlauke: i'd avoid linking directly to the html spec, as we generally try to be technology-agnostic (unless it's in a technology-specific technique, or it's given as a specific example for a particular technology). long story short, i'd keep "inert" but without the link around it.

understanding/20/no-keyboard-trap.html
patrickhlauke: same comment about tech agnostic/removing link

understanding/20/focus-order.html
adampage: Edited, thank you!

understanding/20/no-keyboard-trap.html
adampage: Ditto ditto.

understanding/20/focus-order.html
patrickhlauke: thanks @adampage ... looking good

understanding/20/no-keyboard-trap.html
patrickhlauke: ditto^3

patrickhlauke:
@fstrr should we split the difference and maybe add a parenthetical somehow that reintroduces this? a la 

```
all web page content outside the dialog becomes inert and cannot receive focus
(though, depending on implementation, the focus cycle might still include user agent controls)
```

understanding/20/focus-order.html
```diff
@@ -132,9 +132,10 @@ <h2>Examples of Focus Order</h2>
             element.
          </li>
          
-         <li>A web page implements modal dialogs via scripting. When the trigger button is activated,
+         <li>A web page implements modal dialogs. When the trigger button is activated,
```
fstrr, line 135: If you change this to remove "via scripting", you should likely remove the same phrase from [the previous example on the page](https://www.w3.org/WAI/WCAG22/Understanding/focus-order.html#:~:text=A%20web%20page%20implements%20modeless%20dialogs%20via%20scripting).

patrickhlauke: good catch, done

understanding/20/focus-order.html
```diff
@@ -124,17 +124,18 @@ <h2>Examples of Focus Order</h2>
...
             button. When the dialog is closed, the focus order goes from the button to the following
             element.
          </li>
          
-         <li>A web page implements modal dialogs via scripting. When the trigger button is activated,
+         <li>A web page implements modal dialogs. When the trigger button is activated,
             a dialog opens and focus is set within the dialog. As
-            long as the dialog is open, focus is limited to the elements of the dialog. When the
+            long as the dialog is open, all web page content outside the dialog becomes 
+            inert and cannot receive focus. When the
```
patrickhlauke:
```suggestion
            inert and cannot receive focus (though, depending on implementation, the focus cycle
            might still include user agent controls). When the
```

understanding/20/no-keyboard-trap.html
```diff
@@ -64,9 +64,9 @@ <h2>Examples of No Keyboard Trap</h2>
                prior to the applet as well as within the applet itself.</dd>
          <dt>A modal dialog box</dt>
          <dd>A web application brings up a dialog box. At the bottom of the dialog are two buttons,
-               Cancel and OK. When the dialog has been opened, focus is trapped within the dialog;
-               tabbing from the last control in the dialog takes focus to the first control in the
-               dialog. The dialog is dismissed by activating the Cancel button or the OK button.</dd>
+               Cancel and OK. When the dialog has been opened, all web page content outside the dialog becomes 
+               inert and cannot receive focus.
```
patrickhlauke:
```suggestion
               inert and cannot receive focus (though, depending on implementation, the focus cycle
               might still include user agent controls).
```

patrickhlauke:
> @fstrr should we split the difference and maybe add a parenthetical somehow that reintroduces this? a la
> 
> ```
> all web page content outside the dialog becomes inert and cannot receive focus
> (though, depending on implementation, the focus cycle might still include user agent controls)
> ```

as discussed in today's WCAG 2.x backlog call, I've made that change @fstrr

fstrr (approved):

understanding/20/focus-order.html
```diff
@@ -124,17 +124,19 @@ <h2>Examples of Focus Order</h2>
...
             button to the elements of the dialog, then to the interactive element following the
             button. When the dialog is closed, the focus order goes from the button to the following
             element.
          </li>
          
-         <li>A web page implements modal dialogs via scripting. When the trigger button is activated,
+         <li>A web page implements modal dialogs. When the trigger button is activated,
             a dialog opens and focus is set within the dialog. As
-            long as the dialog is open, focus is limited to the elements of the dialog. When the
+            long as the dialog is open, all web page content outside the dialog becomes
```
iadawn, line 137: Does this section say something slightly different to the [APG modal dialogs pattern keyboard interaction](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/#keyboardinteraction) section?

My personal view is that the APG needs to be changed to avoid focus being locked into the modal.

understanding/20/no-keyboard-trap.html
```diff
@@ -64,9 +64,10 @@ <h2>Examples of No Keyboard Trap</h2>
                prior to the applet as well as within the applet itself.</dd>
          <dt>A modal dialog box</dt>
          <dd>A web application brings up a dialog box. At the bottom of the dialog are two buttons,
-               Cancel and OK. When the dialog has been opened, focus is trapped within the dialog;
-               tabbing from the last control in the dialog takes focus to the first control in the
-               dialog. The dialog is dismissed by activating the Cancel button or the OK button.</dd>
+               Cancel and OK. When the dialog has been opened, all web page content outside the dialog becomes 
+               inert and cannot receive focus (though, depending on implementation, the focus cycle
```
iadawn, line 68: Same issue as above

understanding/20/focus-order.html
patrickhlauke: APG is only a best practice suggestion, one way to implement a modal, and generally what the keyboard behaviour would be. it's not binding, and the revised wording here takes account of the fact that there are other valid implementations - like a `<dialog>` element - where browser UI will also still be within the focus order.

otherwise, this may suggest that native `<dialog>` fails WCAG

adampage: I’ve heard anecdotally that one of the reasons the WHATWG chose _not_ to trap focus in the `<dialog>` element is because that behavior has been used maliciously to extract sensitive information from vulnerable users.

Because of that, I’ve been tempted to submit a PR to APG updating its ARIA dialog implementation to follow suit. There’s an open (but 3 years stale) discussion about this here: w3c/aria-practices#1772. I’ve looked around for a paper trail on that WHATWG decision but haven’t been able to find anything.

mbgower:
@adampage please take this up with the APG https://github.com/w3c/wcag/pull/4381#discussion_r2246753367

PR #4382, opened 2025-05-05
Add security mention to Timing Adjustable and tweak Re-Authentication

### LINKED ISSUE #1424: Is a time-based token a security concern (timing adjustable)
Hi

Can you please clarify what is meant by "security concerns" in the context of [timing adjustable](https://www.w3.org/TR/WCAG21/#timing-adjustable)? More specifically, is there a difference between security concerns set by the content and a 2-step authentication that relies on time-based tokens?

The "Notes regarding server time limits" section in [the understanding doc for 2.2.1](https://www.w3.org/WAI/WCAG21/Understanding/timing-adjustable.html) says:

> This Success Criterion applies only to time limits that are set by the content itself. For example, if a time limit is included to address **security concerns**, it would be considered to have been set by the content because it is designed to be part of the presentation and interaction experience for that content.

To illustrate where I get confused imagine a scenario where after the login page, the user needs to authenticate again by providing a code that is sent to the user's mobile phone. A token is generated, and it's valid only for a few minutes. If the user doesn't enter the correct code within the time limit, then the next click on any UI component redirects the user to the login page, and the process starts again (the token expires). There is no chance to either turn off, adjust or extend the session. 

As I currently understand it, the understanding doc for timing adjustable seems to say that it is a failure because the time-based token is automatically generated, and it was implemented due to security concerns. However, there is an argument to be made that it is essential because extending it would invalidate the activity.

It doesn't help that examples of this issue do not seem to be consistent. In the [understanding doc for re-authentication](https://www.w3.org/WAI/WCAG21/Understanding/re-authenticating), the example of a questionnaire does not seem to meet any of the requirements of timing adjustable and it's referring to re-authenticating. It makes me believe that there is a difference between security concerns and authentication.

Is not being able to extend, adjust, or turn off the time limit of the time-based token used for 2-step authentication failure of the timing adjustable criterion? 

Clarity on this issue would be much appreciated!

Many thanks
aron

patrickhlauke:
* adds mention of security exceptions (under "essential") for timing adjustable
* tweaks examples in re-authentication (they appear to violate timing adjustable at the moment) and cross-reference to timing adjustable

closes #1424

### DIFF
--- a/understanding/20/re-authenticating.html
+++ b/understanding/20/re-authenticating.html
@@ -36,7 +36,12 @@
          they be given the ability to re-authenticate and continue with the transaction without
          the loss of any data already entered.
       </p>
+      <p class="note">Sites that implement session time limits and re-authentication requests are
+         still subject to the requirements of other criteria, such as
+         <a href="timing-adjustable">2.2.1 Timing Adjustable</a>.
+      </p>
    </section>
+
    <section id="benefits">
       <h2>Benefits of Re-authenticating</h2>
 
--- a/understanding/20/re-authenticating.html
+++ b/understanding/20/re-authenticating.html
@@ -59,10 +59,11 @@
 
       <dl>
          <dt>A shopping site checkout</dt>
-         <dd>A user with extremely limited use of the hands is logged into a shopping site. It
-               takes so long to enter credit card information into the application that a time limit
-               occurs while the user is performing the checkout process. When the user returns to
-               the checkout process and submits the form, the site returns a login screen to re-authenticate.
+         <dd>A user is logged into a shopping site. While in the middle of the checkout process, the user
+               is interrupted and has to leave their computer. While they are away, the site prompts the user
+               that the process is about to time out, and offers the ability to extend the timeout – but with
+               the user away, the timeout is not extended and the user is logged out. When the user returns to
+               the computer, they have to re-authenticate.
                After the user logs in, the check out process is restored with the same information
                and at the same stage. The user did not lose any data because the server had temporarily
                accepted and stored the submission even though the session had timed out and restored
@@ -74,7 +75,7 @@
                remains intact and, after re-authentication, the user may send that data.</dd>
          <dt>A questionnaire with a time limit</dt>
          <dd>A long questionnaire provided within a single web page has information at the beginning
-               that indicates that the session will time out after 15 minutes. The user is also informed
+               that indicates that the session will time out after 20 hours. The user is also informed
                that the questionnaire can be saved at any point and completed at a later time. Within
                the web page there are several buttons provided to save the partially completed form.
                In addition, with JavaScript in the list of accessibility-supported content technologies
--- a/understanding/20/re-authenticating.html
+++ b/understanding/20/re-authenticating.html
@@ -1,73 +1,62 @@
 <!DOCTYPE html>
 <html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml">
 <head>
-   <meta charset="UTF-8"></meta>
+   <meta charset="UTF-8"/>
    <title>Understanding Re-authenticating</title>
    <link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove"/>
 </head>
 <body>
    <h1>Understanding Re-authenticating</h1>
-   
+
    <section id="brief">
       <h2>In brief</h2>
       <dl>
          <dt>Goal</dt><dd>Users do not lose information or context due to reauthentication.</dd>
          <dt>What to do</dt><dd>Preserve users' prior activity and data through reauthentication.</dd>
-         <dt>Why it's important</dt><dd>Some people may require additional time to complete an activity.</dd> 
+         <dt>Why it's important</dt><dd>Some people may require additional time to complete an activity.</dd>
       </dl>
 
    </section>
-   
+
    <section id="intent">
       <h2>Intent of Re-authenticating</h2>
-      
-      
+
       <p>The intent of this success criterion is to allow all users to complete authenticated
          transactions that have inactivity time limits or other circumstances that would cause
          a user to be logged out while in the midst of completing the transaction.
       </p>
-      
       <p>For security reasons, many sites implement an authentication time limit after a certain
          period of inactivity. These time limits may cause problems for persons with disabilities
-         because it may take longer for them to complete the activity. 
+         because it may take longer for them to complete the activity.
       </p>
-      
       <p>Other sites will log a person out of a session if a person logs in on the website
          from another computer or if other activities arise that make the site suspicious of
          whether the person is still the same legitimate person who logged in originally. When
          users are logged out while still in the midst of a transaction - it is important that
          they be given the ability to re-authenticate and continue with the transaction without
          the loss of any data already entered.
-         
       </p>
-      
-      
    </section>
    <section id="benefits">
       <h2>Benefits of Re-authenticating</h2>
-      
-      
+
       <ul>
-         
          <li>This success criterion benefits people who may require additional time to complete
             an activity. People with cognitive limitations may read slowly and require additional
             time to read and respond to a questionnaire. Users interacting via a screen reader
-            may need extra time to navigate and complete a complicated form. 
+            may need extra time to navigate and complete a complicated form.
             A person with motor impairments or who navigates with an alternative input device
             may require additional time to navigate through or complete input within a form.
          </li>
-         
          <li>In circumstances where a sign-language interpreter may be relating audio content to
             a user who is deaf, control over time limits is also important.
          </li>
-         
       </ul>
-      
    </section>
-   
+
    <section id="examples">
       <h2>Examples of Re-authenticating</h2>
-      
+
       <dl>
          <dt>A shopping site checkout</dt>
          <dd>A user with extremely limited use of the hands is logged into a shopping site. It
@@ -92,87 +81,54 @@
                that are relied upon, the user can elect to be alerted via a pop-up if the session
                is close to timing out.</dd>
       </dl>
-      
    </section>
-   
+
    <section id="resources">
       <h2>Resources for Re-authenticating</h2>
-      
-      
    </section>
-   
+
    <section id="techniques">
       <h2>Techniques for Re-authenticating</h2>
-      
-      
+
       <section id="sufficient">
          <h3>Sufficient Techniques for Re-authenticating</h3>
-         
-         
+
          <ul>
-            
             <li>
-               
-               <p>
-                  Providing options to continue without loss of data using one of the following techniques:
-               </p>
-               
+               <p>Providing options to continue without loss of data using one of the following techniques:</p>
                <ul>
-                  
                   <li>
-                     											             
                      <a href="../Techniques/general/G105" class="general">Saving data so that it can be used after a user re-authenticates</a>
-                     										           
                   </li>
-                  
                   <li>
-                     											             
                      <a href="../Techniques/general/G181" class="general">Encoding user data as hidden data in re-authorization page</a>
-                     										           
                   </li>
-                  
                </ul>
-               
             </li>
-            
          </ul>
-         
+
          <div class="note">
-            
-            <p>Refer to 
+            <p>Refer to
                <a href="timing-adjustable#techniques" class="understanding">Techniques for Addressing Success Criterion 2.2.1</a> for techniques related to providing notifications about time limits.
-               
             </p>
-            
          </div>
-         
       </section>
-      
+
       <section id="advisory">
          <h3>Additional Techniques (Advisory) for Re-authenticating</h3>
-         
-         
       </section>
-      
+
       <section id="failure">
          <h3>Failures for Re-authenticating</h3>
-         
-         
+
          <ul>
-            
             <li>
-               									         
                <a href="../Techniques/failures/F12" class="failure">Failure due to having a session time limit without a mechanism for saving user's input
                   and re-establishing that information upon re-authentication
                </a>
-               								       
             </li>
-            
          </ul>
-         
       </section>
-      
    </section>
-   
 </body>
 </html>
--- a/understanding/20/timing-adjustable.html
+++ b/understanding/20/timing-adjustable.html
@@ -69,6 +69,12 @@
             <a href="interruptions" class="guideline">2.2.4 Interruptions</a> and
             <a href="re-authenticating" class="guideline">2.2.5 Re-Authentication</a> may also apply.
          </li>
+         <li>Certain time limits implemented for security reasons, such as time-based / time-limited
+            two-factor authentication tokens, can be considered <em>essential</em>, and are
+            exempt from this criterion. However, other criteria may apply, such as
+            <a href="redundant-entry">3.3.7 Redundant Entry</a>,
+            <a href="accessible-authentication-minimum">3.3.8 Accessible Authentication (Minimum)</a>, and
+            <a href="accessible-authentication-enhanced">3.3.9 Accessible Authentication (Enhanced)</a> may also apply.</li>
          <li>Ten times the default was chosen based on clinical experience and other guidelines.
             For example, if 15 seconds is allowed for a user to respond and hit a switch, 150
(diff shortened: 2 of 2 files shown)


fstrr (approved):

understanding/20/timing-adjustable.html
```diff
@@ -34,59 +33,55 @@ <h2>Intent of Timing Adjustable</h2>
...
-            <a href="re-authenticating" class="guideline">2.2.5</a> may also apply.)
+            and are covered. Success Criteria
+            <a href="no-timing" class="guideline">2.2.3 No Timing</a>,
+            <a href="interruptions" class="guideline">2.2.4 Interruptions</a> and
+            <a href="re-authenticating" class="guideline">2.2.5 Re-Authentication</a> may also apply.
          </li>
-         
+         <li>Certain time limits implemented for security reasons, such as time-based / time-limited
+            two-factor authentication tokens, can be considered <em>essential</em>, and are
+            exempt from this criterion. However, other criteria may apply, such as
```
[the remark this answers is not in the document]
patrickhlauke: > as long as the timeout was not chosen arbitrarily. For example, a timeout of 30 seconds for Time-based One-Time passcodes is common and a minimum timeout for other time-limited two-factor authentication tokens or actions of 5-10 minutes is also common.

however WCAG is not a security spec, and i'd be careful making pass/fail dependent on "knowing whether a timeout was chosen arbitrarily or not". an auditor is generally not in a position to determine that. an auditor that feels strongly that what an author/site has implemented is ridiculously too short can still fail it but be prepared for a discussion/argument.

perhaps tweaking the wording slightly from "are exempt" to "may be exempt" is in order though, without having to go into lengthy detail.

understanding/20/timing-adjustable.html
```diff
@@ -34,59 +33,55 @@ <h2>Intent of Timing Adjustable</h2>
...
-            <a href="interruptions" class="guideline">2.2.4</a> and 
-            <a href="re-authenticating" class="guideline">2.2.5</a> may also apply.)
+            and are covered. Success Criteria
+            <a href="no-timing" class="guideline">2.2.3 No Timing</a>,
+            <a href="interruptions" class="guideline">2.2.4 Interruptions</a> and
+            <a href="re-authenticating" class="guideline">2.2.5 Re-Authentication</a> may also apply.
          </li>
-         
+         <li>Certain time limits implemented for security reasons, such as time-based / time-limited
+            two-factor authentication tokens, can be considered <em>essential</em>, and are
```
patrickhlauke, line 73:
```suggestion
            two-factor authentication tokens, can be considered <em>essential</em>, and may be
```

understanding/20/timing-adjustable.html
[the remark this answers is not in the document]
mfairchild365: Thanks @patrickhlauke. Agree that WCAG is not a security spec, but it has to be balanced with security. As you mention, a WCAG auditor is (likely) not a security SME, and thus won't be able to make an informed decision about when the essential exception for timeouts like these kicks in. I see a lot of "well, engineering says it's for security, so it's exempt", but a lot of that has little to no basis in reality or evidence to back it up - it's just being too cautious at the cost of blocking people with certain disabilities. I'm lucky enough to work in a large company with security SMEs that I can consult with on stuff like this, but others are not so lucky.

I'm saying this because I do think there is opportunity (and need) for some sort of documentation to help a WCAG auditor make an informed decision here. That sort of documentation / recommendation should have a cross-discipline participation though. That's likely a whole other project tho.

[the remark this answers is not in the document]
mbgower: As per the two issues that this PR attempts to address, I think there is general agreement that a timed authentication process meets the intent of the essential exception, especially one that follows a published security spec:
> The time limit is [essential](https://www.w3.org/WAI/WCAG21/Understanding/timing-adjustable#dfn-essential) and extending it would invalidate the activity; or

If authors are following their company's security policy, and that company has adopted a published specification that stipulates time limits, that seems entirely sufficient to me for the exception. it seems unreasonable to put the burden on the author to prove the spec is essential.

If someone believes an existing standard's time-out policy is unnecessarily short -- or feels that a standard's timeout should able to be extended 20 times, that feels like an issue that should be opened against the security standard.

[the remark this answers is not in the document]
mbgower: Patrich has made updates to reduce the strength of the language, including changing "are" to "may be" so that it now reads:
> Certain time limits implemented for security reasons, such as time-based / time-limited two-factor authentication tokens, can be considered <em>essential</em>, and may be exempt from this criterion.

Given your comments and these changes, I'm going to mark this as resolved. Feel free to add more comments if you'd like more attention on it.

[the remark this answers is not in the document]
mfairchild365: I like that wording. Thank you @GreggVan.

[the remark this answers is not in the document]
patrickhlauke: @greggvh that suggested rewording doesn't quite make grammatical sense ("if such as ... are essential"). want to take another stab at that?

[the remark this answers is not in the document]
patrickhlauke: I'm guessing your intention was:

```
Certain time limits may be implemented for security reasons. If such time-based / time-limited two-factor authentication tokens are essential, they would be exempt from this criterion.
```

However, this seems more limited and reductive than my original wording. In my wording, I gave "time-based / time-limited two-factor authentication tokens" as examples, as a non-exhaustive list, of time limits implemented for security. This rewording now limits the scope of the note and says that only when THOSE are essential, they would be exempt. If there are other kinds of time limits implemented for security that are similar to those tokens, they would not be covered by the note.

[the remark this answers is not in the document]
patrickhlauke: Longer, but turning things back around:

> Certain time limits, such time-based / time-limited two-factor authentication tokens, may be implemented for security reasons. If these time limits are essential, they would be exempt from this criterion.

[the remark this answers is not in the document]
GreggVan: > @GreggVan that suggested rewording doesn't quite make grammatical sense ("if such as ... are essential"). want to take another stab at that?

Yep  -- had an extra word in there -- I  fixed it.   

But I  like your last version  better. 

> Certain time limits, such time-based / time-limited two-factor authentication tokens, may be implemented for security reasons. If these time limits are essential, they would be exempt from this criterion.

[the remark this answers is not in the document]
patrickhlauke: I'd just add that that last "they would be" seems to still be hedging its bets a bit. if they *are* essential, then they *are* (per normative wording of the SC) exempt.

understanding/20/re-authenticating.html
```diff
@@ -85,94 +80,61 @@ <h2>Examples of Re-authenticating</h2>
                remains intact and, after re-authentication, the user may send that data.</dd>
          <dt>A questionnaire with a time limit</dt>
          <dd>A long questionnaire provided within a single web page has information at the beginning
-               that indicates that the session will time out after 15 minutes. The user is also informed
+               that indicates that the session will time out after 20 hours. The user is also informed
```
mbgower, line 83: I feel like there was a possible intention in the original that although this timed out after 15 minutes, the javascript function at the end of this long list of features was intended to allow it to pass due to the "adjust" method (although as written it doesn't).
My feeling is that we've got a LOT of detritus now, when the only thing the author needs to do is set it to time out after 20 hours. The rest is superfluous and risks confusion what is required.
Does it make any sense to either trim it down?

mbgower: In fact, there isn't even a requirement to _inform_ the user of the time out, so just the first sentence would exceed what is necessary.

understanding/20/re-authenticating.html
```diff
@@ -1,79 +1,74 @@
 <!DOCTYPE html>
 <html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml">
 <head>
-   <meta charset="UTF-8"></meta>
+   <meta charset="UTF-8"/>
```
patrickhlauke:
```suggestion
   <meta charset="UTF-8">
```

understanding/20/re-authenticating.html
```diff
@@ -1,79 +1,74 @@
 <!DOCTYPE html>
 <html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml">
```
patrickhlauke:
```suggestion
<html lang="en">
```

understanding/20/re-authenticating.html
```diff
@@ -1,79 +1,74 @@
 <!DOCTYPE html>
-<html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml">
+<html lang="en">
 <head>
-   <meta charset="UTF-8"></meta>
+   <meta charset="UTF-8">
    <title>Understanding Re-authenticating</title>
    <link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove"/>
```
patrickhlauke:
```suggestion
   <link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove">
```

understanding/20/timing-adjustable.html
```diff
@@ -1,13 +1,13 @@
 <!DOCTYPE html>
 <html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml">
 <head>
-   <meta charset="UTF-8"></meta>
+   <meta charset="UTF-8"/>
```
patrickhlauke:
```suggestion
   <meta charset="UTF-8">
```

understanding/20/timing-adjustable.html
```diff
@@ -1,13 +1,13 @@
 <!DOCTYPE html>
 <html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml">
```
patrickhlauke:
```suggestion
<html lang="en">
```

understanding/20/timing-adjustable.html
```diff
@@ -1,13 +1,13 @@
 <!DOCTYPE html>
-<html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml">
+<html lang="en">
 <head>
-   <meta charset="UTF-8"></meta>
+   <meta charset="UTF-8">
    <title>Understanding Timing Adjustable</title>
    <link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove"/>
```
patrickhlauke:
```suggestion
   <link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove">
```

understanding/20/timing-adjustable.html
```diff
@@ -100,33 +95,21 @@ <h2>Intent of Timing Adjustable</h2>
...
             the device should return the user to the exact condition that existed before it asked
             the question.
-            
-         </li>
-         
-         <li>20 hours was chosen as an upper limit because it is longer than a full waking day.
-            
          </li>
-         
+         <li>20 hours was chosen as an upper limit because it is longer than a full waking day.</li>
```
gundulaniemann, line 99: Some users need to rest or even sleep before they can continue and then might meet the need for some errands or appointments. I feel 30 hours meets the need better.
We cannot change WCAG 2 in this point. Can you remind me where the parking lot for WCAG 3 is located?

mbgower: @gundulaniemann https://github.com/w3c/wcag/wiki/WCAG-3-parking-lot-considerations-from-2.x

understanding/20/re-authenticating.html
patrickhlauke: Note to self: need to revisit why I made this change...

understanding/20/timing-adjustable.html
bruce-usab: @gundulaniemann we have a [WCAG 3 parking lot in the wiki](../wiki/WCAG-3-parking-lot-considerations-from-2.x) for this repo.

Back in the day, one of the reasons we settled on 20 hours is that it was less than the 24 hour day &mdash; which would be blocker for many technologies.  Tweny hours is also 10&times; as long as two hours which is a long enough periods of time that most long process could be expected to have a "save and continue later" type of functionality.

bruce-usab:
Discussed on backlog call 7/18 reaffirming long-standing position that just asserting "security" or "server timeout" is not enough (absent other information) to warrant the essential exception.

bruce-usab (approved):

bruce-usab:
Diffs of deploy previews: 
- [Understanding 2.2.1 Timing Adjustable](https://services.w3.org/htmldiff?doc1=https://www.w3.org/WAI/WCAG22/Understanding/timing-adjustable.html&doc2=https://deploy-preview-4382--wcag2.netlify.app/understanding/timing-adjustable)
- [Understanding 2.2.5 Re-authentication](https://services.w3.org/htmldiff?doc1=https://www.w3.org/WAI/WCAG22/Understanding/re-authenticating.html&doc2=https://deploy-preview-4382--wcag2.netlify.app/understanding/re-authenticating)

understanding/20/re-authenticating.html
patrickhlauke: @mbgower right, I remember now ... I chose to change this to 20 hours to align it unequivocally with the timing adjustable SC, as *this* is about the re-authentication SC ... so wanted to focus the thing on the reauthentication aspect, without creating weird doubts about "wouldn't this pass re-authentication but then still fail the timing SC?"

understanding/20/timing-adjustable.html
patrickhlauke: note that this is existing wording, not changed by this SC (barring the change in whitespace)

understanding/20/re-authenticating.html
mbgower: I find it a little interesting that in this scenario, the system is not actually saving the data. It is left to the user to do that manually. I'm not sure it even meets the SC wording? 
Maybe we should just pull the example?

understanding/20/timing-adjustable.html
patrickhlauke: resolving this conversation here, as the 20 hours are normative and this is about changing it in wcag 3

understanding/20/re-authenticating.html
mbgower: On TF call, agreed to strip out this example entirely

understanding/20/timing-adjustable.html
```diff
@@ -34,59 +33,55 @@ <h2>Intent of Timing Adjustable</h2>
...
             set externally to content, such as by the user agent or by factors intrinsic to the
             Internet are not under the author's control and not subject to WCAG conformance requirements.
             Time limits set by web servers should be under the author's/organization's control
-            and are covered. (Success Criteria 
-            <a href="no-timing" class="guideline">2.2.3</a>, 
-            <a href="interruptions" class="guideline">2.2.4</a> and 
-            <a href="re-authenticating" class="guideline">2.2.5</a> may also apply.)
+            and are covered. Success Criteria
+            <a href="no-timing" class="guideline">2.2.3 No Timing</a>,
+            <a href="interruptions" class="guideline">2.2.4 Interruptions</a> and
```
dbjorge: nit: our style guide (via the Chicago Manual of Style) prefers serial commas

```suggestion
            <a href="interruptions" class="guideline">2.2.4 Interruptions</a>, and
```

mbgower:
@bruce-usab will be generating a preview link.
@patrickhlauke will be removing the example

bruce-usab:
Diff preview for  Understanding [Timing Adjustable](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FUnderstanding%2Ftiming-adjustable&doc2=https%3A%2F%2Fdeploy-preview-4382--wcag2.netlify.app%2Funderstanding%2Ftiming-adjustable) and [Re-Authentication](https://services.w3.org/htmldiff?doc1=https://www.w3.org/WAI/WCAG22/Understanding/re-authenticating.html&doc2=https://deploy-preview-4382--wcag2.netlify.app/understanding/re-authenticating) (corrected).

patrickhlauke:
@mbgower @bruce-usab removed the quiz example from re-authenticating, as discussed in today's TF call

PR #4386, opened 2025-05-07
Chore: Remove invalid closing `</meta>` and `</link>`, self-closing slashes, `xml:lang` and `xmlns`

patrickhlauke:
Removes all sorts of unnecessary XHTML cruft that we kept carrying around due to the old XSLT-based publication pipeline which is now obsolete


patrickhlauke:
Assume we *could* expand this to also remove the `xml:lang="en" xmlns="http://www.w3.org/1999/xhtml"`

kfranqueiro:
The `link` tags look like they weren't modified correctly; there is now a doubled closing bracket in every `link` tag.

Also, I would point out that the build system is fixing these (without the self-closing tag). If I look for these instances in the live pages, I don't see them in the code that is output. So I'm wondering if this isn't worth the effort (and potential for conflicts) to merge, somewhat akin to #3927 which we closed last week after your suggestion.

fstrr:
> Assume we could expand this to also remove the xml:lang="en" xmlns="http://www.w3.org/1999/xhtml"

Oh yeah. I've been wanting to yoink them for ages but wasn't sure if we were keeping them around for some reason.

patrickhlauke:
> The link tags look like they weren't modified correctly; there is now a doubled closing bracket in every link tag.

oops, fat finger typo. I corrected it and force-pushed

Regarding potential for conflicts: I think this sort of change *should* be fine to make, as it would only collide in cases where a PR already made some changes to those bits of the pages...

patrickhlauke:
Controversial perhaps, but: although this crud may be getting automagically removed at actual build time, it would be nice to just rip this plaster off swiftly once and decrap at source. Again, this should not present a big risk of merge conflicts - possibly just a handful of manual tweaks needed for any PRs currently in flight if they affected any images where we're now removing the self-closing slash, for instance

guidelines/relative-luminance.html
```diff
@@ -242,9 +242,9 @@ <h1>MathML version of the relative luminance definition</h1>
               </msub>
               <mo>,</mo>
               <mrow>
-                <mspace width="1ex"/>
+                <mspace width="1ex">
                 <mtext>and</mtext>
-                <mspace width="1ex"/>
+                <mspace width="1ex">
```
fstrr, line 247:
```suggestion
                <mspace width="1ex"/>
```

guidelines/relative-luminance.html
```diff
@@ -242,7 +242,7 @@ <h1>MathML version of the relative luminance definition</h1>
               </msub>
               <mo>,</mo>
               <mrow>
-                <mspace width="1ex"/>
+                <mspace width="1ex">
```
fstrr, line 245:
```suggestion
                <mspace width="1ex"/>
```

fstrr: Put back a trailing slash for the MathML code sample

kfranqueiro, line 245: Looks like this only ended up fixed for one and not the other?
```suggestion
                <mspace width="1ex"/>
```

kfranqueiro:
I have spent the past ~45 minutes reviewing these changes to check for any unintentional changes, especially in the self-closing tag cleanup. (Thanks Francis for spotting the MathML instances.) I didn't find anything else unintentionally caught in that crossfire, but I did find some more `<html>` tags to clean up.

I would point out this is changing files we typically have no interest in changing, so I'm wondering how we feel about them being included in this PR:

- `conformance-challenges/index.html`
- `requirements/21/index.html` and `requirements/22/index.html`
- Many files under `wcag20/` which we generally avoid touching; I would particularly consider backing out changes here, as I am unsure if whatever build process they were involved in requires XHTML

If we want to back any of these out, I can do so quickly.

In addition, note this touches `guidelines/index.html` and `guidelines/relative-luminance.html`, though not in ways that are at all consequential to the written content.

patrickhlauke:
thank you for the assist @kfranqueiro ... think that's now ok to do/send to AGWG/whatever

mbgower:
@kfranqueiro I believe there were a few files you were going to remove or tweak before putting in front of working group?

kfranqueiro:
Addressing Mike's last comment, which had since been resolved in other discussions: I had already reverted the files I'd mentioned, and there's nothing warranting raising an erratum since this changes markup in ways that do not impact any rendered output.

PR #4361, opened 2025-04-25
Update seizures-and-physical-reactions.html

scottaohara:
closes #4328

changes wcag 2.0 to just wcag 2

removes list of a single bullet: 'Ensuring that content does not violate spatial pattern thresholds (future link)' that was not created.

Issue #4328, opened 2025-04-09
Understanding Guideline 2.3: Seizures and Physical Reactions


caztcha:
Understanding Guideline 2.3: Seizures and Physical Reactions | WAI | W3C
https://www.w3.org/WAI/WCAG22/Understanding/seizures-and-physical-reactions

This page contains the description "WCAG 2.0" in the "Intent" section. Because the 2.0 is outdated currently, shouldn't it be "WCAG 2" instead?

> The objective of this guideline is to ensure that content that is marked as conforming to WCAG 2.0 avoids the types of flash that are most likely to cause seizure when viewed even for a second or two.

Kind regards,

patrickhlauke:
I *think* we held off updating every mention of 2.0 in https://github.com/w3c/wcag/pull/3588 because of the mention in https://www.w3.org/TR/WCAG22/#background-on-wcag-2

> Where this document refers to WCAG 2 it is intended to mean any and all versions of WCAG that start with 2.

Happy to reconsider though - there are quite a few other mentions of 2.0 in various understanding docs beyond that one.

stevefaulkner:
> I think we held off updating every mention of 2.0 in https://github.com/w3c/wcag/pull/3588 because of the mention in https://www.w3.org/TR/WCAG22/#background-on-wcag-2

Isn't that a reason to change it to "WCAG 2" rather than specifically refer to an outdated version?

bruce-usab:
I suspect that Understanding for Guidelines (as opposed to Success Criteria) have fallen through the cracks.

When someone volunteers for this, please also delete:
> Ensuring that content does not violate spatial pattern thresholds (future link)

scottaohara:
i made pr #4361 for this and have added it to the drafted column in the wcag 2.x project.  so removing this issue from the project.

PR #4387, opened 2025-05-07
Replace mentions of "WCAG 2.0" with "WCAG 2"

patrickhlauke:
Follow-up to https://github.com/w3c/wcag/pull/3588 where we held off updating mentions because, if i recall correctly, at the time older versions of the spec were also being generated from the same sources. As this has now been addressed, it may make  sense to revisit this.

https://github.com/w3c/wcag/issues/4328#issuecomment-2794872091

This is a more comprehensive update compared to https://github.com/w3c/wcag/pull/4361


scottaohara (approved):

fstrr (approved):

bruce-usab:
Discussed on backlog call 6/6.  Can move to Ready for Approval after one additional double-check.

mbgower:
Based on the latest notes, I'm moving this to Ready for Approval.

Issue #3806, opened 2024-04-29
Creating a new sufficient technique using audio ducking to create audio descriptions


mbgower:
The fairly long discussion in #1768 surfaces the idea of using audio ducking to provide more opportunities for audio descriptions within a prerecorded video.

Introduction of the same technique is also proposed through PR #1790  to the existing language of 1.2.5 

Although this may work better as a discussion, I thought I would initiate it as an issue.

mbgower:
@mraccess77 made the following comment in 1768:
> I worry about audio ducking reducing the volume of audio and playing two audio tracks at once for those who are blind/low vision who are also hard of hearing. I would think that any non-overwritable ducking should be limited to non-spoken audio or important sounds.

patrickhlauke:
regarding @mraccess77 's comment: absolutely, i thought it was implicit that you'd only duck parts that don't have spoken words/content. worth making it explicit though, just in case

bruce-usab:
Noting that this (open) issue relates to conversation on 3/28 backlog call.
- Having reread the threads, and since we have this more narrow thread, I recommend closing #1768 as resolved.
- Rather than a new sufficient technique, I (still) recommend a new _Failure_ around not ducking non-speech audio (in order to dub additional narration).

mbgower:
@bruce-usab and @patrickhlauke I've created a draft of the first failure technique for 1.2.5 https://github.com/w3c/wcag/pull/4390
I'm wondering if we could incorporate the concept of audio ducking into that failure?

I don't feel like we need a separate failure for audio ducking, as it logically seems to be part of a failure called "Failure of Success Criterion 1.2.5 due to not using available pauses in dialogue to provide audio descriptions of important visual content"

I do still feel like audio ducking should be part of a _sufficient_ technique, and based on what I see in the existing techniques, I believe it should be a separate technique that can be listed as one of the techniques these other grosser techniques (like using a second audio track) employ.

mbgower:
Currently, the failure has a note that reads
> Not all pauses are usable for audio descriptions. If the pauses in dialogue are too short (for instance, less than 2 seconds), or do not occur in proximity to the visual content that needs to be described, they may not be appropriate for audio descriptions.

I suggest removing the note styling and making that be a paragraph, and then adding an additional paragraph to read something like:
[...]

PR #4327, opened 2025-04-07
Meeting audio description through narration

mbgower:
- Added the word "additional" to note 1 of Understanding documents for 1.2.3 and 1.2.5 (This makes it match note 3 in the audio description definition, which reads "Where all of the video information is already provided in existing audio, no additional audio description is necessary.")
- Added a new technique "Providing audio descriptions by incorporating narration into the soundtrack"

[Preview of technique](https://raw.githack.com/w3c/wcag/AudioDescriptionPreexisting/techniques/general/G226.html)

PR #4390, opened 2025-05-09
Audio description failure technique

mbgower:
Closes https://github.com/w3c/wcag/issues/3806
Closes https://github.com/w3c/wcag/issues/1768 (in combination with the other previously merged changes outlined in https://github.com/w3c/wcag/issues/1768#issuecomment-2784911795)

Creates a failure technique to show the conditions under which a video fails 1.2.5 due to important visual information not being described in pauses in the dialogue

[Preview](https://deploy-preview-4390--wcag2.netlify.app/techniques/failures/f113)


techniques/failures/F113.html
```diff
@@ -0,0 +1,26 @@
...
+   <h1>Failure of Success Criterion 1.2.5 due to not using available pauses in dialogue to provide audio descriptions of important visual content</h1>
+   <section class="meta"><p class="id">ID: F113</p><p class="technology">Technology: failures</p><p class="type">Type: Failure</p></section>
+   <section id="applicability"><h2>When to Use</h2>
+      <p> Applies to all synchronized media. </p>
+   </section>
+   <section id="description"><h2>Description</h2>
+      <p>This describes a failure condition for all techniques involving audio descriptions. Audio descriptions can either be provided as part of the soundtracks's original narration or be added to audio during pauses in existing dialogue. If actions, characters, scene changes, and on-screen text are only conveyed visually, appropriate pauses in dialogue must be used to provide this information as audio descriptions.</p>
```
mbgower, line 8:
```suggestion
      <p>This describes a failure condition for all techniques involving audio descriptions. Audio descriptions can either be provided as part of the soundtrack's original narration or be added to audio during pauses in existing dialogue. If actions, characters, scene changes, and on-screen text are only conveyed visually, appropriate pauses in dialogue must be used to provide this information as audio descriptions.</p>
```

techniques/failures/F113.html
```diff
@@ -0,0 +1,26 @@
...
+   <h1>Failure of Success Criterion 1.2.5 due to not using available pauses in dialogue to provide audio descriptions of important visual content</h1>
+   <section class="meta"><p class="id">ID: F113</p><p class="technology">Technology: failures</p><p class="type">Type: Failure</p></section>
+   <section id="applicability"><h2>When to Use</h2>
+      <p> Applies to all synchronized media. </p>
+   </section>
+   <section id="description"><h2>Description</h2>
+      <p>This describes a failure condition for all techniques involving audio descriptions. Audio descriptions can either be provided as part of the soundtrack's original narration or be added to audio during pauses in existing dialogue. If actions, characters, scene changes, and on-screen text are only conveyed visually, appropriate pauses in dialogue must be used to provide this information as audio descriptions.</p>
```
mbgower, line 8:
```suggestion
      <p>This describes a failure condition for all techniques involving audio descriptions. Audio descriptions can either be provided as part of the soundtrack's original narration or be added to audio during pauses in existing dialogue. If important actions, characters, scene changes, and on-screen text are only conveyed visually, appropriate pauses in dialogue must be used to provide this information as audio descriptions.</p>
```

techniques/failures/F113.html
```diff
@@ -0,0 +1,26 @@
...
+   <h1>Failure of Success Criterion 1.2.5 due to not using available pauses in dialogue to provide audio descriptions of important visual content</h1>
+   <section class="meta"><p class="id">ID: F113</p><p class="technology">Technology: failures</p><p class="type">Type: Failure</p></section>
+   <section id="applicability"><h2>When to Use</h2>
+      <p> Applies to all synchronized media. </p>
+   </section>
+   <section id="description"><h2>Description</h2>
+      <p>This describes a failure condition for all techniques involving audio descriptions. Audio descriptions can either be provided as part of the soundtrack's original narration or be added to audio during pauses in existing dialogue. If important actions, characters, scene changes, and on-screen text are only conveyed visually, appropriate pauses in dialogue must be used to provide this information as audio descriptions.</p>
```
scottaohara, line 8: staying away from the sacred word "must".  adding "might" to account for some of the "what about when pauses are meaningful" discussion from today's call. IMO, this also helps align what's said here to square with the following note that says "not all pauses are usable..."

i know this says "appropriate pauses" - but since there is no definition of "appropriate"...  my thought is this also helps align with the note that kind of outlines what might be considered appropriate or not.

```suggestion
      <p>This describes a failure condition for all techniques involving audio descriptions. Audio descriptions can either be provided as part of the soundtrack's original narration or be added to audio during pauses in existing dialogue. If important actions, characters, scene changes, and on-screen text are only conveyed visually, appropriate pauses in dialogue might need to be used to provide this information as audio descriptions.</p>
```

scottaohara (commented):
just one suggestion.  otherwise this makes sense to me  - and the other delving into the applicability or essential nature of a pause can be handled in content to be added to the understanding doc

techniques/failures/F113.html
mbgower: @scottaohara 
> staying away from the sacred word "must".

IMO, the context of the use of "must" here is appropriate for a failure technique. The scenario is:
1. important visual information is present
2. appropriate pauses in the dialogue exist

So, the idea is that it has already been determined that the pauses are appropriate. In such a case, I do not believe there is any way to provide meet 1.2.5 except by adding audio descriptions to narrate missing important visuals?

If you agree with this sentiment, but do not feel that is clear from this wording, I'll attempt a rewrite.
If you disagree with this sentiment, maybe you can say what level of containment we need to determine a failure.

Note that I have tried to use the phrase "appropriate pauses" consistently in the title, wording and check. Not a defined term, but my hope is consistent use within the failure can give clarity.

scottaohara: naw i get where you're coming from.  i'd still rather we not use the word 'must' per everything that's been drilled into me about not using such terms in informative docs.  wcag is far looser with must and should than any other doc i've worked on.

techniques/failures/F113.html
```diff
@@ -0,0 +1,26 @@
...
+       <div class="note">
+        <p>Not all pauses are usable for audio descriptions. If the pauses in dialogue are too short (for instance, less than 2 seconds), or do not occur in proximity to the visual content that needs to be described, they may not be appropriate for audio descriptions.</p>
```
scottaohara, line 10: so issue #3806 said to maybe remove this note and use the following instead:

>In situations where important visual information is being conveyed at the same time as non-spoken audio (such as music and sound effects), the technique of "audio ducking" can be used. This involves dropping the overall sound level so that it is easier to distinguish the narration that is added during pauses in dialogue.
>
>This technique can work well with background music and sounds, but audio ducking has the potential to mask important audio information. Often such audio information can convey much of the sense of what is visually happening. There may be some pauses in dialogue where non-spoken audio information is so important that the addition of audio descriptions is unnecessary or inappropriate.

I wonder if all the content shouldn't just be used?    The note could be revised to segue into the first paragraph of the proposed addition.

mbgower: To clarify, it is the word "note" and the note styling that is to be removed. The existing wording is retained, and these_ two paragraphs were in _addition_ to that wording. Just no "Note:".

techniques/failures/F113.html
mbgower, line 8:
```suggestion
      <p>This describes a failure condition for all techniques involving audio descriptions. Audio descriptions can either be provided as part of the soundtrack's original narration or be added to audio during pauses in existing dialogue. If important actions, characters, scene changes, and on-screen text are only conveyed visually, appropriate pauses in dialogue need to be used to provide this information as audio descriptions.</p>
```

techniques/failures/F113.html
mbgower: I've made those changes.

techniques/failures/F113.html
mbgower: When in doubt, take it out :)
I took your suggested rewording but removed "might".
If it looks fine now, please mark the conversation as resolved.

scottaohara (approved):

techniques/failures/F113.html
```diff
@@ -0,0 +1,28 @@
...
+        <p>This technique can work well with background music and sounds, but audio ducking has the potential to mask important audio information. Often such audio information can convey much of the sense of what is visually happening. There may be some pauses in dialogue where non-spoken audio information is so important that the addition of audio descriptions is unnecessary or inappropriate.</p>
+
+</p>
+   </section>
+    
+   <section id="tests"><h2>Tests</h2>
+      <section class="procedure"><h3>Procedure</h3>
+         <p>For each occurrence of synchronized time-based media:</p>
+         <ol>
+            <li> Check that all important visual information has been conveyed in the narration.</li>
```
awkawk, line 20: Check that all visual information necessary to understand the content is conveyed by audio in the synchronized media.

techniques/failures/F113.html
```diff
@@ -0,0 +1,28 @@
...
+            <li> Check that narration has been added in all appropriate pauses in the dialogue. </li>
```
awkawk, line 21: Check that audio description is provided for any visual information necessary to understand the content which is not conveyed by audio in the synchronized media.

techniques/failures/F113.html
```diff
@@ -0,0 +1,28 @@
...
+        <p>In situations where important visual information is being conveyed at the same time as non-spoken audio (such as music and sound effects), the technique of "audio ducking" can be used. This involves dropping the overall sound level so that it is easier to distinguish the narration that is added during pauses in dialogue.</p>
+        <p>This technique can work well with background music and sounds, but audio ducking has the potential to mask important audio information. Often such audio information can convey much of the sense of what is visually happening. There may be some pauses in dialogue where non-spoken audio information is so important that the addition of audio descriptions is unnecessary or inappropriate.</p>
+
+</p>
+   </section>
+    
+   <section id="tests"><h2>Tests</h2>
+      <section class="procedure"><h3>Procedure</h3>
+         <p>For each occurrence of synchronized time-based media:</p>
```
awkawk, line 18: For each occurrence of synchronized media containing video:

techniques/failures/F113.html
```diff
@@ -0,0 +1,28 @@
...
+        <p class="note">Not all pauses are usable for audio descriptions. If the pauses in dialogue are too short (for instance, less than 2 seconds), or do not occur in proximity to the visual content that needs to be described, they may not be appropriate for audio descriptions.</p>
+        <p>In situations where important visual information is being conveyed at the same time as non-spoken audio (such as music and sound effects), the technique of "audio ducking" can be used. This involves dropping the overall sound level so that it is easier to distinguish the narration that is added during pauses in dialogue.</p>
+        <p>This technique can work well with background music and sounds, but audio ducking has the potential to mask important audio information. Often such audio information can convey much of the sense of what is visually happening. There may be some pauses in dialogue where non-spoken audio information is so important that the addition of audio descriptions is unnecessary or inappropriate.</p>
```
awkawk, line 11: Having trouble following the point of this paragraph. Is it that ducking can obscure needed audio information or that sometimes the non-spoken audio can eliminate the need for some descriptions?

mbgower: I assume you mean the last one?

There is material about _not_ filling every pause with more information if the audio is actually more important than the video 
For instance, see [example 2](https://dcmp.org/learn/captioningkey/617#7:~:text=Let%20pauses%20or%20quiet%20moments%20pass%20without%20a%20description%20when%20appropriate%20to%20support%20the%20mood.%20Do%20not%20try%20to%20fill%20every%20pause.%20Allow%20atmosphere%20and%20background%20sound%20to%20come%20through.)

Some_ examples would probably be useful, although maybe that would be better in a sufficient technique than in a failure technique.

techniques/failures/F113.html
mbgower, line 20: i think we need to be careful with the wording here. The normative language says "important visual details that cannot be understood from the main soundtrack alone". I know "important" is not, itself, defined, but I think it's better to stick with "important" rather than introducing a concept that is inferred but never stated explicitly: 
```suggestion
            <li> Check that all important visual information  that cannot be understood from the main soundtrack alone has been conveyed in the narration.</li>
```

techniques/failures/F113.html
mbgower: This second check is there to make sure any available appropriate pauses are used.
The first check has already validated that the audio descriptions either cover all the material, or they don't.
 Here, we're testing that all the pauses have been used for narration.
 Together they give us the gates for the failure:
#1 passes= no failure
 #1 fails and #2 passes=no failure
 1 fails, 2 fails=failure

techniques/failures/F113.html
awkawk: What is "the narration"? I thought this first line was about what was in the main soundtrack audio?

techniques/failures/F113.html
mbgower, line 18:
```suggestion
         <p>For each occurrence of synchronized media containing video:</p>
```

techniques/failures/F113.html
awkawk: The original working is actually checking that all possibly useful pauses are used, which may not be appropriate.

techniques/failures/F113.html
mbgower: It could either be in the main soundtrack or in the added audio description (or both).

awkawk: So I guess I don't understand what you are doing here. I'll re-read.

mbgower: I added in a variation (below) to try to address this.

techniques/failures/F113.html
mbgower: incorporated

awkawk:
Here's what I was thinking this was testing:

For all synchronized media with video:
1) Check that all important visual information in the video is conveyed by the main soundtrack. 
2) Check that any important visual information in the video that isn't conveyed in the main soundtrack is conveyed by audio description.

Is this what you are saying?

techniques/failures/F113.html
mbgower: One of the challenges we've been given with the Audio Description SC is that obviously if someone has done their work well, an instructional or educational video can have good narration without the need for the kind of intervention envisioned in this SC (adding a second audio track, etc). However, there is little scope in the SC wording for how an author could incorporate Audio Description into the pre-existing soundtrack (either while it is still in production by modifying the script, or in what I'll call post-production by adding more narrative into the main soundtrack.
The SC wording does not do a good job of considering adjustments made to a video that do _not_ involve the 2 main techniques that currently exist (adding a second, independently playable audio track; adding a separate version of the video with audio descriptions).
The definition of audio description is "narration added to the soundtrack".  That is the sense in which "narration" is intended to be used.

mbgower: I think we actually need a third check which is something like 'there is another version of this video which passes 1 and 2"

awkawk: If you get into CAVs in all failure techniques you will need to write them all.

mbgower:
> Here's what I was thinking this was testing:
> 
> For all synchronized media with video:
> 
> 1. Check that all important visual information in the video is conveyed by the main soundtrack.
> 2. Check that any important visual information in the video that isn't conveyed in the main soundtrack is conveyed by audio description.
> 
> Is this what you are saying?

Nope.
1. Check that all important visual information in the video is conveyed ~by the main soundtrack~.
2. Check that all pauses are used for audio desriptions

techniques/failures/F113.html
awkawk: > However, there is little scope in the SC wording for how an author could incorporate Audio Description into the pre-existing soundtrack (either while it is still in production by modifying the script, or in what I'll call post-production by adding more narrative into the main soundtrack.

What about the definition of audio description?
"narration added to the soundtrack to describe important visual _details that cannot be understood from the main soundtrack alone_"

awkawk:
> Nope.
> 
> 1. Check that all important visual information in the video is conveyed ~by the main soundtrack~.
> 2. Check that all pauses are used for audio descriptions
> 
> If you fail both, you fail.
> 
> I was _hoping_ that was obvious from the title "Failure of Success Criterion 1.2.5 due to not using available pauses in dialogue to provide audio descriptions of important visual content"

Ha! Hope is dashed! :)

The title is not the test procedure, you can't rely on it.

So you want to combine my 1 and 2 into 1. How about:
1. Check that all important visual information in the video is conveyed by the main soundtrack or by an audio description track.
2. Check that the main soundtrack has pauses that would be appropriate to provide audio description of important visual content.

both true: pass
both false: pass
1 true, 2 false: pass
1 false, 2 true: fail

techniques/failures/F113.html
mbgower: You're forgetting note 3
> Where all of the video information is already provided in existing audio, no additional audio description is necessary.

If the only way we can fix this is to revise the unfortunate original definition of audio description, we can go there. I was hoping that succeeded in somewhat plugging this hole by the raft of changes we've got queued up to take to the WG, including https://github.com/w3c/wcag/pull/4327

awkawk:
> Can you explain where you are primarily snagging on this now?

You really seem to want to have audio description in all pauses, but this is definitely the snag.

mbgower:
> > Can you explain where you are primarily snagging on this now?
> 
> You really seem to want to have audio description in all pauses, but this is definitely the snag.

I would not say that. What the test is doing is seeing if all the video is described. If it's all described, then unused pauses are immaterial, it passes. 
If all important visuals are not described, then check that all appropriate pauses have been used. If the meaningful visuals aren't all described and if all the appropriate pauses aren't utilitized, then it fails.

awkawk:
@mbgower I emailed with Bryan Gould, who is at WGBH-NCAM and worked for many years as a description editor at WGBH-DVS.

Bryan replied to me (quoted sections are from my email to Bryan, non-quoted are his replies):
I’ve responded to your comments below:

> My concern is the second check, which suggests that the right way to handle audio description is to fill every available pause with audio description if there is any information deemed important that is missing. I worry that we will get into a situation where well-intentioned people will think that there is some important information from the video that isn’t voiced and therefore must fill every pause.

I agree. For 2, the word “appropriate” provides enough ambiguity that this check may very well be interpreted as a “fill-every-pause” mandate.
 
If that is the case, the next question is, what constitutes a pause? A lot of information can be conveyed with a single well-placed word, for example, “Later” or “Elsewhere.”  These words can be vital to understanding but very few one-second breaks in dialogue require description.
 
The practice of filling every pause with description may also be detrimental. In order to reduce cognitive load, the viewer may filter out the “excess”  description.  As a result, the viewer may miss the actually important information.

> I suspect that even the best-described video might fail this evaluation, as there is way too much visual information to completely reproduce in AD, and there are times where a description editor needs to make a judgement about the impact on the mood of the content and viewer’s experience. 
 
Correct. A description editor’s main role is to apply context to all visual information in order to deem what is “important” and thus requiring of description.  For example, a 10-second view of an orchid requires a different description approach depending on whether the context is botany, photography, or part of a montage set to music. There are also instances when a pause in dialogue is intended for the viewer to reflect on what was just said, this is especially true in education and training videos. In other words, there is almost always too much visual information to describe and the description editor is always deciding what is important. Finally, dedication to context helps to prevent the cognitive load issue raised above.

mbgower:
Thanks for providing that context and outside, informed opinion in your latest [comment](https://github.com/w3c/wcag/pull/4390#issuecomment-2877821323), @awkawk.

There's nothing stated there that is new to me, and I share all those concerns and observations based on my own limited experience in this field. However, this is pretty similar, philosophically, to problems that exist in 1.1.1 for images. In the case of 1.1.1, we realistically have  a gross litmus test: "is there an alt (and if not, is this purely decorative)?" This is the basic check every automated checker currently does. And then we have a qualitative test: "is it equivalent?" 

We've all seen some pretty awful alt text. But concerns with the quality of the alt has not deterred us from insisting that alt text must exist.

For audio descriptions, we only have the text we were given to work. I thought a gross test for 1.2.5 could be "Is there any audio description?" That seems entirely supported by the normative text. But several people have pushed back against that idea.

The only other gross tests we seem to have at our disposal with the normative wording are:
- are there important visuals with no audio equivalent?
- are there pauses which could be used to insert audio descriptions?

Both of these are _highly_ subjective. I sympathize with the concerns Bryan states, but the reality is that as a percentage of total videos on the web, the number that are audio described is much closer to 0% than 1%.  I think we _have_ to provide some kind of gross test for audio descriptions _before_ we turn to worrying about whether the descriptions are good or not.

Rejecting usable pauses as a metric for 1.2.5 for fear someone creates crappy audio descriptions seems unfortunate to me.

mbgower:
@awkawk a compromise is possibly to remove "all" from both of the checks. I'm not keen on it (since it really lowers the bar for passing)  BUT it would create a higher bar for failing.

Check that important visual information has been conveyed in the narration.
Check that narration has been added in appropriate pauses in the dialogue.

mbgower:
Check that all important visual information has been conveyed in the narration.
Check that narration has been added in appropriate pauses in the dialogue.

mbgower:
Moved to ready for approval; will move back depending on @awkawk response.

bruce-usab (changes requested):
I don't think the 2nd step of test procedure tracks closely enough to the failure name.  I recommend:

For each occurrence of synchronized media containing video:
1. Check that all important visual information has been conveyed in the narration.
2. Check that there are available pauses in dialogue where narration could be added.

If check 1 is false and check 2 is true, then this failure condition applies and the content fails the success criterion.

techniques/failures/F113.html
```diff
@@ -0,0 +1,28 @@
...
+
+</p>
+   </section>
+    
+   <section id="tests"><h2>Tests</h2>
+      <section class="procedure"><h3>Procedure</h3>
+         <p>For each occurrence of synchronized media containing video:</p>
+         <ol>
+            <li> Check that all important visual information has been conveyed in the narration.</li>
+            <li> Check that narration has been added in all appropriate pauses in the dialogue. </li>
```
mbgower, line 21:
```suggestion
            <li> Check that audio descriptions have been added in pauses in the dialogue, where appropriate. </li>
```

techniques/failures/F113.html
mbgower:
```suggestion
            <li>Check that all important visual information that cannot be understood from the main soundtrack alone has been conveyed in audio descriptions.</li>
```

techniques/failures/F113.html
mbgower:
```suggestion
        <p>This technique can work well with background music and sounds, but audio ducking has the potential to mask important audio information. Sometimes such audio information can convey much of the sense of what is visually happening. </p>
        <p class="note">There may be some pauses in dialogue where non-spoken audio information is so important that the addition of audio descriptions is inappropriate.</p>
```

mbgower:
@bruce-usab 
> I don't think the 2nd step of test procedure tracks closely enough to the failure name and it's not as simple as it could be.

The PR was updated after your comment, and I believe may have at least partially addressed. Please review and comment

bruce-usab:
> Check that audio descriptions have been added in pauses in the dialogue, where appropriate.

Is not as clear as:
> Check that there are available pauses in dialogue where narration could be added.

1. The "where appropriate" is more ambiguous than "are available pauses".
2. Still retains the words "audio description" which has overloaded implications that there is a second audio track or distinct voice actor for the addition narration.

GreggVan:
@bruce-usab   Can you clarify your point?  I'm not sure I  understand.

To help - my confusion is around

>1. The "where appropriate" is more ambiguous than "are available pauses".

the "where appropriate" relates to whether or not Audio  Description is needed at all at that point in the video.  The "are available pauses"  relates to the opportunity to add it even if there was a need.     

>2. Still retains the words "audio description" which has overloaded implications that there is a second audio track or distinct voice actor for the addition narration.

Audio description is the normal term - and talks about adding description in gaps.   Like captions - it can be the only audio track or an additional audio track (it is usually the latter).       Do you think there is another term that should be used?


PS   on a different topic (Sort of)    I  like your "test steps" suggestion further above

techniques/failures/F113.html
```diff
@@ -0,0 +1,28 @@
...
+      <section class="procedure"><h3>Procedure</h3>
+         <p>For each occurrence of synchronized media containing video:</p>
+         <ol>
+            <li>Check that all important visual information that cannot be understood from the main soundtrack alone has been conveyed in audio descriptions.</li>
+            <li> Check that audio descriptions have been added in pauses in the dialogue, where appropriate. </li>
+         </ol>
+      </section>
+      <section class="results"><h3>Expected Results</h3>
+         <ul>
+            <li>If checks 1 and 2 are false, then this failure condition applies and the content fails the success criterion.</li>
```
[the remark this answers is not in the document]
patrickhlauke: Mentioned in call that having that second step worded this proposed way, it feels weirdly back to front. Suggest rewording the second bullet as "IF not, check that..."

and then the expected results can remain as a single "if both fail"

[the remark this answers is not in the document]
kfranqueiro: Suggestion of alternative wording for the second check:

Check that there are pauses in dialogue where it would be appropriate to add audio descriptions.

bruce-usab (approved):
I am okay with PR as-is.

I think readability would be improved by using AD in the singular.

Current:
> - Check that all important visual information that cannot be understood from the main soundtrack alone has been conveyed in audio descriptions.
> - Check that there are pauses in dialogue where it would be appropriate to add audio descriptions.

Suggestion:
> - Check that all important visual information that cannot be understood from the main soundtrack alone has been conveyed by audio description.
> - Check that there are pauses in dialogue where it would be appropriate to add audio descriptions.

bruce-usab:
Discussed and some changes approved on backlog call 5/30.  There was still some concern about "if check 1 is false and check 2 is true" being awkward.  If check 1 is true, then check 2 is not necessary.  Remains in For Discussion for now.

kfranqueiro:
> Discussed and some changes approved on backlog call 5/30. There was still some concern about "if check 1 is false and check 2 is true" being awkward. If check 1 is true, then check 2 is not necessary. Remains in For Discussion for now.

I was poking around for anything similar, and so far the most similar existing procedure I have found is in [F83](https://www.w3.org/WAI/WCAG22/Techniques/failures/F83#tests), where the first check is given a name that is then referenced below. Interestingly, the condition is included both in the 2nd step and in the expected results.

If we do this, I might suggest something like "Preliminary check" rather than "Quickcheck"?

techniques/failures/F113.html
```diff
@@ -0,0 +1,29 @@
...
+
+</p>
+   </section>
+    
+   <section id="tests"><h2>Tests</h2>
+      <section class="procedure"><h3>Procedure</h3>
+         <p>For each occurrence of synchronized media containing audio and video:</p>
+         <ol>
+            <li>Check that all important visual information that cannot be understood from the main soundtrack alone has been conveyed in audio descriptions.</li>
+            <li>Check that there are pauses in dialogue where it would be appropriate to add audio descriptions.</li>
```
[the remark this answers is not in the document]
patrickhlauke: "A failure should be worded as a statement -- not a series of checks"

See all other existing failure techniques, which do similar? https://www.w3.org/WAI/WCAG22/Techniques/#failures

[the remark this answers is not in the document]
patrickhlauke: > We do not require that all gaps be used

and this PR doesn't seem to suggest that it's required?

techniques/failures/F113.html
GreggVan: This failure test goes beyond the SC.  The SC does not require that ALL important information be conveyed.    Nor does it require that all gaps be used.   There is a good reason for both

1) there are not enough gaps in the right places to describe all important information for many or most video.  So it was not required or else we would have had to make it level AAA because it could not be applied to all (or even most) video.    So the SC only requires that SOME but not ALL important info be provided

2) it does not require that all gaps be used because a gap that appears at a place where there is no important information cannot be used to describe somethig that happened a long time ago.    Also  - some gaps are too small to be used.   

So a failure that implies or tests for  all important information  or all gaps to be used -- is not a valid failure since it is testing for things not in the SC.    In addition it would be something that would fail the best attempts at audio description for many videos.

patrickhlauke: i assume this was in response to my comment here https://github.com/w3c/wcag/pull/4390#discussion_r2116441623 rather than andrew's from 3 weeks ago...

techniques/failures/F113.html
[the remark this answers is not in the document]
mbgower: > a failure should be worded as a statement -- not a series of checks -- unless you are talking about a TEST for a failure.

It is the test, and in failure techniques they _are_ a series of checks, @GreggVan 

I went through the first 20 failure techniques. Almost _all_ of them are a series of one or more checks followed by a "If check #1 is..." Of the four that do not use the word "Check", three of them, F7, F12, and F19 just use the synonym "determine" to exactly the same purpose. The only one I could find that does not actually use either of these words in its steps was F15 -- and even then, ALL 20 used the concluding structure "If check #1 is..." to form the final test..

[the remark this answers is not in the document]
mbgower: @patrickhlauke IMO, your two cited examples are not parallel (F2 has a confusingly worded _opposite_ of the prior check and F44 is simply one check with a preamble, not a set of checks). I also find that when there are only a couple of examples out of dozens, that it sorta makes the point that what you're asking for is uncommon.

However, it is possible to emulate F44 if that's what you want, although it makes for a wordy single check that involves a double negative.

```suggestion
            <li>If all important visual information that cannot be understood from the main soundtrack alone has not been conveyed in audio descriptions, check that there are pauses in dialogue where it would be appropriate to add audio descriptions.</li>
<li>If check 1 is true, then this failure condition applies and the content fails the success criterion.</li>
```

Or maybe you'd prefer it in three steps? I actually think this reads okay, and is maybe in line with what you're seeking? (Note that this isn't a true suggestion, because I haven't included replacing the first bullet, but I think you get the idea.)

```suggestion
            <li>Check that all important visual information can be understood from the main soundtrack alone</li>
            <li>Check that all missing important visual information has been conveyed in audio descriptions</li>
            <li>If not, check that there are pauses in dialogue where it would be appropriate to add audio descriptions.</li>
<li>If Checks 1 and 2 are false and Check 3 is true, then this failure condition applies and the content fails the success criterion.</li>
```

techniques/failures/F113.html
```diff
@@ -0,0 +1,29 @@
+<!DOCTYPE html><html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml"><head><title>Failure of Success Criterion 1.2.5 due to not using available pauses in dialogue to provide audio descriptions of important visual content</title><link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove"/></head><body>
```
mbgower, line 1:
```suggestion
<!DOCTYPE html><html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml"><head><title>Failure of Success Criteria 1.2.3 and 1.2.5 due to not using available pauses in dialogue to provide audio descriptions of important visual content</title><link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove"/></head><body>
```

techniques/failures/F113.html
```diff
@@ -0,0 +1,29 @@
+<!DOCTYPE html><html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml"><head><title>Failure of Success Criteria 1.2.3 and 1.2.5 due to not using available pauses in dialogue to provide audio descriptions of important visual content</title><link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove"/></head><body>
+   <h1>Failure of Success Criterion 1.2.5 due to not using available pauses in dialogue to provide audio descriptions of important visual content</h1>
```
mbgower, line 2:
```suggestion
   <h1>Failure of Success Criteria 1.2.3 and 1.2.5 due to not using available pauses in dialogue to provide audio descriptions of important visual content</h1>
```

understanding/20/audio-description-prerecorded.html
```diff
@@ -253,7 +253,11 @@ <h3>Additional Techniques (Advisory) for Audio Description (Prerecorded)</h3>
       <section id="failure">
          <h3>Failures for Audio Description (Prerecorded)</h3>
          
-         
+         <ul>
+            <li>
+               <a href="../Techniques/failure/F113" class="failure">Failure of Success Criterion 1.2.5 due to not using available pauses in dialogue to provide audio descriptions of important visual content</a>
```
mbgower, line 263:
```suggestion
               <a href="../Techniques/failure/F113" class="failure">Failure of Success Criteria 1.2.3 and 1.2.5 due to not using available pauses in dialogue to provide audio descriptions of important visual content</a>
```

understanding/20/audio-description-prerecorded.html
```diff
@@ -253,7 +253,11 @@ <h3>Additional Techniques (Advisory) for Audio Description (Prerecorded)</h3>
       <section id="failure">
          <h3>Failures for Audio Description (Prerecorded)</h3>
          
-         
+         <ul>
+            <li>
+               <a href="../Techniques/failure/F113" class="failure">Failure of Success Criteria 1.2.3 and 1.2.5 due to not using available pauses in dialogue to provide audio descriptions of important visual content</a>
```
mbgower:
```suggestion
               <a href="../Techniques/failure/F113" class="failure">Failure of Success Criterion 1.2.5 due to not using available pauses in dialogue to provide audio descriptions of important visual content</a>
```

techniques/failures/F113.html
```diff
@@ -0,0 +1,29 @@
+<!DOCTYPE html><html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml"><head><title>Failure of Success Criteria 1.2.3 and 1.2.5 due to not using available pauses in dialogue to provide audio descriptions of important visual content</title><link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove"/></head><body>
```
mbgower:
```suggestion
<!DOCTYPE html><html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml"><head><title>Failure of Success Criterion 1.2.5 due to not using available pauses in dialogue to provide audio descriptions of important visual content</title><link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove"/></head><body>
```

techniques/failures/F113.html
```diff
@@ -0,0 +1,29 @@
+<!DOCTYPE html><html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml"><head><title>Failure of Success Criterion 1.2.5 due to not using available pauses in dialogue to provide audio descriptions of important visual content</title><link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove"/></head><body>
+   <h1>Failure of Success Criteria 1.2.3 and 1.2.5 due to not using available pauses in dialogue to provide audio descriptions of important visual content</h1>
```
mbgower:
```suggestion
   <h1>Failure of Success Criterion 1.2.5 due to not using available pauses in dialogue to provide audio descriptions of important visual content</h1>
```

kfranqueiro (changes requested):
@mbgower your last commits seem to have somehow reverted F113 to an empty file?

awkawk:
What if the third check was:
Check that there are no remaining pauses in dialog where it is appropriate to add audio description.

And then the failure would be triggered when all three checks are false?

techniques/failures/F113.html
```diff
@@ -0,0 +1,30 @@
...
+        <p>In situations where important visual information is being conveyed at the same time as non-spoken audio (such as music and sound effects), the technique of "audio ducking" can be used. This involves dropping the overall sound level so that it is easier to distinguish the narration that is added during pauses in dialogue.</p>
+        <p>This technique can work well with background music and sounds, but audio ducking has the potential to mask important audio information. Sometimes such audio information can convey much of the sense of the video. </p>
+        <p class="note">There may be some pauses in dialogue where non-spoken audio information is so important that the addition of audio descriptions is inappropriate in that pause.</p>
+
+</p>
+   </section>
+    
+   <section id="tests"><h2>Tests</h2>
+      <section class="procedure"><h3>Procedure</h3>
+         <p>For each occurrence of synchronized media containing audio and video:</p>
```
alastc: From the meeting: We could move the first check into the scoping statement, and remove the "if not" bits.

techniques/failures/F113.html
```diff
@@ -0,0 +1,30 @@
...
+         <ol>
+             <li>Check that all important visual information can be understood from the main soundtrack alone.</li>
```
mbgower:
```suggestion
         <p>For each occurrence of synchronized media containing audio and video, where all important visual information cannot be understood from the main soundtrack alone:</p>
         <ol>
```

techniques/failures/F113.html
```diff
@@ -0,0 +1,29 @@
...
+        <p class="note">There may be some pauses in dialogue where non-spoken audio information is so important that the addition of audio descriptions is inappropriate in that pause.</p>
+
+</p>
+   </section>
+    
+   <section id="tests"><h2>Tests</h2>
+      <section class="procedure"><h3>Procedure</h3>
+         <p>For each occurrence of synchronized media containing audio and video, where all important visual information cannot be understood from the main soundtrack alone:</p>
+         <ol>
+	          <li>If not, check that all missing important visual information has been conveyed in audio descriptions.</li>
```
mbgower:
```suggestion
         <p>For each occurrence of synchronized media containing audio and video, where all important visual information cannot be understood from the main soundtrack alone:</p>
         <ol>
	          <li>Check that all missing important visual information has been conveyed in audio descriptions.</li>
```

techniques/failures/F113.html
```diff
@@ -0,0 +1,29 @@
...
+
+</p>
+   </section>
+    
+   <section id="tests"><h2>Tests</h2>
+      <section class="procedure"><h3>Procedure</h3>
+         <p>For each occurrence of synchronized media containing audio and video, where all important visual information cannot be understood from the main soundtrack alone:</p>
+         <ol>
+	          <li>Check that all missing important visual information has been conveyed in audio descriptions.</li>
+	          <li>If not, check that there are pauses in dialogue where it would be appropriate to add audio descriptions.</li>
```
mbgower:
```suggestion
	          <li>Check that there are no remaining pauses in dialog where it is appropriate to add audio description.</li>
```

techniques/failures/F113.html
```diff
@@ -0,0 +1,29 @@
...
+      <section class="procedure"><h3>Procedure</h3>
+         <p>For each occurrence of synchronized media containing audio and video, where all important visual information cannot be understood from the main soundtrack alone:</p>
+         <ol>
+	          <li>Check that all missing important visual information has been conveyed in audio descriptions.</li>
+	          <li>Check that there are no remaining pauses in dialog where it is appropriate to add audio description.</li>
+         </ol>
+      </section>
+      <section class="results"><h3>Expected Results</h3>
+         <ul>
+            <li>If Checks 1 and 2 are false and Check 3 is true, then this failure condition applies and the content fails the success criterion.</li>
```
mbgower:
```suggestion
            <li>If Checks 1 and 2 are false, then this failure condition applies and the content fails the success criterion.</li>
```

mbgower:
> What if the third check was: Check that there are no remaining pauses in dialog where it is appropriate to add audio description.
> 
> And then the failure would be triggered when all three checks are false?

Incorporated.

techniques/failures/F113.html
```diff
@@ -0,0 +1,29 @@
...
+        <p>In situations where important visual information is being conveyed at the same time as non-spoken audio (such as music and sound effects), the technique of "audio ducking" can be used. This involves dropping the overall sound level so that it is easier to distinguish the narration that is added during pauses in dialogue.</p>
+        <p>This technique can work well with background music and sounds, but audio ducking has the potential to mask important audio information. Sometimes such audio information can convey much of the sense of the video. </p>
+        <p class="note">There may be some pauses in dialogue where non-spoken audio information is so important that the addition of audio descriptions is inappropriate in that pause.</p>
+
+</p>
+   </section>
+    
+   <section id="tests"><h2>Tests</h2>
+      <section class="procedure"><h3>Procedure</h3>
+         <p>For each occurrence of synchronized media containing audio and video, where all important visual information cannot be understood from the main soundtrack alone:</p>
```
mbgower:
```suggestion
         <p>For each occurrence of synchronized media containing audio and video where all important visual information cannot be understood from the main soundtrack alone:</p>
```

techniques/failures/F113.html
```diff
@@ -0,0 +1,29 @@
...
+
+</p>
+   </section>
+    
+   <section id="tests"><h2>Tests</h2>
+      <section class="procedure"><h3>Procedure</h3>
+         <p>For each occurrence of synchronized media containing audio and video where all important visual information cannot be understood from the main soundtrack alone:</p>
+         <ol>
+	          <li>Check that all missing important visual information has been conveyed in audio descriptions.</li>
+	          <li>Check that there are no remaining pauses in dialog where it is appropriate to add audio description.</li>
```
mbgower:
```suggestion
	          <li>Check that there are no remaining pauses in dialog where it is appropriate to add audio descriptions.</li>
```

PR #4400, opened 2025-05-17
Amend "contrast ratio" term - move out non-term notes to understanding docs

patrickhlauke:

---

A different take compared to https://github.com/w3c/wcag/pull/1018

While reviewing that PR, it was noted that some of the existing notes in the term are inappropriate and incorrect, and are not necessarily related to the term itself but to SC testing.

This PR removes them altogether (which does mean this is a normative change), and moves the remaining notes to the understanding for contrast minimum, contrast enhanced, and one of them to non-text contrast — where they more logically belong as they don't define the term, but relate to evaluation and testing.

and yes, i'm aware this PR changes both normative and non-normative docs - which is why the check for this fails at the moment. however, as this is a "move content from normative to non-normative" i'd say it makes sense to keep it as a single PR rather than splitting into two separate ones.

Adding the more granular commit messages for context:

**Remove smoothing/anti-aliasing note**
The term now (with 1.4.11) applies to both text and non-text content, but it seems incorrect to refer to specific testing methodology in the normative term rather than in the more specific SCs. 1.4.3 and 1.4.6 understanding (https://www.w3.org/WAI/WCAG22/Understanding/contrast-minimum.html and https://www.w3.org/WAI/WCAG22/Understanding/contrast-enhanced) already have a note about antialiasing, so it's redundant to have here)

**Remove the note about "It is a failure if no background color is specified"**
This note is contentious and lacks context. There is already a failure technique [Failure of Success Criterion 1.4.3, 1.4.6 and 1.4.8 due to specifying foreground colors without specifying background colors or vice versa](https://www.w3.org/WAI/WCAG22/Techniques/failures/F24) that is referenced from 1.4.3, 1.4.6, 1.4.8 (but not 1.4.11), so the idea behind the note is still present (though that failure lacks nuance). But fundamentally, this note has no business being in the definition of "what is *contrast ratio*"

**Remove the note about assuming a white background when none is defined**
This note is in contradiction already with the other note above that was removed, and contradicts the failure technique about missing background, as it currently stands.

**Move note about border for text into understanding**

**Move note about border for text into enhanced understanding**

**Tweak 1.4.11 note about anti-aliasing**
mentioning font smoothing here is irrelevant


understanding/20/contrast-minimum.html
patrickhlauke:
```suggestion
         <p>Because authors do not have control over user settings for font smoothing and anti-aliasing, when evaluating this
```

patrickhlauke:
While this PR touches both normative and non-normative files, it makes sense to keep it as a single PR (as the non-normative changes wouldn't make sense without the normative ones)

patrickhlauke:
Discussed quickly at today's TF meeting. Will give folks more time to digest this. To make the PR clearer, just added the more granular commit messages to the top comment/PR description.

baldino-m:
It seems to me that the border for text (former note) has not got enough prominence in its new location (understanding): since it used to be a note, it might be a note on the understanding page as well.

patrickhlauke:
Updated the description above https://github.com/w3c/wcag/pull/4400#issue-3071092784 giving more context about the notes that were removed altogether

awkawk:
Disagree with this change in the normative documents. This is a major change that we shouldn't be entertaining. If there is a specific issue with the notes we should edit the notes as errata.

patrickhlauke:
> Disagree with this change in the normative documents. This is a major change that we shouldn't be entertaining

Notes are informative, not normative, so I don't believe their removal/moving them to a more appropriate informative location (the understanding docs) changes anything substantially.

detlevhfischer:
I don't care all too much whether these notes sit here (where they are more prominent but, also bulk up the glossary) or in Understanding... Given there is much else to do, I would rather shelve this.

WilcoFiers:
I'm with @awkawk on this. I appreciate that these notes seem like they belong in an understanding document. But by moving them we're downgrading their status from W3C approved content to AGWG documentation with no formal status.

This introduces significant drift between WCAG and its translations. I think these notes are important enough for understanding how to measure this requirement that we should want to keep them in official translations, which we only get if we keep it in the WCAG itself.

bruce-usab (approved):
Preview diff at [glossary term](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FTR%2F2024%2FREC-WCAG22-20241212%2F&doc2=https%3A%2F%2Fdeploy-preview-4400--wcag2.netlify.app%2Fguidelines%2F#dfn-contrast-ratio) in guidelines and at Intent in Understanding [1.4.3](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FUnderstanding%2Fcontrast-minimum&doc2=https%3A%2F%2Fdeploy-preview-4400--wcag2.netlify.app%2Funderstanding%2Fcontrast-minimum#intent), [1.4.8](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FUnderstanding%2Fcontrast-enhanced&doc2=https%3A%2F%2Fdeploy-preview-4400--wcag2.netlify.app%2Funderstanding%2Fcontrast-enhanced#intent), and [1.4.11](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FUnderstanding%2Fnon-text-contrast&doc2=https%3A%2F%2Fdeploy-preview-4400--wcag2.netlify.app%2Funderstanding%2Fnon-text-contrast#intent).&nbsp; The substantive edits to Understanding in this PR are at the end of the Intent sections.&nbsp; (Just before “Rationale for the Ratios Chosen” with Understanding 1.4.3 and 1.4.8, and just before “User Interface Components” in Understanding 1.4.11.)

patrickhlauke:
Discussed in last TF meeting. Agreed to move to the potential errata column for the next CfC

understanding/21/non-text-contrast.html
```diff
@@ -28,9 +28,11 @@ <h2>Intent</h2>
 			</div>
 
 			<div class="note">
+
 				<p>Because authors do not have control over user settings for font smoothing and anti-aliasing, when evaluating this
```
patrickhlauke:
```suggestion
				<p>Because authors do not have control over user settings for anti-aliasing, when evaluating this
```

Issue #902, opened 2019-09-17
exception for logos (1.4.3, 1.4.11)


JAWS-test:
* 1.4.3 has the exception for logos in the normative SC
* 1.4.11 has no exception for logos in the normative SC for "User Interface Components".
* 1.4.11 has the "essential" exception for "Graphical Objects" in normative SC
* "essential" is clearly defined: "if removed, would fundamentally change the information or functionality of the content, and information and functionality cannot be achieved in another way that would conform ". In my opinion, logos do not fall under this definition, at least not if they convey information or serve as interactive elements. 
* The Understanding declares that logos fall under the "essential" exception on the one hand, and there is a fail example for logos on the other (Figure 14). I consider this to be contradictory, even if the logos have been changed in colour, it is not clear that this is the only reason why it is a fail. Why are logos essential? Why is logo exception at 1.4.3 in normative SC and at 1.4.11 only in non-normative Understanding? I understand that logos cannot be changed so easily, but if a logo does not have sufficient contrasts, it should not be used without visible alternative text.
* what is if a logo is both a "Graphical Object" and a "User Interface Component" (link with the logo as the only label)? Does the exception "essential" apply or not?

My suggestion:
* for WCAG 2.1 and 2.2: Answer the above questions in the Understanding.
* for WCAG 3.0: Remove the exception for logos. Alternatively, leave the exception only if the logos do not convey information or are interactive elements.

See also the discussion, starting from comment: https://github.com/w3c/wcag/issues/901#issuecomment-532126775

alastc:
> "essential" is clearly defined... In my opinion, logos do not fall under this definition, at least not if they convey information or serve as interactive elements.

In my opinion it does, it is such a well established exception from WCAG 2.0 that during the discussions on non-text contrast we decided not to call it out separately.

It is currently called out in the understanding doc. I can put it to the group if you like, but I suspect the answer would be that it is fine as-is.

JAWS-test:
> In my opinion it does, it is such a well established exception from WCAG 2.0 that during the discussions on non-text contrast we decided not to call it out separately.

I don't think that's a good decision. If WCAG 2.1 supplements WCAG 2.0, it would have been better to maintain the way in which exceptions are described so that SCs remain consistent.

What's more important:  
> what is if a logo is both a "Graphical Object" and a "User Interface Component" (link with the logo as the only label)? Does the exception "essential" apply or not?

That's relevant because
[...]

PR #4402, opened 2025-05-18
Add note about logo/logotypes contrast to 1.4.3, 1.4.6, and 1.4.11 understanding

### LINKED ISSUE #1275: No exception regarding contrasts for logos if they are interactive (SC 1.4.3, 1.4.11)

### LINKED ISSUE #1739: Logo excepted from 1.4.3 Contrast (Minimum), but does it fail as a link for 1.4.11 Non-text Contrast?

> 
> **Graphical Objects**

> Required for Understanding
> 

Further on it again says:

> Essential Exception
> 

Anything we can do about it?

Example here:

### LINKED ISSUE #1742: Logo still excepted 1.4.11 Non-text Contrast when colors are not the original brand colors?
When logo's are used as a way to communicate they are the brands present / available for the product of the page, and needed to understand what choices you have.

but...

The Graphical Objects used to provide this information happen to be logo's BUT the color is changed, to have a visual match for all logo's (basically make the intent of the except non applicable...).

May we FAIL this? (I think we should)

And going into this direction, does the except for logo's only apply if the reason for the except also apply (the 'essential'?

Example is at a Deque demo page, where the logo's in the middle of the page are intentionally made light grey, basically taking away the 'essential' part

### LINKED ISSUE #4376: SC 1.4.11 - Solid colour logo as link

## The case

## My impression
> **Graphical Objects**

"Testing Principles" states:

## My conclusion

Possible solutions:

patrickhlauke:
While exempting text or non-text that is part of a logo/logotype makes sense (due to some companies' strict corporate identity requirements), it's nonetheless problematic when these logos/logotypes act as links of buttons.

These notes try to clarify the situation ... for 1.4.3 and 1.4.6 we can only suggest a best practice of - if possible - choosing an alternative that is still allowed by the CI/brand guidelines, or at the very least an additional link/control with the same purpose and with sufficient contrast.

This part of the original PR has been split out into a separate PR https://github.com/w3c/wcag/pull/5113
> For 1.4.11 there is a wrinkle here because the SC distinguishes between user interface controls and graphical objects. Logos on their own count as graphical objects, and if the CI/brand guidelines stipulated non-contrasting colours, then that is "essential". However, the "user interface control" part of a link/button with a logo is still subject to the requirements of the UIC needing to be identifiable.

For 1.4.11, this PR adds the same best practice suggestion about using a logo variant or alternative control.

Lastly, this PR addresses the situation where logos have been used, but it's the author's choice - rather than a corporate identity/brand requirement - that the logos have insufficient contrast.

Closes https://github.com/w3c/wcag/issues/1742

x-ref https://github.com/w3c/wcag/issues/902 https://github.com/w3c/wcag/issues/4376 https://github.com/w3c/wcag/issues/1275 https://github.com/w3c/wcag/issues/1739

Previews:

* https://deploy-preview-4402--wcag2.netlify.app/understanding/contrast-minimum
* https://deploy-preview-4402--wcag2.netlify.app/understanding/contrast-enhanced
* https://deploy-preview-4402--wcag2.netlify.app/understanding/non-text-contrast


erikkroes (approved):

understanding/21/non-text-contrast.html
```diff
@@ -396,6 +396,19 @@ <h4>Required for Understanding</h4>
 					<li> The information is available in another form, such as in a table that follows the graph, which becomes visible when a "Long Description" button is pressed.</li>
 					<li> The graphic is part of a logo or brand name (which is considered &quot;essential&quot; to its presentation).</li>
 				</ul>
+
+				<div class="note">
+					<p>Logos and logotypes are exempted from contrast requirements when they are purely used as
+						<em>graphical objects</em>, under the assumption that they must comply with stricter color choices mandated
+            by corporate identity guidelines.
+						However, this is not the case when they also act as <em>user interface components</em>
+						(such as links or other interactive controls). In these cases, authors should choose
```
erikkroes, line 405: I think people might read your writing as best practice or advice. "If the logo is a user interface component, you should ...."
While I think you meant to say something like: "When logos are used as interface components don't meet contrast requirements, you could do one of the following..."
I think a more explicit disctinction between the clarifcation (user interface components are not excempt) and the possible solutions would be more clear than the current proposal.

erikkroes (changes requested):

understanding/21/non-text-contrast.html
patrickhlauke: i intended it in the RFC sense

> [3](https://datatracker.ietf.org/doc/html/rfc2119#section-3). SHOULD   This word, or the adjective "RECOMMENDED", mean that there
   may exist valid reasons in particular circumstances to ignore a
   particular item, but the full implications must be understood and
   carefully weighed before choosing a different course.

understanding/21/non-text-contrast.html
```diff
@@ -396,6 +396,19 @@ <h4>Required for Understanding</h4>
...
+						a variant of the logo or logotype that has sufficient contrast, if allowed by the
+						corporate identity guidelines.
+						Alternatively, authors should provide an equivalent <em>user interface component</em>
+						which serves the same purpose and does meet the contrast requirements.</p>
```
erikkroes, line 409:
```suggestion
					<p>Logos and logotypes are exempted from contrast requirements when they are purely used as <em>graphical objects</em>, under the assumption that they must comply with stricter color choices mandated by corporate identity guidelines. They are not excempt when they also act as <em>user interface components</em> (such as links or other interactive controls). 
            When logos and logotypes are used as <em>user interface components</em>, authors should choose a variant of the logo or logotype that has sufficient contrast, if allowed by the  corporate identity guidelines. Alternatively, authors should provide an equivalent <em>user interface component</em> which serves the same purpose and does meet the contrast requirements.</p>
```

I understand the value and meaning of should in RFCs. I hope this slight rephrasing and reformatting makes the text more explicit and understandable.

fstrr: Visually, the code looks like you want that update as two paragraphs, but there's only one set of `<p>` tags. Should it be one or two?

patrickhlauke: I'm assuming the salient bit was the addition of this in the first para:

`They are not exempt when they also act as <em>user interface components</em> (such as links or other interactive controls).`

(with correction of `excempt` to `exempt` as well)

patrickhlauke: Noting that making this categorical statement will no doubt ruffle some feathers, as it "seems" to introduce a normative failure by the backdoor that some will argue wasn't as cut and dry there before...which is why I was treading lightly (in one of the linked issues there was some mention of "exempting logos was on purpose as it would lead to too many failures" or words to that effect)

patrickhlauke: happy to take it as additional suggestion to the WCAG 2.x backlog group though for further discussion

patrickhlauke: @erikkroes I updated your suggestion above for clarity (keeping the line breaks where they were, so it's clearer what your change is), and assuming you did want the first para to be actually split into two as well

erikkroes: I've opened this issue for clarification, so maybe I tread less lightly 😄

The split might not be needed indeed. I think it helped make my point in this conversation, but it should be fine without. 
Thanks for picking this up and taking it further!

understanding/21/non-text-contrast.html
gundulaniemann: In fact, if the logo is visible by some contrasting parts, it should be sufficient if the text inside is given by an alternative like a tooltip. Where the tooltip should also explain the functionality.

kfranqueiro:
HTML diffs:

- [Contrast (Enhanced)](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fwcag%2Funderstanding%2Fcontrast-enhanced&doc2=https%3A%2F%2Fdeploy-preview-4402--wcag2.netlify.app%2Funderstanding%2Fcontrast-enhanced)
- [Contrast (Minimum)](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fwcag%2Funderstanding%2Fcontrast-minimum&doc2=https%3A%2F%2Fdeploy-preview-4402--wcag2.netlify.app%2Funderstanding%2Fcontrast-minimum)
- [Non-text Contrast](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fwcag%2Funderstanding%2Fnon-text-contrast&doc2=https%3A%2F%2Fdeploy-preview-4402--wcag2.netlify.app%2Funderstanding%2Fnon-text-contrast#required-for-understanding)

understanding/20/contrast-enhanced.html
```diff
@@ -126,7 +126,19 @@ <h2>Intent of Contrast (Enhanced)</h2>
...
+      <div class="note">
+         <p>Text used as part of a logo or logotype is exempted from contrast requirements,
+            under the assumption that logos/logotypes must comply with stricter color choices mandated
+            by corporate identity guidelines.
+            However, this can be problematic when the logo or logotype also acts as a
+            <em>user interface component</em> (such as a link or other interactive control).
+            In these cases, authors should choose a variant of the logo or logotype that has sufficient
+            contrast, if allowed by the corporate identity guidelines.
+            Alternatively, authors should provide an equivalent <em>user interface component</em>
+            which serves the same purpose and does meet the contrast requirements.</p>
```
alastc, line 139:
```suggestion
         <p>Text used as part of a logo or logotype is exempted from contrast requirements, under the assumption that logos/logotypes must comply with stricter color choices mandated by corporate identity guidelines. However, this can be problematic when the logo or logotype also acts as a <em>user interface component</em> (such as a link or other interactive control). Some part of the control must meet the 3:1 contrast ratio so users can identify that there is a control there. Authors could choose a variant of the logo or logotype that has sufficient contrast, include a border with contrast, or provide an equivalent <em>user interface component</em> which serves the same purpose and does meet the contrast requirements.</p>
```

alastc: A suggestion to avoid "should".

patrickhlauke:
> This informal change would invalidate a long practice of exempting logos from contrast requirements

I'd argue that for graphical/non-text logos at least, that practice was ... misguided, considering the distinction in 1.4.11 between graphical object and user interface component. for those, i'd posit that it's not a normative change, just a clarification of what should have been the practice all along

patrickhlauke:
> well, it is a practice covered in 1.1.1 for much longer than 1.4.11 exists 

where does 1.1.1 exempt logos?

assuming you meant 1.4.3 / 1.4.6, it only covers "Text that is part of a logo or brand name" so doesn't exempt any non-text parts of a logo

understanding/21/non-text-contrast.html
patrickhlauke: note that generally, you're not allowed to modify logos along the lines of "just put an outline around the logo's shapes"

understanding/20/contrast-enhanced.html
```diff
@@ -126,7 +126,11 @@ <h2>Intent of Contrast (Enhanced)</h2>
          in charts, graphs, diagrams, and other non-text-based information, which is covered by
          <a href="non-text-contrast">Success Criterion 1.4.11 Non-Text Contrast</a>.         
       </p>
-      
+
+      <div class="note">
+         <p>Text used as part of a logo or logotype is exempted from contrast requirements, under the assumption that logos/logotypes must comply with stricter color choices mandated by corporate identity guidelines. However, this can be problematic when the logo or logotype also acts as a <em>user interface component</em> (such as a link or other interactive control). Some part of the control must meet the 3:1 contrast ratio so users can at least identify that there is a control. Authors could choose a variant of the logo or logotype that has sufficient contrast, include a border with contrast (though this may still not help users identify <em>what</em> the control's label is), or provide an equivalent <em>user interface component</em> which serves the same purpose and does meet the contrast requirements.</p>
```
alastc, line 131:
```suggestion
         <p class="wcag21">Text used as part of a logo or logotype is exempted from contrast requirements, under the assumption that logos/logotypes must comply with stricter color choices mandated by corporate identity guidelines. However, this can be problematic when the logo or logotype also acts as a <em>user interface component</em> (such as a link or other interactive control). There are also contrast requirements for <a href="non-text-contrast">Success Criterion 1.4.11 Non-text Contrast</a>, where some part of the control must have contrast, or an accessible alternative.</p>
```

GreggVan: yes  this also needs to be worded differently.  This language puts and exception and rationale in a note -- which is not allowable.   If it is an exception the exception needs to be in the provision and the rationale needs to use wording in the provision. 

Eg something like this. 

Note: Text used as part of a logo or logotype falls under the "essential" exemption because logos/logotypes often must comply with stricter color choices mandated by corporate identity guidelines.  

Note:  When the logo or logotype also acts as a <em>user interface component</em> (such as a link or other interactive control) other contrast requirements apply. See  <a href="non-text-contrast">Success Criterion 1.4.11 Non-text Contrast</a>.

and then include the proper note on logos as controls there.

bruce-usab:
Discussed on backlog call 5/30, so far so good, but additional feedback is welcomed.

mbgower:
A few points, as a broad follow up to today's TF call.
1. I think the space we want to largely tackle the logo requirements, in regard to 1.4.11, has to be within the context of User Interface Components (not Graphical Objects).
2. I think we can put in some language on best practices, including the idea that most corporate logos exist in various forms, including black and white renditions which tend to have better contrast. However, there is already specific wording in the Understanding document addressing how logos are exempt via the essential part of the graphical objects clause.
3. Likewise, we should think about how that best practices wording could be reflected in some updates to the text contrast SCs, bearing in mind the logotype exception in place in 1.4.3/6.
4. I agree we can provide some possible techniques as examples within the Understanding document. But I also think this _maybe_ would work better as a separate technique called something like "G###: Improving contrast for logos when used as user interface components". The reason I think this is that the Understanding document is already very long. With a distinct technique, we could mention it each time the logo exception is mentioned in the existing understanding document, and have a fairly small new section added. And then elaborate in detail in the technique. @gundulaniemann if you are game for this, I think you could run with creating a first draft of this techique, and have @patrickhlauke be your initial reviewer.
5. @patrickhlauke I'd like to get your thoughts on ways we can try to get more agreement on an approach _**in the issue**_(s) before moving to PR discussions? I ask this because when we spin these PR discussions into dozens (or a hundred) of comments, it makes it hard for someone reviewing the PR to quickly grasp the essentials. We can obviously address to some degree by putting in a summary in the main description after the fact (before it gets reviewed by the WG), but I wonder if there are ways of chasing nuance in the issue, where conversations tend to be better preserved?
On the other hand, I get that if the topic only gets attention when a PR is generated, it makes for a conundrum!

patrickhlauke:
Reworded/reworked this into a new suggestion, seeing the pushback the original suggestion got:

* for non-text contrast, clarified what the absolute minimum requirement seems to be - having "something" that at least helps indicate that there IS a UIC (even if you can't tell WHAT it is)
* still added clarification of when a logo may NOT be exempted (if the presentation of the logo is down to author choice, not corporate identity/brand guidelines)
* for text contrast, can only make it a best practice suggestion

patrickhlauke:
for clarity, the "author choice" thing would be things like this mockup here ... say a conference site chooses (because it looks "sexy") to have sponsor logos super faint (and then on hover, they become properly visible)

image

that presentation is not "essential", it's not because the corporate id mandated it ... it's because the author chose to do that. and THAT is not exempt, I would claim.

patrickhlauke:
Direct links to the previews

* https://deploy-preview-4402--wcag2.netlify.app/understanding/contrast-minimum
* https://deploy-preview-4402--wcag2.netlify.app/understanding/contrast-enhanced
* https://deploy-preview-4402--wcag2.netlify.app/understanding/non-text-contrast

detlevhfischer:
> If logos are presented with an insufficient contrast, but their presentation was an author choice rather than
being mandated by corporate identity or brand guidelines, then that particular low contrast presentation is
_not_ "essential",	and the logo is _not_ exempt from the contrast requirements.

Author choice or not? I fear this will often be an impossible call to make, at least for auditors, when they have no direct access to corporate style guides or policies, or designers' deliberations. 

I still do not see how being linked (which logos very often are, and have been, as part of site navigation) invalidates the essential exception in 1.4.11 for graphical objects. If it should invalidate it, that should be spelled out normatively in the exception, not sneaked in via a note. The fact that UIE and graphic objects stand separate in 1.4.11 and the first does not explicitly limit the exception of the second leaves ambiguity, but in the face of such ambiguity I prefer avoiding a stealth normative change unless such a change is agreed by the WG (even as substantive errata) and embedded in the normative text (in which case I fully support it).

As a mere side comment, im many cases the larger size and the overall gestalt of the header with the conventional top left position of the logo as well as the general rule of thumb that activating "that blob top left" takes you to some landing / start page, all contextually support logo identification and seem to reduce the practical user impact of low contrast in logos.

patrickhlauke:
> I still do not see how being linked (which logos very often are, and have been, as part of site navigation) invalidates the essential exception in 1.4.11 for graphical objects.

the graphical object is exempted, but the user interface component that wraps the logo isn't exempted. so you can keep the faint logo, but there needs to be *some* way for the user to understand the presence of the actual UIC (i'd love to be able to go further and actually invalidate the exemption for the graphical object part in this case, but ... at the very least, users will know that there IS a UIC there, even if they still can't discern WHAT it is)

patrickhlauke:
> As a mere side comment, im many cases the larger size and the overall gestalt of the header with the conventional top left position of the logo

you seem to be fixated on the idea that this is aimed at somehow punishing the logo in the top left of the header. when really this is more about cases where authors have gleefully done "design" on logos on the main page, like those "here's a list of all our sponsors" / "these are the clients that use our product" / etc

patrickhlauke:
> Author choice or not? I fear this will often be an impossible call to make, at least for auditors, when they have no direct access to corporate style guides or policies, or designers' deliberations.

ok, so you do an audit, you as an auditor think "surely this visual treatment of the logo is not mandated by their CI, so i'll flag it". the client, upon receipt of the assessment, tells you "actually, our CI forces us to do this" and you change the fail to a pass...

and if you're an author (the other audience for these guidelines), you are told here that just because it's a logo, it doesn't mean you now have carte blanche to do whatever you want.

detlevhfischer:
@patrickhlauke lets just see what the rest of the group will say—it is not a molehill I intend to suffer on

bruce-usab:
Discussed on backlog call 6/20.  Everyone liked the idea of adding an [example described](#issuecomment-2970679410) and @patrickhlauke is volunteered to write that up.

detlevhfischer:
> Discussed on backlog call 6/20. Everyone liked the idea of adding an [example described](https://github.com/w3c/wcag/pull/4402#issuecomment-2970679410) and @patrickhlauke is volunteered to write that up.

So is this now just on the level of best practice / example? Or does it require, on the part of the auditor, to determine if a logo has bad contrast because it follows the CI or because the designer thought it looked fancy in low contrast? I wouldn't like to have a back-and-forth with site owners on such a question, necessitating re-opening a test and changing pass/fail judgements later on because the site owner said: "that's actually part of our CI". But if all we are doing is adding an example and a best practice recommendation, I see no problem.

patrickhlauke:
> Or does it require, on the part of the auditor, to determine if a logo has bad contrast because it follows the CI or because the designer thought it looked fancy in low contrast?

My initial intention is the latter, ideally. But if that receives too much pushback now, I could imagine it being tempered into advice that goes beyond the requirement, but I'd preface it accordingly - explaining that the exemption exists under the assumption that logos use mandated CI colours and can't be (easily) changed, but unfortunately that's not been made immediately explicit, so any logos pass even if you as an author decided to show them at 0.1 opacity or something because you think it looks fancier... but that's really very weak

understanding/20/contrast-minimum.html
```diff
@@ -127,6 +127,18 @@ <h2>Intent of Contrast (Minimum)</h2>
          <a href="contrast-enhanced">1.4.6 Contrast (Enhanced)</a>.
       </p>
 
+      <div class="note">
+         <p>Text used as part of a logo or logotype is exempted from contrast requirements,
+            under the assumption that logos/logotypes must comply with stricter color choices mandated
+            by corporate identity or brand guidelines. However, this can be problematic when
+            the logo or logotype also acts as a <a>user interface components</a> (such as a link or
```
fstrr: need to tweak this to deal with the plural definition.
Also: not sure if "also act" adds some complexity and removing "also" would be simpler.
```suggestion
            by corporate identity or brand guidelines. However, this can be problematic when
            logos or logotypes also act as <a>user interface components</a> (such as a link or
```

patrickhlauke: i believe it also works in singular, but sure. and can probably drop the also, yes. i need to revisit this at some point (just didn't get around to it in a while)

understanding/20/contrast-enhanced.html
```diff
@@ -126,7 +126,19 @@ <h2>Intent of Contrast (Enhanced)</h2>
          in charts, graphs, diagrams, and other non-text-based information, which is covered by
          <a href="non-text-contrast">Success Criterion 1.4.11 Non-Text Contrast</a>.         
       </p>
-      
+
+     <div class="note">
+         <p>Text used as part of a logo or logotype is exempted from contrast requirements,
+            under the assumption that logos/logotypes must comply with stricter color choices mandated
+            by corporate identity or brand guidelines. However, this can be problematic when
+            the logo or logotype also acts as a <a>user interface components</a> (such as a link or
```
patrickhlauke:
```suggestion
            logos or logotypes act as <a>user interface components</a> (such as a link or
```

understanding/20/contrast-minimum.html
```diff
@@ -127,6 +127,18 @@ <h2>Intent of Contrast (Minimum)</h2>
          <a href="contrast-enhanced">1.4.6 Contrast (Enhanced)</a>.
       </p>
 
+      <div class="note">
+         <p>Text used as part of a logo or logotype is exempted from contrast requirements,
+            under the assumption that logos/logotypes must comply with stricter color choices mandated
+            by corporate identity or brand guidelines. However, this can be problematic when
+            logos or logotypes also act as <a>user interface components</a> (such as a link or
```
patrickhlauke:
```suggestion
            logos or logotypes act as <a>user interface components</a> (such as a link or
```

understanding/20/contrast-minimum.html
```diff
@@ -127,6 +127,18 @@ <h2>Intent of Contrast (Minimum)</h2>
...
 
+      <div class="note">
+         <p>Text used as part of a logo or logotype is exempted from contrast requirements,
+            under the assumption that logos/logotypes must comply with stricter color choices mandated
+            by corporate identity or brand guidelines. However, this can be problematic when
+            the logo or logotype also acts as a <a>user interface components</a> (such as a link or
+            other interactive control). In these cases, as a best practice, choose a variant of
+            the logo or logotype that has sufficient text contrast, if allowed by the
+            corporate identity or brand guidelines. Alternatively, consider providing an equivalent
+            <a>user interface components</a> which serves the same purpose and does meet
```
fstrr: tweak to deal with plural definition
```suggestion
            corporate identity or brand guidelines. Alternatively, consider providing equivalent
            <a>user interface components</a> which serve the same purpose and do meet
```

understanding/21/non-text-contrast.html
```diff
@@ -401,6 +401,23 @@ <h4>Required for Understanding</h4>
 					<li> The information is available in another form, such as in a table that follows the graph, which becomes visible when a "Long Description" button is pressed.</li>
 					<li> The graphic is part of a logo or brand name (which is considered &quot;essential&quot; to its presentation).</li>
 				</ul>
+
+				<div class="note">
+					<p>Logos are exempted from contrast requirements when they are purely used as
+						<em>graphical objects</em>, under the assumption that they must comply with stricter color choices mandated
+            by corporate identity or brand guidelines.	When they also act as <a>user interface components</a>
```
patrickhlauke:
```suggestion
            by corporate identity or brand guidelines.	When they act as <a>user interface components</a>
```

understanding/20/contrast-enhanced.html
```diff
@@ -126,7 +126,19 @@ <h2>Intent of Contrast (Enhanced)</h2>
...
-      
+
+     <div class="note">
+         <p>Text used as part of a logo or logotype is exempted from contrast requirements,
+            under the assumption that logos/logotypes must comply with stricter color choices mandated
+            by corporate identity or brand guidelines. However, this can be problematic when
+            logos or logotypes act as <a>user interface components</a> (such as a link or
+            other interactive control). In these cases, as a best practice, choose a variant of
+            the logo or logotype that has sufficient text contrast, if allowed by the
+            corporate identity or brand guidelines. Alternatively, consider providing an equivalent
```
patrickhlauke:
```suggestion
            corporate identity or brand guidelines. Alternatively, consider providing equivalent
```

understanding/20/contrast-enhanced.html
```diff
@@ -126,7 +126,19 @@ <h2>Intent of Contrast (Enhanced)</h2>
...
+            <a>user interface components</a> which serves the same purpose and does meet
```
patrickhlauke:
```suggestion
            <a>user interface components</a> which serve the same purpose and meet
```

understanding/20/contrast-minimum.html
```diff
@@ -127,6 +127,18 @@ <h2>Intent of Contrast (Minimum)</h2>
...
 
+      <div class="note">
+         <p>Text used as part of a logo or logotype is exempted from contrast requirements,
+            under the assumption that logos/logotypes must comply with stricter color choices mandated
+            by corporate identity or brand guidelines. However, this can be problematic when
+            logos or logotypes act as <a>user interface components</a> (such as a link or
+            other interactive control). In these cases, as a best practice, choose a variant of
+            the logo or logotype that has sufficient text contrast, if allowed by the
+            corporate identity or brand guidelines. Alternatively, consider providing equivalent
+            <a>user interface components</a> which serve the same purpose and do meet
```
patrickhlauke:
```suggestion
            <a>user interface components</a> which serve the same purpose and meet
```

understanding/21/non-text-contrast.html
```diff
@@ -401,6 +401,23 @@ <h4>Required for Understanding</h4>
 					<li> The information is available in another form, such as in a table that follows the graph, which becomes visible when a "Long Description" button is pressed.</li>
 					<li> The graphic is part of a logo or brand name (which is considered &quot;essential&quot; to its presentation).</li>
 				</ul>
+
+				<div class="note">
+					<p>Logos are exempted from contrast requirements when they are purely used as
+						<em>graphical objects</em>, under the assumption that they must comply with stricter color choices mandated
+            by corporate identity or brand guidelines.	When they act as <a>user interface components</a>
+						(such as links or other interactive controls), authors must nonetheless make sure that there is an aspect of
+						the user interface component itself that allows	users to identify the component and its state.
```
patrickhlauke:
```suggestion
						the user interface components that allows users to identify the components and their state.
```

understanding/21/non-text-contrast.html
```diff
@@ -401,6 +401,23 @@ <h4>Required for Understanding</h4>
...
 				</ul>
+
+				<div class="note">
+					<p>Logos are exempted from contrast requirements when they are purely used as
+						<em>graphical objects</em>, under the assumption that they must comply with stricter color choices mandated
+            by corporate identity or brand guidelines.	When they act as <a>user interface components</a>
+						(such as links or other interactive controls), authors must nonetheless make sure that there is an aspect of
+						the user interface components that allows users to identify the components and their state.
+						If possible, and if allowed by the relevant corporate identity or brand guidelines, authors should choose a
+						variant of the logo that has sufficient contrast.	Alternatively, authors should provide an additional
```
patrickhlauke:
```suggestion
						variant of the logo that has sufficient contrast. Alternatively, authors should provide an additional
```

understanding/21/non-text-contrast.html
```diff
@@ -401,6 +401,23 @@ <h4>Required for Understanding</h4>
 					<li> The information is available in another form, such as in a table that follows the graph, which becomes visible when a "Long Description" button is pressed.</li>
 					<li> The graphic is part of a logo or brand name (which is considered &quot;essential&quot; to its presentation).</li>
 				</ul>
+
+				<div class="note">
+					<p>Logos are exempted from contrast requirements when they are purely used as
+						<em>graphical objects</em>, under the assumption that they must comply with stricter color choices mandated
+            by corporate identity or brand guidelines.	When they act as <a>user interface components</a>
```
patrickhlauke:
```suggestion
            by corporate identity or brand guidelines. When they act as <a>user interface components</a>
```

understanding/21/non-text-contrast.html
```diff
@@ -401,6 +401,23 @@ <h4>Required for Understanding</h4>
...
+						(such as links or other interactive controls), authors must nonetheless make sure that there is an aspect of
+						the user interface components that allows users to identify the components and their state.
+						If possible, and if allowed by the relevant corporate identity or brand guidelines, authors should choose a
+						variant of the logo that has sufficient contrast. Alternatively, authors should provide an additional
+						visual element with sufficient contrast that helps identify the	user interface component or its state – such as
+						additional text, or an outline or border around the component. Lastly, consider providing an equivalent
+						<a>user interface components</a> which serves the same purpose and does meet the contrast requirements.</p>
+					<p>If logos are presented with an insufficient contrast, but their presentation was an author choice rather than
+						being mandated by corporate identity or brand guidelines, then that particular low contrast presentation is
+						<em>not</em> "essential",	and the logo is <em>not</em> exempt from the contrast requirements.</p>
```
patrickhlauke:
```suggestion
						<em>not</em> "essential", and the logo is <em>not</em> exempt from the contrast requirements.</p>
```

patrickhlauke:
Added an example of logos being shown with low contrast as an author choice, which does not exempt them as that presentation is then not "essential" but just an aesthetic decision of the designer/author

see the second note at the end of https://deploy-preview-4402--wcag2.netlify.app/understanding/non-text-contrast#required-for-understanding

fstrr:
I do like the idea of dealing with those common blocks of faded out logos, but am concerned that more than that might be going a little too far. For me, the second note would suffice. In the meantime, I have these questions:

> When they act as user interface components (such as links or other interactive controls), authors must nonetheless make sure that there is an aspect of the user interface components that allows users to identify the components and their state.

1. Do you have any examples of "an aspect"? It's pretty vague and, to use one of your examples from conversations on focus visible, as written, this could be a single pixel.
2. Is that a normative "must"?
3. Using the [Orange](https://travel.orange.com/) logo as an example, are you suggesting that the company cannot use its decades-old branded logo as a link to the homepage because the white and orange has a contrast of less than 3:1?
4. If so, what "aspect" might the company do here? A different logo for just the web site header?
5. I'm unsure why state comes into it. Are you suggesting that colors need to change on hover / focus?

-----

Finally, this feels like an important enough update to nor have it in two note sections.

patrickhlauke:
> 1. Do you have any examples of "an aspect"? It's pretty vague and, to use one of your examples from conversations on focus visible, as written, this could be a single pixel.

The same way that 1.4.11 doesn't currently get any more specific about which part of a graphic is necessary to understand it, this part is intentionally left vague. Trying to be more specific would introduce a lot more normative requirements here that...shouldn't happen.

> 2. Is that a normative "must"?

In my view, yes. Again, I contend that this was always present already in the normative wording when considering logos when they act as UICs.

> 3. Using the [Orange](https://travel.orange.com/) logo as an example, are you suggesting that the company cannot use its decades-old branded logo as a link to the homepage because the white and orange has a contrast of less than 3:1?

The "aspect" of the orange logo that is needed to recognise that there is something there that can be interacted with, as a UIC, is arguably the fact that the orange has sufficient contrast against the black header background. A user can at least perceive that there is something there. Now, if there were lots of other logos that are the same square shape, and then have very low contrast bits inside them to distinguish them, that might get trickier...

Also, in the specific case of the Orange logo there ... the UIC contains both that logo, and the big text "Orange Travel", so in this case, it's a moot point ... even if a user couldn't perceive that logo at all, the text next to it would still let them know that a) there is something there and b) what it is.

> 4. If so, what "aspect" might the company do here? A different logo for just the web site header?

They could also do something like a partial or full border around it, or similar "external" interventions.

> 5. I'm unsure why state comes into it. Are you suggesting that colors need to change on hover / focus?

Ah, good catch. I admit at this point, I can't remember where the "and their state" came from. That can be removed. EDIT: ah, state came straight from the normative wording of the SC: "Visual information required to identify [user interface components](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html#dfn-user-interface-component) and [states](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html#dfn-state)"

> Finally, this feels like an important enough update to nor have it in two note sections.

I may be using the "Note" (and its highlighting) here as a way to draw attention, rather than to give the impression that it's some aside / less important aspect. Happy to make this part of the regular text rather than a note.

understanding/21/non-text-contrast.html
```diff
@@ -401,6 +401,29 @@ <h4>Required for Understanding</h4>
 					<li> The information is available in another form, such as in a table that follows the graph, which becomes visible when a "Long Description" button is pressed.</li>
 					<li> The graphic is part of a logo or brand name (which is considered &quot;essential&quot; to its presentation).</li>
 				</ul>
+
+				<div class="note">
+					<p>Logos are exempted from contrast requirements when they are purely used as
+					<em>graphical objects</em>, under the assumption that they must comply with stricter color choices mandated
+					by corporate identity or brand guidelines. When they act as <a>user interface components</a>
+					(such as links or other interactive controls), authors must nonetheless make sure that there is an aspect of
+					the user interface components that allows users to identify the components and their state.
```
patrickhlauke:
```suggestion
					the user interface components that allows users to identify the components.
```

understanding/21/non-text-contrast.html
```diff
@@ -401,6 +401,29 @@ <h4>Required for Understanding</h4>
...
+
+				<div class="note">
+					<p>Logos are exempted from contrast requirements when they are purely used as
+					<em>graphical objects</em>, under the assumption that they must comply with stricter color choices mandated
+					by corporate identity or brand guidelines. When they act as <a>user interface components</a>
+					(such as links or other interactive controls), authors must nonetheless make sure that there is an aspect of
+					the user interface components that allows users to identify the components.
+					If possible, and if allowed by the relevant corporate identity or brand guidelines, authors should choose a
+					variant of the logo that has sufficient contrast. Alternatively, authors should provide an additional
+					visual element with sufficient contrast that helps identify the	user interface component or its state – such as
```
patrickhlauke:
```suggestion
					visual element with sufficient contrast that helps identify the user interface component  – such as
```

patrickhlauke:
@fstrr 

> Finally, this feels like an important enough update to nor have it in two note sections

Good point, I tried a different approach: an actual "Logos" section https://deploy-preview-4402--wcag2.netlify.app/understanding/non-text-contrast#logos

bruce-usab (approved):
I agree that 1.4.11 should not exempt logos, and this is a helpful set of improvements.

understanding/21/non-text-contrast.html
```diff
@@ -401,6 +401,29 @@ <h4>Required for Understanding</h4>
...
+				<p>Logos are exempted from contrast requirements when they are purely used as
+				<em>graphical objects</em>, under the assumption that they must comply with stricter color choices mandated
+				by corporate identity or brand guidelines. When they act as <a>user interface components</a>
+				(such as links or other interactive controls), authors must nonetheless make sure that there is an aspect of
+				the user interface components that allows users to identify the components.
+				If possible, and if allowed by the relevant corporate identity or brand guidelines, authors should choose a
+				variant of the logo that has sufficient contrast. Alternatively, authors should provide an additional
+				visual element with sufficient contrast that helps identify the user interface component  – such as
+				additional text, or an outline or border around the component. Lastly, consider providing an equivalent
+				<a>user interface components</a> which serves the same purpose and does meet the contrast requirements.</p>
```
adampage:
```suggestion
				<a>user interface component</a> which serves the same purpose and does meet the contrast requirements.</p>
```

understanding/21/non-text-contrast.html
```diff
@@ -401,6 +401,29 @@ <h4>Required for Understanding</h4>
...
+
+				<p>If logos are presented with an insufficient contrast, but their presentation was an author choice rather than
```
adampage, line 415:
```suggestion
				<p>If logos are presented with insufficient contrast, but their presentation was an author choice rather than
```

understanding/21/non-text-contrast.html
```diff
@@ -453,7 +473,9 @@ <h4>Essential Exception</h4>
 
 					<p>Graphical objects do not have to meet the contrast requirements when &quot;a particular presentation of graphics is essential to the information being conveyed&quot;. The Essential exception is intended to apply when there is no way of presenting the graphic with sufficient contrast without undermining the meaning. For example:</p>
 					<ul>
-						<li><strong>Logotypes and flags</strong>: The brand logo of an organization or product is the representation of that organization and therefore exempt. Flags may not be identifiable if the colors are changed to have sufficient contrast.</li>
+						<li><strong>Logotypes</strong>: The brand logo of an organization or product is the representation of that organization, and is often
```
patrickhlauke:
```suggestion
						<li><strong>Logos and logotypes</strong>: The brand logo of an organization or product is the representation of that organization, and is often
```

patrickhlauke:
As suggested at the last WCAG 2.x TF meeting, I've split out the more controversial part - when logos are also acting as user interface components, requiring that the UIC have a sufficiently contrasting element https://github.com/w3c/wcag/pull/5113.

Instead, in this PR, I added the best practice suggestion.

https://github.com/w3c/wcag/pull/4402/changes#diff-365e883aab3c4ffb86e62f1de22de162e3dc48ce1c181d2f790705b56a3bec17R408

understanding/21/non-text-contrast.html
```diff
@@ -401,6 +400,27 @@ <h4>Required for Understanding</h4>
...
+				</section>
+
+				<section>
+				<h4>Logos</h4>
+
+				<p>Logos are exempted from contrast requirements, under the assumption that they must comply with stricter color choices mandated
+				by corporate identity or brand guidelines. However, this can be problematic when logos act as <a>user interface components</a>
+				(such as a link or other interactive control). In these cases, as a best practice, authors should consider choosing a variant of
+				the logo that has sufficient contrast, if allowed by the corporate identity or brand guidelines.
+				Alternatively, consider providing an equivalent <a>user interface component</a> which serves the same purpose and meets
```
patrickhlauke:
```suggestion
				Alternatively, authors should consider providing an equivalent <a>user interface component</a> which serves the same purpose and meets
```

detlevhfischer:
To me it still feels a bit dubious to see the essential exception for logotypes be modified by the 1.4.11 requirement for UICs, a requirement that in the normative text stands separate from that that for "Graphical Objects". The normative text simply does not clarify if any essential exception granted to Graphical Objects will be null once it is also a UIC. The 1.4.11 understanding text explicitly mentions logos es essential and thus exempted from the contrast requirements for graphical objects. Saying logos once linked now fall under the UIC requirements, thus invalidating the exemption, is an argument that can be made, but to me this is non an unambiguous conclusion from reading the normative text. I believe this point simply was not sufficiently recognised and clarified at the time 1.4.11 was set free in WCAG 2.1.

patrickhlauke:
> Saying logos once linked now fall under the UIC requirements, thus invalidating the exemption

Nuance: it's not the that the logos themselves are not exempt. It's that the *UIC* is NOT exempt, so the UIC still needs to have something that makes it visible/perceivable. whether you do that by choosing a more contrasty version of the logo, or you add a border, or some extra contrasting text, or an underline, or at a stretch at least another UIC with sufficient contrast that does the same thing, is up to you as an author

also, your comment is probably more apt for the other PR #5113

PR #4403, opened 2025-05-18
Suggested additional note for focus-not-obscured (AAA)

patrickhlauke:
AAA version of the note that was added for AA in https://github.com/w3c/wcag/pull/4104


patrickhlauke:
@dbjorge you mentioned in our call that the quote bit needs to be updated "no part of the component is hidden" ... but that *is* the wording used in the AAA version https://www.w3.org/TR/WCAG22/#focus-not-obscured-enhanced ... so i think we're good with this?

understanding/22/focus-not-obscured-enhanced.html
```diff
@@ -23,6 +23,7 @@ <h2>Intent of Focus Not Obscured (Enhanced)</h2>
...
+  <p class="note">This criterion evaluates the focused <em>component</em>, rather than the focus <em>indicator</em>. The component itself does not include the focus indicator when checking that "no part of the component is hidden" - unless the focus indicator is inside the component, or focus is indicated by a change to the component itself. Although users benefit from both the component and the focus indicator (if external to the component) not being obscured when tracking the focus, for the purposes of this criterion only checking the component provides a clearer metric. However, if the focus indicator is obscured, it would likely fail <a href="focus-visible">2.4.7 Focus Visible</a>.</p>
```
mbgower: Slight wording changes to improve readability and to match the wording to the context of the AAA requirement. 
```suggestion
  <p class="note">This criterion evaluates the focused <em>component</em>, rather than the focus <em>indicator</em>. The component itself does not include the focus indicator when checking that "no part of the component is hidden" — unless the focus indicator is inside the component, or focus is indicated by a change to the component itself. Users benefit from both the component and the focus indicator (if external to the component) not being obscured when tracking the focus; however, for the purposes of this criterion, only checking the component provides a clearer metric. If the focus indicator is fully obscured, it would likely fail <a href="focus-visible">2.4.7 Focus Visible</a>.</p>
```

patrickhlauke: I don't think it makes sense to change that here. It matches the wording we already agreed on and used in the AA note https://github.com/w3c/wcag/pull/4104 and doesn't need changing for the AAA

mbgower: If you don't think it's more readable with the changes, okay. But if you think the mild editorial tweaks improve readability, maybe we should upgrade the AA at the same time. I think we _should_ definitely consider the final change for "fully", as this is only likely to cause a failure of 2.4.7 if the focus indicator is fully obscured, no?

patrickhlauke: the AA note didn't include "fully" https://github.com/w3c/wcag/pull/4104/files so yes that would need to be patched there too to keep them in line.

patrickhlauke: done the "fully" to both AA and AAA, tweaked the PR description accordingly

PR #4424, opened 2025-06-03
Update errata/21.html for previous publications

kfranqueiro:
This documents the errata that occurred since each of the 2 previous (not current) publications, i.e. the publications in September 2023 and December 2024.

This directly references commit hashes on the WCAG-2.1 branch and their respective dates (which occurred later), rather than referencing the original pull requests and the dates they were merged into main.

https://deploy-preview-4424--wcag2.netlify.app/errata/21

### DIFF
--- a/errata/21.html
+++ b/errata/21.html
@@ -52,7 +52,52 @@
       </section>
     </section>
 
-{%- assign trDate = "2024-12-12" -%}
+    {%- assign trDate = "2024-12-12" -%}
+    {%- capture trUrl -%}
+      https://www.w3.org/TR/{{ trDate | split: "-" | first }}/REC-WCAG{{ page.fileSlug }}-{{ trDate | replace: "-", "" }}/
+    {%- endcapture -%}
+    <section id="since-{{ trDate }}">
+      <h2>Errata since <a href="{{ trUrl }}">{{ trDate | date: "%d %B %Y" }} Publication</a></h2>
+      <section id="substantive-{{ trDate }}">
+        <h3>Substantive Errata</h3>
+        <p>No substantive errata have been recorded at present.</p>
+      </section>
+      <section id="editorial-{{ trDate }}">
+        <h3>Editorial Errata</h3>
+        <ul>
+          <li>
+            2025-03-20:
+            Making editorial changes to improve consistent use of the terms
+            "success criteria/criterion", "web", "website", and "web page"
+            ({% gh "d920ac2" %})
+          </li>
+          <li>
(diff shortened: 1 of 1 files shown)


mbgower (approved):

mbgower (dismissed):
I suggest we reduce the TOC to only the H2s (i.e., leave out the secondary entries for each publication date.
Screenshot 2025-06-09 at 9 14 41 AM

errata/21.html
```diff
@@ -51,6 +51,74 @@ <h3>Editorial Errata</h3>
...
+        </ul>
+      </section>
+    </section>
+
+    {%- assign trDate = "2023-09-21" -%}
+    {%- capture trUrl -%}
+      https://www.w3.org/TR/{{ trDate | split: "-" | first }}/REC-WCAG{{ page.fileSlug }}-{{ trDate | replace: "-", "" }}/
+    {%- endcapture -%}
+    <section id="since-{{ trDate }}">
+      <h2>Errata since <a href="{{ trUrl }}">{{ trDate | date: "%d %B %Y" }} Publication</a></h2>
```
bruce-usab, line 105: I suggest adding something like ", first publication as final W3C Technical Report" but I defer to your and Kevin's judgment on that.

bruce-usab (approved):
I am okay with PR as-is.
1. Some references are to GitHub PRs instead of commits.  Would it be easy (and desirable) for those to _all_ be commits?  This is mostly for the sake of consistency.
2. Is it not possible to pick a date for the last (i.e., bottom-most) item?
> In the Status of This Document, removing the first instance of the repeated paragraph beginning "This document has been reviewed by W3C Members...".

errata/21.html
kfranqueiro: I am confused by this comment:

- The 2023 republication, which this is commenting on, is not the first publication of the recommendation
- Is "Technical Report" distinct from "Recommendation"?

kfranqueiro:
> 1. Some references are to GitHub PRs instead of commits.  Would it be easy (and desirable) for those to _all_ be commits?  This is mostly for the sake of consistency.

My thought process is that linking to PRs is preferable when possible, because they provide more context. Typically this will only not be feasible for modern 2.1 backports (which in the most recent case I performed via a single PR but then intentionally rebase-merged rather than squashed), in which case the commit message does still include a link to the PR.

> 2. Is it not possible to pick a date for the last (i.e., bottom-most) item?

When I first migrated this page into the repo, that was the only change I was not able to trace to a particular commit.

PR #4431, opened 2025-06-05
Update error-identification.html

detlevhfischer:
Addition of a section on HTML error validation.

### DIFF
--- a/understanding/20/error-identification.html
+++ b/understanding/20/error-identification.html
@@ -47,7 +47,25 @@
          <li>the user enters alphabetic characters or parentheses into their phone number field that only accepts numbers;</li>
          <li>the user enters a bid that is below the previous bid or the minimum bid increment.</li>
       </ul>
-
+<h3>Note on using HTML form validation</h3>
+<p>With HTML form validation, the use of the <code>required</code> attribute on a field has the consequence that 
+     when trying to submit the form with this field empty, the user agent will display a generic text message that a value 
+  in the field is required, and set the focus to the first field missing required input.
+  It will also scroll the page to this field is in the viewport.</p> 
+  <p>In most common user agent and screen reader combinations, the screen reader will read the error message 
+    and the programmatic name of the required field. 
+    While this in principle meets the requirements of this Success Criterion, 
+    it should be noted that there are several disadvantages related to this approach:</p>
+     <ul>
+         <li>Depending on the user agent, the message may be non-permanent, or fail to scroll with the page</li>
+<li>Depending on the user agent, zoomed-in (magnified) content may display the error message unmagnified; 
+  the message may may be too small for users to read</li>
+<li>HTML validation messages for fields with type=text will be unspecific 
+  and will neglect to provide helpful suggestions where these may be needed </li>
+<li>If several errors are present, only the first error is exposed. 
+  Once the user has provided an input that conforms to the type of field and resubmits the form, 
+  the next error (if present) will be exposed. This means that repeated resubmissions and corrections may be required.</li>
+     </ul>
       <div class="note">
          <p>If a user enters a value that is too high or too low, and the coding on the page automatically
             changes that value to fall within the allowed range, the user's error would still


mbgower:
@patrickhlauke to re-check.

understanding/20/error-identification.html
```diff
@@ -47,7 +47,24 @@ <h2>Intent of Error Identification</h2>
...
+  in the field is required, and set the focus to the first field missing required input.
+  The user agent will generally also scroll the page so that the field in error and the generated error message will be visible in the viewport.</p> 
+  <p>In most common user agent and screen reader combinations, the screen reader will read the generic error message 
+    and the programmatic name of the required field. 
+    While this in principle meets the requirements of this Success Criterion, 
+    it should be noted that there are several disadvantages related to this approach:</p>
+     <ul>
+         <li>Depending on the user agent, the message may be non-permanent, or fail to scroll with the page</li>
+         <li>Depending on the user agent, zoomed-in (magnified) content may display the error message unmagnified; 
+  the message may may be too small for users to read</li>
```
patrickhlauke:
```suggestion
  the message may be too small for users to read</li>
```

understanding/20/error-identification.html
```diff
@@ -47,7 +47,24 @@ <h2>Intent of Error Identification</h2>
          <li>the user enters alphabetic characters or parentheses into their phone number field that only accepts numbers;</li>
          <li>the user enters a bid that is below the previous bid or the minimum bid increment.</li>
       </ul>
-
+<h3>Note on using native HTML form validation</h3>
+<p>With native HTML form validation, the use of the <code>required</code> attribute on a field has the consequence that
```
patrickhlauke: not just the `required` attribute. you can trigger browser-native form validation with other things like just `pattern`, `minlength`/`maxlength`, `min`/`max`

understanding/20/error-identification.html
```diff
@@ -47,7 +47,24 @@ <h2>Intent of Error Identification</h2>
          <li>the user enters alphabetic characters or parentheses into their phone number field that only accepts numbers;</li>
          <li>the user enters a bid that is below the previous bid or the minimum bid increment.</li>
       </ul>
-
+<h3>Note on using native HTML form validation</h3>
```
patrickhlauke: i'd drop the "Note on" part

understanding/20/error-identification.html
```diff
@@ -47,7 +47,24 @@ <h2>Intent of Error Identification</h2>
...
+     <ul>
+         <li>Depending on the user agent, the message may be non-permanent, or fail to scroll with the page</li>
+         <li>Depending on the user agent, zoomed-in (magnified) content may display the error message unmagnified; 
+  the message may be too small for users to read</li>
+         <li>HTML validation messages for fields with <code>type="text"</code> will be unspecific, i.e. they will not provide helpful suggestions where these may be needed</li>
+         <li>If several errors are present, only the first error is exposed. 
+  Once the user has provided an input that conforms to the type of field and resubmits the form, 
+  the next error (if present) will be exposed. This means that repeated resubmissions and corrections may be required.</li>
+     </ul>
       <div class="note">
```
patrickhlauke, line 50: with the current structure, this makes it look like the next two notes are "sub-notes" that belong under the native form validation section. To avoid this, it may be worth moving the new section *after* the second note here

patrickhlauke (changes requested):
suggest renaming the section, moving it after the two existing notes, and tweaking the language (as it's not just the `required` attribute that triggers these things).

may also be worth couching the description a bit more into "in most user agents" type language (though from testing, the statements here hold true for chrome, edge, firefox, safari, probably others, from my testing)

patrickhlauke:
@detlevhfischer let me know if you're happy for me to make some of my proposed more fundamental tweaks here for you, or if you want to do them yourself (don't want to just barge in and rewrite your PR without your approval)

detlevhfischer:
Very happy for you to step in!

patrickhlauke (approved):

patrickhlauke:
@detlevhfischer pushed some changes ... have a look and see if this still captures what you were trying to do

mbgower:
Note that "success criterion/criteria" is almost always lower case now. Please see our [style guide](https://github.com/w3c/wcag/wiki/WCAG-2-style-guide)

understanding/20/error-identification.html
```diff
@@ -21,7 +21,7 @@ <h2>In brief</h2>
    <section id="intent">
       <h2>Intent of Error Identification</h2>
 
-      <p>The intent of this success criterion is to ensure that users are aware that an error
+      <p>The intent of this Success Criterion is to ensure that users are aware that an error
```
patrickhlauke:
```suggestion
      <p>The intent of this success criterion is to ensure that users are aware that an error
```

understanding/20/error-identification.html
```diff
@@ -51,8 +51,8 @@ <h2>Intent of Error Identification</h2>
       <div class="note">
          <p>If a user enters a value that is too high or too low, and the coding on the page automatically
             changes that value to fall within the allowed range, the user's error would still
-            need to be described to them as required by the success criterion. Such an error description
-            telling the person of the changed value would meet both this success criterion (Error
+            need to be described to them as required by the Success Criterion. Such an error description
+            telling the person of the changed value would meet both this Success Criterion (Error
```
patrickhlauke:
```suggestion
            need to be described to them as required by the success criterion. Such an error description
            telling the person of the changed value would meet both this success criterion (Error
```

understanding/20/error-identification.html
```diff
@@ -61,7 +61,7 @@ <h2>Intent of Error Identification</h2>
          that user agents or assistive technologies can use to identify an error and provide
          error information to the user. For example, certain technologies can specify that
          the user's input must not fall outside a specific range, or that a form field is required.
-         This type of programmatic information is not required for this success criterion, but may be covered
+         This type of programmatic information is not required for this Success Criterion, but may be covered
```
patrickhlauke:
```suggestion
         This type of programmatic information is not required for this success criterion, but may be covered
```

understanding/20/error-identification.html
```diff
@@ -87,7 +107,7 @@ <h2>Benefits of Error Identification</h2>
             Providing information about input errors in text allows users who are blind, have low vision, or have color vision deficiency to perceive the fact that an error occurred.
          </li>
          <li>
-            This success criterion may help people with cognitive, language, and learning disabilities
+            This Success Criterion may help people with cognitive, language, and learning disabilities
```
patrickhlauke:
```suggestion
            This success criterion may help people with cognitive, language, and learning disabilities
```

understanding/20/error-identification.html
```diff
@@ -108,7 +128,7 @@ <h2>Examples of Error Identification</h2>
                the user which field or fields were missing or incorrect.</p>
 
             <div class="note">
-               <p>This success criterion does not mean that color or text styles cannot be used to indicate
+               <p>This Success Criterion does not mean that color or text styles cannot be used to indicate
```
patrickhlauke:
```suggestion
               <p>This success criterion does not mean that color or text styles cannot be used to indicate
```

patrickhlauke:
Apologies @mbgower .... removed my over-eager "capitalise ALL THE SUCCESS CRITERION mentions" faux-pas. @adampage this is now clean and ready for review. for reference, when I checked that the behaviour described here actually happens, i tested in the following combinations: windows chrome, firefox, edge, with JAWS and NVDA; ios/safari/VO; android/chrome/talkback; macos/safari/VO

detlevhfischer:
@patrickhlauke read through "Files changed", looks good to me!

gundulaniemann:
It does not become clear whether the method to use the user agent's ability leads to a conformant page. One hand it says: "While this meets the requirements of this Success Criterion, it should be noted that there are several disadvantages related to this approach", while further down it states that the error message might not show with enlarged text (which is a violation of text resizing), and further down it states that the user might need to resubmit while in the other sections of the understanding it clearly suggests that all errors should be displayed at once for convenient correction.

gundulaniemann:
To rephrase my last comment: I think the new text somehow contradicts itself concerning the point whether this approach leads to a conformant page or not.
Instead, it should be clearly stated whether it does.
In fact I think it doesn't.

understanding/20/error-identification.html
```diff
@@ -78,6 +78,26 @@ <h2>Intent of Error Identification</h2>
...
+         <li>Depending on the user agent, the message may not be permanent, or fail to scroll with the page.</li>
+         <li>Depending on the user agent, even if a user has zoomed-in (magnified) the content, the error messages will not appear magnified,
+            as the text in the validation message will be displayed at the same size as the user agent interface; the message may be too small for users to read.</li>
+         <li>The default HTML validation error messages are generally quite generic, and they may not provide sufficiently helpful or specific suggestions to the user
+            that would conform to <a href="error-suggestion">3.3.3 Error Suggestion</a>.
+         </li>
+         <li>If several errors are present, only the first error message is exposed; once the user has provided an input that conforms to the type of field,
+            and resubmits the form, the next error (if present) will be exposed. This means that repeated resubmissions and corrections may be required.</li>
+      </ul>
+
```
gundulaniemann: What about adding a sentence: "So, while leading to a formally compliant form, we do not recommend this method."

baldino-m:
I think that the browser native validation approach is not recommended, but if this violates another SC, it is still a pass for "Error identification". The only concern is about the back and forth with submissions-showing a single error-submit again-show another error, since the system is technically aware of all errors when the form is submitted the first time (and not showing all the messages for the identified errors), so it might represent a violation. However, since this is a native behaviour, we can discuss if this can be forgiven.

TestPartners:
> I think that the browser native validation approach is not recommended, but if this violates another SC, it is still a pass for "Error identification". 

The problem with native validation in Chrome and Firefox is that the error message can disappear or become dissociated from the form control that is in error. It therefore fails SC 3.3.1 even if the page only contains one form control.

patrickhlauke:
> The problem with native validation in Chrome and Firefox is that the error message can disappear

I don't think normatively it's required that error messages be persistent.

> or become dissociated from the form control that is in error

Is there an example of this? I've tried to replicate this with various zoom levels, scrolling around, etc, but it seems solid enough from my cursory testing. Is it an issue on mobile/small screen browsers, more than large screen/desktop ones?

TestPartners:
In Chrome, the error message disappears if you scroll the page such that the error message moves out of the viewport. The error message does not reappear when you scroll the page back to the original position, so there is no indication there was ever an error message.

In Firefox, the error message remains in the same place in the viewport when the page is scrolled, so the visual relationship between the form control and the error message is lost. If the page contains multiple form controls, it is impossible to know which one the error message is associated with.

I regard both behaviours as being non-conformant.

patrickhlauke:
ah, somehow missed the firefox behaviour. that's certainly a bug with firefox (which would impact "accessibility support" of that technique while it's not fixed)

understanding/20/error-identification.html
patrickhlauke:
```suggestion
    <p>As these problems relate to user agent behavior, developers will need to carefully consider if native browser validation is <a href="https://www.w3.org/TR/WCAG22/#dfn-accessibility-supported">accessibility supported</a>.</p>
```

bruce-usab:
Discussed on backlog call 10/24.  TF members on call concur that only concern is for bug in FireFox.  Moved to Ready for Review.

understanding/20/error-identification.html
```diff
@@ -78,6 +78,26 @@ <h2>Intent of Error Identification</h2>
...
 
+      <h3>User agent native HTML form validation</h3>
+
+      <p>When using native HTML <a href="https://html.spec.whatwg.org/multipage/forms.html#client-side-form-validation">client-side form validation</a>,
+         user agents will automatically prevent the submission of incomplete or invalid forms, and display generic error messages to the user.
+         The user agent will generally set focus back to the first form field that is in error, and as a result scroll the page
+         so that the field in error and the generated error message will be visible in the viewport.</p> 
+      <p>In most common user agent and screen reader combinations, the screen reader will announce the error message
+         and the programmatic name of the focused field. 
+         While this meets the requirements of this Success Criterion, it should be noted that there are several disadvantages related to this approach:</p>
```
mbgower:
```suggestion
         While this meets the requirements of this success criterion, it should be noted that there are several disadvantages related to this approach:</p>
```

PR #4440, opened 2025-06-11
Update abbreviation.html

### LINKED ISSUE #4436: Inconsistent indentation of Notes
In  [5.3.1 Required Components of a Conformance Claim](https://www.w3.org/TR/2024/REC-WCAG22-20241212/#conformance-required), Note 3 has a different indentation than Note 1 and Note 2.  Based on how Notes are treated in the previous section, it _might_ be the case that the formatting of Note 1 and Note 2 are the ones needing correction.  It might also be the case that all three notes are indented correctly, per the style guide.  Regardless, in context of that section, it looks very odd.  WCAG 2.1 has the same [formatting of this section](https://www.w3.org/TR/2025/REC-WCAG21-20250506/#conformance-required), and the items flagged below are also applicable to 2.1.

[Note 3 for abbreviation](https://www.w3.org/TR/2024/REC-WCAG22-20241212/#dfn-abbreviations) in the glossary I _think_ is indented correctly, but it looks odd in context.

The [Keyboard Interface definition](https://www.w3.org/TR/2024/REC-WCAG22-20241212/#dfn-keyboard-interface) is the only instance of an Example within a Note.  This seems like a mistake as other places in glossary intermix Notes and Examples.

In the glossary, _most_ example boxes have consistent formatting.  These do not use the word "example" _except_ for the leading label for the box.  It's very nice!  A minority of the  glossary terms use both the formatting and the label, but then include phrases like "for example" or "examples are".  These outliers are a little bit confusing (but _only_ a little bit) and should probably be made consistent with the others.

mbgower:
Closes #4436 

Removed note class on first note so that it is now part of the definition.

The numbering of notes appears to be done by code, so I have just left the other two notes to be autonumbered, although I maintain that note numbering should not span nested material, especially where a subsequent note is actually referring to the hierarchically higher material.

### DIFF
--- a/guidelines/terms/20/abbreviation.html
+++ b/guidelines/terms/20/abbreviation.html
@@ -5,7 +5,7 @@
       of the language
    </p>
    
-   <p class="note">This includes initialisms and acronyms where:</p>
+   <p>This includes initialisms and acronyms where:</p>
    
    <ol>
       
@@ -15,7 +15,7 @@
             syllables contained in that name or phrase
          </p>
          
-         <p class="note">Not defined in all languages.</p>
+         <p class="note">Initialisms are not defined in all languages.</p>
          
          <aside class="example"><p>SNCF is a French initialism that contains the initial letters of the <span lang="fr">Société Nationale des Chemins de Fer</span>, the French national railroad.
          </p></aside>


kfranqueiro:
Superseded by #4452 based on discussion.

PR #4452, opened 2025-06-16
Fix note structure in abbreviation definition

### LINKED ISSUE #4436: Inconsistent indentation of Notes

kfranqueiro:
Fixes #4436.

This fixes the markup for the first top-level note in the abbreviations definition to use a `div`, so that it can properly wrap the ordered list that is supposed to exist inside of it (as it did in [WCAG 2.0](https://www.w3.org/TR/WCAG20/#abbreviationsdef)).

This also relegates the nested note to an inline annotation in a paragraph (suggested by Alastair), to avoid needing to worry about an unimplemented nesting edge case.

The diff will be easier to view with whitespace ignored (`?w=1`).

@netlify /guidelines/#dfn-abbreviations

### DIFF
--- a/acknowledgements/ag-wg-active.html
+++ b/acknowledgements/ag-wg-active.html
@@ -17,7 +17,7 @@
 		<li>Thaddeus Cambron (Invited Expert)</li>
 		<li>Alastair Campbell (Nomensa)</li>
 		<li>Laura Carlson (Invited Expert)</li>
-    	<li>Sukriti Chadha (Invited Expert)</li>
+		<li>Sukriti Chadha (Invited Expert)</li>
 		<li>Rafal Charlampowicz (AccessibilityOZ)</li>
 		<li>Michael Cooper (W3C)</li>
 		<li>Jennifer Delisi (Invited Expert)</li>
@@ -87,7 +87,8 @@
 		<li>Kathleen Wahlbin (Invited Expert)</li>
 		<li>Léonie Watson (TetraLogical)</li>
 		<li>Jason White (Educational Testing Service)</li>
-    <li>White, Kevin (W3C Staff)</li>
+		<li>White, Kevin (W3C Staff)</li>
 		<li>Mark Wilcock (Unify Software and Solutions)</li>
+		<li>Todd Libby (Invited Expert)</li>
 	</ul>
 </section>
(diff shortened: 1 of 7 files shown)


bruce-usab (approved):

mbgower:
@alastc out of an abundance of caution, I am putting this into the quarterly errata column, pending a discussion with you.
We are making no normative change here -- it is all in Notes, and we've established those were properly formatted notes in 2.0.
But since it will be listed in the errata document (according to the table we have in https://github.com/w3c/wcag/wiki/WCAG-2-Task-Force-process#Errata-changes) I think it makes sense to include in the quarterly.

bruce-usab:
[diff between 2.0 and this PR](https://services.w3.org/htmldiff?doc1=https://www.w3.org/TR/WCAG20/&doc2=https://deploy-preview-4452--wcag2.netlify.app/guidelines/#dfn-abbreviations)

[diff between 2.2 and this PR](https://services.w3.org/htmldiff?doc1=https://www.w3.org/TR/2024/REC-WCAG22-20241212/&doc2=https://deploy-preview-4452--wcag2.netlify.app/guidelines/#dfn-abbreviations)

PR #4362, opened 2025-04-25
Add note about transaction-amount to 1.3.5 understanding

patrickhlauke:
Follow-up to https://github.com/w3c/wcag/pull/3539

Adds a note that tries to square the circle on why "transaction amount" is included.

Issue #3777, opened 2024-04-05
Input Purposes for User Interface Components is an appendix, so should be marked as non-normative


fstrr:
Discussed in the backlog task force meeting on 5th April, [Input Purposes for User Interface Components](https://www.w3.org/TR/WCAG22/#input-purposes) is an appendix, so should be [marked as non-normative](https://www.w3.org/TR/WCAG22/#interpreting-normative-requirements).

The trick to this seems to be adding a `class="informative"` attribute to the containing `section` element, but building the site locally doesn't create the homepage, so can't check this.

patrickhlauke:
you beat me there by 5 minutes...was just about to push up a PR 🏃

fstrr:
Well, you took that merge conflict off my hands, so I had a little time to look at something else

mbgower:
[As noted in the PR](https://github.com/w3c/wcag/pull/3539#issuecomment-2042989256), I have located the original official WG response showing that this was clearly intended to be an appendix.
Given the wording, 
>Introductory material, appendices, sections marked as "non-normative", diagrams, examples, and notes are [informative](https://www.w3.org/TR/WCAG22/#dfn-informative) (non-normative).

alastc:
My suggestion from the meeting:

Add appendix heading about change log, and mark those and appendix a/b/c.

Leave section 7 as it is (removing transction-amount separately), and it essentially is normative by that marking, removing the ambiguity.

dbjorge:
I maintain that *regardless of the original intent*, the fact that this section is incorporated by reference as a required part of specifying SC 1.3.5 means that either the section must be normative (unless we were to update SC 1.3.5 to say something like “for example” when referencing it)

GreggVan:
It looks like there might be a confusion between something "being" normative and something "containing" normative information.   

"Sections" cannot be normative.  Only the specific clauses that are required for conformance are normative.  

- the titles of provisions are not even normative  (only the requirement itself)
- notes are never normative (they should only clarify what is normative )
- examples are never normative -- and they should never fall outside of or expand the requirement language

What is normative?

- the text of the provision itself - including exceptions if and only if they are included in the provision language itself
- definitions
- the conformance section/requirements

bruce-usab:
[...]

PR #4458, opened 2025-06-20
Remove contradictory normative "must" from Section 7 preamble (first step to address `transaction-amount` issue in 1.3.5)

patrickhlauke:
* this is not a success criterion - it should not contain normative *must* statements that create specific "new" requirements by the backdoor
* the *must* here is contradicted in the sentence immediately following, where it starts with "Where applicable"?

This is related to https://github.com/w3c/wcag/pull/4362#issuecomment-2932418242

For info, this may also be relevant https://github.com/w3c/wcag/issues/3777

### DIFF
--- a/acknowledgements/ag-wg-active.html
+++ b/acknowledgements/ag-wg-active.html
@@ -17,7 +17,7 @@
 		<li>Thaddeus Cambron (Invited Expert)</li>
 		<li>Alastair Campbell (Nomensa)</li>
 		<li>Laura Carlson (Invited Expert)</li>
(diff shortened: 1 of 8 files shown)


bruce-usab (approved):
Briefly discussed on backlog call 6/30.

dbjorge:
I'm fine with this update, personally; even if I still disagree about calling this an "appendix", I do agree that it's not the thing that should be creating a "must" requirement like this, that's the SC's job.

hidde:
great catch, thumbed up!

Issue #4473, opened 2025-06-30
T1: Incorrect text highlighting


momdo:
https://www.w3.org/WAI/WCAG22/Techniques/text/T1#examples

As the screenshot below shows, some words are incorrectly identified as keywords of a programming language.

![Image](https://github.com/user-attachments/assets/4569d48c-0913-4fbb-9711-e206c3aba4e8)

patrickhlauke:
Good spot

PR #4475, opened 2025-06-30
Reformat T1, T2, T3 and suppress highlight.js

patrickhlauke:
Prevents unnecessary/unwanted `highlight.js` highlighting

Closes https://github.com/w3c/wcag/issues/4473

Note: I went for a "belt and braces" (intended redundancy) approach here with using `no-highlight language-html`.
In theory (from reading around the subject), the first one should be sufficient, but apparently not always the case, so the latter - while technically maybe not exact - also stops any highlighting (as there's no angle brackets anywhere in the preformatted blocks)

Preview links:

* https://deploy-preview-4475--wcag2.netlify.app/techniques/text/t1
* https://deploy-preview-4475--wcag2.netlify.app/techniques/text/t2
* https://deploy-preview-4475--wcag2.netlify.app/techniques/text/t3

### DIFF
--- a/.github/workflows/check-files.yaml
+++ b/.github/workflows/check-files.yaml
@@ -42,6 +42,13 @@ jobs:
             echo "::notice title=Mixed TR and informative docs changes::Please submit changes to TR space separately from changes to informative docs."
             echo "CHECK_FAILED=1" >> "$GITHUB_ENV"
           fi
+      - name: Check TypeScript
+        run: |
+          npm ci
+          if test $(npx tsc | wc -l) -gt 0; then
+            echo '::notice title=Branch contains TypeScript errors::Run `npm run check` locally for details.'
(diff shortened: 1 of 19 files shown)


patrickhlauke:
@kfranqueiro not sure why, but the preview build has been running for 8 minutes at this point, with no sign of concluding. might just be a temporary glitch, or something more fundamentally borked in the pipeline (I'll keep an eye on this and see if it eventually completes)

patrickhlauke:
@kfranqueiro as you were...after a few force pushes, it seems to have cleared whatever blockage there was in the preview build

patrickhlauke:
* https://deploy-preview-4475--wcag2.netlify.app/techniques/text/t1
* https://deploy-preview-4475--wcag2.netlify.app/techniques/text/t2
* https://deploy-preview-4475--wcag2.netlify.app/techniques/text/t3

PR #4506, opened 2025-07-28
add meta charset to all techniques

fstrr:
closes #4505

(1. also remove links to pre-current-build-system css files.
2. I started doing this manually, which was taking ages, so just did a find and replace for the rest. This has the benefits of me not adjusting the whitespace…)

PR #4513, opened 2025-07-31
Add missing meta charset and remove editors stylesheet at build time

### LINKED ISSUE #4505: missing `<meta charset>` in Techniques
We need to insert `<meta charset="UTF-8">` into each Techniques page.

kfranqueiro:
Fixes #4505.

This is an alternative to #4506 that resolves the same issues without changing source files, to avoid potential conflicts with other open PRs.

I diffed the build before and after this PR, and it results in effectively the same output as #4506, plus also catches a few instances under `understanding`.

I did update the technique and understanding template files, to try to avoid the same mistakes happening on future additions.


mbgower:
Check Language of Parts for potential issues resulting from this.

PR #4520, opened 2025-08-05
[Techniques] remove external highlight.js

momdo:
Remove unnecessary `<link>` and `<script>` tags.


patrickhlauke:
Are these the only techniques that currently include this for no useful reason?

kfranqueiro:
> Are these the only techniques that currently include this for no useful reason?

Yes, I just had a look. It's in exactly these 4 files, and otherwise `highlightAll` is already called from a script that is added to every page in the build.

patrickhlauke (approved):

Issue #771, opened 2019-06-07
Editorial change Example 5: Piano - 1.3.4: Orientation


jake-abma:
https://www.w3.org/WAI/WCAG21/Understanding/orientation.html#examples

Example 5 talk about room for "enough of the keys", what is enough of the keys? Seems to suggest if you have room for a certain number, it's fine, like 8 keys, and not for others?

> An example where orientation is essential could be a piano app that requires the device to be in landscape mode to allow room for enough of the piano keys to be functionally usable.

I guess we want them all to have enough room and probably wanted to say it this way:

> An example where orientation is essential could be a piano app that requires the device to be in landscape mode allowing the piano keys to have enough room to be functionally usable.

jake-abma:
#787 is closed but I do not see a change in Understanding docs.... do I miss something or look at the wrong place?

PR #787, opened 2019-06-16
Reword piano example

patrickhlauke:
Closes https://github.com/w3c/wcag/issues/771

PR #4528, opened 2025-08-11
1.3.4 Orientation - Update Example 5: Piano App

giacomo-petri:
Closes: #771

Example 5, replaced 

> An example where orientation is essential could be a piano app that requires the device to be in landscape mode to allow room for enough of the piano keys to be functionally usable.

with

> An example where orientation is essential could be a piano app that requires the device to be in landscape mode allowing the piano keys to have enough room to be functionally usable.


patrickhlauke:
a resurrection (of sorts) of https://github.com/w3c/wcag/pull/787 :)

understanding/21/orientation.html
```diff
@@ -62,7 +62,7 @@ <h2>Examples</h2>
 				<li><strong>Example 4: Check deposit in banking app</strong><br>
 				An example where orientation is essential could be a banking app that requires the device be in landscape mode to easily and accurately capture an image of a check for deposit. These paper forms are typically about twice as wide as they are high.</li>
 				<li><strong>Example 5: Piano app</strong><br>
-				An example where orientation is essential could be a piano app that requires the device to be in landscape mode to allow room for enough of the piano keys to be functionally usable. Since a piano app is emulating a physical piano keyboard that needs to retain relative physical characteristics between keys, either too few keys would be available, or the keys would be much too narrow.</li>
+				An example where orientation is essential could be a piano app that requires the device to be in landscape mode allowing the piano keys to have enough room to be functionally usable. Since a piano app is emulating a physical piano keyboard that needs to retain relative physical characteristics between keys, either too few keys would be available, or the keys would be much too narrow.</li>
```
patrickhlauke: i think it's ok to leave "to allow" here. otherwise, it may need a comma before "allowing", as otherwise it reads a bit stilted

giacomo-petri:
```suggestion
				An example where orientation is essential could be a piano app that requires the device to be in landscape mode, allowing the piano keys to have enough room to be functionally usable. Since a piano app is emulating a physical piano keyboard that needs to retain relative physical characteristics between keys, either too few keys would be available, or the keys would be much too narrow.</li>
```

Issue #2977, opened 2023-01-29
Technique H91's example 7b's code should be changed


ShamarYarde:
Example 7b of the [H91: Using HTML form controls and links](https://www.w3.org/WAI/WCAG22/Techniques/html/H91) technique currently reads as the following:

 `<textarea id="ta_1" title="Type your speech here" >Four score and seven years ago</textarea>`

I think it should read as the follows, because it's current value is "Four score and seven years ago" instead of the empty value that the preceding text is implying:

`<textarea id="ta_1" title="Type your speech here" ></textarea>`

PR #4539, opened 2025-08-12
Updated the description for the textarea example B H91

giacomo-petri:
Closes: https://github.com/w3c/wcag/issues/2977

Updated the description for the textarea example B so it accurately matches the current code.


bruce-usab:
Discussed on backlog call, needs double check against issue.

patrickhlauke:
@giacomo-petri while you're here, maybe also worth removing the `id` from line 138 https://github.com/w3c/wcag/pull/4539/files#diff-27009fe1d6b86fdb387c91f9ceb8ef9a102054ffeec99c677ef5f17b794f8eddR138 as it's not needed anymore there (example 7b in https://www.w3.org/WAI/WCAG22/Techniques/html/H91)

patrickhlauke:
(won't open a whole can of worms by wondering if it's kosher for us to have examples without visible labels that would possibly fail 3.3.2 Labels or Instructions)

mbgower:
@patrickhlauke to create new PR for ID and label issue.

Issue #3343, opened 2023-08-15
Word spacing in Understanding SC 1.4.12


xfq:
https://www.w3.org/WAI/WCAG22/Understanding/text-spacing.html#results

> In languages that do not have words (e.g. Japanese) applying word spacing had no effect.

Japanese is not a language that does not have words. Some languages do not use spaces, but still have the concept of a word. Chinese and Japanese rarely use spaces, but there is a concept of word in these two languages (although the definition of a word is subjective). For example, word boundaries are identified when users double-click the text. Heuristics and dictionaries are needed to identify the boundaries of words.

Also, spaces are occasionally used in these two languages in order to reduce the burden of reading for lower grade children, see https://www.w3.org/TR/clreq/#group-ruby for an example in Chinese.

It might be useful to change the text to something like:

> In languages that do not usually leave spaces between words [...]

giacomo-petri:
Note: The suggestion is part of the understanding doc rather than the normative content. I do not believe the ErratumRaised label is necessary.

PR #4542, opened 2025-08-12
Text spacing - Addressed the concern that Japanese does have words

giacomo-petri:
Closes: https://github.com/w3c/wcag/issues/3343

Addressed the concern that Japanese does, in fact, have words and may include spacing, rewording a bit that sentence.


patrickhlauke (approved):

bruce-usab:
Reviewed on backlog call.

Issue #3891, opened 2024-06-04
Technique G65: Providing a breadcrumb trail (position).


rscano:
Some customers place breadcrumbs at end of page. 
This is annoying because an user shall find it asap and not at end of page.
Shall we put a "shall" note on g95 page for request to make it reachable?

patrickhlauke:
best not to add RFC2119 "shall" in a non-normative technique. and ideally, couching it in more "we recommend"/"as a best practice" type language

PR #4545, opened 2025-08-13
G65 - breadcrumbs position

giacomo-petri:
Closes https://github.com/w3c/wcag/issues/3891

Adding a sentence to recommend positioning the breadcrumb near the top of the page.


giacomo-petri:
During our last TF meeting, we agreed to update the examples by using aria-labelledby to reference the h2 titled "You are here".

Honestly, I think this creates an unrealistic scenario, since breadcrumbs typically aren't visually labeled. The APG group, in fact, provides an example that uses aria-label instead: [APG Breadcrumb Example](https://www.w3.org/WAI/ARIA/apg/patterns/breadcrumb/examples/breadcrumb/).

In my opinion, since this PR is addressing a different problem, we should open a separate issue to discuss this. Anyway, I believe the best approach would be to remove the visible label and follow the APG pattern.

patrickhlauke:
agree with @giacomo-petri here. let's keep this PR tightly focused on just some extra advice, and revisit the fundamental technique in general separately

Issue #4215, opened 2025-02-04
Changing the test procedure for F1


JAWS-test:
[Technique F1: Failure of Success Criterion 1.3.2 due to changing the meaning of content by positioning information with CSS](https://www.w3.org/WAI/WCAG22/Techniques/failures/F1): 

> Remove the style information from the document or turn off use of style sheets in the user agent.

If I turn off CSS, this can have undesirable side effects, such as display:none content being shown or CSS-generated content being hidden, which has nothing to do with the F1 technique, but can still lead to incorrect test results. I therefore suggest changing the test procedure. Instead of turning off CSS, the reading order should be tested with CSS, e.g. with a screen reader or in the Dev Tools, Accessibility tab (i.e. checking the accessibility tree)

PR #4547, opened 2025-08-13
F1 update - Updated procedure and clarified scope

giacomo-petri:
Closes: https://github.com/w3c/wcag/issues/4215

With the goal of closing https://github.com/w3c/wcag/issues/4215 and prevent people from removing all styles, which can (and probably does) lead to invalid testing, I've clarified also the scope of the procedure. 

The key clarification is that the content in scope (which was not clear if it was the single item or the whole page) must be checked in relation to the surrounding page or context. Previously, it could be interpreted that only the “component” or "specific content" needed to be checked in isolation. The intent is to ensure that the reading sequence is correct within the full context of surrounding elements, not just internally within the component.

### DIFF
--- a/techniques/failures/F1.html
+++ b/techniques/failures/F1.html
@@ -117,10 +117,16 @@
       <section class="procedure"><h3>Procedure</h3>
          <p>For content which uses CSS for positioning: </p>
          <ol>
-            <li>Remove the style information from the document or turn off use of
-                                    style sheets in the user agent.</li>
-            <li>Check that the reading order of the content is correct and the
-                                    meaning of the content is preserved.</li>
+            <li>Identify the content elements that use CSS for positioning.</li>
+            <li>Check that the reading order of the content is correct and the meaning of the content is preserved in relation to the surrounding page or context, using one of the following methods:
+                <ul>
+                  <li>Code inspection: review the HTML code to determine the logical reading sequence.</li>
+                  <li>Accessibility tree inspection: examine the accessibility tree to confirm the reading order.</li>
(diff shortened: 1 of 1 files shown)


bruce-usab:
Discussed on TF call 9/19/2025 and moved to Ready for Approval.

bruce-usab (approved):
Preview [diff for Technique F1](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FTechniques%2Ffailures%2FF1&doc2=https%3A%2F%2Fdeploy-preview-4547--wcag2.netlify.app%2Ftechniques%2Ffailures%2Ff1): Failure of Success Criterion 1.3.2 due to changing the meaning of content by positioning information with CSS

techniques/failures/F1.html
```diff
@@ -117,19 +117,24 @@
...
+                </ul>
+            </li>
+           
          </ol>
       </section>
       <section class="results"><h3>Expected Results</h3>
          <ul>
-            <li>If check #2 is false, then this failure condition applies and the
-                                    content fails this Success Criterion.</li>
+           <li>If any options in check #2 is false, then this failure condition applies and the content fails the Success Criterion.</li>
```
brianelton: Minor grammatical error - "If any options in check # 2 is false" should either be "If any option in check # 2 is false" or "If any options in check # 2 are false"
(space between # and 2 intentional to avoid linking)

bruce-usab: Thank you @brianelton !  I change options to option.

Issue #3635, opened 2024-01-11
WCAG 2.2 Errata: wrong reference to understanding URL (WCAG21 instead o WCAG22)


rscano:
Inside WCAG 2.2 REC there are other wcag 2.1 understanding references:

[Accessibility supported definition](https://www.w3.org/TR/WCAG22/#dfn-accessibility-supported) 

note 1:
`<a href="https://www.w3.org/WAI/WCAG21/Understanding/conformance#support-level">Level of Assistive Technology Support Needed for "Accessibility Support"</a>`

note 5:
`<p class="">One way for authors to locate uses of a technology that are accessibility supported
      would be to consult compilations of uses that are documented to be accessibility supported.
      (See <a href="https://www.w3.org/WAI/WCAG21/Understanding/conformance#documented-lists">Understanding Accessibility-Supported Web Technology Uses</a>.) Authors, companies, technology vendors, or others may document accessibility-supported
      ways of using Web content technologies. However, all ways of using technologies in
      the documentation would need to meet the definition of accessibility-supported Web
      content technologies above.
   </p>`

[Conforming alternate version](https://www.w3.org/TR/WCAG22/#dfn-conforming-alternate-versions) (last paragraph)
`<p>See <a href="https://www.w3.org/WAI/WCAG21/Understanding/conformance#conforming-alt-versions">Understanding Conforming Alternate Versions</a></p>`

[Text alternative definition](https://www.w3.org/TR/WCAG22/#dfn-text-alternative) (note already in #3486 )
`<p class="">Refer to <a href="https://www.w3.org/WAI/WCAG21/Understanding/conformance#text-alternatives">Understanding Text Alternatives</a> for more information.
   </p>`

bruce-usab:
Mentioned on 1/19 call, needs input from Kevin White.

fstrr:
Take a look at this [PR from October 2023](https://github.com/w3c/wcag/pull/3481) where I've gone through and edited a lot of absolute links to make them relative. It's been there long enough that it's now got some merge conflicts that I think @alastc is trying to untangle.

WWJ30:
> Inside WCAG 2.2 REC there are other wcag 2.1 understanding references:
> 
> [Accessibility supported definition](https://www.w3.org/TR/WCAG22/#dfn-accessibility-supported)
> 
[...]

PR #4509, opened 2025-07-30
Reimplement associated techniques as data

kfranqueiro:
This replaces the HTML for each Understanding page's Techniques section with a data representation of the same information, to improve consistency, reduce likelihood of mistakes, and eventually enable restoring the structured `techniques` field to success criteria in the JSON export (see the first list item in #4393).

All associated techniques are now defined in `understanding/understanding.11tydata.js` under the `associatedTechniques` field, in a way that should make the simplest cases easier to define than they were before. [Documentation](https://github.com/w3c/wcag/blob/main/11ty/README.md#associated-techniques-data) is added in this PR. In addition, various common mistakes are checked for at runtime with specific error messages, and typings are also defined, enabling some amount of autocomplete and feedback in editors that support TypeScript, such as Visual Studio Code.

Most of the changes in this PR involve the build system and centralized templates; the only changes to the Understanding pages themselves are to use the new template rather than spell out the related techniques within HTML.

However, this _does_ visibly impact some pages within the informative docs. I have outlined notable changes to focus TF/WG review on below.

## Changes to Understanding pages

This list describes changes to the Techniques section within respective SC pages.

- Swaps out "technology-specific techniques below" phrasing seen in the following 2 pages, to standardize on the more common "the following techniques", since technology specificity is already indicated by the technique ID:
  - [1.2.2: Captions (Prerecorded)](https://deploy-preview-4509--wcag2.netlify.app/understanding/captions-prerecorded#techniques) (compare to [current 1.2.2](https://www.w3.org/WAI/WCAG22/Understanding/captions-prerecorded#techniques))
  - [4.1.2: Name, Role, Value](https://deploy-preview-4509--wcag2.netlify.app/understanding/name-role-value#techniques) (compare to [current 4.1.2](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value#techniques))
- Appends "techniques" in cases of "using one of the following" 
  - [1.2.3: Audio Description or Media Alternative (Prerecorded)](https://deploy-preview-4509--wcag2.netlify.app/understanding/audio-description-or-media-alternative-prerecorded#techniques) (compare to [current 1.2.3](https://www.w3.org/WAI/WCAG22/Understanding/audio-description-or-media-alternative-prerecorded#techniques))
[...]

PR #4564, opened 2025-08-19
Replace straggler WAI/WCAG21 links to WAI/WCAG22

patrickhlauke:
Closes https://github.com/w3c/wcag/issues/3635


patrickhlauke:
Ah, this PR currently fails the "Reflect status" check, but no idea what that is (and the error it shows in the log is not very helpful in that regard). is this the check that wants to avoid a mix of normative and non-normative files being changed in the same PR? and if so, should i bifurcate this PR into two separate ones? (and also, can the error messaging in the check https://github.com/w3c/wcag/actions/runs/17083042062/job/48441437816?pr=4564#step:9:7 be made a bit more human-understandable?)

mbgower:
@kfranqueiro is Patrick's analysis correct? Is this just flagging him on a mixed normative/non-normative PR?

kfranqueiro:
If you scroll to the top of the check and expand the Annotations section (or click Summary and scroll down which shows the same annotations), you will see the human-readable version:

> **Check for isolation of changes to TR space or informative docs**
> Please submit changes to TR space separately from changes to informative docs.

This isn't reflected as clearly within the log in order to allow running all of the checks at once rather than short-circuiting as soon as one fails. IIRC I also tried separating each to its own job but that proved less readable for other reasons.

patrickhlauke:
do we want/need to separate this PR into two then?

patrickhlauke:
Assuming so, done https://github.com/w3c/wcag/pull/4586

kfranqueiro:
Mentioned on backlog call on 5 Sep 2025: This PR's changes under `understanding` are to the Techniques subsection, which will be covered by #4509. Removing those from this PR will avoid conflicts later.

patrickhlauke:
@kfranqueiro reverted the changes to understanding

PR #4573, opened 2025-08-22
Replace raster images with SVG in Understanding SC 1.4.11: Non-Text Contrast

### LINKED ISSUE #4549: Some images in Understanding SC 1.4.11 are small and difficult to perceive
Some of the example images in [Understanding SC 1.4.11](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html) are small and difficult to perceive, particularly in narrow viewports.

A partial screencap of an example table from Understanding SC 1.4.11, in whic a very small image is annotated by an emoji of a face with a monocle and skeptical expression.

adampage:
Closes #4549.

Replaces many of the raster images in [Understanding Non-Text Contrast](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html) with inlined SVG assets. This gives the example visuals higher fidelity in varying viewport sizes, user zoom levels, and high-density displays.

Please compare [before](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html) and [after](https://deploy-preview-4573--wcag2.netlify.app/understanding/non-text-contrast).

⚠️ While doing this work, I also discovered that the current Understanding page has horizontal scrolling overflow in ~`320px` viewports because of the non-responsive tables like [“_Passing User Interface Component Examples_”](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html#:~:text=Passing%20User%20Interface%20Component%20Examples). I’ll create a separate issue & PR to see about resolving that. #4641


patrickhlauke:
alternatively...just use regular `` element pointing to the `.svg`s like I did here? might be more straightforward, perhaps https://github.com/w3c/wcag/pull/3058

kfranqueiro:
> * Okay to embed `svg` assets as partials?
> * The svg `<title>` elements are being overwritten with the page `<title>` on build?

#4575 fixes the title issue.

You should be able to shorten the includes from `understanding/21/img/...` to `./img/...`. (Interestingly, support for relative paths is immediately visible in [Eleventy's docs](https://www.11ty.dev/docs/languages/liquid/#supported-features), whereas I didn't see implicitly-relative-to-project-root paths mentioned, though they clearly work on your branch.)

> * Okay to organize assets in new subdirectories?

This one has some strings attached, but tl;dr I am inclined to say "no" for reasons explained below.

Currently, the build does not copy static assets in further subdirectories under `img`. Furthermore, it flattens all of `20/img`, `21/img`, `22/img` to a top-level `img` directory in the output (behavior carried over from the XSLT build, which the pages have been written to expect).

I could fix the build to also copy subdirectories, but it is still going to fully flatten, so it will still be a good idea to have a distinguishable prefix on filenames. (I see that some of the SVG files in the PR start with `ntc` for non-text contrast, but not all of them do.) I'd be a bit hesitant to encourage subdirectories when it could hide this detail and cause surprises in the build output.

Because you are inlining the SVGs, this isn't really a problem for this PR, but it would be something to keep in mind in general for files under `img`.

> * Outline text as `<path>` or leave as `<text>`?

Discussed on backlog call on August 22; definitely +1 to avoiding the bloat that paths would add, but some checking will need to be done to ensure text remains within expected boundaries, etc.

> * Where to put new styles?

`base.css` (where you have it) seems like the right place.

adampage:
Wow, thanks so much for #4575 and your incredibly speedy review, @kfranqueiro.

I’ll put the new `svg` files back in the same directory as the `png` files they’ll be replacing, and I‘ll also incorporate a distinguishable prefix in their filenames. 

I’ll also do some more benchtesting to make sure the `<text>` elements work well in different browsing environments before my next push.

Onward and upward!

iadawn:
Looks good to me. Thanks for the work @adampage

adampage:
> Looks good to me. Thanks for the work @adampage

Thanks, @iadawn! I re-illustrated the [source artwork in Figma](https://www.figma.com/design/fGNqnfK5uRXxRzX6idzCWv/Understanding-example-images?node-id=3-48&t=m2siKquhxvYVXzMn-1) and wrote a custom Figma plugin to sanitize the exported SVG code and to structure each image’s alt text inside a `<title>` layer. Does the W3C have any sort of asset management for assets like these? I’d be glad to migrate the source so that other contributors will have access.

understanding/20/contrast-enhanced.html
```diff
@@ -236,7 +236,7 @@ <h3>Inactive User Interface Components</h3>
 
          <p>User Interface Components that are not available for user interaction (e.g., a disabled control in HTML) are not required to meet contrast requirements. An inactive user interface component is visible but not currently operable. An example would be a submit button at the bottom of a form that is visible but cannot be activated until all the required fields in the form are completed.</p>
          <figure id="figure-grey-button-and-text">
-            <img src="img/inactive-button.png" alt="Grey button with non-contrasting grey text." width="120">
+            {% include "../21/img/1.4.11-ntc-button-inactive.svg" %}
```
adampage, line 239: Heya @kfranqueiro, I found references to one of the updated assets in 2 other Understanding docs from WCAG 2.0: Contrast Enhanced and Contrast Minimum.

I went ahead and updated those to point explicitly to the `../21/img` directory. Is that okay?

mbgower:
@kfranqueiro to take a look and move to Ready for Approval if fine.

understanding/20/contrast-enhanced.html
kfranqueiro: Discussed on Backlog call on 26 Sep, yes, this is fine; the version subfolders are purely an artifact of how the files were first organized, and crossing subfolders is not an issue.

kfranqueiro:
I reviewed and see no problems with the SVG replacements themselves.

However, I'm not sure how I feel about deferring the reflow problem to a separate issue, because I notice that this PR makes it significantly worse. Before, horizontal scrolling would be incurred around 412px; on this branch, it is incurred around 632px, which I would expect to affect more viewports.

I presume this is because the `svg` elements have explicit widths, whereas the previous `img` elements were allowed to scale down. In its current state, this PR doesn't feel like it's closing #4549 so much as trading one problem for the other (#4641).

Is there a reasonable way to avoid making the reflow problem worse in this PR?

css/base.css
```diff
@@ -444,6 +444,11 @@ figure svg {
 	border: 1px solid var(--line-grey);
 }
 
+figure svg[role*="image"] {
+    display: block;
+    max-inline-size: 100%;
+}
```
adampage, line 451: > Is there a reasonable way to avoid making the reflow problem worse in this PR?

A screen cap of the production Understanding Non-Text Content page showing Chrome’s open DevTools and an annotation highlighting the max-width: 100% rule that is in effect for all image elements within the main landmark.


@kfranqueiro, you bet. The current `png` images on production are scaling down because of ☝🏻 this `max-inline-size: 100%` rule. A quick workaround would be to extend that rule to the `svg` elements. What’s odd is I can’t find where in the repo that style rule is getting set. 🤔

In the near term, we could just add the selector to the `figure svg[role*="image"]` rule I’ve already added to `base.css` in this PR, since the `max-inline-size: 100%` property will achieve the same outcome:

```suggestion
main table svg[role*="image"],
figure svg[role*="image"] {
    display: block;
    max-inline-size: 100%;
}
```

What do you think?

kfranqueiro: Oh, that img style is inherited from the main WAI stylesheet which is outside of this repo. So yes, if you're okay with expanding the selector in this PR at least for the time being until tackling #4641, that would probably be fine.

adampage: Absolutely — done. 👍🏻

Issue #3376, opened 2023-09-07
Minimum bounding box and line wrapping - unclear definition


dd8:
It's unclear what the following sentence means for line wrapped text: 

>For components which wrap onto multiple lines as part of a sentence or block of text (such as hypertext links), the bounding box is based on how the component would appear on a single line.

https://www.w3.org/TR/WCAG22/#dfn-bounding-boxes

Does this mean 

a) It's calculated as if the text is truncated at the line wrap - the red box below

<svg width=300 hight=100>
<rect x=0 y=0 width=50 height=20 fill="#eee"></rect>
<text x=0 y=15 width=50 height=20>1 of 3</text>
<rect x=60 y=0 width=50 height=20 fill="#eee"></rect>
<text x=60 y=15 width=50 height=20>2 of 3</text>
<rect x=120 y=0 width=10 height=20 fill="#f00"></rect>
<text x=120 y=15 width=10 height=20>3</text>
<rect x=0 y=22 width=40 height=20 fill="#eee"></rect>
<text x=0 y=37 width=40 height=20>of 3</text>
</svg>

a) It's calculated as if `white-space: nowrap` is specified and the text overflows into the wider red box:

<svg width=300 hight=100>
<rect x=0 y=0 width=50 height=20 fill="#eee"></rect>
<text x=0 y=15 width=50 height=20>1 of 3</text>
<rect x=60 y=0 width=50 height=20 fill="#eee"></rect>
<text x=60 y=15 width=50 height=20>2 of 3</text>
<rect x=120 y=0 width=50 height=20 fill="#f00"></rect>
<text x=120 y=15 width=10 height=20>3</text>
<rect x=0 y=22 width=40 height=20 fill="#eee"></rect>
<text x=0 y=37 width=40 height=20>of 3</text>
</svg>

alastc:
Discussion from the WCAG 2.x backlog meeting:

- This shouldn't come up often due to the Inline exception.
- Our interpretation is (b) from the original post. 
- It would be useful to add something to the understanding doc for this.

dd8:
If the interpretation is b) doesn't that mean the center of the bounding box could be off the edge of the device? (i.e. if the wrap was caused by hitting the edge of the viewport).

Maybe I''m missing something, but seems more likely folks would try and hit the center of the box [3] in interpretation a) since they would have no idea where the box center is in interpretation b)

mbgower:
**Draft Working Group Response**

The bounding box definition reads as follow:

[...]

PR #4584, opened 2025-08-25
Target size update

mbgower:
Partially addresses discussion emerging from #3376

### DIFF
--- a/guidelines/sc/22/target-size-minimum.html
+++ b/guidelines/sc/22/target-size-minimum.html
@@ -8,7 +8,7 @@
     <p>The size of the <a>target</a> for <a>pointer inputs</a> is at least 24 by 24 <a>CSS pixels</a>, except when:</p>
     <dl>
         <dt>Spacing</dt>
-        <dd>Undersized targets (otherwise unexcepted targets less than 24 by 24 CSS pixels) are positioned so that if a 24 CSS pixel diameter circle is centered on the <a>bounding box</a> of each, the circles do not intersect another target or the circle for another undersized target;</dd>
+        <dd>Undersized targets (those less than 24 by 24 CSS pixels which are not excepted under other bullets) are positioned so that if a 24 CSS pixel diameter circle is centered on the <a>bounding box</a> of each, the circles do not intersect another target or the circle for another undersized target;</dd>
         <dt>Equivalent</dt> 
         <dd>The function can be achieved through a different control on the same page that meets this criterion;</dd>
         <dt>Inline</dt> 
--- a/guidelines/sc/22/target-size-minimum.html
+++ b/guidelines/sc/22/target-size-minimum.html
@@ -8,7 +8,7 @@
     <p>The size of the <a>target</a> for <a>pointer inputs</a> is at least 24 by 24 <a>CSS pixels</a>, except when:</p>
     <dl>
         <dt>Spacing</dt>
-        <dd>Undersized targets (those less than 24 by 24 CSS pixels) are positioned so that if a 24 CSS pixel diameter circle is centered on the <a>bounding box</a> of each, the circles do not intersect another target or the circle for another undersized target;</dd>
+        <dd>Undersized targets (otherwise unexcepted targets less than 24 by 24 CSS pixels) are positioned so that if a 24 CSS pixel diameter circle is centered on the <a>bounding box</a> of each, the circles do not intersect another target or the circle for another undersized target;</dd>
         <dt>Equivalent</dt> 
         <dd>The function can be achieved through a different control on the same page that meets this criterion;</dd>
         <dt>Inline</dt> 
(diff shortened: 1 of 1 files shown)


patrickhlauke:
This does get tricky though, because the other target may be excepted not just under the inline clause, but any of the other ones. So for example, you may have a target (target 1) that is too small, but lacks enough spacing to another target - and this second target (target 2) is also too small, but this one has an *equivalent*. so, while it's fine that THAT one (target 2) has an equivalent, this still doesn't solve the fact that our original target (target 1) is too small and too close to the other target (target 2).

mbgower:
> This does get tricky though, because the other target may be excepted not just under the inline clause, but any of the other ones. So for example, you may have a target (target 1) that is too small, but lacks enough spacing to another target - and this second target (target 2) is also too small, but this one has an equivalent. so, while it's fine that THAT one (target 2) has an equivalent, this still doesn't solve the fact that our original target (target 1) is too small and too close to the other target (target 2).

But if there is sufficiently sized equivalent target to a small target 2 , it doesn't matter if you can't hit target 2; that's the whole point of excepting a small target with an equivalent.  

This is the case for all the excepted items. They are beyond either the author's ability or need to control. 
If you go back in the history of this SC's language, the spacing was part of the normative wording (but read very awkwardly). That is, the target was assessed for size or spacing _unless_ it was excepted. In an attempt to make it more readable, we moved the spacing in as a bullet, but the clear intent was that you assessed targets for size (or spacing)  _except_ for inline, user agent, redundant (equivalent) or essential.

> our original target (target 1) is too small and too close to the other target (target 2).

Also, we _are_ assessing spacing for target 1. It will be spaced sufficiently  (e.g., its circle space check will not intersect with target 2).

patrickhlauke:
> But if there is sufficiently sized equivalent target to a small target 2 , it doesn't matter if you can't hit target 2; that's the whole point of excepting a small target with an equivalent.

but target 1 is also small, but your edit suggests says that spacing can't take target 2 into account because THAT one is excepted already...and target 1 is just small, and has no equivalent, etc

EDIT: ah, hang on, i seem to have misread the edit you're proposing... thought your parenthetical was referring to which targets to check against. ignore me...

EDIT 2: well, re-re-reading it...I think my original confusion is due to the fact that the wording almost seems to suggest that targets exempted under any other bullet are excluded from the list of comparisons. because it may not be obvious that you need to perform the check for two targets "both ways". once, considering target 1. then, considering target 2. it's not just one single comparison (in which case exempting one would obviously make it pass for both of them). think it's down to the overall awkward original wording...

patrickhlauke:
maybe a more convoluted, but more explicit rewrite that avoids some potential confusion (though it does become admittedly unwieldy):

> Each undersized target (those less than 24 by 24 CSS pixels) which does not already come under any of the other exceptions for this criterion is positioned so that if a 24 CSS pixel diameter circle is centered on the <a>bounding box</a> of each target, the circle for the target does not intersect any other target or its circle;

moving the exception bit out of the parenthetical removes the suggestion that the "undersized target" definition includes that stipulation

guidelines/sc/22/target-size-minimum.html
```diff
@@ -8,7 +8,7 @@ <h4>Target Size (Minimum)</h4>
     <p>The size of the <a>target</a> for <a>pointer inputs</a> is at least 24 by 24 <a>CSS pixels</a>, except when:</p>
     <dl>
         <dt>Spacing</dt>
-        <dd>Undersized targets (those less than 24 by 24 CSS pixels) are positioned so that if a 24 CSS pixel diameter circle is centered on the <a>bounding box</a> of each, the circles do not intersect another target or the circle for another undersized target;</dd>
+        <dd>Undersized targets (those less than 24 by 24 CSS pixels which are not excepted under another bullet) are positioned so that if a 24 CSS pixel diameter circle is centered on the <a>bounding box</a> of each, the circles do not intersect another target or the circle for another undersized target;</dd>
```
mbgower, line 11:
```suggestion
        <dd>Each undersized target (those less than 24 by 24 CSS pixels) which does not already come under any of the other exceptions for this criterion is positioned so that if a 24 CSS pixel diameter circle is centered on the bounding box, the circle for the target does not intersect any other target or its circle;</dd>
```

patrickhlauke:
Discussing this in today's backlog, one possible proposal I suggested, which seems to have potential:

* move the "Spacing" exemption to the end (after "Essential") to make it clear that it's the last resort exemption
* start the prose of the exemption with something like "If the target does not already fall under any other exemptions, ..." and then just have the existing wording (minus the parenthetical "smaller than 24x24" because you wouldn't even BE in that list of exemptions if it wasn't, so it's redundant)

Will try to make an actual commit to this PR branch if that's ok, @mbgower (as this is a bit more involved than making an in-situ correction, and would get messy trying to do in the GitHub interface itself)

patrickhlauke:
Also, if you don't mind, I've changed this to a "Draft" issue here for now, as we're still panel-beating the structure of the change

guidelines/sc/22/target-size-minimum.html
detlevhfischer: Hmm. I dont see the point of adding "which are not excepted under another bullet"  because it is a common format in amny SCs to first spell out the general requirement and then list exceptions to which it does not apply. I find this additinon more confusing than helpful.

patrickhlauke: see https://github.com/w3c/wcag/pull/4584#issuecomment-3237590961

Issue #3864, opened 2024-05-15
incorporate suggestions from restoration of Target Size (Enhanced) Understanding


mbgower:
During the restoration of accidentally removed content from the Understanding document, https://github.com/w3c/wcag/pull/3638 a number of people suggested improvements to the content.
We kept the scope to only restoring existing content, but this issue is intended to capture those suggestions so that an improved Understanding doc can be created.

PR #3638, opened 2024-01-15
Restore Target Size (Enhanced) Understanding Doc content

giacomo-petri:
Closes: #3617

This PR restores content in the 2.5.5 Target Size (Enhanced) understanding document that was lost during the transition from 2.1 (when it was just "Target Size") to 2.2 (after the 2.5.8 Minimum version was added, and 2.5.5 become "Enhanced").
Specifically, it adds back:
- exceptions in the intent section
- additional examples

Some code indentations were also modified to improve code readability. These do not affect the resulting HTML document.

The resulting Understanding document (without CSS styling) can be viewed [here](https://raw.githack.com/w3c/wcag/giacomo-petri-target-size-enhanced-patch/understanding/21/target-size-enhanced.html)

PR #4589, opened 2025-08-28
Incorporate suggestions from restoration of Target Size (Enhanced) Understanding

filippo-zorzi:
Closes https://github.com/w3c/wcag/issues/3864

Adding the suggestion presented in this PR https://github.com/w3c/wcag/pull/3638 by @detlevhfischer, @frankie-wolf and @bruce-usab.


understanding/21/target-size-enhanced.html
```diff
@@ -17,19 +17,19 @@ <h2>In brief</h2>
...
+      <p><strong>Inline:</strong> Content displayed is often designed to reflow based on the screen width available. This is known as responsive design and makes it easier to read since you do not need to scroll both horizontally and vertically. In reflowed content, the targets can appear anywhere on a line and can change position based on the width of the available screen. Since targets can appear anywhere on the line, the size cannot be larger than the available text and spacing between the sentences or paragraphs, otherwise the targets could overlap. It is for this reason targets which are contained within one or more sentences are excluded from the target size requirements.</p>
```
patrickhlauke: while this is original text, i'm not a super-fan of the awkward "Content displayed". wonder if just dropping "displayed" would make this sound less clunky. or, if it must retain the distinction (so it's not confused with audio content, for instance), maybe changing it to "Visible content" or even at the very least "Displayed content"

understanding/21/target-size-enhanced.html
```diff
@@ -17,19 +17,19 @@ <h2>In brief</h2>
...
+      <p>The issue can be further complicated for responsive and mobile sites that need to accommodate different types of fine and coarse inputs (for example, a site that can be accessed on a traditional desktop or laptop with a mouse, as well as on a tablet or mobile phone with a touch screen).</p>
       <p>While this criterion defines a minimum target size, it is recommended that larger sizes are used to reduce the possibility of unintentional actions. This is particularly relevant if any of the following are true:</p>
   	  <ul>
         <li>the control is used frequently;</li>
         <li>the result of the interaction cannot be easily undone;</li>
         <li>the control is positioned where it will be difficult to reach, or is near the edge of the screen;</li>
         <li>the control is part of a sequential task.</li>
   	  </ul>
-      <p>The targets on a screen can have different purposes and uses, and this success criterion specifies how each is to be handled.</p>
+      <p>The targets on a screen can have different purposes and uses, and this Success Criterion specifies how each is to be handled.</p>
```
mbgower: we do not put this in upper case when used generically
```suggestion
      <p>The targets on a screen can have different purposes and uses, and this success criterion specifies how each is to be handled.</p>
```

understanding/21/target-size-enhanced.html
mbgower: The inline description in https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum is significantly different, and doesn't have this awkward phrase. As a lite change, I suggest a slight riff on its first sentence:
```suggestion
      <p><strong>Inline:</strong> The success criterion does not apply to inline targets in sentences or blocks of text. Content is often designed to reflow based on the screen width available. This is known as responsive design and makes it easier to read since you do not need to scroll both horizontally and vertically. In reflowed content, the targets can appear anywhere on a line and can change position based on the width of the available screen. Since targets can appear anywhere on the line, the size cannot be larger than the available text and spacing between the sentences or paragraphs, otherwise the targets could overlap. It is for this reason targets which are contained within one or more sentences are excluded from the target size requirements.</p>
```

mbgower:
@filippo-zorzi can you please check for any rationale for why example 3 was removed?

filippo-zorzi:
In the PR [#3638](https://github.com/w3c/wcag/pull/3638) @bruce-usab made a comment on the third esample:

> I do not agree that the ability to scroll is justification for small targets. I am open to be swayed. I think hypertext typically is excepted by being part of a sentence. If the target is part of in-page menu, would it not need to meet target size?

Since I wanted to bring over all the comments from the other PR, I had to find a way to highlight and remember the needed revision of the third example.

For sure this method is not the proper way to bring over a discussion given the fact that I forgot about that on last Friday meeting.

That commit is made to be changed.

bruce-usab (approved):

understanding/21/target-size-enhanced.html
patrickhlauke: this was thumbsed-up last time, but then the convo here was resolved without committing. do we want to or not?

PR #4604, opened 2025-09-09
Converge on default pre > code highlight.js usage

kfranqueiro:
## Background

When highlight.js was first added to this repository, and from that point onward, it has used the selector `pre`, deviating from the default of `pre code`. Despite that, most of our usage has indeed followed the `<pre><code>` pattern.

Our customized configuration has a few adverse effects:

- Forcing the syntax via `<pre><code class="language-...">` _does not work_ (see example in [C8](https://www.w3.org/WAI/WCAG22/Techniques/css/C8#the-html) - if you look at the source, you'll see `class="language-html"`, but if you inspect the DOM, you'll see that highlight.js applied `language-css`)
- Instances that _only_ use `<pre>` (without nested `<code>`) with the intent of not being highlighted get processed anyway unless they explicitly add `class="no-highlight"`
  - Examples where this was intentionally worked around: [T1](https://www.w3.org/WAI/WCAG22/Techniques/text/T1), [T2](https://www.w3.org/WAI/WCAG22/Techniques/text/T2), [T3](https://www.w3.org/WAI/WCAG22/Techniques/text/T3)
  - Examples where this was overlooked: [F33](https://www.w3.org/WAI/WCAG22/Techniques/failures/F33#example-1) (incorrectly highlighted as CSS), [F72](https://www.w3.org/WAI/WCAG22/Techniques/failures/F72#examples) (incorrectly highlighted as Markdown, which also sporadically introduces italics)

## Changes

This PR...

- Removes our customization of the highlight.js config
- Adds `<code>` in cases where it is warranted but was missing
  - This PR intentionally does not change e.g. F33 and F72 mentioned above, which are now fixed by resetting the configuration
- Removes `no-highlight` and the `<code>` tag in its entirety in T1, T2, and T3, which don't need highlight.js to run at all
- Copies and fixes references to the highlight.js script and theme for Understanding documents, so it will work for the code block in Understanding Conformance

Note: Silverlight and Flash techniques use only `<pre>`, but I left them alone, since they are all obsolete, and our build of highlight.js does not support e.g. C# and ActionScript anyway.

### DIFF
--- a/techniques/html/H46.html
+++ b/techniques/html/H46.html
@@ -37,7 +37,7 @@ obsoleteSince: 20
 &lt;noembed&gt;
   &lt;img alt="Still from Movie" src="moviename.gif" 
     width="100" height="80" /&gt;
-&lt;/noembed&gt;;</code></pre>
+&lt;/noembed&gt;</code></pre>
       </section>
    </section><section id="tests"><h2>Tests</h2>
       <section class="procedure"><h3>Procedure</h3>


techniques/html/H46.html
```diff
@@ -21,23 +21,23 @@
...
 							     <code class="el">noembed</code> is provided beside an <code class="el">embed</code>
 						   </h3>
-         <pre xml:space="preserve">&lt;embed src="moviename.swf" width="100" height="80"
+         <pre><code>&lt;embed src="moviename.swf" width="100" height="80"
   pluginspage="http://example.com/shockwave/download/" /&gt;
 &lt;noembed&gt;
   &lt;img alt="Still from Movie" src="moviename.gif" 
     width="100" height="80" /&gt;
-&lt;/noembed&gt;;</pre>
+&lt;/noembed&gt;;</code></pre>
```
fstrr:
```suggestion
&lt;/noembed&gt;</code></pre>
```

PR #4610, opened 2025-09-12
Remove unnecessary class attributes from code elements in TR space

kfranqueiro:
These classes were added in #3380, but they do not do anything, both because inline `code` elements (not inside `pre`) are never touched by highlight.js, and because highlight.js is not present in the normative document.

This PR causes no visible change to the document.


mbgower:
@mbgower to check whether this needs to go in as an erratum

mbgower:
Based on discussions, I have updated our process document, adding a section, which reads:
> **Normative invisible bug fix**: Any proposed change to the TR space that does NOT cause any visible change to the wording or presentation of the specification itself is tagged as both "Bug Fixes" and "Normative". However, it is not tagged with "ErratumRaised", because the change is not presented to users. As such it does not need to be included in the errata list for WCAG, and these changes can be incorporated without delay. These are fairly rare, and almost exclusively motivated by resolving technical debt.

As such, I am moving this PR out of the Errata for Quarterly Review column and placing it in Done.

PR #4611, opened 2025-09-12
Remove unnecessary code class and pre xml:space attributes from informative docs

kfranqueiro:
This removes:

- All `xml:space` attributes from `pre` elements, which are no longer needed since the XSLT build system was retired
- All `att`, `el`, `obj`, `prop`, and `propval` classes from `code` elements, which don't have any associated styles
- All `language-*` classes from inline `code` elements, which have no effect because highlight.js does not run on them
- Most `language-*` classes on `code` elements inside `pre`, except where highlight.js guesses incorrectly without it

This fixes the following highlight.js instances, which had been manually overridden to the wrong values:

- `html` incorrectly overridden to `css` in [C35](https://www.w3.org/WAI/WCAG22/Techniques/css/C35) (2x), [C36](https://www.w3.org/WAI/WCAG22/Techniques/css/C36), [F2](https://www.w3.org/WAI/WCAG22/Techniques/failures/F2), [F100](https://www.w3.org/WAI/WCAG22/Techniques/failures/F100)
- `html` incorrectly overridden to `javascript` in [SCR14](https://www.w3.org/WAI/WCAG22/Techniques/client-side-script/SCR14), [F3](https://www.w3.org/WAI/WCAG22/Techniques/failures/F3), [F37](https://www.w3.org/WAI/WCAG22/Techniques/failures/F37), [F50](https://www.w3.org/WAI/WCAG22/Techniques/failures/F50)
- `css` incorrectly overridden to `html` in [ARIA24](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA24) (2x), [G188](https://www.w3.org/WAI/WCAG22/Techniques/general/G188)
- `php-template` incorrectly overridden to `html` in [C29](https://www.w3.org/WAI/WCAG22/Techniques/css/C29) (2x)
- `php-template` (with HTML also highlighted) incorrectly overridden to `php` (HTML not highlighted) in [SVR3](https://www.w3.org/WAI/WCAG22/Techniques/server-side-script/SVR3), [SVR4](https://www.w3.org/WAI/WCAG22/Techniques/server-side-script/SVR4)

All other instances are unaffected, as auto-detection matches the removed override.

Note this will also benefit from #4520, which will further prevent slightly-inaccurate-but-mostly-harmless auto-detection of CSS as Less (which is present in the external highlight.js loaded in those files, but not present in our own `highlight.min.js`).

## How I determined where language-* classes were required

1. Add some logging to highlight.js before running `highlightAll`:
   ```js
[...]

PR #4624, opened 2025-09-18
Fix incorrect highlight.js overrides

kfranqueiro:
This fixes instances where we had specified the wrong `language-...` class in `<pre><code class="...">`, causing highlight.js to highlight code incorrectly.

(Originally fixed/reported in #4611, which is being split up and closed)

### DIFF
--- a/techniques/client-side-script/SCR14.html
+++ b/techniques/client-side-script/SCR14.html
@@ -28,7 +28,7 @@
         <section id="the-javascript">
 				<h4>The JavaScript</h4>
          
-<pre xml:space="preserve"><code class="language-javascript">&lt;script&gt;
+<pre xml:space="preserve"><code class="language-html">&lt;script&gt;
 document.addEventListener("DOMContentLoaded", function (e) {
   const stockBox = document.querySelector("#stock-box");
   const stockMovement = document.querySelector("#stock-movement");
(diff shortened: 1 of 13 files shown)


techniques/client-side-script/SCR14.html
```diff
@@ -28,7 +28,7 @@ <h3>Updating screen reader settings for stock price update alerts</h3>
         <section id="the-javascript">
 				<h4>The JavaScript</h4>
          
-<pre xml:space="preserve"><code class="language-javascript">&lt;script&gt;
+<pre xml:space="preserve"><code class="language-html">&lt;script&gt;
```
iadawn, line 31: Wouldn't this be better as `language-javascript`? Seems there is much more JS than HTML

kfranqueiro: What matters isn't how much there is, but that there's a `<script>` tag around it. This inherently makes this an HTML block.

Same for instances of CSS surrounded by `<style>`. The xml language support is capable of understanding JS and CSS within appropriate contexts, but the reverse is not true.

- [JS support](https://github.com/highlightjs/highlight.js/blob/5697ae5187746c24732e62cd625f3f83004a44ce/src/languages/xml.js#L170-L186)
- [CSS support](https://github.com/highlightjs/highlight.js/blob/5697ae5187746c24732e62cd625f3f83004a44ce/src/languages/xml.js#L150-L169)

Issue #3808, opened 2024-05-01
SC 3.3.2: Labels or Instructions and a <select> element lacks a visible label and rely on default option


sarasuri:
Hello,

I have a question on the SC 3.3.2 and a dropdown list that is missing a visible label. In a Web application, an international phone number field has a country code dropdown list associated with it and has an accessible name provided via the aria-label="Country code". However, there is no visible label presented to all users except the default option.

Would this be considered as a failure for SC 3.3.3 that it is missing the visible label the dropdown list?

![image](https://github.com/w3c/wcag/assets/12256067/51db462d-5247-4d29-be5e-68a7b378da49)

![image](https://github.com/w3c/wcag/assets/12256067/68b4e747-c0cf-4211-ba59-5c7be7d59628)

Thanks

patrickhlauke:
I would personally say that the intention/purpose of the dropdown is clear from context (from the nearby/overarching "Phone number" and the country flag that's visible for the currently active/selected code)

giacomo-petri:
Proposal (happy to submit a PR if the group agrees):

In 2.4.6 headings and labels, we currently have:

> Labels of form controls are usually text-based. In some cases, images can serve as descriptive labels without additional text. In these cases, authors should ensure that the image and its use as a label (in context) are widely understood.

Should we include a similar note in 3.3.2 labels or instructions? I believe the scenario highlighted by @sarasuri is quite common, and there seems to be a lot of differing opinions on it. Clarifying this in the understanding could be very helpful.

amberhinds:
Another example of this that doesn't include an image:

<img width="358" height="58" alt="two selects with default options (filter by category and date and a filter button, without visible labels for the selects" src="https://github.com/user-attachments/assets/9127bf2b-5112-4412-aaf8-e8bd9f808670" />

```
<div class="alignleft actions">
				<label for="filter-by-date" class="screen-reader-text">Filter by date</label>
		<select name="m" id="filter-by-date">
			<option selected="selected" value="0">All dates</option>
		<option value="202508">August 2025</option>
		</select>
		<label class="screen-reader-text" for="cat">Filter by category</label><select name="cat" id="cat" class="postform">
	<option value="0">All Categories</option>
	<option class="level-0" value="1">Uncategorized</option>
</select>
<input type="submit" name="filter_action" id="post-query-submit" class="button" value="Filter">		</div>
```

[...]

PR #4626, opened 2025-09-19
3.3.2 - Form control labels using non-text content

giacomo-petri:
Closes https://github.com/w3c/wcag/issues/3808

This PR clarifies the guidance regarding non-text content used as labels for form controls (re-used 2.4.6 wording). For example, in a form field set for phone numbers with country code and number, using a flag as the visual label may be sufficient, provided that a meaningful accessible name is also provided separately in accordance with SC 4.1.2.

**This addresses the initial concern, but doesn’t cover the point raised by @amberhinds. It might be worth opening a separate issue, or we could discuss it first and decide on the best way to address it.**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
         "gray-matter": "^4.0.3",
         "highlight.js": "^11.11.1",
         "liquidjs": "^10.21.0",
-        "lodash-es": "^4.17.21",
+        "lodash-es": "^4.17.23",
         "tsx": "^4.20.3"
       },
       "devDependencies": {
@@ -2885,9 +2885,10 @@
       "license": "MIT"
     },
     "node_modules/lodash-es": {
-      "version": "4.17.21",
-      "resolved": "https://registry.npmjs.org/lodash-es/-/lodash-es-4.17.21.tgz",
-      "integrity": "sha512-mKnC+QJ9pWVzv+C4/U3rRsHapFfHvQFoFB92e52xeyGMcX6/OlIl78je1u8vePzYZSkkogMPJ2yjxxsb89cxyw=="
+      "version": "4.17.23",
+      "resolved": "https://registry.npmjs.org/lodash-es/-/lodash-es-4.17.23.tgz",
+      "integrity": "sha512-kVI48u3PZr38HdYz98UmfPnXl2DXrpdctLrFLCd3kOx1xUkOmpFPx7gCWWM5MPkL/fD8zb+Ph0QzjGFs4+hHWg==",
+      "license": "MIT"
     },
     "node_modules/lru-cache": {
       "version": "10.2.2",
@@ -3721,9 +3722,9 @@
       "integrity": "sha512-ARDJmphmdvUk6Glw7y9DQ2bFkKBHwQHLi2lsaH6PPmz/Ka9sFOBsBluozhDltWmnv9u/cF6Rt87znRTPV+yp/A=="
     },
     "node_modules/undici": {
-      "version": "6.21.3",
-      "resolved": "https://registry.npmjs.org/undici/-/undici-6.21.3.tgz",
-      "integrity": "sha512-gBLkYIlEnSp8pFbT64yFgGE6UIB9tAkhukC23PmMDCe5Nd+cRqKxSjw5y54MK2AZMgZfJWMaNE4nYUHgi1XEOw==",
+      "version": "6.23.0",
+      "resolved": "https://registry.npmjs.org/undici/-/undici-6.23.0.tgz",
+      "integrity": "sha512-VfQPToRA5FZs/qJxLIinmU59u0r7LXqoJkCzinq3ckNJp3vKEh7jTWN589YQ5+aoAC/TGRLyJLCPKcLQbM8r9g==",
       "license": "MIT",
       "engines": {
         "node": ">=18.17"
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "gray-matter": "^4.0.3",
     "highlight.js": "^11.11.1",
     "liquidjs": "^10.21.0",
-    "lodash-es": "^4.17.21",
+    "lodash-es": "^4.17.23",
     "tsx": "^4.20.3"
   },
   "devDependencies": {


understanding/20/labels-or-instructions.html
```diff
@@ -39,6 +39,9 @@ <h2>Intent of Labels or Instructions</h2>
          The goal is to make certain that enough information is provided for the user to accomplish
          the task without undue confusion or navigation.
       </p>
+
+      <p>Labels of form controls are usually text-based. In some cases, images can serve as descriptive labels without additional text.
```
mbgower:
```suggestion
      <p>Labels of form controls are usually text-based. Images with appropriate text alternatives can serve as labels without additional text. 
```

understanding/20/labels-or-instructions.html
```diff
@@ -39,6 +39,9 @@ <h2>Intent of Labels or Instructions</h2>
          The goal is to make certain that enough information is provided for the user to accomplish
          the task without undue confusion or navigation.
       </p>
+
+      <p>Labels of form controls are usually text-based. In some cases, images can serve as descriptive labels without additional text. 
+        In these cases, authors should ensure that the image and its use as a label (in context) are widely understood.</p>
```
patrickhlauke: Suggest merging the two sentences just into one?

```suggestion
      <p>Images (with an appropriate text alternative) can serve as descriptive labels without additional text. In these cases, authors should ensure that the image and its use as a label (in context) are widely understood.</p>
```

gundulaniemann: I prefer to point out "Labels of form controls are usually text-based.".
It makes clear that using labels is the exception, and that we are talking about forms here.

mbgower: I feel like this "widely understood" sentence would probably be better in Headings and Labels. Although it is goes beyond a simple discussion on descriptive text, the qualitative concepts of Headings and Labels seems like a better fit than the basic question of 3.3.2 which is: is there a label for the input?

Some task force members ended up having a long post-meeting discussion on the nuances of the telephone example in the original issue. It would be good to try to capture some of those concepts in our guidance, in some way.

patrickhlauke: I still fundamentally disagree that we need to make any kind of "this is common, this is the exception" statement here. That seems to add some sort of judgement on our part that has no business being here. Per the actual definition of label, it's "text or other component with a text alternative". No "but mostly text" or anything else. Adding qualifiers about what's "an exception" seems inappropriate to me.

patrickhlauke: agree, the qualitative part should be moved to headings and labels (and, at a stretch, then cross-referenced from here)

giacomo-petri:
After our after-hours chat last Friday, I've put together several codepen examples where the legend, combined with the visual context, should be (IMO) sufficient or where a visible label for each input may not be strictly necessary (though an accessible name is still required): https://codepen.io/giacomo-usablenet/full/vELaXpN

We can review them together during our next meeting if you'd like.

Note: These examples are only partially functional (if something doesn't work, just ignore it; the goal is to illustrate the concept).

mbgower:
@giacomo-petri thanks for those examples.
One of my take aways is that for the most part, the examples have a group label/legend that is doing double-duty. There's nothing wrong with that, and I spent quite a bit of time covering similar considerations in various parts of Label in Name, such as https://www.w3.org/WAI/WCAG22/Understanding/label-in-name.html#identifying-label-text-for-components

I just think that's a slightly different use case from an input with _no_ text that could be considered its label. I also think  some of your examples do not actually use an image as a label. The "visual content", as you put it, is part of the UIC, not an image serving as a label. So it feels like this is drifting away from the question about an image as a label, which is I thought was the focused of the discussion?

I still feel like it is _totally_ valid -- and important -- to say that most labels are text-based. That is the reality.

Maybe it's also worth pointing out the historical example still part of the current Understanding document, which somewhat covers some of the considerations in your examples:
> A U.S. phone number separates the area code, exchange, and number into three fields. Parentheses surround the area code field, and a dash separates the exchange and number fields. While the punctuation provides visual clues to those familiar with the U.S. telephone number format, the punctuation is not sufficient to label the fields. The single "Phone number" label also cannot label all three fields. To address this, the three fields are grouped in a fieldset with the legend "Phone number". Visual labels for the fields (beyond the punctuation) cannot be provided in the design, so invisible labels are provided with the "title" attribute to each of the three fields. The value of this attribute for the three fields are, respectively, "Area Code", "Exchange", and "Number".

understanding/20/labels-or-instructions.html
mbgower: @giacomo-petri I suggest that line 44 come out, and use some of this discussion to create a new PR for Heading and Labels.

bruce-usab:
Briefly discussed on Backlog call 10/31.  Left in drafted pending incorporation of feedback.

giacomo-petri:
Sorry, I was focusing only on examples where each fieldset has an associated legend, but there are many cases where there's no legend and the input still makes sense on its own. I've put together a bunch of examples here: https://codepen.io/giacomo-usablenet/full/yyeGYLx

patrickhlauke:
suggested in the call just now: while i'm not a fan of the "most labels are text based, but images are fine", if we do think it's valuable as a context for authors, then let's couch it into something like a "best practice", and then provide something actionable, along the lines of "As a best practice, note that the majority of labels are text based. Using images is fine, but care should be taken that the images are widely understood by the intended target audience. You may consider adding additional hints like text-based tooltips on the images..."

giacomo-petri:
I've revised the paragraph trying to incorporate everything we've discussed at our last TF meeting

understanding/20/labels-or-instructions.html
```diff
@@ -39,6 +39,11 @@ <h2>Intent of Labels or Instructions</h2>
          The goal is to make certain that enough information is provided for the user to accomplish
          the task without undue confusion or navigation.
       </p>
+
+      <p>As a best practice, note that the majority of form control labels are text-based. 
+         Using images as labels is fine, but care should be taken to ensure that the images are widely understood 
+         by the intended target audience. Authors may consider providing additional hints,
```
gundulaniemann, line 45: In fact, by Success Criterion 1.1.1, the author needs to add a text alternative to comply. Therefore I suggest to say:
"Note that by Success Criterion 1.1.1 non-text content, the author needs to provide a text alternative. For equal support of all users, this alternative should be visible, like a tooltip or supplementary text, to support clarity when using non-text labels.

understanding/20/labels-or-instructions.html
```diff
@@ -39,6 +39,11 @@ <h2>Intent of Labels or Instructions</h2>
          The goal is to make certain that enough information is provided for the user to accomplish
          the task without undue confusion or navigation.
       </p>
+
+      <p>As a best practice, note that the majority of form control labels are text-based. 
+         Using images as labels is fine, but care should be taken to ensure that the images are widely understood
```
gundulaniemann: I suggest to say "images or icons". 
With the term 'image' many people connect photos or paintings, while an icon (technically often realized in a font) does not ring a bell when reading the term 'image'.

understanding/20/labels-or-instructions.html
patrickhlauke: the "additional hints" part is intended to be in addition to things. of course 1.1.1 applies. all other SCs apply. not sure we need to explicitly call 1.1.1 out (why not also mention 1.4.11 etc otherwise?)

understanding/20/labels-or-instructions.html
patrickhlauke:
```suggestion
         Using images/icons as labels is fine, but care should be taken to ensure that these are widely understood 
```

understanding/20/labels-or-instructions.html
```diff
@@ -39,6 +39,11 @@ <h2>Intent of Labels or Instructions</h2>
          The goal is to make certain that enough information is provided for the user to accomplish
          the task without undue confusion or navigation.
       </p>
+
+      <p>As a best practice, note that the majority of form control labels are text-based.
```
patrickhlauke: i'd still prefer at least dropping the "As a best practice" and making this a statement of fact, rather than suggesting that the use of text labels is a best practice.
Maybe the best practice part is the "care should be taken..." part in the next sentence

understanding/20/labels-or-instructions.html
gundulaniemann: I feel the phrase "Authors may consider" is too weak, as indeed it is a must.

detlevhfischer: @gundulaniemann wrote
> I feel the phrase "Authors may consider" is too weak, as indeed it is a must.

Where does it say additional hints are a must? Beyond ensuring that a field labelled with an icon/image is sufficiently conventional for visual users and has a descriptive accessible name for non-visual users, I see no "must" here for additional hints. Can you point to a normative requirement that supports your claim?

patrickhlauke: exactly. the must part is having alt text. that is not disputed, but it's also not mentioned here at all because it goes without saying (all other SCs still apply, even when not explicitly mentioned ... we sometimes cross-reference things when they're not immediately obvious, but having an alternative to non-text content is pretty much a cornerstone of WCAG that doesn't need to be repeated).

additional hints here are about going beyond just the alt text - some form of tooltip, some extra adjacent text, etc. and as @detlevhfischer notes, *that* is not a *must*

patrickhlauke: possibly change "may" to a more non-RFC term

understanding/20/labels-or-instructions.html
patrickhlauke: suggestion to reword "is fine" to be more formal

alastc:
```suggestion
         Using images as labels can meet the requirements of the criterion, but care should be taken to ensure that the images are widely understood 
```

understanding/20/labels-or-instructions.html
```diff
@@ -39,6 +39,11 @@ <h2>Intent of Labels or Instructions</h2>
          The goal is to make certain that enough information is provided for the user to accomplish
          the task without undue confusion or navigation.
       </p>
+
+      <p>As a best practice, note that the majority of form control labels are text-based. 
+         Using images as labels can meet the requirements of the criterion, but care should be taken to ensure that the images are widely understood
```
```suggestion
         Using images as labels meets the requirements of the criterion, but care should be taken to ensure that the images are widely understood 
```

understanding/20/labels-or-instructions.html
patrickhlauke:
```suggestion
      <p>Note that the majority of form control labels are text-based. 
```

understanding/20/labels-or-instructions.html
```diff
@@ -39,6 +39,11 @@ <h2>Intent of Labels or Instructions</h2>
          The goal is to make certain that enough information is provided for the user to accomplish
          the task without undue confusion or navigation.
       </p>
+
+      <p>Note that the majority of form control labels are text-based. 
+         Using images as labels meets the requirements of the criterion, but care should be taken to ensure that the images are widely understood 
+         by the intended target audience. Authors may consider providing additional hints,
```
patrickhlauke, line 45:
```suggestion
         Using images as labels meets the requirements of the criterion, but as a best practice care should be taken to ensure that the images are widely understood 
         by the intended target audience. Authors may consider going beyond the normative requirements in these cases, and providing additional hints,
```

gundulaniemann: I suggest to rather recommend. Pointing out this goes beyond the normative requirement may be perceived as that a visible alternative is not needed, while it is needed specifically by users with cognitive or learning issues. Having an invisible  text alternative but no visible one disadvantages non-screen reader users.

patrickhlauke: but that was never a requirement. you're trying to force a new normative requirement in here that never existed.

gundulaniemann: > but that was never a requirement. you're trying to force a new normative requirement in here that never existed.

I understand a visible text alternative is not normatively required.
Nevertheless, a recommendation is not normative. I suggested to recommend it to give more motivation.

patrickhlauke: you "while it is needed" gave the impression that you're saying it *is* a normative requirement.

so, changing "Authors may consider" to something like "We recommend" is what you're saying? A bit more loaded, but i could just about live with that.

patrickhlauke:
Moving to discussion. Hopefully we can resolve this at last (though it'll need to go through another round for "sent to AWGW")

understanding/20/labels-or-instructions.html
gundulaniemann: > you "while it is needed" gave the impression that you're saying it _is_ a normative requirement.
> 
> so, changing "Authors may consider" to something like "We recommend" is what you're saying? A bit more loaded, but i could just about live with that.

Yes, exactly. Sorry for being not clear.

As I am sure it is needed by some users, yet I understand it is not normatively requested, I suggest to recommend it. Hope that is clearer now.

giacomo-petri:
I want to acknowledge that the current version reflects what the TF collectively agreed on. Everyone included in this thread was present, and we reached alignment on this together.

Of course, we can make adjustments if needed, but out of respect for everyone's time, I strongly believe that once we define (together) an acceptable solution that improves the current understanding document, we should move forward and present it to the main group, without introducing new requirements that were not agreed upon.

Additionally, based on our TF meeting, we explicitly agreed not to introduce quality levels (such as "best practices"). In the scenario described (and demonstrated in my codepen) there is no need for certain labels, as they are universally recognized as standard. For example, we would not require a "MENU" label under a hamburger menu icon. Is it a best practice? Maybe, but I don't see it reflected in the real world web.

The current wording may not be perfect, but it represents a solid step forward and something we already aligned on. My concern is that we risk spending time repeatedly revisiting points that were already discussed and agreed upon in meetings held two sessions ago. I would really like to avoid going in circles and instead focus on making concrete progress from where we collectively landed.

patrickhlauke:
Discussed at facilitator meeting today. Agreed with @alastc @kfranqueiro @bruce-usab that this PR as is already improves the wording. Decided to merge as is, but @gundulaniemann feel free to suggest a follow-up advisory technique that shows how labels with image+text can be better (for most/all users, but particularly those with cognitive disabilities)

PR #4627, opened 2025-09-19
highlight.js: Add missing overrides, update version, and run at build time

kfranqueiro:
Changes in this PR:

- Adds `language-*` or `no-highlight` classes to code elements inside pre elements where they were missing, to provide stability across highlight.js upgrades in the event of changes to auto-detection logic
  - This intentionally does not go through the trouble/noise of adding syntax highlighting to obsolete Flash and Silverlight techniques
- Adds a build-time check which logs in dev / fails in CI if any code blocks have no class specified
  - Example of message logged for each missing class found:
    `./techniques/css/C45.html missing "language-*" or "no-highlight" class on <pre><code> starting with &lt;!DOCTYPE html&gt;\n&lt;html lang="en...`
  - Example of message printed when build fails (for CI, in addition to the above message):
    ```
    [11ty] 1. Having trouble rendering liquid template ./techniques/css/C45.html (via TemplateContentRenderError)
    [11ty] 2. Please ensure all code blocks have a highlight.js class (language-* or no-highlight).
    ```
- Updates to use the latest version of highlight.js
- Switches to invoking `highlight.js` at build time and including it as an npm dependency
  - This means no more `highlight.min.js` bundle taking up more space in the repo on each update
  - This also means syntax highlighting no longer involves running client-side JS on every visit to every page, better optimizing for sustainability

Before switching from client-side to build-time, I verified (using the script I'd posted in #4611) that updating the bundle does not cause any syntax highlighting changes, other than adding it for the previously-missing languages.

I also performed screenshot diffs and have verified that the only changes caused by the update are fixes to syntax highlighting (primarily in terms of recognized CSS property names).

I performed screenshot diffs again after switching to build-time highlight.js, and there were only two diffs, in examples that specifically intended to include `<span lang="...">` elements inside of `<pre><code>`. These have been adjusted to apply `lang` to the `pre` tag instead.

**Note:** The failure on the Check PR files check is an extremely rare false positive because I had updated it to consider `script` normative, but technically only `script/wcag.js` is involved in TR space; `script/highlight.min.js` is not. This should never be an issue again, since I'm removing `highlight.min.js`.

### DIFF
--- a/11ty/CustomLiquid.ts
+++ b/11ty/CustomLiquid.ts
@@ -153,6 +153,22 @@ export class CustomLiquid extends Liquid {
       // Add charset to pages that forgot it
       if (!$("meta[charset]").length) $('<meta charset="UTF-8">').prependTo("head");
 
+      const $missingHljsOverrides = $("pre code:not([class])");
+      if ($missingHljsOverrides.length) {
+        $missingHljsOverrides.each((_, el) => {
+          const code = $(el).html()!.replace(/\n/g, "\\n");
+          const excerpt = `${code.slice(0, 40)}${code.length > 40 ? "..." : ""}`;
+          console.log(
+            `${filepath} missing "language-*" or "no-highlight" class on <pre><code> starting with ${excerpt}`
+          );
+        });
+        if (process.env.ELEVENTY_RUN_MODE === "build") {
+          throw new Error(
+            "Please ensure all code blocks have a highlight.js class (language-* or no-highlight)."
+          );
+        }
+      }
+
       const prependedIncludes = ["header"];
       const appendedIncludes = ["wai-site-footer", "site-footer"];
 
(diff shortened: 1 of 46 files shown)


11ty/CustomLiquid.ts
```diff
@@ -153,6 +153,20 @@ export class CustomLiquid extends Liquid {
...
+      const missingHljsOverrides$ = $("pre code:not([class])");
+      if (missingHljsOverrides$.length) {
+        missingHljsOverrides$.each((_, el) => {
+          const code = $(el).html()!.replace(/\n/g, "\\n");
+          const excerpt = `${code.slice(0, 40)}${code.length > 40 ? "..." : ""}`;
+          console.log(`${filepath} <pre><code> missing hljs override: ${excerpt}`);
+        });
+        if (process.env.ELEVENTY_RUN_MODE === "build") {
+          throw new Error(
+            "Some code blocks are missing highlight.js classes (language-* or no-highlight); please fix."
```
iadawn: ANy opportunity to give people a clue as to what code is missing the class?

kfranqueiro: The `console.log` printed before this (L161) will provide context. This error is thrown in addition only in build mode, to stop/fail the build. (Edited the PR description to show examples)

Issue #3803, opened 2024-04-26
Success Criterion 2.4.2 -- Disconnect in the examples 


lhenze:
Hello, 

On https://www.w3.org/WAI/WCAG21/Understanding/page-titled#examples

I see 

> The title of [Understanding WCAG 2.1 ](https://www.w3.org/WAI/WCAG21/Understanding/)is "Understanding WCAG 2.1."

But when I go to the page and view source, I notice that the title is currently 

>  <title>Understanding WCAG 2.1 | WAI | W3C</title>

I am researching the importance and expectation of including the site title in each page title; I understand that it is not an accessibility requirement to do so, but the disconnect seems a bit disorienting.

Thanks!

filippo-zorzi:
Hi @lhenze,
adding "| WAI | W3C" is a Sufficient technique ([G127](https://www.w3.org/WAI/WCAG21/Techniques/general/G127)) to meet the [SC 2.4.8](https://www.w3.org/WAI/WCAG22/Understanding/location).

"Understanding WCAG 2.1." meets the [SC 2.4.2](https://www.w3.org/WAI/WCAG21/Understanding/page-titled) and explains the requirements of the SC.
I think it was a choice to not add "| WAI | W3C" to not make two Success Criteria mix. 

For you research I suggest you to check the SC 2.4.8 as well as 2.4.2 and the Advisory and Sufficient techniques

Let me know if you have any other question.
Thanks

giacomo-petri:
Picking this up again, as the example has become even more confusing now.
- Page titles now reflect the structure of the flow ($page_title | WAI | W3C).
- Appendices are no longer separate pages; they are anchors within the main page and have no titles of their own.

The example works if viewed on its own, but when compared to the WCAG website, it no longer matches the current structure, leading to confusion and misunderstandings.

Should we update it to match the current website structure, or create a "static" example with placeholder data to avoid having to revise it every time the site changes?

patrickhlauke:
> Should we update it to match the current website structure, or create a "static" example with placeholder data to avoid having to revise it every time the site changes?

I'd go for the latter. Never been a fan of self-referential examples (which then go out of date)

lhenze:
Thank you; this is clearer!

PR #4631, opened 2025-09-22
Updated "A document collection" example since obsolete - SC 2.4.2

giacomo-petri:
Closes https://github.com/w3c/wcag/issues/3803

Description:
The current example does not accurately reflect the WCAG title structure and causes considerable confusion. Additionally, self-referential examples are difficult to maintain. For these reasons, I have replaced it with a static example.

### DIFF
--- a/understanding/20/page-titled.html
+++ b/understanding/20/page-titled.html
@@ -96,15 +96,23 @@
          <dt>An HTML web page</dt>
          <dd>The descriptive title of an HTML web page is marked up with the &lt;title&gt; element so
                that it will be displayed in the title bar of the user agent.</dd>
-         <dt>A document collection</dt>
+         <dt>A guide on authoring practices</dt>
          <dd>
-            <p>The title of <a href="./">Understanding WCAG 2.2</a> is "Understanding WCAG 2.2".</p>
+            <p>The title of the landing page is "ARIA Authoring Practices Guide | APG | WAI | W3C"</p>
             <ul>
-               <li>The <a href="../Understanding/intro">Introduction to Understanding WCAG</a> page has the title "Introduction to Understanding WCAG".</li>
-               <li>Major sections of the document collection are pages titled "Understanding Guideline X" and "Understanding Success Criterion X."</li>
-               <li>Appendix A has the title "Glossary."</li>
-               <li>Appendix B has the title "Acknowledgements."</li>
-               <li>Appendix C has the title "References."</li>
+              <li>
+                <span>The patterns list page has the title "Patterns | APG | WAI | W3C"</span>
+                <ul>
+                  <li>
+                    <span>Specific patterns are pages titled "X Pattern | APG | WAI | W3C" (e.g., "Alert and Message Dialogs Pattern | APG | WAI | W3C")</span>
+                    <ul>
+                      <li>Specific examples for each pattern are pages titled "X Example | APG | WAI | W3C" (e.g., "Alert Dialog Example | APG | WAI | W3C")</li>
+                    </ul>
+                  </li>
+                </ul>
+              </li>
+              <li>Practices page has the title "Practices | APG | WAI | W3C"</li>
+              <li>Index page has the title "Index | APG | WAI | W3C"</li>
             </ul>
          </dd>
          <dt>A web application</dt>


understanding/20/page-titled.html
```diff
@@ -96,15 +96,23 @@ <h2>Examples of Page Titled</h2>
...
-               <li>The <a href="../Understanding/intro">Introduction to Understanding WCAG</a> page has the title "Introduction to Understanding WCAG".</li>
-               <li>Major sections of the document collection are pages titled "Understanding Guideline X" and "Understanding Success Criterion X."</li>
-               <li>Appendix A has the title "Glossary."</li>
-               <li>Appendix B has the title "Acknowledgements."</li>
-               <li>Appendix C has the title "References."</li>
+              <li>
+                <span>The patterns list page has the title "Patterns | APG | WAI | W3C"</span>
+                <ul>
+                  <li>
+                    <span>Specific patterns are pages titled "X Pattern | APG | WAI | W3C" (e.g., "Alert and Message Dialogs Pattern | APG | WAI | W3C")</span>
```
patrickhlauke: any reason for the `<span>`?

giacomo-petri: Nope, it was only there as a non-semantic container in case we needed to style the content separately. I'm removing it

bruce-usab (changes requested):
I prefer the current example because it's simpler and uses the generic heading of "a document collection".  I agree it should be improved, and adding "landing page for…" is nice.

bruce-usab:
I was just looking at this page in the context of WCAG2ICT and thought the "document collection" example needed improvement.  I am okay with using APG as that example, and it's nice to have an example with the vertical bars | since that is common nowadays.

[preview diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FUnderstanding%2Fpage-titled&doc2=https%3A%2F%2Fdeploy-preview-4631--wcag2.netlify.app%2Funderstanding%2Fpage-titled)

understanding/20/page-titled.html
```diff
@@ -96,15 +96,23 @@ <h2>Examples of Page Titled</h2>
          <dt>An HTML web page</dt>
          <dd>The descriptive title of an HTML web page is marked up with the &lt;title&gt; element so
                that it will be displayed in the title bar of the user agent.</dd>
-         <dt>A document collection</dt>
+         <dt>A guide on authoring practices</dt>
```
patrickhlauke:
```suggestion
         <dt>A multi-page guide on authoring practices</dt>
```

understanding/20/page-titled.html
```diff
@@ -96,15 +96,23 @@ <h2>Examples of Page Titled</h2>
...
+            <p>The title of the landing page is "ARIA Authoring Practices Guide | APG | WAI | W3C"</p>
             <ul>
-               <li>The <a href="../Understanding/intro">Introduction to Understanding WCAG</a> page has the title "Introduction to Understanding WCAG".</li>
-               <li>Major sections of the document collection are pages titled "Understanding Guideline X" and "Understanding Success Criterion X."</li>
-               <li>Appendix A has the title "Glossary."</li>
-               <li>Appendix B has the title "Acknowledgements."</li>
-               <li>Appendix C has the title "References."</li>
+              <li>
+                The patterns list page has the title "Patterns | APG | WAI | W3C"
+                <ul>
```
patrickhlauke, line 105: I think it's unnecessary and confusing to use nested (and then nested again, 3 levels) bullet points here. I get that the intention here is to also show hierarchy, but a simple straightforward set of bullet points will work best here I'd say

understanding/20/page-titled.html
bruce-usab: We still need the simple titular category of "a document collection" or "set of web pages"

bruce-usab:
@giacomo-petri I took the liberty to move this back to "In Progress" but please be encouraged to put it under "For Discussion" if you would like feedback from the group.

bruce-usab:
I think the top level bullet should be something generic, if not "document collection" then "set of web pages".  The APG example could follow.  I agree it's helpful to have longer real-world code.

Edit to mention that a simple edit to line 99 addresses my concern.  Just change "A guide on authoring practices" to "A set of web pages".  There isn't a need to provide the title/name of the set.

PR #4633, opened 2025-09-22
Update name.html

mbgower:
An attempt to make a clearer connection between "name" and "accessible name", which we currently have stated in Understanding documents such as https://www.w3.org/WAI/WCAG22/Understanding/label-in-name, but not in a note to the defined term.

### DIFF
--- a/guidelines/terms/20/name.html
+++ b/guidelines/terms/20/name.html
@@ -8,5 +8,7 @@
    </p>
    
    <p class="note">This is unrelated to the name attribute in HTML.</p>
+
+   <p class="note">The term "accessible name", is a synonym for "name". The term is also defined and used in the <a href="https://www.w3.org/TR/accname-1.2/#dfn-accessible-name">Accessible Names and Description Computation</a></p>
    
 </dd>


bruce-usab:
I love the prospect of being explicit that [name](https://www.w3.org/TR/WCAG22/#dfn-name) and [accessible-name](https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/) are synonymous.  But this crystallized for me a problem with 1.1.1.

>  If non-text content is a control or accepts user input, then it has a name that describes its purpose.

It would be counterproductive for accessible-name to be an analog to `longdesc` was or [page titles](https://www.w3.org/TR/WCAG22/#page-titled).  All interactive elements need an `accessible-name` but not all need `accessible-description` (and the latter defaults to the former).

I submit that the “controls, input“ portion of [non-text content](https://www.w3.org/TR/WCAG22/#non-text-content) should be:
>  If non-text content is a control or accepts user input, then it has a name that provides descriptive identification.

1. In practice, that's how the requirement is understood.
1. That’s the phrasing used for the next three exceptions (time-based media, test, sensory experience).
1. Text that describes its purpose &lt; text that provides descriptive information.

For example, with a ”submit button” the word “submit” doesn‘t “describe” the button’s purpose, it _is_ the purpose of the button.  The purpose provides the descriptive identification.  It serves no one for accessible-name to be further conflated with accessible-description (which “describes its purpose” is doing).  Alternatively, we use phrasing closer to that is used by the APG:

>  If non-text content is a control or accepts user input, then it has a name that distinctively conveys purpose or intent.

PR #4666, opened 2025-10-03
Aria2 as advisory technique for 3.3.2 and 3.3.1

### LINKED ISSUE #246: Moving ARIA2 technqiue from 3.3.3 and putting in 3.3.2; adding ARIA21 to 3.3.3
Recommendation: [ARIA2: Identifying a required field with the aria-required property](ARIA2: Identifying a required field with the aria-required property ) is removed as a sufficient technique for SC 3.3.3, and added to SC 3.3.2.
As a second recommendation, [ARIA21](https://www.w3.org/TR/WCAG20-TECHS/ARIA21.html) should be added to 3.3.3.
Rationale:
[SC 3.3.3 Error Suggestions](https://www.w3.org/TR/UNDERSTANDING-WCAG20/minimize-error-suggestions.html) is concerned with responding to a detected error in a user form interaction, while SC [3.3.2 Labels or Instructions](https://www.w3.org/TR/UNDERSTANDING-WCAG20/minimize-error-cues.html) covers providing guidance when content needs user input.

[ARIA2: Identifying a required field with the aria-required property](http://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/ARIA2) is more appropriate as a label or instruction technique to guide user input than as a technique only to be used in response to an error, which is what is currently implied (unless it is the WG's position that aria-required should only be set to true after a user has failed to fill in a required field).
For 3.3.3, a better aria technique to include would be [ARIA21: Using Aria-Invalid to Indicate An Error Field](https://www.w3.org/TR/WCAG20-TECHS/ARIA21.html).

mbgower:
Closes #246


baldino-m:
I don't get why, in the first line, "provide" has been replaced with "enhance": if a form field is visually identifiable as required (with an asterisk or whatever) and `aria-required` is _required_ (pun not intended), then no programmatic information is given, so `aria-required` does actually **provide** some. 

My understanding is also that, if the accessible name of the field contains some "required" wording, such as "First name (required)", then `aria-required` is redundant.

techniques/aria/ARIA2.html
```diff
@@ -17,12 +17,12 @@ <h2>When to Use</h2>
...
+         <p>This technique is set as advisory for meeting 3.3.2 Labels or Instructions, because use of <code>aria-required="true"</code> alone, without any visual indicator, would not be perceivable to most users.</p>
+         <p>However, when an asterisk or other text symbol is associated with the field, using the <code>aria-required</code> property in addition to the visual presentation makes it much easier for user agents to pass on this important information to the user in a user agent-specific manner. Refer to <a href="https://www.w3.org/TR/html-aria/">ARIA in HTML</a> for information on how to provide WAI-ARIA States and Properties with HTML. WAI-ARIA States and Properties are compatible with other languages as well; refer to documentation in those languages. </p>
```
adampage:
```suggestion
         <p>However, when an asterisk or other text symbol is associated with the field, using the <code>aria-required</code> property in addition to the visual presentation enables user agents to pass on this important information to the user in a user agent-specific manner. Refer to <a href="https://www.w3.org/TR/html-aria/">ARIA in HTML</a> for information on how to provide WAI-ARIA States and Properties with HTML. WAI-ARIA States and Properties are compatible with other languages as well; refer to documentation in those languages. </p>
```

mbgower:
> I don't get why, in the first line, "provide" has been replaced with "enhance": if a form field is visually identifiable as required (with an asterisk or whatever) and `aria-required` is _required_ (pun not intended), then no programmatic information is given, so `aria-required` does actually **provide** some.

@baldino-m the technique has been changed from sufficient to advisory for 3.3.1, and it has also been added as advisory for 3.3.2. As such, the technique is intentionally not sufficient for any of these requirements. In that context, changing the preamble from "provide" to "enhance" hopefully makes more sense.

Please have a look at the preview of the updated document, including its unmodified examples, and let us know if all the changes together, provide sufficient context to answer your comments.
https://deploy-preview-4666--wcag2.netlify.app/techniques/aria/aria2

One option could be to make it "provided or enhanced"

detlevhfischer:
This is just a very minor concern: The image with the row of `li` elements containing the smaller links shows them offset by a small margin that is not present in the code given below the figure. Given the CSS reproduced, I would expect `li`-elements to sit flush to each other, without margin (there's also no CSS that will cause the list to be displayed inline and with `list-style-type: none`, but that is perhaps something that can be left out). So it may be unclear for users if the 24px value refers to the bounding box of the `li`, or includes the spacing around them. (Taking the margin between boxes of approx. 4px for granted, you would already meet  target size if your `li` max-width was just 20px.)

mbgower:
> The image with the row of `li` elements containing the smaller links shows them offset by a small margin that is not present in the code given below the figure. Given the CSS reproduced, I would expect `li`-elements to sit flush to each other, without margin (there's also no CSS that will cause the list to be displayed inline and with `list-style-type: none`, but that is perhaps something that can be left out). So it may be unclear for users if the 24px value refers to the bounding box of the `li`, or includes the spacing around them. (Taking the margin between boxes of approx. 4px for granted, you would already meet target size if your `li` max-width was just 20px.)

@detlevhfischer I don't see any images, so I'm not sure what you mean. The third example contains a code example that contains two list items elements. These are indented slightly from the parent unordered list element. Is this what you're referring to? Or was this comment intended for a different issue?

detlevhfischer:
@mbgower This was supposed to be a comment to [C42](https://deploy-preview-4685--wcag2.netlify.app/techniques/css/c42). I don't know why I put it  here, sorry. Ignore. (I have posted it in the C42 technique update PR now).

PR #4673, opened 2025-10-06
Add F114: Failure technique for focus obscured by overlay

### LINKED ISSUE #3203: Create a failure technique for Non-text contrast based on a lightbox effect
This need was identified in discussions on Focus Not Obscured and whether a lightbox effect would be considered to partially obscure the item receiving keyboard focus.

At some point the wording of Focus Not Obscured (minimum and enhanced) ditched the word and definition for "obscured" as part of its normative text, and now uses "hidden"; however, the discussion on how a lightbox effect could still lead to a failure of sufficient contrast for the focus indicator should be captured.

Essentially: if a lightbox effect is used on a dialog, and focus is not constrained to the dialog, a keyboard user may be able to tab out of the dialog, resulting in the item with focus being in the 'greyed' part of the screen (the lightbox effect). At that point, the contrast between the focus indicator and the background must still achieve 3:1 or it will fail Non-text Contrast. At some point it will also fail Focus Appearance as well.

venkatvemuri28:
## **Add F114: Failure Technique for Focus Indicators Obscured by Semi-Opaque Overlays**

Closes https://github.com/w3c/wcag/issues/3203

### **Problem**
WCAG 2.2 Understanding document for SC 2.4.11 Focus Not Obscured (Minimum) mentions that semi-opaque overlays can reduce focus indicator visibility, but only describes the problem informally. There was no dedicated Failure Technique (F114) that auditors could reference when testing SC 1.4.11 Non-text Contrast failures where focus indicators are dimmed by overlays.

### **Solution**
Created F114 failure technique with comprehensive testing guidance and a working example that demonstrates the issue.

#### **New Files:**
- `techniques/failures/F114.html` - Complete failure technique documentation
- `working-examples/css-focus-obscured-overlay/index.html` - Interactive working example

#### **Modified Files:**
- `understanding/understanding.11tydata.js` - Added F114 to SC 1.4.11 failure techniques list

#### **F114 Technique Includes:**
- Clear applicability criteria for semi-transparent overlays
- Detailed description of the failure condition
- Practical examples with CSS code
- Step-by-step testing procedure
- Expected results criteria
- Related techniques (C40, G195)

#### **Working Example Features:**
- Form with focusable controls demonstrating normal focus indicators
- Semi-transparent overlay (70% opacity) triggered by button
- Clear instructions for testing the failure
- Responsive design following WCAG project patterns
- Refactored CSS for maintainability (simplified form layout, consolidated button styles)

### **Testing**
The working example allows auditors to:
1. Observe normal focus indicators
2. Activate overlay to see contrast reduction
3. Verify focus indicators become difficult to perceive
4. Test with color contrast analyzers

### **Impact**
Provides auditors with a formal reference (F114) for citing SC 1.4.11 failures when focus indicators are obscured by semi-opaque overlays, filling the identified gap in WCAG testing guidance.


fstrr:
[Preview](https://deploy-preview-4673--wcag2.netlify.app/working-examples/css-focus-obscured-overlay/)

scottaohara:
while i realize the primary goal of this new technique is to show how this fails 2.4.11, should it not also mention this example fails other SCs as well?  Other failing techniques similarly list the other SC(s) that are applicable

detlevhfischer:
Hmmm... I think for a failure of 2.4.11, it does not matter whether the obscuring object is opaque or semitransparent. With modal dialogs, focus should not reach the page background, so that is the primary failure (of 2.4.3) in the scenario outlined. It the failure occurs and the focus is not (or not well enough) visible on darkened background as a consequence, this seems at least secondary. So it is difficult to pin down this way because more than one SC fail (1.4.11 also being a candidate). The failure condition also would need to take into account if the dialog was explictily opened by the user (think of a chat window) or can be moved around / repositioned. If so, lack of contrast would more likely be a 1.4.11 failure (not-text contrast).

venkatvemuri28:
@kfranqueiro Any advise on this?

understanding/understanding.11tydata.js
```diff
@@ -541,7 +541,7 @@ export default function (data) {
             techniques: ["G207", "G209"],
           },
         ],
-        failure: ["F78"],
+        failure: ["F78", "F114"],
```
kfranqueiro, line 545: This was added to 1.4.11 instead of 2.4.11 by mistake.

Edit: Nevermind, I think I misread the PR description.

techniques/failures/F114.html
```diff
@@ -0,0 +1,47 @@
...
+}</code></pre>
+         <p>When the overlay is present and an element receives focus, the white outline appears behind the semi-transparent black overlay. The effective contrast of the white outline against the dimmed background may be insufficient, failing 1.4.11.</p>
+         <ul class="working-example">
+            <li><a href="../../working-examples/css-focus-obscured-overlay/">Example of focus indicator obscured by overlay</a></li>
+         </ul>
+      </section>
+   </section><section id="tests"><h2>Tests</h2>
+      <section class="procedure"><h3>Procedure</h3>
+         <ol>
+            <li>Navigate through the page using keyboard navigation (Tab key) to move focus to various interactive elements.</li>
```
alastc:
```suggestion
            <li>Navigate through the page using keyboard navigation to move focus to each interactive element.</li>
```

techniques/failures/F114.html
```diff
@@ -0,0 +1,47 @@
...
+            <li>Check if any semi-opaque overlays or translucent content appear that could cover focus indicators.</li>
+            <li>For each focusable element that could be covered by such an overlay, measure the contrast ratio of the focus indicator while the overlay is present.</li>
```
mbgower:
```suggestion
            <li>For each element receiving focus while obscured by the overlay, measure the contrast ratio of the item with focus or its focus indicator.</li>
```

fstrr:
There's another,  new, design technique that this would cover: on some sites, when a user scrolls down the page, what looks like an automatically opened modal dialog appears, usually with a "subscribe to my blog" message. All of the page's content around the "dialog" is dimmed with a fake [backdrop](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/::backdrop), making the "dialog" stand out. The "dialog" is just a section of content on the page and as the user continues to scroll, the backdrop disappears.

bruce-usab:
On Backlog call, we noted that this is part of Understanding:

> A properly constructed modal dialog will always pass this SC. Even if it appears directly on top of an item with focus, the dialog takes focus on appearance, and thus the item receiving focus -- the dialog or one of its components -- is visible. A properly constructed modal maintains that focus and prevents interaction outside the modal until it is dismissed.
> 
> A dialog-like overlay that does not take focus on appearance and does not either constrain interaction to the overlay or dismiss itself on loss of focus (thus allowing focus to exit into the content behind it) will be at risk of failing this SC, where it is positioned such that it can obscure other focusable items.

fstrr:
[Here's an example of my comment above](https://cdpn.io/pen/debug/ogxVRRj/f62b729eb4acde3d4e1f9987ecf04d53)—this is still not common, but I've seen it on several sites in the past month. When the "subscribe" section comes into view, it's visually presented as a modal dialog, but all that happens is the rest of the page is dimmed out, resulting in links underneath the "backdrop" being visually obscured.

venkatvemuri28:
I really liked the example shared.  I was not the call today.  Is the suggestion to replace the failure technique to scrolling modal pop up ? 

I can update that.

mbgower:
On Friday's WCAG 2 TF call, we got into a debate about ways to fail semi-transparent overlays that are non-modal. I believe @scottaohara, @kfranqueiro and some others were saying 4.1.2 Name, Role, Value can be used to fail these.

I was unsuccessful in describing a few scenarios where I thought this wouldn't work, so I thought I'd supply a bit more info.

First off, lightbox effects -- that is, a rendering of the background of a dialog that degrades its visual appearance, and implies it is inert -- tend to be applied to modal dialogs. But isn't it possible to use a lightbox effect on an html5 `dialog` that is not modal? In such a situation, the user can tab out of the dialog and interact with other parts of the screen. I wouldn't call that great design; but my point is that Name, Role, Value doesn't seem to forestall anyone doing this, does it? We got stuck on this, and I'd like to understand how the SC language of 4.1.2 can be used to fail this.

However, that was not the primary scenario I was trying to address. I was trying to address scenarios where _overlays_ are not fully opaque. I'm not talking about a lightbox effect. I'm talking about the opacity of the 'box' drawn onto the screen in which the overlay appears. This [article on glassmorphism by NN/g](https://www.nngroup.com/articles/glassmorphism/) covers some of what I'm talking about.

Here's an example of one in the wild.Text on overlay reads 'Want to unlock our editors' tips for a bigger and more beautiful home?' with two buttons labelled 'Show me how' and 'My home is already perfect'
The overlay is not fully opaque, so that I can _barely_ see details of the shadows of the drapes and other things in the image behind it.

Ignore the size and specific use case. In this screen snippet it appears to obscure almost all the screen, but say this one is covering 1/8 of the viewable area, and is meant as one of those irritating 'try me' patterns. Say it floats over top of other content, so that it ends up covering controls. Say it appears when the page loads and does not take focus right away. In other words, it is exactly the phenomenon intended to be failed by Focus Not Obscured (Minimum).

The problem is that since it is not fully opaque, it does not meet the language of "entirely hidden". I don't see how we can use 4.1.2 to fail this. So this is the kind of situation I'm saying can reference 1.4.11 Non-text Contrast as a means to fail. Hope that helps give some context. 

I can think of no other criteria that can be brought to bear. Can anyone? @detlevhfischer suggested 2.4.3 Focus Order, but that seems shaky, especially in a situation where the overlay is floating. Some would argue that logically it _should_ be in the tab order following the main content, and regardless of where it appears in the order, it is on a different layer than the rest of the content, and therefore trying to say its position in the focus order constitutes a fail seems tenuous.

detlevhfischer:
Would this work (not sure whether this would qualify as mere errata)?
> When a user interface component receives keyboard focus, the component is not entirely ~hidden~ covered due to author-created content.

patrickhlauke:
> I can think of no other criteria that can be brought to bear.

if i understand your issue correctly... we (TetraLogical) usually fail this under Focus Order, as clearly focus should not go to an element that's behind a semi-transparent overlay that visually hints at the fact that the content that's underneath it is inert at that particular point in time.

mbgower:
> if i understand your issue correctly... we (TetraLogical) usually fail this under Focus Order, as clearly focus should not go to an element that's behind a semi-transparent overlay that visually hints at the fact that the content that's underneath it is inert at that particular point in time.

That interpretation could be proposed as a means of addressing the use of a lightbox effect on a non-modal dialog, but it does not seem to address the problem of a translucent, non-modal overlay that appears on a page.

patrickhlauke:
serious question: what site/designer uses a semi-transparent overlay behind a non-modal dialog? your case here seems a bit "theoretical" @mbgower 

if they're showing a semi-transparent overlay, they're visually indicating that the underlying page is inert. generally, clicking on the overlay is prevented, or it closes the dialog. i've never come across a site that explicitly sets a semi-transparent overlay, but *intends* users to still be able to interact with the page behind the overlay...

patrickhlauke:
> The overlay is not fully opaque, so that I can barely see details of the shadows of the drapes and other things in the image behind it.

that example you have there is not a dialog with an overlay. it's not sitting on top of page content behind it (apart from the background image, which is not a focusable element)

patrickhlauke:
having now caught up with some of the earlier discussions (before the whole modal/non-modal/lightbox side discussion) ... let's keep it simple to something like a sticky header or footer, whose background is not fully opaque but semi-transparent. normatively in AA if the focused component is "not entirely hidden" it passes. Of course, what "entirely hidden" means in practice is debatable. if i can still barely see it (opacity of 99.9999%), is it *entirely* hidden? if not, where's the cut-off point? as painful as it may be, for simplicity "entirely hidden" to me would only apply to fully opaque covering over the element. pass at AA, fail at AAA.

patrickhlauke:
Related to this, filed https://github.com/w3c/wcag/pull/4824 which may be controversial....

patrickhlauke:
Moving this to in-progress for now, as the outcome of https://github.com/w3c/wcag/pull/4824 will likely affect this.

venkatvemuri28:
> Moving this to in-progress for now, as the outcome of #4824 will likely affect this.

I’m revising this failure technique and could use guidance on one point. I plan to:

• Add a scroll‑triggered overlay example as suggested ￼.
• Note that the pattern can also fail 1.4.11 (Non‑text Contrast) and 2.4.3 (Focus Order) ￼.
• Clarify that a proper modal (which takes focus) doesn’t fail, but a pseudo‑modal overlay does ￼.
• Align the text with whatever decision we make in PR #4824 regarding semi‑transparent or blurred overlays ￼.

Should I keep the existing modal dialog example and add the scroll‑triggered overlay example, or replace the modal example entirely?

@patrickhlauke

PR #4682, opened 2025-10-10
Updating language around full screen

### LINKED ISSUE #4664: full-screen window glossary term needs work
On a Backlog TF call discussing [G206](https://www.w3.org/WAI/WCAG22/Techniques/general/G206), the use of “on a full-screen window” was noted as not being grammatically correct.
- The preposition should be **in** _not_ **on**
- It‘s contradictory since, if an app is running full-screen, then there‘s no window!

Unfortunately, [on a full-screen window](https://www.w3.org/TR/WCAG22/#dfn-on-a-full-screen-window) is in the glossary.  Further, the definition is dated to the point of being factually incorrect (emphasis added):
> on the _most common_ sized desktop/laptop display with the viewport maximized

The term appears only once in WCAG, as part of [Success Criterion 1.4.8 Visual Presentation](https://www.w3.org/TR/WCAG22/#visual-presentation):
> Text can be resized without assistive technology up to 200 percent in a way that does not require the user to scroll horizontally to read a line of text <ins>on a full-screen window</ins>.

mbgower:
Closes #4664 by removing the one normative occurrence of "on a full screen" in an SC and removing the defined term.

Replaces with a phrase using the existing defined term "viewport"


gundulaniemann:
The original glossary entry, which now according to discussion on 2025-10-10 is subject of deletion, points to the fact that one one hand, the author may rely on commonly used monitors (excluding monitors from the 1980s) and on the other hand should not rely on high end monitors.
If rewording the sentence implies loosing this information, I would not like to do this change.

mbgower:
I don't understand how the updated definition (without a screen size note) alters the basic outcome of this exercise. Regardless of whether the monitor is 12" or 28", if the user maximizes a window they should not have to scroll sideways to read a line of text.
The text that was in the note of the definition is still in the Understanding document. Non-normative text in a definition is no more 'valid' than non-normative text in an understanding document, so no non-normative guidance has been removed.

If we are going to be concerned about scoping on this, there _is_ language in the understanding document noting that this is not intended for small screens:
> The horizontal scrolling requirement is not intended to apply to small-screen devices where long words may be displayed on a single line and require users to scroll horizontally. For the purposes of this requirement, authors should ensure that content meets this requirement on standard desktop/laptop displays with the browser window maximized. 

I'm not sure we want to open up that can of worms. But if we do want to engage with the note, it may be an idea to understand exactly how we want it to be understood and conveyed. 

This SC simply updates the _normative_ wording to make it clearer, as per the original issue. If there is support to perpetuate the existing note, I'm not going to object.

techniques/general/G206.html
```diff
@@ -15,7 +15,7 @@
       </section>
       <section class="example">
          
-            <p>A financial spreadsheet is online. It includes text explaining changes in the housing market in January. Off-screen to the right, there is a column with an explanation of changes to the market in September. The user can horizontally scroll to the September area and read each line of text without any further scrolling when the window size is maximized. </p>
+            <p>A financial spreadsheet is online. It includes text explaining changes in the housing market in January. Off-screen to the right, there is a column with an explanation of changes to the market in September. The user can horizontally scroll to the September area and read each line of text without any further scrolling when the viewport (i.e., window) is maximized. </p>
```
fstrr, line 18: Do we need "(i.e., window)"? We haven't used it anywhere else.

mbgower: In this case, since "window" is used in the original, I think cueing the inexperienced user that "viewport" is _basically_ a synonym for _window_ here helps.

patrickhlauke:
This PR will need to be split into normative and non-normative - unless we make an exception for it (which probably would make sense, as this changes/removes a normative term, and then updates non-normative references to it)

bruce-usab (approved):

patrickhlauke:
taking a step back ... in essence, this is an early prototypical light-touch version of what we now have with the Reflow SC. if we're making this a 2.3 change, rather than a 2.2 erratum, I wonder if we should tighten the point to just

```
Text can be resized without <a>assistive technology</a> up to 200 percent in a way that does not require the user to scroll horizontally to read a line of text.
```

and drop the vague/subjective/variable bit about maximised window/viewport/etc altogether.

while we work on this, may also be worth to change the note - `Content is not required to use these values.` makes no sense for that last point about scrolling...it's not a value.

patrickhlauke:
might even need more nuance, as the "horizontal scrolling" here is clearly just aimed at "western" languages, and excludes vertical text. port similar language from Reflow to this. might also port over minima for width/height in CSS px

bruce-usab:
@patrickhlauke my preference would be to approach this as a 2.2 erratum.  A rewrite of [1.4.8](https://www.w3.org/TR/2024/REC-WCAG22-20241212/#visual-presentation) could be drafted for 2.3.  There are other bullets in 1.4.8 which maybe could/should be less western.

Issue #4193, opened 2025-01-09
Techniques may fail both 1.4.1 Use of Color and 1.3.1 Info and Relationships to convey HTML element names and attributes


charleshalldesign:
All WCAG 2.2 Techniques documents that refer to HTML elements and their attributes use a convention of rendering the text in a <code> element with applied CSS to style the text differently than non code text.

However, the semantic role of <code> is not conveyed in most scenarios, leaving only the break in text elements to convey the text is different than the surrounding text – but not how.

Further, the style applied to this code text includes:
- A change from the body text color of #1d1d1d to a blue color of #036.
- This difference has a luminance contrast of only 1.3:1 which is too subtle for most to perceive and differentiate.
- A change of font from Noto Sans to Noto Sans Mono.
- This difference is too subtle for most to perceive and differentiate.

This leaves some people the possibility to perceive a visible difference and that information and relationships are being conveyed. While there is no perceivable difference when the presentation is audible.

Example:
The word ‘select’ within the first sentence of the second paragraph of the ‘Description’ section of the [H85 Technique](https://www.w3.org/WAI/WCAG22/Techniques/html/H85).

May fail 1.4.1
Because the change in font only affect some pixels of some characters (in the example, only the bottom left edge of the ‘l’ character has any perceivable change), only color is a perceivable change for the entire word.

May fail 1.3.1
Because there is an attempt to convey semantics in the audible presentation by use of <code> element, but not all instances are preceded or followed by text that conveys this by stating “element” or “attribute”.

Suggestions:

1. to better attempt to audibly convey the code element role, try using the roledescription attribute
2. to better attempt to visibly convey the difference in text, try using a different monospace font and a different color hue that has a sufficient luminance contrast difference with both the background and the body text.

giacomo-petri:
** First Draft Working Group Response**

> May fail 1.4.1
> Because the change in font only affect some pixels of some characters (in the example, only the bottom left edge of the ‘l’ character has any perceivable change), only color is a perceivable change for the entire word.

As noted, a different font family (monospace) is used, so the distinction is not based solely on color, meaning 1.4.1 is not failing in this case. Also, the dark blue color is barely perceptible visually (even for users without low vision). Personally, I find the difference in font-family (with its added letter spacing) more helpful than color for recognizing these words.

> May fail 1.3.1
> Because there is an attempt to convey semantics in the audible presentation by use of `<code>` element, but not all instances are preceded or followed by text that conveys this by stating “element” or “attribute”.

[...]

PR #4683, opened 2025-10-10
Tweak inline code and preformatted blocks

patrickhlauke:
Adds a contrasting border to preformatted blocks (mainly for code blocks, but a few others), gives a subtle style to inline code elements in prose

Relates to https://github.com/w3c/wcag/issues/4193 and applies suggestion 1 (the box around the `pre`) and a slightly tweaked variant of suggestion 2b (the "subtle" style for `code`, with some extra selector shenanigans to avoid having the background when it's inside a link or already in a `pre` block)

EDIT: as a quick before/after, see https://www.w3.org/WAI/WCAG22/Techniques/html/H85 and the preview here https://deploy-preview-4683--wcag2.netlify.app/techniques/html/h85


patrickhlauke:
@kfranqueiro do changes to CSS not get reflected in preview? https://deploy-preview-4683--wcag2.netlify.app/techniques/html/h85

kfranqueiro:
Thank you, Mario, but our CSS is in another castle!

`additional.css` is only used by the `wcag20` folder. You probably want `base.css`.

patrickhlauke:
Thanks for the pointer @kfranqueiro ... updated, and the preview now shows up my change :)

adampage:
Hmm, yeah, to @kfranqueiro’s [point](https://github.com/w3c/wcag/issues/4193#issuecomment-3366319294), these [two consecutive code blocks in ARIA19’s Example #1](https://deploy-preview-4683--wcag2.netlify.app/techniques/aria/aria19#example-1-injecting-error-messages-into-a-container-with-role-alert-already-present-in-the-dom) appear to be _one_.

Perhaps some explicit `margin` for `<pre>`s, or try relocating those new styles to `pre > code`?

Or does those two `<code>` elements need to be contained by only one `<pre>`, in that particular case? We could search the repo for adjacent `pre > code + code` instances and tweak them in the markup.

patrickhlauke:
Example of `pre` being used for something OTHER than code blocks https://deploy-preview-4683--wcag2.netlify.app/techniques/text/t1

kfranqueiro:
RE multiple `code`s in one `pre`, I think Patrick had identified and separated some of those in the original big PR that I had split things out of, so yeah we could hunt those down (or make the build system do it). They technically cause no issue for highlight.js, but they do potentially complicate these styles.

I guess there's also still the question of whether the styles are worth the trouble, since I think we had a few votes of ambivalence or actual -1 in the issue.

On that note, some other edge cases to consider for inline code:

- Headings, e.g. h1 and h3 in [H39](https://deploy-preview-4683--wcag2.netlify.app/techniques/html/h39)
- This [SCR20 mappings table](https://deploy-preview-4683--wcag2.netlify.app/techniques/client-side-script/scr20#SCR20_mappings) might turn out to look a little odd, but other cases inside `td`s that are inline with other text look fine
- There are a few in [G208's Examples](https://deploy-preview-4683--wcag2.netlify.app/techniques/general/g208#examples) that effectively look like they should be proper highlighted code blocks, which we could decide to tackle separately

patrickhlauke:
For things like multiple code blocks in a single `pre`, and cases like G208 ... I'd say these new styles actually visually surface issues in the original markup that should be addressed in the markup.

Regarding `code` in headings, I actually intentionally kept that in, but those can be easily excluded by expanding the `:not(:is(pre, a)) > code` selector.

ditto things like `code` as direct descendant of a `td` or `th`

patrickhlauke:
I *think* I caught all instances where we have two `code` blocks inside the same `pre` https://github.com/w3c/wcag/pull/4683/commits/70c4816753754b0639e0b4b4b8a7e67b2d66e2ef

patrickhlauke:
I'll add here that we can also soften the border around the `pre`s a bit, as it's not the *only* visual indication that it's a code block (indentation, margin, different typeface, all provide visual hints). so we could make the border less contrasty (e.g. using `var(--line-grey);` like the border around the "About this technique" at the top of techniques for instance)

EDIT: done that with the next commit

Example (zoomed out to show both the "About this technique" and a newly styled code block):

> screenshot of H85 with the new styles applied

Obviously personal preference on whether or not these styles are even worth pursuing, but for my own part, I find them much more immediately "skimmable" as I read through a technique, seeing exactly where code blocks are, where HTML elements/attributes are being mentioned (rather than only finding them by carefully reading a sentence), etc. Arguably yes, because these styles are not all `3:1`, some users will not be seeing them ... but I think these styles fall more under "eye candy" than actual "required for understanding/meaning"...

patrickhlauke:
thanks for spotting the `</code> ... <code>` variants there @kfranqueiro

PR #4662, opened 2025-10-03
fix pause, stop, hide link

fstrr:
Closes #4661

PR #4693, opened 2025-10-16
Update timing-adjustable.html

### LINKED ISSUE #4661: Broken link on 2.2.1 Timing adjustable Understanding page
The link to the Pause, Stop, Hide Understanding page inside this paragraph ([2.2.1 Timing adjustable](https://www.w3.org/WAI/WCAG22/Understanding/timing-adjustable#intent)):

> These are situations which do include time limits, but the content is still available to the user because it has controls for accessing it, as specified in [2.2.2 Pause, Stop, Hide](https://www.w3.org/WAI/guidelines/sc/20/pause-stop-hide.html).

is broken.

mbgower:
Closes #4661


mbgower:
Dulicate of #4662

PR #4316, opened 2025-04-04
Improve the SC 2.1.2 No Keyboard Trap Understanding

yatil:
Very drafty Friday afternoon start of an improvement here.

PR #4726, opened 2025-11-02
Expand 2.1.2 understanding to make it more ... understandable

patrickhlauke:
Ports over most of the ideas from https://github.com/w3c/wcag/pull/4316, but extends/clarifies them further.

Expands on what "standard exit methods" are (and the fact that WCAG doesn't actually normatively define these, that they're hardware/OS/UA dependent, and that authors/auditors will need to use their judgement here). Also reiterates that even if untrapping involves "non-standard" methods, that's fine as long as users are informed how to do it (but still needs to be using the keyboard).

Also cleans up the markup a bit.

Tweaks the modal dialog example (focusing on the focus trapping, rather than the opposite - which parts of the page don't receive focus anymore) and adds a new example for Rich Text Editors.

Supersedes #4316 (as that PR didn't allow for changes to be pushed to it, including rebasing it to the latest main branch)

Preview: https://deploy-preview-4726--wcag2.netlify.app/understanding/no-keyboard-trap


patrickhlauke:
@kfranqueiro "untrapping" flagged as a spelling error...

kfranqueiro:
> @kfranqueiro "untrapping" flagged as a spelling error...

Slightly amusing, since I had already needed to add "untrap". Add "untrapping" on a new line in `/custom-words.txt` right below that one.

patrickhlauke:
ah, perfect (and yes, chuckling as well) thanks @kfranqueiro

understanding/20/no-keyboard-trap.html
```diff
@@ -11,48 +11,64 @@ <h1>Understanding No Keyboard Trap</h1>
...
+         not fail the requirements of this criterion, as long as the user knows how to "untrap" the focus
+         and leave that component.</p>
+
+      <p>Keyboard focus is not considered trapped when the user can navigate away from a component
+         using only a keyboard interface, and if it only requires unmodified arrow or <kbd>Tab</kbd> keys
+         or other "standard exit methods". This specification does not define what constitutes a
+         "standard exit method" – this is dependent on the user's hardware, user agent, and operating system,
+         and as such will require some interpretation from authors and auditors.
+         In most desktop environments, pressing the <kbd>Esc</kbd> key is a commonly used "standard exit method".
+         On mobile operating systems, the use of a physical "back" button on the device or external keyboard can
```
mbgower: I'm a little leery of this statement. Using the "back" button in this context seems like a user agent method of returning to a prior page. That will let the user go to a different web page, but it has not resolved or addressed the keyboard trap on the current page.

mbgower: Otherwise, looks good!

patrickhlauke: the "back" button (on Android in particular) doesn't (always) map to "previous page", but is used contextually - when you're in a modal dialog, it acts as a method to close the dialog.

However, this bit could probably use some massaging, yes.

adampage (approved):

bruce-usab (approved):
See [diff preview](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FUnderstanding%2Fno-keyboard-trap&doc2=https%3A%2F%2Fdeploy-preview-4726--wcag2.netlify.app%2Funderstanding%2Fno-keyboard-trap#intent) (starting at “Intent”).

understanding/20/no-keyboard-trap.html
```diff
@@ -11,48 +11,63 @@ <h1>Understanding No Keyboard Trap</h1>
...
+
+      <p>Keyboard focus is not considered trapped when the user can navigate away from a component
+         using only a keyboard interface, and if it only requires unmodified arrow or <kbd>Tab</kbd> keys
+         or other "standard exit methods". This specification does not define what constitutes a
+         "standard exit method" – this is dependent on the user's hardware, user agent, and operating system,
+         and as such will require some interpretation from authors and auditors. Generally, in most
+         environments with a physical keyboard, pressing the <kbd>Esc</kbd> key is a commonly
+         used "standard exit method", but other platform-specific methods may be available.</p>
+
+      <p>If untrapping focus requires a different method (rather than unmodified arrow or <kbd>Tab</kbd> keys or
```
patrickhlauke:
```suggestion
      <p>If untrapping focus requires a different method (rather than unmodified arrow keys, the <kbd>Tab</kbd> key, or
```

patrickhlauke:
@detlevhfischer Not sure if you deleted this comment, or if github is having a moment and not displaying it...but just in case, the short answer here is: that's irrelevant from a "can the user untrap focus" point of view. They can by hitting "OK". whether doing so has other consequences is orthogonal to the requirement of the SC.

It would also be really crappy UI design, as if they offer the option to bail out to keyboard users but not pointer users...they should really not use an "OK"-only dialog, but an "OK"/"Cancel" one.

detlevhfischer:
@patrickhlauke I was commenting on the PR before reading it all :)  - below there is then sufficient stuff on "standard exit methods", which is why I deleted my comment.

PR #4738, opened 2025-11-07
wcag.json: Strip role="heading" and aria-level from content

### LINKED ISSUE #4725: WCAG JSON Feedback - one tiny issue? overall very good report
I just wanted to share an experience of using the WCAG22 JSON. I made a little app, started using the JSON, and found some formatting/content issues. I felt sort of excited that I maybe had discovered some minor issues to raise in the Repository to report. After doing a second pass at my app, I realized that the issues were all apparently on my side and the JSON is pretty much blemish-free- except maybe one tiny piece of html aria formatting to be consistent, hardly qualifies as an issue at all. I am learning an incredible amount making the little app, and clicked many links back to WAI generated from JSON. It seemed really approachable to grab the pieces of URLs like the technology ID and link it back to WAI, and it just worked.

Thank you for releasing this form of structured information. The JSON has encouraged creativity for me.

I did find one tiniest detail in my entire exploration of the JSON Document:

This is in the content section of the 2.1.2 criteria:
```
"id": "no-keyboard-trap",
"num": "2.1.2",
...
"content": "<p>If keyboard focus can be moved to a component of the page using a <a href=\"https://www.w3.org/TR/WCAG22/#dfn-keyboard-interface\" title=\"interface used by software to obtain keystroke input\">keyboard interface</a>, then focus can be moved away from that component using only a keyboard interface,\n      and, if it requires more than unmodified arrow or tab keys or other standard exit\n      methods, the user is advised of the method for moving focus away.\n   </p>\n   \n   <div class=\"note\" role=\"note\" id=\"issue-container-generatedID-17\"><div role=\"heading\" class=\"note-title marker\" id=\"h-note-17\" aria-level=\"5\"><span>Note</span></div><p class=\"\">Since any content that does not meet this success criterion can interfere with a user's\n      ability to use the whole page, all content on the web page (whether it is used to\n      meet other success criteria or not) must meet this success criterion. See <a href=\"https://www.w3.org/TR/WCAG22/#cc5\">Conformance Requirement 5: Non-Interference</a>.\n   </p></div>",
```

Specifically, in the content property value is:

`aria-level=\"5\"`

The effect of this is that inserting some of the data into my application causes it to have a level 5 heading out of order, because of the apparent aria-level. Im not sure how to override or fix this, Im sure I can fix it, it is just the only thing I found in my entire testing of the JSON document to report.

**Suggestion: remove aria level from this content value in the JSON? if it is not needed**

kfranqueiro:
Fixes #4725.

This causes no change to the build's HTML output; its only impact is to remove these two attributes from note markers in `wcag.json` output.

### DIFF
--- a/11ty/guidelines.ts
+++ b/11ty/guidelines.ts
@@ -266,6 +266,10 @@ const loadRemoteGuidelines = async (version: WcagVersion, stripRespec = true) =>
   // so that names parse consistently
   $("bdi").remove();
 
+  // Remove role="heading" + aria-level from notes/issues, as they cause more harm than good
+  // (especially in context of content reuse via wcag.json export)
+  $("[role='note'] .marker").removeAttr("role").removeAttr("aria-level");
+
   if (!stripRespec) return $;
 
   // Re-collapse definition links and notes, to be processed by this build system


iadawn (approved):
Presumably this will need to come out when respec is aligned with our way of thinking? :)

kfranqueiro:
> Presumably this will need to come out when respec is aligned with our way of thinking? :)

It'd become a no-op, so it wouldn't break anything, but it'd make sense to clean up at that point.

PR #4741, opened 2025-11-08
Conform phrasing in exception of 1.4.13

bruce-usab:
Replaces “controlled by the user agent” with “determined by the user agent” because that is the phrasing used by other SC.

SC 1.4.13 Content on Hover or Focus is the only SC using **controlled** by the user agent.  Five other SC phrase this as **determined** by the user agent.  Since we mean the same thing in all five places, we should use the same phrasing in all five places.  The unique phrasing in 1.4.13 implies that there is some kind of difference.

[Preview diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FTR%2F2024%2FREC-WCAG22-20241212%2F&doc2=https%3A%2F%2Fdeploy-preview-4741--wcag2.netlify.app%2Fguidelines%2F#content-on-hover-or-focus) using anchor to SC 1.4.3


patrickhlauke (approved):

guidelines/sc/21/content-on-hover-or-focus.html
```diff
@@ -19,9 +19,9 @@ <h4>Content on Hover or Focus</h4>
 
   </dl>
     	
-    	<p>Exception: The visual presentation of the additional content is controlled by the <a>user agent</a> and is not modified by the author.</p>
+    	<p>Exception: The visual presentation of the additional content is determined by the <a>user agent</a> and is not modified by the author.</p>
```
WilcoFiers, line 22: I agree that "controlled" isn't the best word for this, but determined is worse IMO. Pulling up a dictionary, this is the explanation that bothers me:

> ascertain or establish exactly by research or calculation.

The presentation of the page is calculated by the user agent based on user, author and user agent CSS. So the user agent **always** determines the presentation. 

I think a better word to use here is "set":

```suggestion
    	<p>Exception: The visual presentation of the additional content is set by the <a>user agent</a> and is not modified by the author.</p>
```

WilcoFiers (dismissed):

guidelines/sc/21/content-on-hover-or-focus.html
patrickhlauke: in that case, do the other SCs that use

> determined by the user agent and is not modified by the author

(that this PR tries to harmonise to) also need to be changed to "set"? or is only this one distinct in its meaning compared to those?

1.4.11 Non-text Contrast
2.4.13 Focus Appearance
2.5.5 Target Size (Enhanced)
2.5.7 Dragging Movements
2.5.8 Target Size (Minimum)

dbjorge: @WilcoFiers I disagree; I agree that if you pick the specific definition of "determine" you quoted it's unclear, but I think [most definitions of "determine"](https://www.merriam-webster.com/dictionary/determine) imply "making or settling a decision" moreso than most definitions of "set" do, which is the operative thing here.

I agree with Patrick that the more important piece here is consistency in wording across the SCs, which is the main thing this PR is trying to resolve. I slightly prefer "determined" to "set", but I would strongly prefer "pick either one of them and use it consistently across all SCs" over "use 'set' here and 'determined' elsewhere".

WilcoFiers: Alright, fair enough. This wasn't a strong objection. I can live with the change as is. Withdrawing my change request. @patrickhlauke @dbjorge thank you for the response, and + @bruce-usab for the hard work.

bruce-usab:
Ping to @awkawk per 9 June [Errata and updates for WCAG 2](https://lists.w3.org/Archives/Public/w3c-wai-gl/2026AprJun/0039.html) 9 June email to list with additional context.  Please comment or remove 👎

kfranqueiro:

Issue #4756, opened 2025-11-18
F33 technique - Text content is incorrectly referred to as a "table"


baldino-m:
On the [F33 technique](https://www.w3.org/WAI/WCAG22/Techniques/failures/F33) page, the Example 1 shows a two-column layout obtained with white spaces. The content is referred to as a "table":

> If this table was to be interpreted and spoken by a screen reader it would speak the following lines:

which is not appropriate (it was probably copied and pasted from the very similar [F34 technique](https://www.w3.org/WAI/WCAG22/Techniques/failures/F34) which shows an incorrect table formatting with the same technique).

I propose to re-phrase as follows:

"If this content was to be interpreted and spoken by a screen reader it would speak the following lines:"

PR #4758, opened 2025-11-18
Re-phrased Example 1 description for F33 technique

baldino-m:
closes [issue #4756](https://github.com/w3c/wcag/issues/4756)


patrickhlauke (approved):

techniques/failures/F33.html
```diff
@@ -34,7 +34,7 @@
 range of people with disabilities,        wide variety of assistive technologies.
 </pre>
          
-            <p>If this table was to be interpreted and spoken by a screen reader it
+            <p>If this content was to be interpreted and spoken by a screen reader it
```
kfranqueiro: While we're here, we might want to consider updating the verb tense phrasing:

```suggestion
            <p>If this content were to be interpreted and spoken by a screen reader it
```

techniques/failures/F33.html
```diff
@@ -34,7 +34,7 @@
...
                                 would speak the following lines:</p>
```
patrickhlauke:
```suggestion
            <p>Screen readers announce this content as:</p>
```

why could even just drop the whole hypothetical and just state a simple fact/truth

Issue #4378, opened 2025-05-02
Techniques Of Future Past


fstrr:
There's quite a few "potential future technique" comments in documents, some of which have been there for years. Which have the sands of time running low for them and which are the ones for you and me? If you had the time to lose, an open mind and time to choose, would you care to take a look?

### [Concurrent input mechanisms](https://www.w3.org/WAI/WCAG22/Understanding/concurrent-input-mechanisms.html)
1. Only using high-level, input-agnostic event handlers, such as focus, blur, click, in Javascript
2. Registering event handlers for keyboard/keyboard-like and pointer inputs simultaneously in Javascript; see [Example 1 in Pointer Events Level 2](https://w3c.github.io/pointerevents/#examples)

### [Content On Hover Or Focus](https://www.w3.org/WAI/WCAG22/Understanding/content-on-hover-or-focus)
1. ARIA: Using role="tooltip"
2. CSS: Using hover and focus pseudo classes
3. Failure to make content dismissable without moving pointer hover or keyboard focus
4. Failure to meet by content on hover or focus not remaining visible until dismissed or invalid

### [Label In Name](https://www.w3.org/WAI/WCAG22/Understanding/label-in-name)
1. If an icon has no accompanying text, consider using its hover text as its accessible name
2. Accessible name contains the visible label text, but the words of the visible label are not in the same order as they are in the visible label text
3. Accessible name contains the visible label text, but one or more other words are interspersed in the label

### [Pointer Cancellation](https://www.w3.org/WAI/WCAG2/Understanding/pointer-cancellation)
1. Touch events are only triggered when touch is removed from a control

### [Reflow](https://www.w3.org/WAI/WCAG2/Understanding/reflow)
1. CSS, Reflowing simple data tables
2. CSS, Fitting data cells within the width of the viewport
3. Mechanism to allow mobile view at any time
4. Alternate view supporting Reflow for otherwise excepted content

### [Accessible Authentication (Enhanced)](https://www.w3.org/WAI/WCAG22/Understanding/accessible-authentication-enhanced) and [Accessible Authenication Minimum](https://www.w3.org/WAI/WCAG22/Understanding/accessible-authentication-minimum)
1. Providing WebAuthn as an alternative to username/password
2. Providing a 3rd party login using OAuth
3. Using two techniques to provide 2 factor authentication

### [Redundant Entry](https://www.w3.org/WAI/WCAG22/Understanding/redundant-entry)
1. Not requesting the same information twice

[...]

PR #4770, opened 2025-11-24
New Failure Technique for 1.4.13 Content on hover or focus when popup content cannot be dismissed

detlevhfischer:
I have attempted to write a Failure Technique for 1.4.13 Content on hover and focus.
Preview: https://deploy-preview-4770--wcag2.netlify.app/techniques/failures/f114

The concern I have is whether the Failure Technique should not also allow (and check) for other author-specified and documented ways of dismissing popup content, such as -- to give a morwe or less random example -- a documented "pressing ALT + C closes popup content". This may reside on the page or in generic settings where discoverability is poor. The normative text does not delimit ways to dismiss content. Not sure if this can be added as a third point and how that would be phrased, something like "Check if there are documented methods for dismissing popup content. If this is the case, check if the documented method succeeds in closing popup content called up by pointer hover or keyboard focus." If this is difficult to nail down, this could be an indication that a watertight FAILURE Technique is not feasible here.


mbgower:
@detlevhfischer is this addressing an existing github issue?

mbgower (commented):
@detlevhfischer, failure techniques normally follow a set structure. A succinct description of the scenario that is actually the failure (not a regurgitation of the SC), then an example, and then a test procedure that is in a very structure format. Here's how I'd suggest the description and test procedure could start.
### Description
This describes the failure condition that results when content that appears on hover or focus cannot be dismissed in a way that meets the requirements of 1.4.13. The success criterion lists three conditions that must be true, and one of these is that the content that appears must be dismissible:
> A [mechanism](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html#dfn-mechanism) is available to dismiss the additional content without moving pointer hover or keyboard focus, unless the additional content communicates an [input error](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html#dfn-input-error) or does not obscure or replace other content;"

### Tests
#### Procedure
For content that appears on hover and focus:

detlevhfischer:
@mbgower 
> @detlevhfischer is this addressing an existing github issue?

It is picking up two of the future techniques listed here: https://github.com/w3c/wcag/issues/4378

...which made me believe such techniques should be welcome. I have followed the scenario of description, examples, test that you point at, keeping it short. What do you think is missing?

As to your suggestion to explain the SC (the three things that are covered) I personally think that is the task of the understanding doc. The failures I have drafted cover one of these points individually (dismissability and hoverability of new popup content) so IMHO, dilating into the other aspects does not seem helpful in applying the respective failure technique.

techniques/failures/F114.html
```diff
@@ -0,0 +1,56 @@
...
+   </section>
+  <section id="description">
+      <h2>Description</h2>
+      <p>Content elements may respond to receiving keyboard focus or pointer hover by inserting additional popup content that may now cover other parts of content. For users, especially those that use strong page magnification, this addition of popup content without prior activation can be disorienting in different ways:</p>
+        <ul>
+          <li>the new popup content can cover content that the user may want to access</li>
+          <li>The new popup content may not appear fully in the magnified viewport, necessitating a move of the viewport by panning, tabbing, or pointer focussing in order to access the popup content</li>
+        </ul>
+     <p>Since popup content on hover and focus can unexpectedly cover page content that users may want to access, it is important that there is an easy way to dismiss such popup content. The most common method for that is to press the `ESC` key. Another possible method is activating the pointer at the current location.</p>
```
fstrr, line 17:
```suggestion
     <p>Since popup content on hover and focus can unexpectedly cover page content that users may want to access, it is important that there is an easy way to dismiss such popup content. The most common method for that is to press the <kbd>ESC</kbd> key. Another possible method is activating the pointer at the current location.</p>
```

techniques/failures/F114.html
```diff
@@ -0,0 +1,56 @@
...
+   </section>
+  <section id="examples">
+     <h2>Examples</h2>
+      <ul>
+         <li><strong>Custom tooltip:</strong> When focussing an icon with a question mark that is placed after an input field, a popup window with an explanation of the expected input appears that covers a part of the form. Pressing the ESC key or activating the pointer at the same position fails to close the submenu.</li>
+        <li><strong>Sub-menu:</strong> When the mouse pointer is moved over the entry in a menubar, a sub-menu appears. Pressing the `ESC` key or activating the pointer at the same position fails to close the sub-menu.</li>
```
fstrr, line 23:
```suggestion
        <li><strong>Sub-menu:</strong> When the mouse pointer is moved over the entry in a menubar, a sub-menu appears. Pressing the <kbd>ESC</kbd> key or activating the pointer at the same position fails to close the sub-menu.</li>
```

techniques/failures/F114.html
```diff
@@ -0,0 +1,56 @@
...
+         <li><strong>Sub-menu over decorative content or whitespace:</strong> When the mouse pointer is moved over the entry in a menubar, a sub-menu appears over a decorative image or whitespace. Across different viewport widths, the sub-menu never covers information. This is not a failure.</li>
+      </ul>
+   </section>
+   <section id="tests"><h2>Tests</h2>
+      <section class="procedure"><h3>Procedure</h3>
+         <ol>
+           <li>
+             <ul>
+                <li>Using the tab key, navigate through interactive content. </li>
+                <li>If popup content appears that hides information, press the `ESC` key and check whether this closes the popup content</li>
```
fstrr, line 33:
```suggestion
                <li>If popup content appears that hides information, press the <kbd>ESC</kbd> key and check whether this closes the popup content</li>
```

techniques/failures/F114.html
```diff
@@ -0,0 +1,56 @@
...
+                    <li>Activate the pointer device at the current pointer position and check whether this closes the popup content</li>
+                  </ul>
+             </ul>
+           </li>
+         </ol>
+      </section>
+      <section class="results">
+        <h3>Expected Results</h3>
+         <ul>
+            <li>If popup content on keyboard focus does not disappear when pressing `ESC` or popup content on hover does not disappear when pressing `ESC` or activating the pointer at the same position then this failure
```
fstrr, line 51:
```suggestion
            <li>If popup content on keyboard focus does not disappear when pressing <kbd>ESC</kbd> or popup content on hover does not disappear when pressing <kbd>ESC</kbd> or activating the pointer at the same position then this failure
```

techniques/failures/F114.html
```diff
@@ -0,0 +1,56 @@
...
+        <ul>
+          <li>the new popup content can cover content that the user may want to access</li>
+          <li>The new popup content may not appear fully in the magnified viewport, necessitating a move of the viewport by panning, tabbing, or pointer focussing in order to access the popup content</li>
+        </ul>
+     <p>Since popup content on hover and focus can unexpectedly cover page content that users may want to access, it is important that there is an easy way to dismiss such popup content. The most common method for that is to press the `ESC` key. Another possible method is activating the pointer at the current location.</p>
+   </section>
+  <section id="examples">
+     <h2>Examples</h2>
+      <ul>
+         <li><strong>Custom tooltip:</strong> When focussing an icon with a question mark that is placed after an input field, a popup window with an explanation of the expected input appears that covers a part of the form. Pressing the ESC key or activating the pointer at the same position fails to close the submenu.</li>
```
fstrr, line 22:
```suggestion
         <li><strong>Custom tooltip:</strong> When focussing an icon with a question mark that is placed after an input field, a popup window with an explanation of the expected input appears that covers a part of the form. Pressing the <kbd>ESC</kbd> key or activating the pointer at the same position fails to close the submenu.</li>
```

techniques/failures/F114.html
```diff
@@ -0,0 +1,56 @@
...
+                <li>Using the tab key, navigate through interactive content. </li>
+                <li>If popup content appears that hides information, press the `ESC` key and check whether this closes the popup content</li>
+            </ul>
+          </li>
+          <li>
+             <ul>
+                <li>Move the pointer over interactive elements. </li>
+                <li>If popup content appears that hides information:
+                  <ul>
+                    <li>press the ESC key and check whether this closes the popup content</li>
```
fstrr, line 41:
```suggestion
                    <li>press the <kbd>ESC</kbd> key and check whether this closes the popup content</li>
```

fstrr (commented):
I think the backticks around some ESC references were meant to be markdown, which the WCAG 2 repo doesn't support, so I replaced them with `kbd` elements. Also added some more of those that seemed to be missing.

Issue #2639, opened 2022-08-25
Use term "natural language" instead of "human language" in glossary?


himorin:
WCAG 2.2 spec defines a term 'human language' in glossary, which is similar to 'natural language' in i18n glossary

https://w3c.github.io/wcag/guidelines/22/#dfn-human-language-s
> human language
> language that is spoken, written or signed (through visual or tactile means) to communicate with humans 

https://www.w3.org/TR/i18n-glossary/#def_natural_language
> Natural Language (sometimes just language) refers to the spoken, written, or signed communications used by human beings. See also localizable text and syntactic content. 

i18n would like to hear feeling whether to replace with more generally used term "natural language".

bruce-usab:
+1 for natural language but does anyone know how people who advocate for Plain Language distinguish between programming languages and human languages?  My first impression is that "natural language" is comp sci jargon.

alastc:
It is before my time (WCAG 2.0), but I assume it was chosen as a better way of differentiating from programming languages. Or perhaps it raises the question: What is an un-natural language? Human vs programming seems more concrete to me.

It seems like a small change, but it would have a ripple effect to a lot of other documentation, so changing the term is unlikely to get support from the group.

bruce-usab:
Yes, point was to differentiate from programming languages.

Am I correct to that WCAG 2.0 definition predated i18n definition?

cstrobbe:
The term "human language" was chosen instead of "natural language" not primarily to differentiate them from programming languages but for a different reason. The term "natural language" may be defined as "any language that has evolved naturally in humans through use and repetition without conscious planning or premeditation" (I'm quoting [Wikipedia's](https://en.wikipedia.org/wiki/Natural_language) definition for convenience). When I was studying linguistics, a "natural language" was also defined as one that people learn as their native language.
So the term "natural language" excludes constructed languages in which web content may also be created, unless there are people who learn them as their native language. [Esperanto](https://en.wikipedia.org/wiki/Esperanto) is said to have native speakers (though not necessarily as their *only* native language) but most don't (or have very few, such as [Ido](https://en.wikipedia.org/wiki/Ido)). These may seem like edge cases, but their are Wikipedia editions in [Esperanto](https://eo.wikipedia.org/wiki/Vikipedio:%C4%88efpa%C4%9Do) and [Ido](https://io.wikipedia.org/wiki/Frontispico), for example.

[...]

PR #4787, opened 2025-12-03
Tweak definition for "human language"

patrickhlauke:
Closes https://github.com/w3c/wcag/issues/2639 which raised a concern about the mismatch of terms in WCAG ("human language") and the i18n glossary ("natural language"). Keep the actual term as "human language" (as it's too pervasive in normative and non-normative documents to just change), while adding the mention of "natural" (as well as "constructed" - for instance, Esperanto) in the definition itself.


bruce-usab:
Discussed on TF call 9 Jan 2026 call and we had concurrence with _not_ making a change to the glossary term.  Among other points raised, there seems some movement within I8n for favoring “human language” over “natural language”.

PR #4794, opened 2025-12-10
Update F74 not labeling a media alternative

### LINKED ISSUE #4734: 1.2.8 Media Alternative (Prerecorded) - Failure technique F74 incorrectly applied to this SC
Think that the application of Failure technique F74 is misplaced on SC 1.2.8. 

In reading the understanding document for 1.2.8 Media Alternative - this SC is about requiring a text-based alternative for synchronized media to address the user needs of someone who has difficulty see and hearing the content in the media. 

However, F74's failure is about labeling "synchronized media alternative to text" specifically. Not about properly labeling a text version type of alternative to synchronized media.  F74 needs to cover both cases or a new failure technique added to cover the case where text is being provided as the alternative to synchronized media.

detlevhfischer:
* Attempt to extend the scope of the Failure technique by generalising content and media alternative
* Addition of an example with a text alternative to audio files

May need a qualification if we think that this failure would not apply to pages where there is just one type of content immediately followed by its alternative. In these cases, there may be no difficulty in associating content and alternative.

Closes #4734 

Preview: https://deploy-preview-4794--wcag2.netlify.app/techniques/failures/f74

### DIFF
--- a/11ty/techniques.ts
+++ b/11ty/techniques.ts
@@ -399,7 +399,7 @@ export async function getTechniquesByTechnology(guidelines: FlatGuidelinesMap) {
       technology,
       title,
       titleHtml,
-      truncatedTitle: title.replace(/\s*\n[\s\S]*\n\s*/, " … "),
+      truncatedTitle: title.trim().replace(/\s*\n[\s\S]*\n\s*/, " … "),
     });
   }
 
--- a/techniques/aria/ARIA1.html
+++ b/techniques/aria/ARIA1.html
@@ -28,7 +28,7 @@
          <h3>Using <code>aria-describedby</code> to associate instructions with form fields</h3>
          
             <p>Sample form field using <code>aria-describedby</code> to associate instructions with form fields while there is a form label.</p>
-<pre xml:space="preserve"><code class="language-html">&lt;form&gt;
+<pre><code class="language-html">&lt;form&gt;
   &lt;label for="fname"&gt;First name&lt;/label&gt;
   &lt;input aria-describedby="int2" autocomplete="given-name" id="fname" type="text"&gt;
   &lt;p id="int2"&gt;Your first name is sometimes called your "given name".&lt;/p&gt;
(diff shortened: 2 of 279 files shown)


techniques/failures/F74.html
```diff
@@ -1,9 +1,13 @@
...
+         
+            <p>A page contains several podcast audio files each of which has a title indicating its topic. Further down on the same page, there is a section with text transcriptions of these podcasts. The headings of the text transcriptions just contain the dates of the podcasts without referring to the podcast titles. It is therefore difficult or impossible for users to identify clearly which transcription belongs to which podcast.</p>
+         
          <h3>Synchronized media alternatives provided elsewhere on page</h3>
          
             <p>A page with instructions to complete a tax form provides a prose description of the fields to complete, data to provide, etc. Additionally, a synchronized media alternative provides spoken instructions, with video of a person completing the section being discussed in the audio. Although both versions are provided on the page, the synchronized media version is provided elsewhere on the page and is not clearly labeled with the part of the text for which it is a substitute. This makes it difficult for users encountering the text to find it, and users encountering the video do not know where its text alternative is.</p>
```
patrickhlauke, line 13: I know this is original text, not touched by this SC, but ... that's some seriously confusing/awkward wording here. 

> Although both versions are provided on the page, the synchronized media version is provided elsewhere on the page

On first reading, this makes it sound like there's a copy of the synchronized media "elsewhere" on the page, that there's two of them.

I think this was *trying* to say that the synchronized media and the prose are not in proximity, close together, near to each other. Might be worth tweaking this as well (not that it fixes the fundamental problem I have with this failure, and the SC)

patrickhlauke:
the fundamental problem I still have with this failure in general (and the normative wording and understanding for 1.2.2 in particular) is that the failure technique (or anything else, unless i'm mistaken) doesn't actually give any example of how alternatives SHOULD be labeled. what does "clearly labeled" mean? what does it look like, in practice? F74 manages to even skirt that by, in essence, saying: here's examples of alternatives that aren't clearly labeled. they fail because they're not clearly labeled. ok, great...

we might want to hold this back/work on it a bit more to somehow at least acknowledge/address this fundamental problem

PR #4797, opened 2025-12-10
WCAG2ICT recommendation: change section / region for 1.3.6 Identify Purpose (AAA)

bruce-usab:

---

WCAG2ICT task force is closely reviewing AAA success criteria.

<abbr title="Bottom Line Up Front">BLUF</abbr>:  WCAG2ICT recommends replacing “region” in SC 1.3.6 with “section” and removing the former from the glossary.

1.3.6 Identify Purpose (AAA):
> In content implemented using markup languages, the purpose of [user interface components](https://www.w3.org/TR/WCAG22/#dfn-user-interface-components), icons, and [regions](https://www.w3.org/TR/WCAG22/#dfn-regions) can be [programmatically determined](https://www.w3.org/TR/WCAG22/#dfn-programmatically-determinable).

> **region:** perceivable, [programmatically determined](https://www.w3.org/TR/WCAG22/#dfn-programmatically-determinable) [section](https://www.w3.org/TR/WCAG22/#dfn-section) of content
> NOTE: In HTML, any area designated with a landmark role would be a region.

> **section:** a self-contained portion of written content that deals with one or more related topics or thoughts
> NOTE: A section may consist of one or more paragraphs and include graphics, tables, lists and sub-sections.

The TF found the term “regions” problematic to recontextualize for non-web documents and software.  The TF also found the definition to be ambiguous.  The TF could not come up with a non-web example which is an analog to _landmark role_ provided in the NOTE of the glossary.

Regardless that the definition for _section_ is also ambiguous, since _region_ depends upon it, nothing is gained by having the additional term.  Both the SC and the glossary term for _region_ require _programmatically determinability_ so the SC is redundant on that point.  Nothing is achieved by having 1.3.6 use _region_ (instead of section) and nothing is lost by by substituting _section_ for _region_.

1.3.6 would then read:
> In content implemented using markup languages, the purpose of user interface components, icons, and sections can be programmatically determined.

The only reference to _region_ is in SC 1.3.6, so with the substitution, the term can (and should be) removed from the glossary.

Please see [WCAG2ICT Issue 538](https://github.com/w3c/wcag2ict/issues/538) for additional context.


patrickhlauke:
not 100% sure the current definition of "section" is up to snuff either

> a self-contained portion of written content that deals with one or more related topics or thoughts

what if it's not written content, but images, videos, form controls, or anything else that's not "written content"? wonder if dropping the word "written" there would help generalise the definition

bruce-usab:
I agree that ”written” is an unnecessary adjective.  OTOH, removing that word (maybe) tips this PR from Class 2 to Class 3?

Edit to note that it’s irrelevant if “written section” is more specific than “section” — because an author (or auditor) would not make that distinction.

bruce-usab:
Discussed on TF call 6 February and moved to Ready for approval.  The TF had consensus that this is a change that does not functionally affect interpretation of WCAG2, consistent with the Process Document for [correction classes](https://www.w3.org/policies/process/#correction-classes).

awkawk:
The rationale for making this change is not great. To help WCAG2ICT avoid writing a note to explain that in a technology that doesn't have regions they should treat it in a particular way, we are going to open up the possibility that every section element in a web document needs to have aria-label. 

We should not approve this. Not Editorial.

bruce-usab:
> we are going to open up the possibility that every section element in a web document needs to have aria-label.

@awkawk for AAA conformance, between [2.4.10](https://www.w3.org/TR/2024/REC-WCAG22-20241212/#section-headings) and 1.3.6 _as-is_ it is already the case that HTML `section` element must have an `accessible-name` which is descriptive (or an `accessible-description`).  Do you disagree?

Do we need add a note (as we do elsewhere where an html element overlaps with a word in common use) that [section](https://www.w3.org/TR/2024/REC-WCAG22-20241212/#dfn-section) is used more generically and is not limited to `section` element?

That would be a new Issue/PR of course.

Link to [1.3.6 in diff preview](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FTR%2F2024%2FREC-WCAG22-20241212%2F&doc2=https%3A%2F%2Fdeploy-preview-4797--wcag2.netlify.app%2Fguidelines%2F#identify-purpose).

awkawk:
> for AAA conformance, between [2.4.10](https://www.w3.org/TR/2024/REC-WCAG22-20241212/#section-headings) and 1.3.6 as-is it is already the case that HTML section element must have an accessible-name which is descriptive (or an accessible-description). Do you disagree?

Not all sections have or need a heading. A section with a heading or aria-label is a region landmark, but the opposite isn't necessarily true.

If the issue is that this is confusing for people who are using WCAG outside of the web I think that it would have a smaller impact to write a note in WCAG2ICT than to update the normative text and deal with updating translations and every tool that references this SC.

PR #4812, opened 2025-12-23
Expand title of H54 - "defining instance of a word *or phrase*"

patrickhlauke:
Closes https://github.com/w3c/wcag/issues/2505

### DIFF
--- a/techniques/aria/ARIA8.html
+++ b/techniques/aria/ARIA8.html
@@ -8,7 +8,7 @@
          <h3>Describing the purpose of a link in HTML using aria-label.</h3>
          
             <p>In some situations, designers may choose to lessen the visual appearance of links on a page by using shorter, repeated link text such as "read more". These situations provide a good use case for <code>aria-label</code> in that the simpler, non-descriptive "read more" text on the page can be replaced with a more descriptive label of the link. The words "read more" are repeated in the <code>aria-label</code> (which replaces the original anchor text of "Read more") to allow consistent communication between users.</p>
-        <p class="note">The Success Criterion <a href="../../Understanding/label-in-name">2.5.3 Label in Name</a> requires that the visible label is included as part of the 'accessible name', which is generally set by the <code>aria-label</code>. The examples below meet this requirement.</p>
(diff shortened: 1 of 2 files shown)


fstrr (approved):

bruce-usab (approved):

techniques/html/H54.html
```diff
@@ -22,11 +22,11 @@ <h2>Description</h2>
             defining instance of the enclosed term. In other words, it marks the occurrence of the
             term where the term is defined. Note that it encloses the term, not the definition. This
             technique would be used in combination with <a href="../general/G112">Using
-              inline definitions</a> to provide the definition.</p>
+            inline definitions</a> to provide the definition.</p>
    </section><section id="examples"><h2>Examples</h2>
       <section class="example">
          
-         <pre xml:space="preserve"><code class="language-html">&lt;p&gt;The Web Content Accessibility Guidelines require that non-text content
+         <pre><code class="language-html">&lt;p&gt;The Web Content Accessibility Guidelines require that non-text content
```
patrickhlauke, line 29:
```suggestion
         <pre xml:space="preserve"><code class="language-html">&lt;p&gt;The Web Content Accessibility Guidelines require that non-text content
```

bruce-usab:
Discussed and approved on TF call 9 Jan 2026.  Moved to Ready for Approval.

fstrr:
@patrickhlauke H54 is referenced in [Unusual Words](https://www.w3.org/WAI/WCAG22/Understanding/unusual-words.html), so needs updating. Not sure if the build system would now fix that or whether it needs to be a manual change. If it's a manual change, it makes sense to do it as part of this PR.

patrickhlauke:
@fstrr unless i'm mistaken (@kfranqueiro can confirm), all the technique listings are now automated with 11ty. the actual file for `unusual-words.html` doesn't even contain that section anymore directly, and `understanding.11tydata.js` just refers to techniques by their number rather than duplicating their title. long story short: I think just changing the technique itself is all that's required to then have it magically resolved everywhere else.

fstrr:
@patrickhlauke That does look like the case. The only other place I can find a hard-coded reference to it is in [wcag2.json](https://github.com/w3c/wcag/blob/main/wcag20/sources/misc/wcag2.json), but that's the old xml stuff, so I assume that we leave that alone. Or do we?

kfranqueiro:
Correct, we generally don't touch files under `wcag20`.

Issue #1771, opened 2021-04-28
Are there exceptions for 3.1.4: Abbreviations?


jake-abma:
When abbreviations become their own word, **or** when used as technical terms, do they still all need to have the definitions all the time in order to pass this SC?

Things like:

- DDR3
- 3GB
- 5400RPM
- Wifi
- PC
- etc...
- 
OR

- FBI
- NATO
- etc...

OR (abbreviated names)

Pen | Penelope
- Ab (Abraham)
- Alex (Alexander)
- Pen (Penelope)
- Theo (Theodore)
- etc..

OR

- pic, picture
- photo, photograph
- phone (telephone)
- fax (facsimile)
- etc...

OR

- AM (ante meridiem)
- PM (post meridiem)

patrickhlauke:
i guess this currently hinges on *who decides* if "the abbreviation has not become part of the language" (per https://www.w3.org/TR/WCAG/#dfn-abbreviations), and if that means purely "it's present in some official dictionary for that language" (like the German "Duden" or some other authoritative resource), or more a gut feeling of "this is now a very commonly understood abbreviation" (and in many cases, some abbreviations are even understood and commonly used by people even if they don't actually know what the expanded form actually is... e.g. people will be familiar with "KGB" without knowing that it stands for "komitet gosudarstvennoy bezopasnosti").

jake-abma:
that's the hard part, how will I decide, for sure, that it is a common word 'for all', so the definite line is: "it is part of a dictionary"?

patrickhlauke:
we've had this handwaving before in this context, haven't we? (if i recall, in the context of 3.1.2 language of parts and deciding when a foreign word has become part of the vernacular/host language or not)

jake-abma:
this one? https://github.com/w3c/wcag/issues/1174

fstrr:
This issue looks like it can be closed. If it needs to be re-opened, please do that and convert it to a Discussion.

patrickhlauke:
i know it's VERY late...but as i'm going through some of my own backlog of "that was an interesting discussion that i should revisit at some point" ... I filed https://github.com/w3c/wcag/pull/4815 to hopefully clarify this

PR #4815, opened 2025-12-23
Add note to abbreviations understanding about exemptions

patrickhlauke:
Follow-up to the discussion in https://github.com/w3c/wcag/issues/1771 to make the outcome explicit

Preview: see the second paragraph in "Intent" https://deploy-preview-4815--wcag2.netlify.app/understanding/abbreviations


fstrr:
TIL that "laser" is an initialism.
update to note the irony of using "TIL" in this thread.

fstrr (approved):

understanding/20/abbreviations.html
```diff
@@ -25,6 +25,11 @@ <h2>Intent of Abbreviations</h2>
       <p>The intent of this success criterion is to ensure that users can access the expanded
          form of abbreviations.
       </p>
+
+      <p>Note that, per the definition, if an <a>abbreviation</a> has <q>become part of the language</q>, it is exempt from this requirement.
+         For instance, while <q>laser</q> is an initialism, it has become a word in its own right in most languages.
```
patrickhlauke:
```suggestion
         For instance, while "laser" is an initialism, it has become a word in its own right in most languages.
```

detlevhfischer:
> Note that, per the definition, if an abbreviation

The wording sounds a bit stilted.

Side note: Should we add something on proper names that are abbreviations but need not be given in extended form, like IBM or NASA?

understanding/20/abbreviations.html
```diff
@@ -25,6 +25,11 @@ <h2>Intent of Abbreviations</h2>
       <p>The intent of this success criterion is to ensure that users can access the expanded
          form of abbreviations.
       </p>
+
+      <p>Note that, per the definition, if an <a>abbreviation</a> has <q>become part of the language</q>, it is exempt from this requirement.
+         For instance, while "laser" is an initialism, it has become a word in its own right in most languages.
+         Generally, for an abbreviation to be considered <q>part of the language</q>, it would need to be listed as a word/noun in an authoritative dictionary for that language.
+         Beyond this, the criterion does not exempt abbreviations even if they are "common" or familiar.</p>
```
chaals: Address @detlevhfischer's [comment](https://github.com/w3c/wcag/pull/4815#issuecomment-4116646813) on stilted wording:

"most languages" is an unnecessarily strong claim that is beside the point. Removing it to focus on the core statements.

reword the grounds for considering a word part of the language.

```suggestion

      <p>If an <a>abbreviation</a> has <q>become part of the language</q>, it is exempt from this requirement.
         For instance "laser" was an initialism (and many years ago commonly presented as one, e.g. printing it in capital letters as "LASER"). Now it is most commonly used as a word in its own right. In english, people are far more likely to understand "a laser", and likely to be confused by describing "a device for light amplification by stimulated emission of radiation". So it is generally not necessary, and often unhelpful, to mark "laser" as an abreviation.

<p>If a term is listed in a common dictionary as a word, that is reasonable grounds to claim it is <q>part of the language</q>.</p>
         
 <p>Beyond this, the criterion does not exempt abbreviations even if they are "common" or familiar.</p>
```

bruce-usab (commented):
+1 to @chaals suggestion

link to intent section of [dIff preview ](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FUnderstanding%2Fabbreviations&doc2=https%3A%2F%2Fdeploy-preview-4815--wcag2.netlify.app%2Funderstanding%2Fabbreviations#intent)

understanding/20/abbreviations.html
patrickhlauke: oof, the whole history of laser may be a bit overly specific/dense here, in my view... might be worth copy-editing the proposed rewording a bit more, i think...

patrickhlauke:
note that this would actually be even nicer to add to the definition for abbreviation itself, but ... as errata of this kind seem to immediately hit resistance in the group, opting for a clarification here in understanding for now

chaals:
I think group agreement is the only real barrier to updating examples even 
in a REC. (Not that it is an insignificant barrier - rightly, it requires 
making the case and convincing the group).

bruce-usab (approved):

patrickhlauke:
> Side note: Should we add something on proper names that are abbreviations but need not be given in extended form, like IBM or NASA?

Do we have a definitive line here though? The thing about having become part of a language was always in the definition implicitly, this addition here just makes it clearer/move obvious. Not sure if we have a good unambiguous measure for abbreviations/acronyms/initialisms that are commonly undestood/are proper names...

chaals:
> Not sure if we have a good unambiguous measure for abbreviations/acronyms/initialisms that are commonly undestood/are proper names...

For English, there isn't one. For "regulated" languages (many European languages, for example), there is a definitive dictionary that would work for ordinary words.  

There are useful approaches providing evidence for a claim that something is a recognised term:
- looking in a dictionary,
- trademarks,
- a simple search on how many occurrences there are of a word compared to expressing it as an abbreviation
- listen to how automated speech systems handle it, on the basis that *sometimes* evidence that there is a recognised pronunciation is evidence that a term is recognised as a word in its own right.

There is no normative definition of the english language, and pretending otherwise doesn't make it so. The famous "Man on the Clapham ominbus/Bourke Street tram/..." (a hypothetical "reasonable person", which is a genuine and important legal term) - in other words, a sensible judgement, guided by whatever advice we want to give (e.g. "be very conservative", or "any trademark is obviously a word", or whatever we decide) is as good as it gets, and as good as it is worth us trying to get. 

Perhaps a bit tangentially, I really think there should be some conversation with the i18n folks about how to provide useful advice relevant to different languages, particularly given the number of times WCAG gets translated and applied to people in other languages.

patrickhlauke:
riffed on @chaals' suggestion, and also added a new bit about company/organisation names and trademarks based on @detlevhfischer's question. at that stage though, it feels like we're essentially making a normative call here ... which may mean we should back this out of understanding and instead push for a change/expansion to the normative definition of "abbreviation" then?

patrickhlauke:
or...we could land this in understanding first, and make a separate PR suggesting a move of these additions as a note to the normative definition as an erratum ... at least if the latter gets bogged down, we have something written down in the understanding

understanding/20/abbreviations.html
```diff
@@ -25,8 +25,13 @@ <h2>Intent of Abbreviations</h2>
       <p>The intent of this success criterion is to ensure that users can access the expanded
          form of abbreviations.
       </p>
-      
-      
+
+      <p>If an <a>abbreviation</a> has <q>become part of the language</q>, it is exempt from this requirement.
+         For instance <q>>laser</q> was an initialism (and many years ago commonly presented in capital letters as <q>LASER</q>). Now it is most commonly used as a word in its own right, and marking it as an abbreviation (expanding it to <q>Light Amplification through the Stimulated Emission of Radiation</q>) would likely be confusing and unhelpful for people.</p>
```
patrickhlauke:
```suggestion
         For instance <q>laser</q> was an initialism (and many years ago commonly presented in capital letters as <q>LASER</q>). Now it is most commonly used as a word in its own right, and marking it as an abbreviation (expanding it to <q>Light Amplification through the Stimulated Emission of Radiation</q>) would likely be confusing and unhelpful for people.</p>
```

understanding/20/abbreviations.html
```diff
@@ -25,8 +25,13 @@ <h2>Intent of Abbreviations</h2>
       <p>The intent of this success criterion is to ensure that users can access the expanded
          form of abbreviations.
       </p>
-      
-      
+
+      <p>If an <a>abbreviation</a> has <q>become part of the language</q>, it is exempt from this requirement.
+         For instance <q>laser</q> was an initialism (and many years ago commonly presented in capital letters as <q>LASER</q>). Now it is most commonly used as a word in its own right, and marking it as an abbreviation (expanding it to <q>Light Amplification through the Stimulated Emission of Radiation</q>) would likely be confusing and unhelpful for people.</p>
+      <p>If an <a>abbreviation</a> is listed in a common dictionary as a word, that is reasonable grounds to claim it has <q> become part of the language</q>.</p>
```
kfranqueiro:
```suggestion
      <p>If an <a>abbreviation</a> is listed in a common dictionary as a word, that is reasonable grounds to claim it has <q>become part of the language</q>.</p>
```

understanding/20/abbreviations.html
```diff
@@ -25,8 +25,13 @@ <h2>Intent of Abbreviations</h2>
...
          form of abbreviations.
       </p>
-      
-      
+
+      <p>If an <a>abbreviation</a> has <q>become part of the language</q>, it is exempt from this requirement.
+         For instance <q>laser</q> was an initialism (and many years ago commonly presented in capital letters as <q>LASER</q>). Now it is most commonly used as a word in its own right, and marking it as an abbreviation (expanding it to <q>Light Amplification through the Stimulated Emission of Radiation</q>) would likely be confusing and unhelpful for people.</p>
+      <p>If an <a>abbreviation</a> is listed in a common dictionary as a word, that is reasonable grounds to claim it has <q>become part of the language</q>.</p>
+      <p>Another signal that an <a>abbreviation</a> can be considered to have <q>become part of the language</q> are well-known company/organization names or product trademarks, such as <q>IBM</q> or <q>NASA</q>, as these are generally used as words in their own right.</p>
+      <p>Beyond this, the criterion does not exempt abbreviations even if they are "common" or familiar.</p>
```
patrickhlauke:
```suggestion
      <p>Beyond this, the criterion does not exempt abbreviations even if they are common or familiar.</p>
```

understanding/20/abbreviations.html
```diff
@@ -25,8 +25,13 @@ <h2>Intent of Abbreviations</h2>
       <p>The intent of this success criterion is to ensure that users can access the expanded
          form of abbreviations.
       </p>
-      
-      
+
+      <p>If an <a>abbreviation</a> has <q>become part of the language</q>, it is exempt from this requirement.
+         For instance <q>laser</q> was an initialism (and many years ago commonly presented in capital letters as <q>LASER</q>). Now it is most commonly used as a word in its own right, and marking it as an abbreviation (expanding it to <q>Light Amplification through the Stimulated Emission of Radiation</q>) would likely be confusing and unhelpful for people.</p>
+      <p>If an <a>abbreviation</a> is listed in a common dictionary as a word, that is reasonable grounds to claim it has <q>become part of the language</q>.</p>
+      <p>Another signal that an <a>abbreviation</a> can be considered to have <q>become part of the language</q> are well-known company/organization names or product trademarks, such as <q>IBM</q> or <q>NASA</q>, as these are generally used as words in their own right.</p>
```
chaals:
```suggestion
      <p>Another signal that an <a>abbreviation</a> can be considered to have <q>become part of the language</q> are well-known company/organization names or product trademarks, such as <q>IBM</q> or <q>NASA</q>, that are generally used as words in their own right.</p>
```

Slight tweak. The key is that the trademark is used as a word, not just well-known as an abbreviation.

understanding/20/abbreviations.html
```diff
@@ -25,8 +25,13 @@ <h2>Intent of Abbreviations</h2>
...
+      <p>Beyond this, the criterion does not exempt abbreviations even if they are common or familiar.</p>
```
chaals, line 33: There's a small class of common abbreviations that are generally recognised as such, but are far more commonly written in their abbreviated forms. Things like "dr", "etc" and "mrs" are I think common across english varieties, while there are local expressions such as "Plc" (UK - doesn't translate so easily to other countries) or "Jr" and "Sr" used in names far more in the US than other large countries in the anglosphere.

I think the intent is to cover these abbreviations as allowable parts of the language that do not require an expansion to be provided. 

I suspect it is also intended to cover certain terms from chemistry that are geenrally used as words, such as CFC, PVC, PFA, and maybe very common biological terms like DNA or RNA.

"Rep." is commonly used in the US, to mean either "representative" as in a member of the US House, or "republican" as in a member of the Republican party. given the potential ambiguity I think it's clear this isn't intended to be accepted as an exemption.

I suspect the same applies to "MP", used in commonwealth countries to mean "Member of Parliament". 

I believe the intention is that organisational terms (Lt., SVP, PFC, Ass., Sr.), and technical acronyms used in commerce (FHE, P/L, RSJ, ROI, VPAT, RBAC, TVL, ABS), are not exempt.

What about such terms as "MLK", "JFK" or "FDR"? Although they are essentially local US terms, they have very high recognition...

understanding/20/abbreviations.html
patrickhlauke: Sure, good catch

understanding/20/abbreviations.html
patrickhlauke: I believe at that stage, there will need to be some further mention of intended target audience that should be taken into consideration (incidentally, similar to what I added in this other PR https://github.com/w3c/wcag/pull/4841 for "unusual words").

This may go beyond the scope of this small targeted tweak I was trying to make to try and explain the vague "has not become part of the language" of the term - and for some that you're mentioning, like MLK etc, I think it may become increasingly more difficult to make an argument for when they've become part of the langugae or not. Suggest we land this first, as I believe the examples provided so far are less debatable, and circle back in another PR, rather than trying to solve this for all scenarios. https://github.com/w3c/wcag/issues/5044

chaals:
Agree that we should land this and then work out what else is there.

I think the "intended audience" approach tends to be used to say "but everyone who is meant to read this understands P/L, TVL and FHE ... ", which I don't think is the point.

It is more that there are a handful of abbreviations recognised as such, but used for so long and so commonly that people know what they mean, but don't know how to write them in full ("Mrs" which dates back centuries or "Ms" which is about as old as me so has been used since before most people alive learned to read and write - how do you write either of these in unabbreviated form?).

More concrete guidance on what *isn't* intended to be exempt would be helpful. Happy if that's bumped to a different PR.

PR #4817, opened 2025-12-23
Add extra note to pause-stop-hide understanding about multiple instances

### LINKED ISSUE #4444: G186, but many, many videos
In [G186: Using a control in the web page that stops moving, blinking, or auto-updating content](https://www.w3.org/WAI/WCAG22/Techniques/general/G186), the following sentence is not entirely accurate:

> A single control may stop all moving or blinking content on the page, or there may be separate controls for separate parts of the content.

This might be correct for large separate pieces of content, but if I have multiple of such content on the page, hunting for each pause button will be tedious. Imagine a shop overview page where each item has an animation, and those animations can only be individually paused/stopped/hidden.

I would consider this a violation of [5.2.5 Non-interference](https://www.w3.org/TR/WCAG22/#cc5) as the user’s ability to remove the barrier and enact a pause will be impacted by other non-paused elements.

I wonder if there is a way to phrase the sentence quoted in a way that clarifies that 2 or 3 individually stoppable things are maybe OK, but that there are limits.

patrickhlauke:
Closes https://github.com/w3c/wcag/issues/4444

Preview: https://deploy-preview-4817--wcag2.netlify.app/understanding/pause-stop-hide (last note before the "Benefits" section)


understanding/20/pause-stop-hide.html
```diff
@@ -120,6 +120,15 @@ <h2>Intent of Pause, Stop, Hide</h2>
...
       </div>
 
+      <div class="note">
+         <p>As a best practice, if a page contains more than one moving, blinking, scrolling, or auto-updating element,
+            we recommend providing a <em>single</em> <a>mechanism</a> to pause. stop, hide, or (in the case of auto-updating content)
+            control the update frequency that affects all these elements at the same time.</p>
+         <p>While having separate controls for each of the elements <em>does</em> satisfy the normative requirements
+            of this Success Criterion, this won't result in a good user experience for users, who will be forced to manually
+            pause, stop, hide, or control each moving, blinking, scrolling, or auto-updating element in isolation.</p>
+      </div>
```
detlevhfischer, line 129: Should we also add something that makes it clear that when there are several regions that move/auto-update, stopping one of them indidually must not (re-)start movement of another? May be difficult to express succinctly (and I'm not even entirely sure this would clearly count as a normative FAIL since the normative text does not account for regarding several independently moving areas...)

detlevhfischer (approved):

understanding/20/pause-stop-hide.html
patrickhlauke: that sounds like a very unusual/odd scenario. is this something you actually ever come across?

and as it's so specific and unusual, not sure it warrants explicitly mentioning this?

detlevhfischer: I may have (scratches head), but I rest my case. :)

understanding/20/pause-stop-hide.html
```diff
@@ -119,6 +119,15 @@ <h2>Intent of Pause, Stop, Hide</h2>
          </p>
       </div>
 
+      <div class="note">
+         <p>As a best practice, if a page contains more than one moving, blinking, scrolling, or auto-updating element,
+            we recommend providing a <em>single</em> <a>mechanism</a> to pause. stop, hide, or (in the case of auto-updating content)
```
adampage:
```suggestion
         <p>As a best practice, if a page contains more than one moving, blinking, scrolling, or auto-updating element,
            we recommend providing a <em>single</em> <a>mechanism</a> to pause, stop, hide, or (in the case of auto-updating content)
```

understanding/20/pause-stop-hide.html
```diff
@@ -119,6 +119,15 @@ <h2>Intent of Pause, Stop, Hide</h2>
...
+            control the update frequency that affects all these elements at the same time.</p>
+         <p>While having separate controls for each of the elements <em>does</em> satisfy the normative requirements
+            of this Success Criterion, this won't result in a good user experience for users, who will be forced to manually
```
adampage:
```suggestion
            of this success criterion, this won't result in a good user experience for users, who will be forced to manually
```

adampage (approved):

understanding/20/pause-stop-hide.html
```diff
@@ -119,6 +119,15 @@ <h2>Intent of Pause, Stop, Hide</h2>
          </p>
       </div>
 
+      <div class="note">
+         <p>As a best practice, if a page contains more than one moving, blinking, scrolling, or auto-updating element,
+            we recommend providing a <em>single</em> <a>mechanism</a> to pause, stop, hide, or (in the case of auto-updating content)
```
detlevhfischer: In case the "we recommend" phrase is unusual (cannot recall it from anywhere else) we might rephrase to
> if a page contains more than one moving, blinking, scrolling, or auto-updating element, it is best practice to provide a <em>single</em> <a>mechanism</a> to pause, stop, hide etc.

detlevhfischer (approved):

understanding/20/pause-stop-hide.html
patrickhlauke: yeah was thinking of simplifying this even further, along the lines of just:

```suggestion
            provide a <em>single</em> <a>mechanism</a> to pause, stop, hide, or (in the case of auto-updating content)
```

bruce-usab (approved):
See [preview diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FUnderstanding%2Fpause-stop-hide&doc2=https%3A%2F%2Fdeploy-preview-4817--wcag2.netlify.app%2Funderstanding%2Fpause-stop-hide) at end of “Intent” (right before “Benefits”).

understanding/20/pause-stop-hide.html
patrickhlauke:
```suggestion
            and which doesn't provide a way to pause, stop, or hide, will fail this criterion, and potentially
```

bruce-usab (approved):

Issue #4660, opened 2025-10-02
Consider rewording technical information in F3


mbgower:
During discussions in [response to an existing issue](https://github.com/w3c/wcag/pull/1035#issuecomment-3120228052) it was pointed out by some members of the WCAG 2 Task Force that the second sentence of the description may no longer be accurate, and that the failure technique may contain other technical inaccuracies, which should be addressed.
> The CSS background-image property was designed for decorative purposes and it is not possible to associate text alternatives with images that are included via CSS.

patrickhlauke:
dropping this here from the old issue #80 

> randomly ... F3 has reared its head again in a recent conversation https://eldritch.cafe/@juliemoynat/115355236855876491
>
> would be good if we could try and de-crappify it

juliemoynat:
What I understand from reading the #80 issue is that this part of the third point of the F3 test procedure "and is also available when the CSS image is not displayed" should not be in a technique associated to SC 1.1.1 but 1.3.1.

So as it is associated to SC 1.1.1, this point should be removed from the technique. That's logical. And I think it should be done because the technique would be clearer for everyone.

But there is a question: should we create another technique to address the point which can be an accessibility issue (high contrast mode, users disabling background images)?

There are issues about these questions: #623 and #2889.

And there is the [F24](https://www.w3.org/WAI/WCAG22/Techniques/failures/F24) that says ([pointed in this comment](https://github.com/w3c/wcag/issues/80#issuecomment-548186222)):

> Background color may also be specified using a background image with the CSS property 'background-image' or with the CSS property 'background' (with the URI of the image, e.g., 'background: url("images/bg.gif")'). With background images, it is still necessary to specify a background color, because users may have images turned off in their browser. But the background image and the background color need to be checked. 

I think that if we care about text contrasts while deactivating background images, we should care about alternative texts being visible while deactivating background images.

mbgower:
@juliemoynat 

The point of this new issue was to determine if the quoted statement early in the technique document is _technically_ true, and if the technique contains statements that are similarly challengeable with current technologies.

I assume you are talking specifically about the wording in the latter part of this check (which I have put in bold):
> If an image does convey important information, the information is provided to assistive technologies and is **also available when the CSS image is not displayed**

[...]

PR #4823, opened 2025-12-24
Remove note in F24 about css background images

patrickhlauke:
This short "in passing" mention in a note about CSS background images appears to introduce a whole new normative requirement by the backdoor, in a technique note, which effectively fails sites that don't provide both a background colour AND background image in their CSS, because users MAY have changed their browser not to load images.

This is not covered anywhere else in WCAG, and in many other places WCAG explicitly does NOT care about how users may have modified/changed their browser settings. Not even 1.1.1 mentions the scenario as a particular reason/concern. This note oversteps its remit.

EDIT: reworked this PR to only remove/amend the apparent normative requirement to define BOTH a background image AND background colour - replaced with a note about contrast testing. Further, this PR now corrects the inaccurate section about inherited color/background (background is not inherited - if not defined, an element has a *transparent* background, so what counts is that at least one of the ancestors of the element has a background

x-ref https://github.com/w3c/wcag/issues/4660#issuecomment-3393328930

preview: https://deploy-preview-4823--wcag2.netlify.app/techniques/failures/f24


scottaohara (approved):

mgifford:
You also made a good clarification about foreground and background colors using images.

Is it possible to leave in a recommendation to check for if CSS is disabled or images fail to load? The Robust part of POUR...

patrickhlauke:
> Is it possible to leave in a recommendation to check for if CSS is disabled or images fail to load?

that would make any use of CSS considerably more difficult to implement. and then would fail further if JS was also unavailable, so do we need to warn about that too?

detlevhfischer:
> which effectively fails sites that don't provide both a background colour AND background image in their CSS, because users MAY have changed their browser not to load images.

If the substance of F24 remains (that each and every element must have foreground and background colour specified) it would presumably also apply to elements where a background image is used? Or where is it written that in that case, background colour need not be specified? Checking contrast for text on background images, something that cannot be unambiguously tested automatically, certainly not across all viewport widths) would remain necessary even where the defined foreground/background colour pair passes. So I do not see anything in this note that introduces a new normative requirement by the backdoor. And therefore do not see the need to remove it.

patrickhlauke:
> I do not see anything in this note that introduces a new normative requirement

This part, arguably: "With background images, it is still necessary to specify a background color"

[EDIT to make it clearer] This part is implying that it is necessary. That an author MUST do it. Which is an overreach, as it's not true. And the rationale given "because users may have turned off images" is not an explicit WCAG concern.

I'd be amenable to keep that note, but to remove that specific bit, and expand on how testing would need to be done by checking the specific colours of the background image. (the fact that it can't be "unambiguously tested automatically" is interesting, but just shows how automated testing if flawed/falls short)

patrickhlauke:
as an aside, F24 is still a bit of a mess because it confuses things

```
Since CSS color properties inherit from ancestor elements, it is sufficient if both foreground and background colors are defined either directly or through inheritance by the time that color is applied to a given element.
```

the background is not "inherited". if not defined, it's transparent, so the background of the parent/ancestor shows through.

detlevhfischer:
> "With background images, it is still necessary to specify a background color"

Hm. So would that become "With background images, it is still necessary to specify a background color if a foreground color has been defined"? Because if the default is being used, nothing would need to be specified?

techniques/failures/F24.html
```diff
@@ -85,8 +85,6 @@ <h3>Specifying foreground color of link text with CSS</h3>
             <p>
                Color and background color may be specified at any level in the cascade of preceding selectors, by external stylesheets or through inheritance rules.
             </p>
-            <p>
-               Background color may also be specified using a background image with the CSS property 'background-image' or with the CSS property 'background' (with the URI of the image, e.g., 'background: url("images/bg.gif")'). With background images, it is still necessary to specify a background color, because users may have images turned off in their browser. But the background image and the background color need to be checked. </p>
```
bruce-usab, line 89: Suggestion for last sentence:

> <s>But</s> <u>Both</u> the background image and the background color need to be checked.

bruce-usab (changes requested):

techniques/failures/F24.html
bruce-usab: Nevermind, I misunderstood that we are eliminating this paragraph!

bruce-usab (approved):

patrickhlauke:
> > "With background images, it is still necessary to specify a background color"
> 
> Hm. So would that become "With background images, it is still necessary to specify a background color if a foreground color has been defined"? Because if the default is being used, nothing would need to be specified?

no my point with the "it is necessary" is that THAT'S what that paragraph is implying. it's not though. so that para is overreaching, making it sound like you MUST define a color in addition to an image if you're using an image.

patrickhlauke:
Revisited/expanded this: reworked this PR to only remove/amend the apparent normative requirement to define BOTH a background image AND background colour - replaced with a note about contrast testing. Further, this PR now corrects the inaccurate section about inherited color/background (background is not inherited - if not defined, an element has a *transparent* background, so what counts is that at least one of the ancestors of the element has a background

techniques/failures/F24.html
```diff
@@ -6,7 +6,7 @@
...
-      <p>It is not necessary that the foreground and background colors both be defined on the same CSS rule. Since CSS color properties inherit from ancestor elements, it is sufficient if both foreground and background colors are defined either directly or through inheritance by the time that color is applied to a given element.</p>
+      <p>It is not necessary that the foreground and background colors both be defined on the same element. Foreground text <code>color</code> is inherited from ancestor elements. If an element's <code>background</code> color has not been defined, the element will have a transparent background, so authors need to ensure that at least one the element's ancestors has a defined background color.</p>
```
detlevhfischer: What about the (admittedly unlikely) case that neither fore- nor background colours are defined and the page just uses the browser default?

patrickhlauke: then that falls outside of the scope of what the failure technique is demonstrating: "Failure of Success Criterion 1.4.3, 1.4.6 and 1.4.8 due to specifying foreground colors without specifying background colors or vice versa"

patrickhlauke: because if the page doesn't specify neither foreground nor background, it's up to the user agent to have acceptably contrasting defaults / up to the user to have picked their own preferred combination in browser settings

detlevhfischer: Yes, true - I forgot about the context.

adampage:
```suggestion
      <p>It is not necessary that the foreground and background colors both be defined on the same element. Foreground text <code>color</code> is inherited from ancestor elements. If an element's <code>background</code> color has not been defined, the element will have a transparent background, so authors need to ensure that at least one of the element's ancestors has a defined background color.</p>
```

adampage (approved):

techniques/failures/F24.html
```diff
@@ -85,23 +85,28 @@ <h3>Specifying foreground color of link text with CSS</h3>
             <p>
                Color and background color may be specified at any level in the cascade of preceding selectors, by external stylesheets or through inheritance rules.
             </p>
-            <p>
-               Background color may also be specified using a background image with the CSS property 'background-image' or with the CSS property 'background' (with the URI of the image, e.g., 'background: url("images/bg.gif")'). With background images, it is still necessary to specify a background color, because users may have images turned off in their browser. But the background image and the background color need to be checked. </p>
+         </div>
+         <div class="note">
+            <p>Background color may also be specified using a background image with the CSS property <code>background-image</code> or with the CSS <code>background</code> shorthand (with the URI of the image, e.g., <code>background: url("images/bg.gif")</code>). With background images, authors will need to check text contrast against the various pixels of the background image that will be behind the text.</p>
```
patrickhlauke:
```suggestion
            <p>Background color may also be specified using a background image with the CSS property <code>background-image</code> or with the CSS <code>background</code> shorthand (with the URI of the image, e.g., <code>background: url("images/bg.gif")</code>). With background images, authors will need to check text contrast against the various colors of the background image that will be behind the text, and may also want to consider contrast when images are not displayed.</p>
```

PR #4824, opened 2025-12-25
Focus obscured and semi-transparent/blurred overlays - reword, and DO make it fail at AAA

### LINKED ISSUE #3203: Create a failure technique for Non-text contrast based on a lightbox effect

patrickhlauke:
Initially, this PR intends to reword the explanation of what "light boxes" are, which is currently quite confused/confusing (or rather, it talks about light boxes while actually referring to the semi-transparent overlay that usually accompanies them) - as the conversations in https://github.com/w3c/wcag/pull/4673 started to veer off at a tangent because of it (as it seems some folks thought "light box" equated to glass/blur type effects). This PR rewrites the section to hopefully be clearer and more direct.

However, from the discussion, it seems clear that some folks would argue that focused components that are covered somehow, even by a semi-transparent only overlay, do normatively fail the requirements of componanent having "no part [...] hidden". While for AA it's probably ok to be lenient and handwave that components aren't *entirely* hidden when behind such overlays, it gets more difficult to justify for the stricter AAA "no part" requirement.

Closes https://github.com/w3c/wcag/issues/3203

/cc @scottohara

Previews:

* https://deploy-preview-4824--wcag2.netlify.app/understanding/focus-not-obscured-minimum
* https://deploy-preview-4824--wcag2.netlify.app/understanding/focus-not-obscured-enhanced


patrickhlauke:
Note that this proposed change in the AAA understanding DOES effectively change the normative interpretation for this SC, so will need careful consideration.

To take it to the extreme though, I can't imagine justifying a component that is completely covered with a 99.99% opaque overlay, or a super-strongly-blurring overlay, having "no part [...] hidden", when they can barely be seen/discerned.

mgifford:
Nice edits.

understanding/22/focus-not-obscured-enhanced.html
```diff
@@ -22,7 +22,7 @@ <h2>Intent of Focus Not Obscured (Enhanced)</h2>
...
-  <p>Another form of obscuring can occur where light boxes or other semi-opaque effects overlap the item with focus. This form of obscuring is <em>not</em> in scope for this success criterion. While less than 100 percent opacity is not causing the component to be <q>visually hidden</q>, such semi-opaque overlaps may cause a failure of <a href="non-text-contrast">1.4.11 Non-text Contrast</a> and/or <a href="focus-appearance">2.4.13 Focus Appearance</a>. When a focus indicator can be covered by a semi-opaque component, the focus indicator should be assessed against 1.4.11 Non-text Contrast and 2.4.13 Focus Appearance. The intention in both situations is that the component receiving focus should never be obscured to the point a user cannot tell which item has focus.</p>
+  <p>Another form of obscuring can occur where semi-transparent overlays, or overlays that apply a blur or other dimming effect, overlap the item with focus. This form of obscuring <em>is</em> in scope for this success criterion, as it will make it more difficult for users to see/understand which component on the page currently has focus.</p>
```
adampage:
```suggestion
  <p>Another form of obscuring can occur when semi-transparent overlays, or overlays that apply a blur or other dimming effect, overlap the focused item. This form of obscuring <em>is</em> in scope for this success criterion, as it will make it more difficult for users to identify which component on the page currently has focus.</p>
```

Super editorial.

understanding/22/focus-not-obscured-minimum.html
```diff
@@ -29,7 +29,7 @@ <h2>Intent of Focus Not Obscured (Minimum)</h2>
...
+      <p>Another form of obscuring can occur where semi-transparent overlays, or overlays that apply a blur or other dimming effect, overlap the item with focus. While the focused component may not be <q>entirely obscured</q> by these overlays, passing this criterion, they may fail <a href="../understanding/non-text-contrast.html">1.4.11 Non-text Contrast</a>. When a focus indicator can be covered by these overlays, the focus indicator must be evaluated against the contrast requirements of 1.4.11 when affected by the overlays. The intention in both situations is that the component receiving focus should never be obscured to the point a user cannot tell which item has focus.</p>
```
adampage:
```suggestion
      <p>Another form of obscuring can occur when semi-transparent overlays, or overlays that apply a blur or other dimming effect, overlap the focused item. While the focused item may not be <q>entirely obscured</q> by these overlays, thereby passing this criterion, it may separately fail <a href="../understanding/non-text-contrast.html">1.4.11 Non-text Contrast</a>. In such cases, the focus indicator must be evaluated against the contrast requirements of 1.4.11 while its appearance is affected by the overlays. The intention is for an item’s focused state to be clearly identifiable at all times.</p>
```

This suggestion started confidently and then fell down a slippery slope. 🥴  Please disregard if you don’t agree that it’s easier to read and still accurate.

adampage (approved):
LGTM, just left 2 soft suggestions.

bruce-usab (approved):

giacomo-petri:
As noted today, since we are referencing related success criteria it may also be worth mentioning 2.4.3 focus order in cases of semi-transparent overlays, as it is somewhat related. This could either be addressed together in this PR or handled separately in a different one.

I'll try to draft a proposal by early next week

awkawk:
For purposes of considering this in the context of what we actually see on the web, there is no reason to think that a designer will expect anyone to interact with content that is covered with a layer with only 10-20% of the original visibility. So, I think that this issue will be covered by 1.4.11 or even 2.1.1 in most cases, and when these exist it is because the content is created incorrectly. I do agree that if the opacity is 0% then the focus would be obscured, but don't see much value in piling on another SC when 2.1.1 (and 1.4.3) applies.

I don't think that we intended "no part" to include a luminosity layer as "part" of the component. Note 1 alludes to this by referring to the positioning of the component:

> Where content in a configurable interface can be repositioned by the user, then only the initial positions of user-movable content are considered for testing and conformance of this success criterion.

Are there any scenarios covered by this reinterpretation where there isn't a failure against another SC? I can't think of any and for these reasons, -1.

patrickhlauke:
i don't believe "a problem would already be covered by another SC anyway" is a solid reason not to specify that something fails the normative wording of an SC

hidde:
Great stuff! I feel this rewrite and moving away from “light box” is a great catch and quite an improvement, and moves it from something many web developers today would not understand (if born/started working after “light box” was a thing) to something that would be universally understood. And likely remains so in the future.

+1!

understanding/22/focus-not-obscured-enhanced.html
```diff
@@ -22,7 +22,7 @@ <h2>Intent of Focus Not Obscured (Enhanced)</h2>
...
-  <p>Another form of obscuring can occur where light boxes or other semi-opaque effects overlap the item with focus. This form of obscuring is <em>not</em> in scope for this success criterion. While less than 100 percent opacity is not causing the component to be <q>visually hidden</q>, such semi-opaque overlaps may cause a failure of <a href="non-text-contrast">1.4.11 Non-text Contrast</a> and/or <a href="focus-appearance">2.4.13 Focus Appearance</a>. When a focus indicator can be covered by a semi-opaque component, the focus indicator should be assessed against 1.4.11 Non-text Contrast and 2.4.13 Focus Appearance. The intention in both situations is that the component receiving focus should never be obscured to the point a user cannot tell which item has focus.</p>
+  <p>Another form of obscuring can occur when semi-transparent overlays, or overlays that apply a blur or other dimming effect, overlap the focused item. This form of obscuring <em>is</em> in scope for this success criterion, as it will make it more difficult for users to identify which component on the page currently has focus.</p>
```
gundulaniemann: The new wording not only says the contrary of what was said in the old wording, but it also does not make clear when an overlay causes a fail and when it might still be a success.
As for focus not obscured (minimum) it should be stated "In such cases, the focus indicator must be evaluated against the contrast requirements of 1.4.11 while its appearance is affected by the overlays. The intention is for an item’s focused state to be clearly identifiable at all times."

In both requirement it should be made clear how the contrast should be handled. Therefore I suggest to keep the mentioning of  2.4.13 Focus Appearance in focus not obscured (enhanced).

The starting point of the change was, if I got this correctyl, that the focus indicator might be dimmed in some places which makes it harder to identify it. But a dimming could also be a technique to get the desired coloring, still maintaining the needed contrast and consistent identification throughout the page (or set of pages). This should still be allowed.

patrickhlauke: Discussed during the TF call on Friday (12 June). In the context of the very binary/cut-and-dry "any type of obscuring/covering/blurring will fail the AAA requirement, as it doesn't have any nuance (compared to the AA version)" I feel that it's appropriate to essentially be hardline about the fact that any obscuring/blurring is a failure, even if in practice it may still be "fine" for users (same way that covering just 1 pixel of the focused element in theory would fail this criterion, while in practice it's unlikely to cause actual problems to users).

As for re-introducing/keeping the mention of 1.4.11 and 2.4.13, sure, let me see if i can graft them back in.

PR #4829, opened 2025-12-27
Add new failure technique for non-semantic ruby annotation

adampage:
Resolves #4746.

Preview links:

- [Failure of Success Criterion 1.3.1 and 1.3.2 due to visible ruby annotations that are not programmatically determinable and disrupt the reading sequence](https://deploy-preview-4829--wcag2.netlify.app/techniques/failures/f114)
  - [Failure example of misuse of `<abbr title>` to convey ruby annotations](https://deploy-preview-4829--wcag2.netlify.app/working-examples/ruby-annotation/misuse-abbr-title/)
  - [Failure example of misuse of `<sup>` to convey ruby annotations](https://deploy-preview-4829--wcag2.netlify.app/working-examples/ruby-annotation/misuse-sup/)
  - [Failure example of misuse of `<span>` to convey ruby annotations](https://deploy-preview-4829--wcag2.netlify.app/working-examples/ruby-annotation/misuse-span/)

PR #4831, opened 2025-12-27
Normalize `h1` text to permit leading, trailing, and wrapping whitespace

adampage:
While developing #4829, I discovered that [my `h1` text for the new page](https://github.com/w3c/wcag/blob/f71bb0c0860b612f25a7d71234f9112771166cf5/techniques/failures/F999.html#L11) wasn’t successfully retrieved by the build system because I’d hard-wrapped the lines like so:

```html
<h1>
  Failure of Success Criterion 1.3.1 due to visible ruby annotations that
  are not programmatically determinable
</h1>
```

@kfranqueiro, would it be alright to tweak the build system to permit wrapped heading text like this?

### DIFF
--- a/11ty/techniques.ts
+++ b/11ty/techniques.ts
@@ -366,7 +366,12 @@ export async function getTechniquesByTechnology(guidelines: FlatGuidelinesMap) {
     // Isolate h1 from each file before feeding into Cheerio to save ~300ms total
     const h1Match = content.match(/<h1[^>]*>([\s\S]+?)<\/h1>/);
     if (!h1Match || !h1Match[1]) throw new Error(`No h1 found in techniques/${path}`);
-    const $h1 = load(h1Match[1], null, false);
+
+    const normalizedH1 = h1Match[1]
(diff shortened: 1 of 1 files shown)


kfranqueiro:
The build system already supports wrapping whitespace in the level-1 heading of techniques. However, as you point out, it is not handling leading/trailing whitespace correctly.

If you look [a few lines later than where you adjusted](https://github.com/w3c/wcag/blob/5287d753/11ty/techniques.ts#L402), `title` and `titleHtml` both contain the full title; only `truncatedTitle` gets the middle truncated, which is typically used for link text. Believe it or not, this was based on existing (probably accidental) behavior from the older XSLT build system, with the added adjustment of inserting an ellipsis where content was removed from the middle.

This PR, as written, ends up undoing the functionality that I pointed out lower in the file. I have created #4846 to fix this without regressions.

adampage:
Got it — thanks for the explanation and follow-up PR, @kfranqueiro. 🙏🏻

PR #4833, opened 2025-12-27
Expand consistent navigation understanding intent

### LINKED ISSUE #2133: 3.2.3 Consistent Navigation -scope of "relative order", and nature of this order: source order, visual presentation, both?
the current wording for the SC https://www.w3.org/TR/WCAG21/#consistent-navigation - including the clarification of the term "same relative order" - still leaves quite a few vague gaps.

does the "same relative order" only apply to the navigation mechanisms themselves amongst each other (which is what the definition of the term hints at https://www.w3.org/TR/WCAG21/#dfn-same-relative-order? just making sure that links/buttons used for navigation are in the same relative order? or does this also include a broader "and the order that those navigation mechanisms have with the rest of the document/page components"? for instance, if i have a `<nav>` that is always the same (so no issue of relative order for the navigation elements inside it), but on some pages it comes before the `<main>`, while on others it inexplicably comes after the `<main>`, is that a failure or 3.2.3 because the relative order of the whole navigation compared to the main content area is not the same?

further, the understanding document https://www.w3.org/WAI/WCAG21/Understanding/consistent-navigation.html starts with

> The intent of this Success Criterion is to encourage the use of consistent presentation and layout

but later talks about the advantage for blind users. So, by implication, this isn't just about visual presentation/layout, but also source/focus order? if so, this should be clarified.

patrickhlauke:
The intent currently leans heavily on visual layout/presentation ... but then almost in passing in the "Benefits..." section just mentions users who are blind. (this is likely because, back when this was written, there was a tacit assumption that of course, visual layout and source order are the same?)

The only way users who are blind would benefit (if that was the intention) is if this is not just about visual order, but also source order. Assuming that that's the case (that this was the intention), this PR makes it much more explicit in the intention.

Closes https://github.com/w3c/wcag/issues/2133

The actual "meat" of this PR is line 35 https://github.com/w3c/wcag/pull/4833/files#diff-ced3417b47e94e145f2e8d5c3937db91b10c17c34ca1dc8df61d76432238755cR35 - the rest is just cleaning up the formatting

Preview: https://deploy-preview-4833--wcag2.netlify.app/understanding/consistent-navigation


bruce-usab (approved):
[diff preview](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FUnderstanding%2Fconsistent-navigation.html&doc2=https%3A%2F%2Fdeploy-preview-4833--wcag2.netlify.app%2Funderstanding%2Fconsistent-navigation#intent)

Issue #4676, opened 2025-10-09
Confusion regarding definition of "status message"


patrickhlauke:
The discussion at https://github.com/w3c/wcag/discussions/4672 is making me think that the definition for status message https://www.w3.org/TR/WCAG22/#dfn-status-messages is indeed a bit flawed.

If the idea is that definitions can be replaced with their longer explanation, then the wording for https://www.w3.org/TR/WCAG22/#status-messages does become (adapting it to the plural form used in the SC)

"In content implemented using markup languages, *changes in content that are not [changes of context](https://www.w3.org/TR/WCAG22/#dfn-change-of-context), and that provides information to the user on the success or results of an action, on the waiting state of an application, on the progress of a [process](https://www.w3.org/TR/WCAG22/#dfn-processes), or on the existence of errors* can be [programmatically determined](https://www.w3.org/TR/WCAG22/#dfn-programmatically-determinable) through [role](https://www.w3.org/TR/WCAG22/#dfn-role) or properties such that they can be presented to the user by [assistive technologies](https://www.w3.org/TR/WCAG22/#dfn-assistive-technologies) without receiving focus." 

This does somehow miss out the idea that status messages are first and foremost that ... messages, explicit notifications. And it makes it sound like the SC applies to any change of content that isn't a change of context, which we explicitly didn't want to have - that's why this was scoped to "status messages" in the first place.

Suggest we either modify the wording of the definition to limit its scope back down, or - if that's less painful - add a note (ideally in the definition, but at a stretch in the understanding doc) that spells it out... that this is strictly about explicit status messages.

DaleenRabe:
I agree with @patrickhlauke idea, the ambiguity is the source of the confusion from the previous discussion.

The current definition is a little confusing because it describes the effect (a change in content that informs the user) rather than the thing we are meant to be marking up (an explicit message).

To prevent this confusion for developers and auditors, I believe his suggestion to modify the definition itself is the best path forward. A note would help, but fixing the definition at its source would be a more robust and permanent solution.

mbgower:
Getting back to our discussion last week, we are only putting things into the TF discussion when there is something drafted. 
If someone wants to work on a revision to the definition and get it into draft -- or compose a draft WG 'response only' -- it will come up through that process.

Personally, I neither think there's a big problem here, nor do I think you are easily going to come up with something to 'fix' it normatively.

patrickhlauke:
Filed https://github.com/w3c/wcag/pull/4840

PR #4840, opened 2025-12-30
Update definition for "status message"

patrickhlauke:
The current definition for "status message" https://www.w3.org/TR/WCAG22/#dfn-status-messages manages to somehow miss out the fact that a status message refers to an actual discrete "message". The current definition makes it sound like any change of content *may* be interpreted as an "intrinsic" status message (see discussion https://github.com/w3c/wcag/discussions/4672).

This PR changes the (normative) definition to make it clear that the WG always intended this to refer to explicit "status messages", and *not* somehow mandate the need for any dynamic content update (such as dynamically updating a list of results) must trigger some sort of SR announcement.

Closes https://github.com/w3c/wcag/issues/4676

Preview (in context): https://deploy-preview-4840--wcag2.netlify.app/understanding/status-messages#dfn-status-message

### DIFF
--- a/guidelines/sc/21/status-messages.html
+++ b/guidelines/sc/21/status-messages.html
@@ -1,9 +1,5 @@
 <section id="status-messages" class="sc">
-   					
    <h4>Status Messages</h4>
-   					
    <p class="conformance-level">AA</p>
-   					
-   <p>In content implemented using markup languages, <a>status messages</a> can be <a>programmatically determined</a> through <a>role</a> or properties such that they can be presented to the user by <a>assistive technologies</a> without receiving focus.</p>
-   				
+   <p><a>Status messages</a> that don't cause a <a>change of context</a> can be <a>programmatically determined</a> such that they can be presented to the user by <a>assistive technologies</a> without receiving focus.</p>
 </section>


patrickhlauke:
Slotting the new definition in place of the term into the normative wording for the SC, results in:

> In content implemented using markup languages, ***discrete pieces of content that are changed/updated dynamically and provide information to the user on the success or results of an action, on the waiting state of an application, on the progress of a process, or on the existence of errors*** can be [programmatically determined](https://www.w3.org/TR/WCAG22/#dfn-programmatically-determinable) through [role](https://www.w3.org/TR/WCAG22/#dfn-role) or properties such that they can be presented to the user by [assistive technologies](https://www.w3.org/TR/WCAG22/#dfn-assistive-technologies) without receiving focus.

(note that the definition is for "status message", singular, while the SC uses "status messages", plural, so I tweaked the replacement accordingly to make it plural)

compare this to the far more open-ended/vague current definition

> "In content implemented using markup languages, ***changes in content that are not [changes of context](https://www.w3.org/TR/WCAG22/#dfn-change-of-context), and that provide information to the user on the success or results of an action, on the waiting state of an application, on the progress of a [process](https://www.w3.org/TR/WCAG22/#dfn-processes), or on the existence of errors*** can be [programmatically determined](https://www.w3.org/TR/WCAG22/#dfn-programmatically-determinable) through [role](https://www.w3.org/TR/WCAG22/#dfn-role) or properties such that they can be presented to the user by [assistive technologies](https://www.w3.org/TR/WCAG22/#dfn-assistive-technologies) without receiving focus."

electrickite:
@patrickhlauke I feel this new definition leaves us in a similar situation. To pressure test it, consider a received chat message. It satisfies "discrete pieces of content that are changed/updated dynamically" and also "provide[s] information to the user on the success or results of an action" and would be a status message under this new definition. Similar arguments can be made for dynamic search results. 

The distinction we are trying to draw here seems to be around author intent. I wonder if something like this would capture it?

> change in content that is not a change of context, the primary purpose of which is to provide information to the user on the success or results of an action, on the waiting state of an application, on the progress of a process, or on the existence of errors

The primary purpose of a chat message is not to "provide information to the user on the success or results of an action", it is to convey the text of the message. Similarly, the primary purpose of a set of search results is to display the links to matching pages.

guidelines/terms/21/status-message.html
```diff
@@ -1,6 +1,4 @@
 <dt><dfn id="dfn-status-messages" data-lt="status messages">status message</dfn></dt>
 <dd>
-   					
-   <p>change in content that is not a <a>change of context</a>, and that provides information to the user on the success or results of an action, on the waiting state of an application, on the progress of a <a>process</a>, or on the existence of errors</p>
-   
+   <p>discrete piece of content that is changed/updated dynamically and provides information to the user on the success or results of an action, on the waiting state of an application, on the progress of a <a>process</a>, or on the existence of errors</p>
```
patrickhlauke:
```suggestion
   <p>discrete piece of content that is changed/updated dynamically to provide information to the user on the success or results of an action, on the waiting state of an application, on the progress of a <a>process</a>, or on the existence of errors</p>
```

patrickhlauke:
@electrickite still think it's a particularly obtuse reading of the words, but ... changed "and provides" to "to provide" which hopefully removes any such misinterpretation

electrickite:
@patrickhlauke My job sometimes requires that I read standards as obtusely as possible, so I appreciate the change! :)

bruce-usab:
I so very much appreciate removing "change of context" from the definition for status message!  Currently, there is some circular logic, but I have some concern that this is a change of meaning.  I am looking forward to discussion on TF call!

bruce-usab:
Reviewed on TF call 1/16 and consensus for change.  Since it is normative, will need to go through CFC.  On call @scottaohara volunteered to update Understanding reflecting the question raised by issue.

patrickhlauke:
Just for comparison, grafting the current definition of status message into the 4.1.3 SC text:

> In content implemented using markup languages, **change in content that is not a change of context, and that provides information to the user on the success or results of an action, on the waiting state of an application, on the progress of a process, or on the existence of errors** can be programmatically determined through role or properties such that they can be presented to the user by assistive technologies without receiving focus.

Grafting the proposed new wording in:

> In content implemented using markup languages, **discrete piece of content that is changed/updated dynamically to provide information to the user on the success or results of an action, on the waiting state of an application, on the progress of a process, or on the existence of errors** can be programmatically determined through role or properties such that they can be presented to the user by assistive technologies without receiving focus.

---

@WilcoFiers

> By removing "change of context" I think you unintentionally added a requirement that even if a status message is focused when it appears, it also has to be programmatically determined.

Looking at the grafted part above, I would think that the "without receiving focus" part in the SC text itself takes care of this. A "status message" itself is not tied to whether or not it does or doesn't receive focus ... the scoping of it needing to be announced when it's NOT receiving focus is part of the SC, not intrinsic to the definition of what a status message is, I'd say

> I also think the wording isn't very clean. The "changed/update" phrase should probably be "changed or updated". 

sure, happy to tweak this

> The word "dynamic" seems redundant. A discrete change or update is always dynamic, no?

This tries to reinforce that it's a change that happens without, say, a page reload - to make sure that this doesn't unintentionally scope itself to "the page reloads, and there's a new count somewhere saying 'Results 11-20 of 100" now, and THAT now needs to be announced on page load"

camusamta:
Really in support of this change, but something about 'discrete pieces of content' isn't sufficiently descriptive or clarifying for me. Is there a way to be clearer still, and more irreducible, perhaps with something like:

> **instructive content that is added or updated dynamically** to provide information to the user on the success or results of an action, on the waiting state of an application, on the progress of a process, or on the existence of errors 

- "instructive content" helps to reaffirm that this intended for 'explicit messages'
- "added" because a status message can appear as well as update

**Edit:** maybe just "content", even, instead of "instructive" or "discrete"?

patrickhlauke:
> Looking at the grafted part above, I would think that the "without receiving focus" part in the SC text itself takes care of this. A "status message" itself is not tied to whether or not it does or doesn't receive focus ... the scoping of it needing to be announced when it's NOT receiving focus is part of the SC, not intrinsic to the definition of what a status message is, I'd say

@WilcoFiers  unless you mean that a "change of context" also covers the scenario where the content is changed under people's focus/feet (as in focus isn't moved per se, the content that is currently focused is) - which i think is a novel interpretation that (although logical once one thinks about it) doesn't come out of the definition for change of context https://www.w3.org/TR/WCAG22/#dfn-change-of-context (as in: don't think this was ever the intended interpretation for having that wording in the original status message definition?)

bruce-usab:
I think the benefits of the change outweigh the cons, even considering the churn necessary for CFC.

This isn’t the only place it happens, but it’s problematic to have too much of the detail (i.e., not a change of context) embedded in the definition — especially when that work is done by the SC.

WilcoFiers:
> @WilcoFiers unless you mean that a "change of context" also covers the scenario where the content is changed under people's focus/feet (as in focus isn't moved per se, the content that is currently focused is) - which i think is a novel interpretation that (although logical once one thinks about it) doesn't come out of the definition for change of context https://www.w3.org/TR/WCAG22/#dfn-change-of-context (as in: don't think this was ever the intended interpretation for having that wording in the original status message definition?)

No that's not what I meant. That is an interesting scenario I'd never considered that. That might be a change of focus though since focus is set back to the body if the focused element is pulled? I'm not sure, but that isn't my main concern.

> discrete pieces of content

This seems vague and raises questions for me that I don't have with the current text.

dbjorge:
I sympathize with wanting to clarify the scenario from the motivating issue, but I don't think this proposed change achieves that goal *and* I agree that it creates new issues like Wilco pointed out.

## 1: It doesn't fix the issue

I don't think the new wording is actually any more clear than the original that it would exclude "a list of search results that appears after clicking a search button" from being considered a "status message". Comparing the wording:

- (old) "change in content that is not a change of context, and that provides information to the user on the success or results of an action"
- (new) "discrete piece of content that is changed/updated dynamically to provide information to the user on the success or results of an action"

...I don't see why someone wouldn't consider the list of results to be "a discrete piece of content", and I think it obviously "provides information... on the results of an action".

## 2: Dropping "change of context" is breaking

I agree with Wilco that dropping the "change of context" wording changes the requirement in a way that's probably not intended, and I don't think we should do that in an erratum.

@patrickhlauke , you said:

>> By removing "change of context" I think you unintentionally added a requirement that even if a status message is focused when it appears, it also has to be programmatically determined.
>
> Looking at the grafted part above, I would think that the "without receiving focus" part in the SC text itself takes care of this. A "status message" itself is not tied to whether or not it does or doesn't receive focus ... the scoping of it needing to be announced when it's NOT receiving focus is part of the SC, not intrinsic to the definition of what a status message is, I'd say

I don't think this is right. Consider a case where a user performs an action and then focus is moved to information about the result. In the old wording, the movement of focus means that this is a "change of context" and thus not a "status message" for the purpose of the requirement. In the new wording, this now *is* a status message, and so now you've added a requirement to announce it (which might require announhcing more than just the change of focus would have done on its own).

## 3: "dynamic" and "discrete" are vague

I agree with @WilcoFiers and @camusamta 's feedback that "dynamic" and "discrete" both feel vague/off, and that it'd make sense to include "added" and not merely "changed/updated".

patrickhlauke:
> 1: It doesn't fix the issue

Maybe I'm looking at this too closely myself, but saying that a "discrete piece of content  that is changed/updated dynamically **to** provide information to the user on the success or results of an action..." to me would indicate that that's its sole purpose, and would not include the actual results/updated table/whatever, whereas "**that** provides information to the user" has more of that ambiguity.

But would welcome alternative suggestions on how to tackle the original issue that motivated this, as it's undisputable that currently the vague definition of status message lends itself to interpreting it in a way that would mandate ANY dynamic change/update as a result of an action needing to be announced explicitly (including things like "in a form, if you select a radio button, a new field appears"), which was not the original intent of the SC.

Maybe expanding to "discrete piece of content whose only purpose is to provide the user..." ... or would that still suggest that the list of results etc can also be seen as that?

Because as far as i remember, we never intended the actual results etc to be covered by this requirement, so somehow we need to make that clear.

> I don't think this is right. Consider a case where a user performs an action and then focus is moved to information about the result. In the old wording, the movement of focus means that this is a "change of context" and thus not a "status message" for the purpose of the requirement. In the new wording, this now is a status message, and so now you've added a requirement to announce it (which might require announcing more than just the change of focus would have done on its own).

i think i'm seeing what you're getting at ... you saying that regardless of the "without receiving focus" at the end of the SC, the normative wording with the new definition inserted would now suggest that it must be (theoretically) possible for this to announce even IF it wasn't receiving focus?

i think similar to the painful confusion we had with the "pointer" definition, this "status message" definition is wrongly overloaded but now necessary because the vague / not precise enough wording of the SC? because to me, a status message is a status message, regardless of whether it receives focus or not. it's only that WHEN it happens and doesn't receive focus, we want to make sure it's programmatically determinable so that it can be announced. now wondering if that could only be fixed with an additional change to the normative wording of the SC itself.

EDIT: *if* we were to also touch the normative SC wording, we'd probably want to change "without receiving focus" to "when it doesn't receive focus"

TestPartners:
Even "the success or results of an action" is problematic because some status messages are not the result of an action by anyone, let alone the user. A session timeout warning would be one example, as is a notification that you have new messages that may have been created automatically by the system.

camusamta:
@patrickhlauke: 

> Maybe expanding to "discrete piece of content whose only purpose is to provide the user..." ... or would that still suggest that the list of results etc can also be seen as that?

I think this is really close. Because I still think 'discrete' raises the possibility for the same misinterpretations you're trying to address, what about something like:

"**information that is added or updated dynamically, whose only purpose is to instruct the user** on the success or results of an action, on the waiting state of an application, on the progress of a process, or on the existence of errors"

**The grafted-in version:**

> "In content implemented using markup languages, i**nformation that is added or updated dynamically, whose only purpose is to instruct the user on the success or results of an action, on the waiting state of an application, on the progress of a process, or on the existence of errors** can be programmatically determined through role or properties such that they can be presented to the user by assistive technologies without receiving focus.

The word 'only' could possibly be swapped for 'explicit', too, if that still reads as having ambiguity.

lilzesh:
> In content implemented using markup languages, information that is added or updated dynamically, whose only purpose is to instruct the user on the success or results of an action, on the waiting state of an application, on the progress of a process, or on the existence of errors** can be programmatically determined through role or properties such that they can be presented to the user by assistive technologies without receiving focus.

This still has the unintentional requirement Wilco was referring to in Patrick's initial proposal right, i.e., if the status message is focused, it would then still be required to be programmatically determined through role/properties? Can we simply slot change of context back in, in a different location, but retain what you have come up with to replace "change in content?"

E.g.:

> In content implemented using markup languages, information that is added or updated dynamically, whose only purpose is to instruct the user on the success or results of an action, on the waiting state of an application, on the progress of a process, or on the existence of errors** can be programmatically determined through role or properties such that they can be presented to the user by assistive technologies without receiving focus, **unless a change of context has occurred**.

patrickhlauke:
> Can we simply slot change of context back in

I would really like to avoid doing that in the glossary/term definition ... if we're already going through the pain of a normative change, I'd be more in favour of slotting it somehow into the normative SC wording, and keeping the definition of what a "status message" is clean (a status is a status, whether it receives focus or not)

bruce-usab:
SC 4.1.3 includes “without receiving focus” so, as Patrick notes, that (redundantly) addresses the “not a change of context” clause of status messages.  The glossary term could drop that, leaving:
> change in content that provides information to the user on the success or results of an action, on the waiting state of an application, on the progress of a process, or on the existence of errors

We could also add note addressing #4676 clarifying than an updated search result (or whatever) does not meet this definition.

bruce-usab:
Discussed on TF call 6 Fed.  @patrickhlauke will take another pass.  From discussion, just changing the term without updated the SC may not be feasible.  It was noted that the current phrasing of 4.1.3 precludes best practice of using ARIA Notify — because 4.1.3 requires role or properties (which is not how ARIA Notify works).  The limiting current scoping to “markup languages” also limits the generalizability of this ICT (e.g., WCAG2ICT).

TestPartners:
> the current phrasing of 4.1.3 precludes best practice of using ARIA Notify

Why do you refer to ARIA Notify as a "best practice"? As far as I can tell, it does not improve the user experience in any way. It may make it easier for developers to implement announcements, but the poor discoverability makes testing much more difficult if you don't know what announcements a page is supposed to make (if any), and what triggers them. We will need to build new testing tools to help with this.

patrickhlauke:
after discussion in the TF meeting two weeks ago, this is now a more fundamental rewrite:

* modifies the definition for status message
* crucially (as we're already in "normative change" territory) also modifies the normative wording of the SC to make it cover new technologies like `ariaNotify` (which would not be covered by the "`role` or properties" part) and non-HTML web content (by removing the markup languages clause, meaning that this SC could then also be applied in principle to native apps in light of WCAG2ICT cases)

Doing the combined "normative SC text with grafted in definition" dance, this PR would land on:

> **Status messages** that don't cause a *change of context* can be *programmatically determined* such that they can be presented to the user by *assistive technologies* without receiving focus.

> **Discrete piece[s] of content that [are] added or updated dynamically, and whose only purpose is to provide information to the user on the success or results of an action, on the waiting state of an application, on the progress or completion of a process, or on the existence of errors**  that don't cause a *change of context* can be *programmatically determined* such that they can be presented to the user by *assistive technologies* without receiving focus.

patrickhlauke:
Closing this PR in favour of its successor https://github.com/w3c/wcag/pull/4952 (mainly because the description, thumbs up/down, and discussion/comments, may be confusing since this morphed considerably since it was first filed).

Issue #3344, opened 2023-08-15
Overlap of 2.5.1 Pointer Gestures and 2.5.7 Dragging Movements


patrickhlauke:
If we have something like a path-based slider (where, as soon as you move your pointer out of the track, the slider "knob" doesn't follow the mouse anymore), 2.5.1 applies...but does 2.5.7? Or is the thinking that because it's path-based, it doesn't count as a pure "dragging movement", hence 2.5.7 does not apply?

and If so, should the definition for "dragging movement" be tweaked to make it clear that path-based movements (that still involve "dragging") are *excluded* from that definition?

https://w3c.github.io/wcag/guidelines/22/#dfn-dragging-movements

or should the normative wording for 2.5.7 be tweaked/clarified (probably too late now)

"All [functionality](https://w3c.github.io/wcag/guidelines/22/#dfn-functionality) that uses a [dragging movement](https://w3c.github.io/wcag/guidelines/22/#dfn-dragging-movements)" > "All [functionality](https://w3c.github.io/wcag/guidelines/22/#dfn-functionality) that uses a [dragging movement](https://w3c.github.io/wcag/guidelines/22/#dfn-dragging-movements) *that is not **path-based***"

bruce-usab:
Bruce will try and find related conversation.  Dragging moving definition is very specific with mouse up/down.  So if following stops without mouse up, might not meet definition.

@patrickhlauke  we discussed on backlog call 8/18 -- can you provide another example?

patrickhlauke:
having a hard time parsing that first sentence of yours, but to be clear: in general, any time you grab something and then have to follow a specific path and then release (sliders being the most obvious ones, when/if they are truly path-based and bail out of the interaction when you leave the slider track). it then is both a path-based interaction, and a dragging interaction, so assume that it then falls under *both* SCs. and if so, should be acknowledged in the respective understanding documents (happy to do a PR, but not until I hear if this take is "correct" based on other folks' understanding of these here)

bruce-usab:
Also from call, adding example to Understanding with using finger as pointer would be helpful to make it clearer that SC is not only for mouse.

bruce-usab:
@patrickhlauke in the conversation today, we focused on [the glossary definition](https://www.w3.org/TR/WCAG22/#dfn-dragging-movements), emphasis added:

> An operation where the pointer engages with an element on the down-event and the element ... follows the pointer **until an up-event** 

[...]

Issue #2684, opened 2022-09-22
2.5.7 Dragging Movements - Carousels - Overflowed Containers


joe-watkins:
**Warning**: Animated Gifs below

Hello :) Excited about 2.2!! 

There is some debate out there about whether or not `2.5.7 Dragging Movements` applies to a typical carousel of items that do not have forward/back buttons. Can someone from the WG let me know if either of the following two scenarios fails `2.5.7 Dragging Movements`? If not, how do the examples satisfy the SC or not fail it?

<details>
<summary>
1. Carousel of products with custom overflowed container requiring a pointer to drag the bar at the bottom to move the carousel forward and back. There are no other controls to perform this same functionality</summary>

![](https://user-images.githubusercontent.com/3695795/191816180-30b334e5-3875-4431-abbc-f71f7d618ad5.gif)
</details>

<details>
<summary>
2. Carousel of products on a touch device where a touch user must swipe/flick left or right to advance through the carousel. There are no nearby controls to perform this functionality. </summary>

![](https://user-images.githubusercontent.com/3695795/191815735-f6309357-1885-4ff9-8d43-7779210389e3.gif)
</details>

bruce-usab:
My own opinion is that neither example above is a <q>dragging movement</q> because swipe left/right is not <q>follows the pointer</q>.  There is only one axis of movement, and the element is limited to its frame.

> [dragging movement](https://www.w3.org/TR/WCAG22/#dfn-dragging-movements):  an operation where the pointer engages with an element on the down event and the element (or a representation of its position) follows the pointer

@a11ydoer Might this be a candidate (as counter-example) for Understanding?

joe-watkins:
@bruce-usab Thank you for your speedy reply. I look forward to others' feedback too!

Confusion may also come from the term "sliders" in the Intent. "sliders" is commonly used as a term for "carousels". What are "sliders" in this case?

> The intent of this Success Criterion is to ensure functionality that uses a dragging movement (e.g., **sliders**, drag-and-drop interfaces) has another [single pointer](https://www.w3.org/WAI/WCAG22/Understanding/dragging-movements#dfn-single-pointer) mode of operation without the need for the dexterity required to drag elements.

yatil:
Looks to me like this is indeed covered by above definition of “Dragging movement”. There is “engagement” on the down event, and the carousel follows the mouse pointer.

(And to my knowledge the intention of the SC is that such interactions have visible or at least keyboard-operable alternatives.)

[...]

Issue #2394, opened 2022-05-12
First time use of 'dragging' in 2.5.1: Pointer Gestures must be a link to the definition


jake-abma:
https://w3c.github.io/wcag/understanding/pointer-gestures

As we have a definition now (which explains dragging movements a bit better than in this Understanding doc) we must 'link' the first instance to the definition.

Sentence starts with:

> Dragging is a movement where the user picks up

Links "Dragging is a movement" to the definition? Or re-write the sentence a bit?

jake-abma:
As we are in Candidate Recommendation do we need to fix this already by now?

bruce-usab:
Flagging as suggested item a backlog conversations  @mbgower @a11ydoer 

Is swipe left/right [2.5.7](https://www.w3.org/TR/WCAG22/#dragging-movements) dragging?  I do not think so, but some placing in Understanding imply otherwise.  (Or it could be, just as easily, that I am mistaken.)

bruce-usab:
From 2.5.1 Understanding:
> Examples of path-based gestures include swiping, sliders and carousels dependent on the direction of interaction

I would not consider any of those three UI experiences to be <q>input that depends on the path of the user's movement and not just the endpoints</q> so, now that we have 2.5.7 dragging, it seems confusing to have those in the same sentence as <q>other gestures which trace a prescribed path such as drawing a specific shape</q>.

patrickhlauke:
swiping is usually a quick "flick" that is highly directional, not a dragging motion (though one can split hairs over exactly where the dividing line here is)

bruce-usab:
And sometimes the speed/intensity of the flick is important.  Speed/intensity is analog and I think worth mentioning under 2.5.1.  I agree the dividing line is hard to nail down.  Regardless, I think the editorial changes needed for 2.2 TR is a bit more than adding a link.

GreggVan:
Agree with Bruce

When we created this  - The purpose of this was specifically to cover places where you can’t say     go from here to here .     Or go from here in this direction.

It was meant to exclude things like  flying a plane through obstacles — in real time.   

Even drawing shapes is not covered  — though  freehand drawing (not regular shapes) could fall into this. 

Gregg Vanderheiden
***@***.***

bruce-usab:
We discussed dragging movement on the backlog call today.  Sliders should not be in the example (.e.g.) list towards the top of Understanding.  Ping to @mbgower because he thought a PR correcting had already been merged.

[...]

PR #4843, opened 2025-12-31
Disambiguation of path-based gesture versus dragging movement

patrickhlauke:
This PR is a long-overdue disambiguation that removes the half-hearted retconning attempt to make pointer gestures apply to sliders at the time before we had dragging movements. (yes, this may mean that for 2.1, sliders MAY not be covered by pointer gestures anymore, but even with the additions we had, they weren't per se...so this doesn't actually change anything normatively in my mind, other than remove the weird points of overlap and confusion).

## Pointer gestures

In understanding for pointer gestures:

* Remove the confusing "intermediate"/"midpoint" patch that was added mostly as a way to retcon/patch in applicability to certain looser swipe gestures and to some types of sliders, and instead explains in much simpler terms that while not strictly "path-based", the fact that the movement is only recognised when the path is "mostly" in a particular direction does make them count as path-based.
* Emphasise that for gestures, the end-point (and sometimes even the start-point) often don't actually matter so much as the direction (and potentially speed)
* Repurpose the "wiggly line from point 1 to point 2" to actually illustrate dragging movement, and move the two illustrations (straight line and wiggly line) to the section that tries to explain the difference between dragging and path-based gestures
* Use some consistency: start-point and end-point (not "start point", or "endpoint" etc, it was all a bit mix and match before)
* Adds a mention about gestures used to control UA/AT (similar to the mention present in the dragging movement understanding)
* Adds a very explicit mention that *sliders* are generally implemented using dragging movement, and are *not* path-based. Removes the very borderline example of a badly implemented slider
* Removes the artificial/stretched example of a kanban board that responds to swipes (which would be unusual) from the pointer gesture examples, rewords it to be an example for dragging movement
* Clarifies the (rare) instance of what happens when something (like a slider) involves *both* dragging *and* is path-based (that it may fail *both* SCs)
* Part of some confusion is how the word "slider" gets interpreted - some people say "slider" when they talk about a "content slider" or "carousel"; even one of the examples in pointer gestures fell into that ambiguity. This PR tries to disambiguate things a bit more and talks about "range slider" and "draggable carousels"
* Adds figures that show a range input slider and a draggable content carousel, to clarify the point
* Removes the paradoxical "input sliders are usually not path-based when used with mouse, but in some cases they're path-based when used with touchscreens" part. First of all, this seems to only happen in Android/Chrome (even Android/Firefox doesn't do this, and iOS doesn't). Secondly, this is a user agent choice, and authors can't be made responsible for it. Thirdly, we can't have these sorts of weird conditional "it's not path-based, EXCEPT on touch" statements...how is an author supposed to deal with that?
* In pointer gestures understanding, corrects mention of "swipes" for scrolling etc, and clarifies that these are on touch screen devices
* Remove the "Challenges for people with disabilities" section - after removing the inaccurate section about the difference between touchscreen and mouse operation for range sliders, the section only contains general points that are far better made in the intent section

Current version: https://www.w3.org/WAI/WCAG22/Understanding/pointer-gestures.html
Preview of the proposed changes: https://deploy-preview-4843--wcag2.netlify.app/understanding/pointer-gestures
[Diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FUnderstanding%2Fpointer-gestures&doc2=https%3A%2F%2Fdeploy-preview-4843--wcag2.netlify.app%2Funderstanding%2Fpointer-gestures)

## Dragging movements

In understanding for dragging movements:

* Remove the mention of "intermediate points"
* Small language tweaks
* Same consistency change (start-point / end-point)
* Adds a variant example of a draggable news item carousel (mirroring the example of a carousel that can be flicked/swiped that exists in pointer gestures)
* Adds figures that show a range input slider and a draggable content carousel, to clarify the point

Current version: https://www.w3.org/WAI/WCAG22/Understanding/dragging-movements.html
Preview of the proposed changes: https://deploy-preview-4843--wcag2.netlify.app/understanding/dragging-movements
[Diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FUnderstanding%2Fdragging-movements&doc2=https%3A%2F%2Fdeploy-preview-4843--wcag2.netlify.app%2Funderstanding%2Fdragging-movements)

---

Closes https://github.com/w3c/wcag/issues/3344
Closes https://github.com/w3c/wcag/issues/2394
Supersedes https://github.com/w3c/wcag/pull/3148
Closes https://github.com/w3c/wcag/issues/2684


patrickhlauke:
It's a draft PR for now, as I want to see if I can also cover/address other aspects such as https://github.com/w3c/wcag/pull/3148 <del>https://github.com/w3c/wcag/issues/2116</del> and https://github.com/w3c/wcag/issues/2684

EDIT: worked some more on this PR, this now addresses these other issues as well

detlevhfischer:
Looks very solid, great work, @patrickhlauke

understanding/22/dragging-movements.html
```diff
@@ -15,60 +15,59 @@ <h2>In brief</h2>
...
 
-      <p>This criterion does not apply to scrolling enabled by the user-agent. Scrolling a page is not in scope, nor is using a technique such as <abbr title="Cascading Style Sheets">CSS</abbr> <code>overflow</code> to make a section of content scrollable.</p>
+      <p>This criterion does not apply to scrolling and dragging gestures enabled by the user agent. Scrolling a page is not in scope, nor is using a technique such as <abbr title="Cascading Style Sheets">CSS</abbr> <code>overflow</code> to make a section of content scrollable, unless content actively suppresses the user agent's own scrolling functionality and implements its own (at which point the dragging gesture is interpreted/processed by the content itself, and thus falls under the responsibility of the content author). Similarly, on touch screen browsers, pages can be refreshed/reloaded using a "drag to refresh" gesture – as this is functionality provided by the browser, rather than implemented by the content, this criterion does not apply to it.</p>
```
detlevhfischer: > is not in scope, nor is using a technique

this is fine content-wise, but the nor-construction might not be that easy to parse especially for non-native speakers. Suggest to make this a separate sentence, starting with something like "The criterion also does not apply to cases where CSS overflow techniques are used to make content sections scrollable...", or similar.

understanding/22/dragging-movements.html
```diff
@@ -15,60 +15,59 @@ <h2>In brief</h2>
...
 
-         <p>This success criterion applies to dragging movements as opposed to pointer gestures, which are covered in <a href="pointer-gestures">Success Criterion 2.5.1 Pointer Gestures</a>. Pointer gestures include directional path-based as well as multi-point gestures. In contrast, for dragging movements, only the start and end point of the movement matters, not the actual path.</p>
+         <p><a href="keyboard">Success Criteria 2.1.1 Keyboard</a> and <a href="keyboard-no-exception">2.1.3 Keyboard (No Exception)</a> require dragging features to be keyboard accessible. However, achieving keyboard equivalence for a dragging operation does not automatically meet this success criterion. It is possible to create an interface that works with dragging and keyboard controls, but still does not work using only clicks or taps. While many designs can be created for a dragging alternative which address both keyboard accessibility and operability by single pointer operation, the two requirements are be assessed independently.</p>
```
detlevhfischer: > achieving keyboard equivalence for a dragging operation does not automatically meet this success criterion

The way this is put invites the interpretation that there are potentially ways by which keyboard operability would meet the criterion -- which I think is not the case (the point being that any extra stuff out in to support keyboard operation would need to be pointer-operable regardless of whether it supports keyboard operation).

detlevhfischer: > It is possible to create an interface that works with dragging and keyboard controls, but still does not work using only clicks or taps.

I'd rather be more explicit and say something like, "Providing functionality for keyboard operation does not satisfy this Success Criterion since it requires that there is a pointer operable alternative. For example, it may be possible to select a draggable element using the space bar and moving it to another position using the arrow keys. This would not help pointer users with no access to, or no capability, to use a keyboard.

detlevhfischer: > the two requirements are be assessed independently.

Assumes an auditing perspective, but may better be worded more generally.

understanding/22/dragging-movements.html
```diff
@@ -15,60 +15,59 @@ <h2>In brief</h2>
...
 
-         <p>Additional examples are selection rectangles that set the first x/y rectangle coordinate at the pointer position via a pointer down-event, and the second x/y coordinate, after a dragging movement, at the next up-event. A similar example is a connecting line drawn between two different items on the screen, as in an allocation test where users are required to draw a line between questions and corresponding answers. In these cases, the dragging movement requires an alternative way to accomplish the same action that does not rely on the dragging movement. For example, two separate single tap or click actions may define the rectangle coordinates or the start and end points of a connecting line.</p>
+         <p>This success criterion applies to dragging movements as opposed to pointer gestures, which are covered in <a href="pointer-gestures">Success Criterion 2.5.1 Pointer Gestures</a>. Pointer gestures include directional path-based as well as multi-point gestures. In contrast, for dragging movements, only the start- and end-point of the movement matters, not the actual path.</p>
```
detlevhfischer: > only the start- and end-point of the movement matters

matters -> matter

understanding/22/dragging-movements.html
```diff
@@ -15,60 +15,59 @@ <h2>In brief</h2>
...
+
+         <p>Additional examples are selection rectangles that set the first x/y rectangle coordinate at the pointer position via a pointer down-event, and the second x/y coordinate, after a dragging movement, at the next up-event. A similar example is a connecting line drawn between two different items on the screen, as in an allocation test where users are required to draw a line between questions and corresponding answers. In these cases, the dragging movement requires an alternative way to accomplish the same action that does not rely on the dragging movement. For example, two separate single tap or click actions may define the rectangle coordinates or the start- and end-points of a connecting line.</p>
```
detlevhfischer: > at the next up-event


at the next pointer up-event? (perhaps too nit-picky...)

detlevhfischer: > as in an allocation test where users are required to draw a line between questions and corresponding answers

Maybe it would be better to separate these cases? Wnile I can't think of a fundamentally different way to do rectangle selection (you't probalby have some instruction and a controll that starts the process, then interpreting the next two clicks taps as setting start and end point, but I have never see the like so far), the more obvious keyboard AND pointer-operable alternative would be to offer a select at each allocation test question allowing to select the corresponding answer.

understanding/22/dragging-movements.html
```diff
@@ -15,60 +15,59 @@ <h2>In brief</h2>
...
 
       </section>
       
       <section>
          <h3>Alternatives for dragging movements on the same page</h3>
-   
+
          <p>Where functionality can be executed via dragging movements and an equivalent option exists that allows for single-pointer access without dragging, this success criterion is passed. It does not have to be the same component, so long as the functionality is equivalent. An example is a color wheel where a color can be changed by dragging an indicator. In addition, text fields for the numerical input of color values allow the definition of a color without requiring dragging movements. (Note that a text input is considered device agnostic; although the purpose is to enter characters, text entry can take place through voice, pointer or keyboard.)</p>
```
detlevhfischer: > this success criterion is _passed_

passes, or is satisfied?

understanding/22/dragging-movements.html
```diff
@@ -15,60 +15,59 @@ <h2>In brief</h2>
...
 
       </section>
       
       <section>
          <h3>Distinguishing dragging movements from path-based pointer gestures</h3>
-      
 
-         <p>Dragging movements covered in this success criterion are pointer interactions where only the start- and endpoints matter. Once the pointer engages with a target, the direction of the dragging movement does not factor into the interaction until the pointer disengages the target. Since the dragging movement does not have an intermediate point, the dragging movement can go in any direction. Path-based gestures are covered in <a href="pointer-gestures.html">Success Criterion 2.5.1 Pointer Gestures</a>.</p> 
+         <p>Dragging movements covered in this success criterion are pointer interactions where only the start- and end-points matter. Once the pointer engages with a target to pick up/grab it, the direction of the dragging movement does not factor into the interaction until the pointer disengages the target. This is different from a path-based gesture, where the direction of the pointer movement matters. Path-based gestures are covered in <a href="pointer-gestures.html">Success Criterion 2.5.1 Pointer Gestures</a>.</p>
```
detlevhfischer: > the direction of the dragging movement does not factor into the interaction until the pointer disengages the target

strictly speaking, the _direction_ never factors in? Maybe too nit-picky again...

detlevhfischer (approved):
Just some editorial comments - the substance is there.

patrickhlauke:
> Just some editorial comments - the substance is there.

thank you @detlevhfischer ... good to get some more of the cruft (most of it was existing text I think) out of those docs. I'll have a look through and make tweaks

understanding/22/dragging-movements.html
patrickhlauke: Done

understanding/22/dragging-movements.html
patrickhlauke: Done

understanding/22/dragging-movements.html
patrickhlauke: Done

understanding/22/dragging-movements.html
patrickhlauke: changed to "evaluated", which seems the preferred term

patrickhlauke: going to leave this as is, as there ARE situations where making things keyboard-accessible does also result in this SC being satisfied ... namely, adding buttons/controls that work for keyboard users (can be focused/activated), but then also let users click/tap those controls. explaining that here would end up diluting the main point too much, i think.

patrickhlauke: for same reason as the previous one, i'm going to leave this as is. think it's still sufficiently clear.

patrickhlauke: did expand the point a tiny bit, but now feels like it's saying the same thing twice...but maybe that's ok.

understanding/22/dragging-movements.html
patrickhlauke: Done

understanding/22/dragging-movements.html
patrickhlauke: done

patrickhlauke: made those examples hopefully clearer, split them out, AND moved them to the actual examples section (there was no point having them "inline" where they were) https://github.com/w3c/wcag/pull/4843/commits/e370b5a0c97ab4aca4ec96cd876418aba0956126

understanding/21/pointer-gestures.html
```diff
@@ -73,7 +81,9 @@ <h3>Challenges for people with disabilities</h3>
...
+			<p>This success criterion <em>does not apply</em> to gestures that involve dragging in any direction because only the start- and end-points matter in a dragging operation. However, such gestures do require fine motor control. Authors are encouraged to provide non-dragging methods, for instance, a drag and drop operation could also be achieved by selecting an item (with a tap or keyboard interaction) and then selecting its destination as a second step.</p>
+
+			<p>This criterion does not apply to gestures required to control the user agent. For instance, on most browsers on touch screen devices, swiping left triggers the browser's "Back" function – as this is functionality of the browser, rather than of the content, this gesture is not covered by the requirements of this criterion. Similarly, the criterion does not apply to gestures used to operate a screen reader on a touch screen device, such as swiping left or right to move the screen reader's cursor to the previous or next element.</p>
```
fstrr:
```suggestion
			<p>This criterion does not apply to gestures required to control the user agent. For instance, on most browsers on touch screen devices, swiping left on the device's screen triggers the browser's "Back" function. As this is functionality of the browser, rather than of the content, this gesture is not covered by the requirements of this criterion. Similarly, the criterion does not apply to gestures used to operate a screen reader on a touch screen device, such as swiping left or right to move the screen reader's cursor to the previous or next element.</p>
```

patrickhlauke:
@alastc as you originally worked on this quite a bit (the retcon about "intermediate point" that originally crammed certain types of "mostly directional to begin with" swipes to pretend they're path-based), would appreciate your thoughts here

understanding/21/pointer-gestures.html
```diff
@@ -73,7 +79,9 @@ <h3>Challenges for people with disabilities</h3>
...
 
-			<p>This success criterion <em>does not apply</em> to gestures that involve dragging in any direction because only the start and end points matter in a dragging operation. However, such gestures do require fine motor control. Authors are encouraged to provide non-dragging methods, for instance, a drag and drop operation could also be achieved by selecting an item (with a tap or keyboard interaction) and then selecting its destination as a second step.</p>
+			<p>This success criterion <em>does not apply</em> to gestures that involve dragging in any direction because only the start- and end-points matter in a dragging operation. However, such gestures do require fine motor control. Authors are encouraged to provide non-dragging methods, for instance, a drag and drop operation could also be achieved by selecting an item (with a tap or keyboard interaction) and then selecting its destination as a second step.</p>
```
patrickhlauke:
```suggestion
			<p>This success criterion <em>does not apply</em> to "free-form" gestures that involve dragging in any direction because only the start- and end-points matter in a dragging operation. However, such gestures do require fine motor control. Authors are encouraged to provide non-dragging methods, for instance, a drag and drop operation could also be achieved by selecting an item (with a tap or keyboard interaction) and then selecting its destination as a second step.</p>
```

understanding/21/pointer-gestures.html
patrickhlauke:
```suggestion
			<p>This success criterion does not require all functionality to be available through pointing devices, but if it is available to pointer devices then it should not require path-based gestures. While content authors generally need to provide keyboard commands or other non-pointer mechanisms that perform actions equivalent to complex gestures (see Success Criterion <a href="keyboard">2.1.1 Keyboard</a>), this is not sufficient to conform to this success criterion. That is because some users rely entirely on pointing devices, or find simple pointer inputs much easier to perform and understand than alternatives. For example, a user relying on a head-pointer would find clicking a control to be much more convenient than activating an on-screen keyboard to emulate a keyboard shortcut, and a person who has difficulty memorizing a series of keys (or gestures) may find it much easier to simply click on a labeled control. Therefore, if one or more pointer-based mechanisms are supported, then their benefits should be afforded to users through simple, single-point actions alone.</p>
```

understanding/21/pointer-gestures.html
```diff
@@ -20,36 +20,40 @@ <h2>In brief</h2>
...
+
 			<p>The intent of this success criterion is to ensure that content can be controlled with a range of pointing devices, abilities, and assistive technologies. Some people  cannot perform gestures in a precise manner, or they may use a specialized or adapted input device such as a head pointer, eye-gaze system, or speech-controlled mouse emulator. Some pointing methods lack the capability or accuracy to perform multipoint or path-based gestures.</p>
 
 			<p>A <strong>path-based gesture</strong> involves an interaction where not just the endpoints matter, but how the pointer moves between these points.</p>
 
-			<p>Swiping is an example of a <strong>path-based gesture</strong>, which is only recognized when the user moves in a (mostly) straight line from the start point to the end point.</p>
+			<p>Swiping is an example of a <strong>path-based gesture</strong>, which is only recognized when the user moves in a (mostly) straight line from the start-point to the end-point.</p>
+
+			<p>For many gestures, the exact start- and end-points themselves are not relevant – what matters is that the overall gesture (mostly) followed a particular path.
```
detlevhfischer:
```suggestion
			<p>For many gestures, the exact start- and end-points themselves are not relevant – what matters is that the overall gesture (mostly) follows a particular path.
```

understanding/21/pointer-gestures.html
```diff
@@ -59,21 +63,25 @@ <h3>Challenges for people with disabilities</h3>
...
+			<p>This success criterion does not require all functionality to be available through pointing devices, but if it is available to pointer devices then it should not require path-based gestures. While content authors generally need to provide keyboard commands or other non-pointer mechanisms that perform actions equivalent to complex gestures (see <a href="keyboard">Success Criterion 2.1.1 Keyboard</a>), this is not sufficient to conform to this success criterion. That is because some users rely entirely on pointing devices, or find simple pointer inputs much easier to perform and understand than alternatives. For example, a user relying on a head-pointer would find clicking a control to be much more convenient than activating an on-screen keyboard to emulate a keyboard shortcut, and a person who has difficulty memorizing a series of keys (or gestures) may find it much easier to simply click on a labeled control. Therefore, if one or more pointer-based mechanisms are supported, then their benefits should be afforded to users through simple, single-point actions alone.</p>
```
detlevhfischer:
```suggestion
			<p>This success criterion does not require all functionality to be available through pointing devices, but if it is available to pointer devices then it should not require path-based gestures. While content authors generally need to provide keyboard operability as alternative for pointer operation, including complex gestures (see <a href="keyboard">Success Criterion 2.1.1 Keyboard</a>), keyboard operability is not sufficient to conform to this success criterion. That is because some users rely entirely on pointing devices, or find simple pointer inputs much easier to perform and understand than alternatives. For example, a user relying on a head-pointer would find clicking a control to be much more convenient than activating an on-screen keyboard to emulate a keyboard shortcut, and a person who has difficulty memorizing a series of keys (or gestures) may find it much easier to simply click on a labeled control. Therefore, if one or more pointer-based mechanisms are supported, then their benefits should be afforded to users through simple, single-point actions alone.</p>
```

understanding/21/pointer-gestures.html
```diff
@@ -59,21 +63,25 @@ <h3>Challenges for people with disabilities</h3>
...
 
-			<p>This success criterion <em>does not apply</em> to gestures that involve dragging in any direction because only the start and end points matter in a dragging operation. However, such gestures do require fine motor control. Authors are encouraged to provide non-dragging methods, for instance, a drag and drop operation could also be achieved by selecting an item (with a tap or keyboard interaction) and then selecting its destination as a second step.</p>
+			<p>This success criterion <em>does not apply</em> to "free-form" gestures that involve dragging in any direction because only the start- and end-points matter in a dragging operation. However, such gestures do require fine motor control. Authors are encouraged to provide non-dragging methods, for instance, a drag and drop operation could also be achieved by selecting an item (with a tap or keyboard interaction) and then selecting its destination as a second step.</p>
```
detlevhfischer: The use of "free-form gestures" here may cause difficulties since it may appear to be conflating dragging movement and pointer gesture. I know these things are difficult to prise apart, but it may be wise not to call a dragging movement a free-form gesture here and limit talk of gestures to 2.5.1 only.

So a rewording suggestion is:

This success criterion <em>does not apply</em> to dragging movements because there, only the start- and end-points matter.

understanding/21/pointer-gestures.html
```diff
@@ -86,12 +94,11 @@ <h2>Benefits</h2>
 		<section id="examples">
 			<h2>Examples</h2>
 			<ul>
-        		<li><p>A website includes a map view that supports the pinch gesture to zoom into the map content. User interface controls offer the operation using plus and minus buttons to zoom in and out.</p></li>
+				<li><p>A website includes a map view that supports the pinch gesture to zoom into the map content. User interface controls offer the operation using plus and minus buttons to zoom in and out.</p></li>
 				<li><p>A website includes a map view that supports the pinch gesture to zoom into the map content. As an single-pointer alternative, the map also allows users to double-tap, hold, and then move the pointer up or down to zoom in or out.</p></li>
```
detlevhfischer: > allows users to double-tap, hold, and then move the pointer up or down to zoom in or out

This sounds fairly complex, gesturewise? Maybe not a good example,

understanding/21/pointer-gestures.html
```diff
@@ -86,12 +94,11 @@ <h2>Benefits</h2>
...
-        		<li><p>A news site has a horizontal content slider with hidden news teasers that can moved into the viewport via a fast horizontal swipe/flicking motion. It also offers forward and backward arrow buttons for single-point activation to navigate to adjacent slider content.</p></li>
+        <li><p>A news site has a horizontal content slider with hidden news teasers that can moved into the viewport via a fast horizontal swipe/flicking motion. It also offers forward and backward arrow buttons for single-point activation to navigate to adjacent slider content.</p></li>
```
detlevhfischer:
```suggestion
        <li><p>A news site has a horizontal content slider with hidden news teasers that can moved into the viewport via a fast horizontal swipe/flicking motion. It also offers forward and backward arrow buttons for single-point activation to navigate to adjacent slider content.</p></li>
      <li><p>A news site has a horizontal content slider with hidden news teasers that can moved into the viewport via a fast horizontal swipe/flicking motion. Alternatively, a single pointer input at the outer vertical edges of the slider navigates to adjacent slider content.</p></li>
```

detlevhfischer: The second news site slider alternative that I have added is sometimes used by devs who will resist adding arrow buttons to their sliders. It does not communicate their affordance like an arrow button does, but normatively, that isn't required? I suspect that many folks would balk at this addition, but I'll float it here nevertheless...

understanding/21/pointer-gestures.html
```diff
@@ -86,12 +94,11 @@ <h2>Benefits</h2>
...
 				<li><p>A kanban widget with several vertical areas representing states in a defined process allows the user to right- or left-swipe elements to move them to an adjacent silo. The user can also accomplish this by selecting the element with a single tap or click, and then activating an arrow button to move the selected element.</p></li>
```
detlevhfischer: The kanban implementations I have encountered would all fall under dragging movements. Do we really want to retain this example? Are there implementations supporting pointer gestures in the way described?

understanding/21/pointer-gestures.html
patrickhlauke: both work (i had "followed" as generally you evaluate a gesture once it's finished)

understanding/21/pointer-gestures.html
patrickhlauke: if a dragging movement is ALSO bound to only follow a very specific path, then it DOES fall under both SCs, so some indication of this needs to still be present...

understanding/21/pointer-gestures.html
patrickhlauke: it's what google maps (and others?) does on a touchscreen device, so thought it needed to be mentioned...

understanding/21/pointer-gestures.html
patrickhlauke: sure, why not...

understanding/21/pointer-gestures.html
patrickhlauke: yeah, the kanban example as written there is a bit...forced and unrealistic/artificial. it would actually be a better example of dragging

understanding/21/pointer-gestures.html
```diff
@@ -20,36 +20,40 @@ <h2>In brief</h2>
...
+				For instance, if a page implemented a horizontal "flick" gesture to move to the previous or next page, the exact point where the user starts the "flick" or ends it is irrelevant, as long as the "flick" gesture is horizontal, is of sufficient length, and is performed above a specific speed.</p>
+
+			<p>Examples of path-based gestures include swiping, "flicking", and other gestures which trace a prescribed path such as drawing a specific shape. Such paths may be drawn with a finger or stylus on a touchscreen, graphics tablet, or trackpad, or with a mouse, joystick, or similar pointer device.</p>
 
-			<figure id="figure-path-based-gesture-1">
-				<img src="img/path-based-gesture-1.png" width="400" alt="Hand showing a starting touch, 1. Moving in a straight line to a second point, 2.">
+      <h3>The difference between Pointer Gestures and Dragging</h3>
+
+			<p>In a path-based gesture, the actual path/direction (and possibly speed) that the pointer is moved matters. For some gestures, even the exact start- and end-points may not matter.</p>
```
adampage:
```suggestion
			<p>In a path-based gesture, what matters is how the pointer moves: its path or direction, and sometimes its speed. For some gestures, even the exact start- and end-points may not matter.</p>
```

adampage: Super editorial suggestion.

patrickhlauke: might put the speed bit still in parentheses, as it's irrelevant for the actual SC (but i wanted to mention it, for completeness)

understanding/21/pointer-gestures.html
patrickhlauke: moved to be an example of dragging, after a subtle rewording

understanding/21/pointer-gestures.html
detlevhfischer: Don't want to be splitting hairs, but if a particular path is required I think that falls squarely under pointer gestures - why would we otherwise define dragging movements as "only start- and endpoint matter"?

understanding/21/pointer-gestures.html
detlevhfischer: Ah, interesting, didn't know that. The point is that double-tap-and-hold followed by moving the finger is a compound / complex gesture - so I struggle to see how it would fit the difinition of "can be operated with a [single pointer](https://www.w3.org/TR/WCAG22/#dfn-single-pointer) without a path-based gesture" - zooming requires directional vertical movement.

patrickhlauke: whether complex or not, how is it not a single pointer without a path-based gesture? it uses a single pointer - a single finger on the touchscreen; the fact that you tap it twice and hold it doesn't preclude it still being something performed with a single pointer; and the dragging up/down to zoom is not path-based, as it's freeform for users

patrickhlauke: " zooming requires directional vertical movement." but the gesture on google maps (and others) doesn't bail if you stray horizontally , so it's not path-based

understanding/21/pointer-gestures.html
patrickhlauke: to me at least, the action of the slider "per se" would be dragging, so because of that it fails dragging. the fact that the developer then also implemented it in a crap way that aborts/bails out once the drag leaves the track then *also* fails pointer gesture because it makes it path-based.

patrickhlauke: i mean mainly, it would feel like overcorrecting the opposite way to say that that badly-implemented slider *isn't* also relying on dragging to me - it would cause confusion back at the other end

patrickhlauke: and looking at the definition of dragging movement, this seems in line with that - it doesn't say that dragging is NOT path-based, it's not an either/or https://www.w3.org/TR/WCAG22/#dfn-dragging-movements

patrickhlauke: 2.5.1 fails things requiring path-based movement, 2.5.7 fails things that involve dragging. both of these can fail simultaneously.

patrickhlauke: this made me think/rethink that particular scenario, so added this here for now for discussion https://github.com/w3c/wcag/pull/4843/changes/c4588accf1c086f29396080c24f62f69ae11fc9e

patrickhlauke:
For those who have already looked at this previously, I've just made a change/addition relating to components (sliders) that involve *both* dragging *and* constrain the drag to be on a particular path.

I believe that based on the definition of [dragging movement](https://www.w3.org/TR/WCAG22/#dfn-dragging-movements) and the actual [wording for 2.5.7](https://www.w3.org/TR/WCAG22/#dragging-movements) this still makes sense, as otherwise it would feel strange saying that that type of slider (which will be very rare, as most of them do capture the pointer and don't constrain the dragging to be along the track/path) does *not* count as being "dragged"

https://github.com/w3c/wcag/pull/4843/changes/c4588accf1c086f29396080c24f62f69ae11fc9e

in short: all sliders involve dragging movement, but not all sliders are ALSO path-based

detlevhfischer:
> In the rare instances where a slider has been implemented in a way where straying from the track "loses" the user's grip on the slider, it is considered to be both a dragging movement (as the action does involve grabbing the slider thumb, moving it, and then releasing it) and a path-based gesture

I struggle to understand this - why will implementing constraints so that the track (or whatever other object that is being dragged) is lost then count also as path-based gesture?

To avoid mudding the waters, we may better consider to remove this section. I’m not sure whether it will really help anyone understanding the difference. The case may be so rare as to be irrelevant.

detlevhfischer:
Another thought: it may be better to change the first graphic for the swipe / flick pointer gesture. The red dots made perfect sense for the dragging movement illustration because here, you pick up an object and move it elsewhere, but the clarification the new text makes about the variation in swipe / flick gestures, i.e. that you don’t need a particular point to start from, creates a dissonance when looking at the graphic with these red points.

patrickhlauke:
> > In the rare instances where a slider has been implemented in a way where straying from the track "loses" the user's grip on the slider, it is considered to be both a dragging movement (as the action does involve grabbing the slider thumb, moving it, and then releasing it) and a path-based gesture
> 
> I struggle to understand this - why will implementing constraints so that the track (or whatever other object that is being dragged) is lost then count also as path-based gesture?

The constraint to a path makes it a path-based gesture.

If you're "grabbing" something and moving it, it's a drag. If it's then also constrained to a specific path, it's also path-based. I think that's clearer than trying to explain some "yes, it's a drag, but *only* when it's free-form, otherwise we don't call it a drag but just a path-based gesture" either/or logic which is where the original overlap/confusion started. They're two distinct aspects that can also happen at the same time.

The definition of dragging movement itself doesn't make that distinction to begin with ... it was always just in the understanding docs that an attempt was made to make these awkwardly either/or with some "when is a drag not a drag? when it's a drag that's constrained to a path" thing.

And the end result is still the same: unless there's a provision/alternative way to operate, it fails WCAG (we're just now arguing over "does it fail just one SC, or both").

> To avoid mudding the waters, we may better consider to remove this section. I’m not sure whether it will really help anyone understanding the difference. The case may be so rare as to be irrelevant.

By ignoring it we'll end up with the same inconsistency in interpretation we're having right here, though.

understanding/21/pointer-gestures.html
```diff
@@ -20,36 +20,42 @@ <h2>In brief</h2>
 
 		<section id="intent">
 			<h2>Intent of this Success Criterion</h2>
+
 			<p>The intent of this success criterion is to ensure that content can be controlled with a range of pointing devices, abilities, and assistive technologies. Some people  cannot perform gestures in a precise manner, or they may use a specialized or adapted input device such as a head pointer, eye-gaze system, or speech-controlled mouse emulator. Some pointing methods lack the capability or accuracy to perform multipoint or path-based gestures.</p>
 
 			<p>A <strong>path-based gesture</strong> involves an interaction where not just the endpoints matter, but how the pointer moves between these points.</p>
```
giacomo-petri:
```suggestion
			<p>A <strong>path-based gesture</strong> involves an interaction where the movement of the pointer along a particular path is significant.</p>
```

The current wording almost implies that the "endpoints" always matter, while in the content below it is clear that sometimes even the endpoints do not matter. To me this reads as a bit contradictory, even though the overall meaning is clear.
For this reason, IMO the first sentence should avoid starting with "not just the endpoints matter" and instead focus on the path being the significant aspect of the gesture. The explanation about startpoints/endpoints is provided in the following paragraphs and when distinguishing path-based from dragging gestures, so mentioning them here is IMO not necessary.

patrickhlauke: great point, love it

giacomo-petri (approved):

understanding/21/pointer-gestures.html
detlevhfischer: Hm... I still think this is over the top and unnecessary, and people will have a ard time understanding this. Not a big issue though. Let's see what other people say.

understanding/21/pointer-gestures.html
```diff
@@ -86,12 +96,11 @@ <h2>Benefits</h2>
...
-				<li><p>A kanban widget with several vertical areas representing states in a defined process allows the user to right- or left-swipe elements to move them to an adjacent silo. The user can also accomplish this by selecting the element with a single tap or click, and then activating an arrow button to move the selected element.</p></li>
-				<li><p>A custom slider requires movement in a strict left/right direction when operated by dragging the thumb control. Buttons on both sides of the slider increment and decrement the selected value and update the thumb position.</p></li>
-      		</ul>
+        <li><p>A news site has a horizontal content slider with hidden news teasers that can moved into the viewport via a fast horizontal swipe/flicking motion. It also offers forward and backward arrow buttons for single-point activation to navigate to adjacent slider content.</p></li>
+        <li><p>A news site has a horizontal content slider with hidden news teasers that can moved into the viewport via a fast horizontal swipe/flicking motion. Alternatively, a single pointer input at the outer vertical edges of the slider navigates to adjacent slider content.</p></li>
```
scottaohara: this example is really just a duplicate of the previous example, but calling out that the buttons on either end of the slider dont' need to be overtly styled as arrow buttons.  calling these "single pointer inputs" instead is more of a head scratcher of "wait what is this referring to" than maybe just revising the previous example to note that there's no one way to style previous/next buttons - heck, they could even be largely invisible hit areas so long as they become visible on focus, or some other keyboard operable equivalent is provided.

patrickhlauke: https://github.com/w3c/wcag/pull/4843/changes/2f6d0a6376aa5209dc82f3179bf4e5289bd739b6

understanding/21/pointer-gestures.html
```diff
@@ -23,7 +23,7 @@ <h2>Intent of this Success Criterion</h2>
 
 			<p>The intent of this success criterion is to ensure that content can be controlled with a range of pointing devices, abilities, and assistive technologies. Some people  cannot perform gestures in a precise manner, or they may use a specialized or adapted input device such as a head pointer, eye-gaze system, or speech-controlled mouse emulator. Some pointing methods lack the capability or accuracy to perform multipoint or path-based gestures.</p>
 
-			<p>A <strong>path-based gesture</strong> involves an interaction where not just the endpoints matter, but how the pointer moves between these points.</p>
+			<p>A <strong>path-based gesture</strong> involves an interaction where the movement of the pointer along a particular path is significant.</p>
```
fstrr: To me, this is less clear than the deleted version—in that version, the reader doesn't have to decide how to  define "significant".

giacomo-petri: significant = matters (in this context)

I believe the sentence should be read in the context of the lines that follow. When you read those lines, with the original version there is a contradiction, which IMO makes it even worse, as I stated in my [previous comment](https://github.com/w3c/wcag/pull/4843#discussion_r2923699722)

patrickhlauke: so if we changed `is significant` to `matters`, would that make it clearer @fstrr ? maybe even changing "particular" to "specific"
```suggestion
			<p>A <strong>path-based gesture</strong> involves an interaction where the movement of the pointer along a specific path matters.</p>
```

fstrr: Well then, let me just delete the comment I was halfway through writing where I was suggesting changing significant to matters :)

patrickhlauke:
dropping this here as an example of a slider that, based on the clarifications in this PR, would explicitly fail the requirements for BOTH pointer gestures (as it's path-based) AND dragging movement (as it involves dragging)

https://codepen.io/patrickhlauke/pen/yMpbXa

these sliders (that don't "capture" the pointer, meaning they bail once you stray out of the track) are very rare.

---

this is a more common slider https://codepen.io/patrickhlauke/pen/eZgqRo - this one is *not* constrained to the path, as the pointer is captured.

this would pass pointer gestures (as it's not path-based), but fail dragging movement

patrickhlauke:
some more tweaking around the edges - in particular, I realised that part of the confusion we often get around these is because people use the overloaded term "slider" to refer to things like carousels, rather than just things like range sliders (think `<input type="range">`). and the example provided in pointer gestures about the news site didn't help matters. so made a few more modifications that hopefully make things even clearer.

understanding/21/pointer-gestures.html
detlevhfischer: @scottaohara wrote

>  heck, they could even be largely invisible hit areas so long as they become visible on focus, or some other keyboard operable equivalent is provided.

Since in 2.5.1 we are discussing pointer alternatives and not keyboard alternatives, the question of invisible hit areas becoming visible on kb focus is not really an issue. It still feels a bit contentious to accept the presence of invisible targets for single pointer input as sufficient alternative (pointer operation alternative) - but this is what I am told developers often go for when wanting to avoid arrow buttons. 

Now reading @patrickhlauke's edit and see that you have included the invisible option. I agree but expect it will be contentious.

understanding/21/pointer-gestures.html
```diff
@@ -20,78 +20,90 @@ <h2>In brief</h2>
...
-				<figcaption>A <strong>free-form gesture</strong> does <strong>not</strong> require any particular path before the end-point, only the start and (optionally) the end point matter. This is <strong>not path-based</strong></figcaption>
-			</figure>
+			<p>Unless the slider is implemented in a way that forces users to precisely follow the path of the track, and straying from the track "loses" the user's grip on the slider thumb, operating a slider is considered a dragging movement, rather than a path-based gesture.</p>
+
+			<p>In the rare instances where a slider <em>has</em> been implemented in a way where straying from the track "loses" the user's grip on the slider thumb, it is considered to be <em>both</em> a dragging movement (as the action does involve grabbing the slider thumb, moving it, and then releasing it) <em>and</em> a path-based gesture. <span class="wcag22">In those cases, the slider may fail the requirements of <em>both</em> this criterion <em>and</em> <a href="dragging-movements">Success Criterion 2.5.7 Dragging Movements</a>.</span></p>
```
patrickhlauke:
```suggestion
			<p>In the rare instances where a slider <em>has</em> been implemented in a way where straying from the track "loses" the user's grip on the slider thumb, it is considered to be <em>both</em> a dragging movement (as the action does involve grabbing the slider thumb, moving it, and then releasing it) <em>and</em> a path-based gesture at the same time. <span class="wcag22">In those cases, the slider may fail the requirements of <em>both</em> this criterion <em>and</em> <a href="dragging-movements">Success Criterion 2.5.7 Dragging Movements</a>.</span></p>
```

understanding/21/pointer-gestures.html
```diff
@@ -20,78 +20,89 @@ <h2>In brief</h2>
...
-				<li><p>A kanban widget with several vertical areas representing states in a defined process allows the user to right- or left-swipe elements to move them to an adjacent silo. The user can also accomplish this by selecting the element with a single tap or click, and then activating an arrow button to move the selected element.</p></li>
-				<li><p>A custom slider requires movement in a strict left/right direction when operated by dragging the thumb control. Buttons on both sides of the slider increment and decrement the selected value and update the thumb position.</p></li>
-      		</ul>
+				<li>A website includes a map view that supports the pinch gesture to zoom into the map content. User interface controls offer the operation using plus and minus buttons to zoom in and out.</li>
+				<li>A website includes a map view that supports the pinch gesture to zoom into the map content. As an single-pointer alternative, the map also allows users to double-tap, hold, and then move the pointer up or down to zoom in or out.</li>
```
detlevhfischer: I still feel this example is not a good one for single pointer alternatives due to the complexity (double-tap and hold, then move up or down). Even if no directionality is required in the move-up-or-down part, this would then still fall under 2.5.7 where in turn single pointer alternatives for dragging movements are required. So it looks like this alternative for 2.5.1 is passing the buck to 2.5.7 where it would then fail unless a still different alternative is present? That seems not ideal, to put it mildly.

patrickhlauke: > this would then still fall under 2.5.7 where in turn single pointer alternatives for dragging movements are required.

just for clarity, 2.5.7 requires single-pointer alternatives *that don't rely on dragging movement* (adding this for the benefit of those who still misunderstand "single pointer" as also covering/being in contrast to dragging on its own)

The reason why I added this example is that this exact question has come up in various actual live assessments, where authors wondered if in the context of 2.5.1 this passes (because they were working to WCAG 2.1, where dragging movement is not present yet).

And to be clear, these are isolated examples of things that pass this particular SC. They're not necessarily endorsements/suggestions of what authors should or shouldn't do.

detlevhfischer: Agree that it would have passed before 2.2 was out - but now it sits very oddly as an example. If retained, it should at least say: "would pass 2.5.1 but fail 2.5.7". But the map view examples are no longer included anyway, if I am not mistaken?

patrickhlauke: > would pass 2.5.1 but fail 2.5.7

doesn't work as is, as the understanding is also used for WCAG 2.1, where 2.5.7 doesn't exist. unless it's wrapped and prefixed with "New in WCAG 2.2" which makes it awkward.

detlevhfischer: > doesn't work as is, as the understanding is also used for WCAG 2.1, where 2.5.7 doesn't exist

Right... So am I right that the example has been removed in this PR?

patrickhlauke: > Right... So am I right that the example has been removed in this PR?

I have no idea which "the example" you're now asking about. Look at https://deploy-preview-4843--wcag2.netlify.app/understanding/pointer-gestures and see for yourself?

detlevhfischer: Sorry for the confusion - I looked at the _Files changed_ tab and failed to notice that it includes the 2.5.7 Understanding doc (while the 2.5.1 doc I was after was collapsed). The example I am taking issue with is still in the PR -- is the one with the map view that lists as alternative "double-tap, hold, and then move the pointer up or down to zoom in or out". Even if this is a valid alternative for WCAG 2.1, i.e., before 2.5.7 was added, it is not a good idea to use it since it would now fail 2.5.7 (using a sequence including a dragging movement as an alternative to a complex pointer gesture, without its own alternative). So I still think this has got to go.

patrickhlauke:
```suggestion
				<li>A website includes a map view that supports the pinch gesture to zoom into the map content. As an single-pointer alternative, the map also allows users to double-tap, hold, and then move the pointer up or down to zoom in or out.<span class="wcag22"> However, note that as this alternative requires dragging, it will fail the requirements of <a href="dragging-movements">Success Criterion 2.5.7 Dragging Movements</a> without a non-dragging alternative.</span></li>
```

As I'm keen to have this, and you're keen not to because of the dragging, let's split the difference then...

detlevhfischer: Hmm... I'am no going to perish on this wardle, but I still think we should look for another less odd example than one that says, here is what you can do, eh, actually no, you can't do it. It doesn't make sense to me, really. Let's see what others think.

understanding/21/pointer-gestures.html
```diff
@@ -20,78 +20,89 @@ <h2>In brief</h2>
...
-				<li><p>A custom slider requires movement in a strict left/right direction when operated by dragging the thumb control. Buttons on both sides of the slider increment and decrement the selected value and update the thumb position.</p></li>
-      		</ul>
+				<li>A website includes a map view that supports the pinch gesture to zoom into the map content. User interface controls offer the operation using plus and minus buttons to zoom in and out.</li>
+				<li>A website includes a map view that supports the pinch gesture to zoom into the map content. As an single-pointer alternative, the map also allows users to double-tap, hold, and then move the pointer up or down to zoom in or out.<span class="wcag22"> However, note that as this alternative requires dragging, it will fail the requirements of <a href="dragging-movements">Success Criterion 2.5.7 Dragging Movements</a> without a non-dragging alternative.</span></li>
```
detlevhfischer: Maybe add a split-tap example instead?
See also my other comment.
As I don't see it now, I sum up again: I think giving an example for an alternative that then fails another SC, 2.5.7 Dragging Movements, unless it has an alternative (the alternative for the alternative?) is odd and not really helpful.

detlevhfischer (dismissed):

understanding/21/pointer-gestures.html
patrickhlauke: but that's the thing, these aren't examples of "here's what you can do", necessarily. it gives examples of applicability/when something passes or fails the SC

patrickhlauke:
Changes based on feedback so far (including some points raised separately by @graemecoleman 

* remove the contentious double-tap and drag example for maps - not worth the fight over it
* some reorganisation in both understanding documents to section out relationship with keyboard more explicitly
* removing some redundant/repeated sentences in pointer gestures that keep saying the same thing

understanding/21/pointer-gestures.html
```diff
@@ -20,78 +20,90 @@ <h2>In brief</h2>
...
-				<li><strong>Using a touch-screen:</strong> If the user puts their finger on the thumb control of the slider and moves upwards more than sideways, the slider may not respond because the browser takes control of the swipe and interprets it as a scroll, and will move the page up and down. Moving left or right on the slider thumb engages the slider and then the user can vary their vertical movement. This implementation has the 3-point requirement to work with a finger on a touch screen device so is a path-based gesture.</li>
-			</ul>
-			<p>As touch screen devices can apply default gestures it is important to test with them if you are unsure whether a particular component does require a path-based gesture.</p>
+			<p>Single pointer operations include taps and clicks, double-taps and double-clicks, long presses, swiping, dragging, and path-based gestures. Gestures such as "pinch to zoom" or two-finger swipes are <em>multipoint</em> gestures, as they require two or more pointer inputs - in this case, two fingers on a touchscreen.</p>
```
detlevhfischer: To me, rereading this, it is not quite clear why this paragraph sits here. It sounds a bit like a definition of single pointer [which we already have in the glossary](https://www.w3.org/TR/WCAG22/#dfn-single-pointer). What does it prove, or lead up to?

understanding/21/pointer-gestures.html
```diff
@@ -20,78 +20,90 @@ <h2>In brief</h2>
...
 
-			<p>Swiping is an example of a <strong>path-based gesture</strong>, which is only recognized when the user moves in a (mostly) straight line from the start point to the end point.</p>
+			<p>A <strong>path-based gesture</strong> involves an interaction where the movement of the pointer along a specific path matters.</p>
 
-			<figure id="figure-path-based-gesture-1">
-				<img src="img/path-based-gesture-1.png" width="400" alt="Hand showing a starting touch, 1. Moving in a straight line to a second point, 2.">
+			<p>Examples of a <strong>path-based gesture</strong> include swiping/"flicking" (which is only recognized when the user moves in a mostly straight line from the start-point to the end-point), having to move a pointer strictly along a specific path, and other gestures which trace a prescribed path such as drawing a specific shape. Such paths may be drawn with a finger or stylus on a touchscreen, graphics tablet, or trackpad, or with a mouse, joystick, or similar pointer device.</p>
```
detlevhfischer: It now appears a bit as if swiping / flicking as opposed to dragging is the only thing that is covered by 2.5.1, while other gestures like split-tap or pinch are just mentioned. However , the changes proposed are good since they disambiguate 2.5.1 and 2.5.7, and more can be added about non-single-pointer gestures in a later and separate PR.

understanding/21/pointer-gestures.html
```diff
@@ -20,78 +20,90 @@ <h2>In brief</h2>
...
+				For instance, if a page implemented a horizontal "flick" gesture to move to the previous or next page, the exact point where the user starts the "flick" or ends it is irrelevant, as long as the "flick" gesture is horizontal, is of sufficient length, and is performed above a specific speed.</p>
+
+      <h3>The difference between Pointer Gestures and Dragging</h3>
+
+			<p>In a path-based gesture, what matters is how the pointer moves: its path or direction (and sometimes its speed). For some gestures, even the exact start- and end-points may not matter.</p>
+
+			<figure id="figure-path-based-gesture">
+				<img src="img/2.5.1-path-based-gesture.png" width="400" alt="Illustration of a stylised finger on a hand touching an initial point 1, then moving in a straight horizontal line to a second point 2.">
 				<figcaption>A <strong>path-based gesture</strong> where pointer movement is only allowed in a straight line from the start-point to the end-point. If the user strays from the straight directional path, the gesture is not recognized, has no effect, or is aborted.</figcaption>
```
detlevhfischer: _is only effective_? or, _only has an effect when_? (just to get rid of "allowed", which sounds a bit odd...)

understanding/21/pointer-gestures.html
```diff
@@ -20,78 +20,90 @@ <h2>In brief</h2>
...
-				<figcaption>A <strong>free-form gesture</strong> does <strong>not</strong> require any particular path before the end-point, only the start and (optionally) the end point matter. This is <strong>not path-based</strong></figcaption>
+			<p>Unless the slider is implemented in a way that forces users to precisely follow the path of the track, and straying from the track "loses" the user's grip on the slider thumb, operating a slider is <em>only</em> considered a dragging movement, rather than a path-based gesture.</p>
+
+			<p>In the rare instances where a slider <em>has</em> been implemented in a way where straying from the track "loses" the user's grip on the slider thumb, it is considered to be <em>both</em> a dragging movement (as the action does involve grabbing the slider thumb, moving it, and then releasing it) <em>and</em> a path-based gesture at the same time. <span class="wcag22">In those cases, the slider may fail the requirements of <em>both</em> this criterion <em>and</em> <a href="dragging-movements">Success Criterion 2.5.7 Dragging Movements</a>.</span></p>
```
detlevhfischer, line 68: To me, this still feels somewhat unnecessary due to the apparent lack of such things in the field, and, unavoidably, hard to understand and likely to puzzle the reader; but I do not object to it if others think this edge case must be called out.

understanding/21/pointer-gestures.html
```diff
@@ -20,78 +20,90 @@ <h2>In brief</h2>
...
 
-			<p>This success criterion applies to gestures in the author-provided content, not gestures defined by the operating system, user agent, or assistive technology. Examples of operating system gestures would be swiping down to see system notifications and gestures for built-in assistive technologies (AT). Examples of user agent-implemented gestures would be horizontal swiping implemented by browsers for navigating within the page history, or vertical swiping to scroll page content.</p>
+			<p>This success criterion applies to gestures in the author-provided content, not gestures defined by the operating system, user agent, or assistive technology. Examples of operating system gestures would be swiping down to see system notifications and gestures for built-in assistive technologies. Examples of user agent-implemented gestures would be horizontal swiping implemented by browsers for navigating within the page history, or vertical swiping to scroll page content.</p>
```
detlevhfischer: Is it worthwhile including mention of content sliders with CSS overflow:auto  that (I think) we consider to be operated by UA gestures and therefore exempt? If so, we would then need a differentiation between those content sliders where the author is involved (e.g. defining lock-in positions) which would require single pointer alternatives, and those where the silder content is merely moved by the UA without author intervention so they are exempt. My hunch is that evaluators will frequently 'din' content in the latter type of sliders we would consider UA-controlled and therefore exempted, so this may be a valuable addition.

understanding/21/pointer-gestures.html
```diff
@@ -20,78 +20,90 @@ <h2>In brief</h2>
...
 		</section>
+
 		<section id="benefits">
 			<h2>Benefits</h2>
 			<ul>
-				<li><p>Users who cannot (accurately) perform path-based pointer gestures - on a touchscreen, or with a mouse - will have alternative means for operating the content.</p></li>
-				<li><p>Users who cannot perform multi-pointer gestures on a touchscreen (for instance, because they are operating the touchscreen with an alternative input such as a head pointer) will have a single-pointer alternative for operating the content.</p></li>
-				<li><p>Users who may not understand the custom gesture interaction intended by the author will be able to rely on simple, frequently used gestures to interact. This can be especially beneficial for users with cognitive or learning disabilities.</p></li>
+				<li>Users who cannot (accurately) perform path-based pointer gestures - on a touchscreen, or with a mouse - will have alternative means for operating the content.</li>
+				<li>Users who cannot perform multi-pointer gestures on a touchscreen (for instance, because they are operating the touchscreen with an alternative input such as a head pointer) will have a single-pointer alternative for operating the content.</li>
```
detlevhfischer, line 87: Is a head pointer constrained to touch screens only? Would it not also move a cursor on a non-touch screen?

understanding/21/pointer-gestures.html
```diff
@@ -20,78 +20,90 @@ <h2>In brief</h2>
...
+				<li>Users who may not understand the custom gesture interaction intended by the author will be able to rely on simple, frequently used gestures to interact. This can be especially beneficial for users with cognitive or learning disabilities.</li>
```
detlevhfischer: just wondering if people will understand that "simple, frequently used gestures" here means clicking or tapping. I think many will think it is anything going beyond simple clicks or taps.

understanding/21/pointer-gestures.html
```diff
@@ -20,78 +20,90 @@ <h2>In brief</h2>
...
-        		<li><p>A news site has a horizontal content slider with hidden news teasers that can moved into the viewport via a fast horizontal swipe/flicking motion. It also offers forward and backward arrow buttons for single-point activation to navigate to adjacent slider content.</p></li>
-				<li><p>A kanban widget with several vertical areas representing states in a defined process allows the user to right- or left-swipe elements to move them to an adjacent silo. The user can also accomplish this by selecting the element with a single tap or click, and then activating an arrow button to move the selected element.</p></li>
-				<li><p>A custom slider requires movement in a strict left/right direction when operated by dragging the thumb control. Buttons on both sides of the slider increment and decrement the selected value and update the thumb position.</p></li>
-      		</ul>
+				<li>A website includes a map view that supports the pinch gesture to zoom into the map content. As a single-pointer alternative, the map also includes plus/minus buttons to zoom in and out.</li>
```
detlevhfischer: possibly add "and reverse pinch / spread gesture" so it maps onto plus/minus

understanding/21/pointer-gestures.html
```diff
@@ -20,78 +20,90 @@ <h2>In brief</h2>
...
+        <li>A news site has a horizontal carousel with different news teasers that can moved into view with a fast horizontal swiping/flicking motion. It also offers forward and backward buttons on the left and right edges of the carousel, to move the carousel to the previous and next item with a simple click/tap. These buttons can be visible (for instance, as large arrow icons) or visually hidden (but still operable with a pointer).</li>
```
detlevhfischer: **If** we add an example of exempt content sliders (see above) this  example might warrant a modification like "...that can moved into view with a fast horizontal swiping/flicking motion **and land (or lock in to) at author-defined positions**", or similar.

understanding/21/pointer-gestures.html
patrickhlauke: Yes, we can add further stuff about multi-point gestures separately. This PR intended to just disambiguate path-based vs dragging

understanding/21/pointer-gestures.html
patrickhlauke: Maybe a matter of taste/opinion, but allowed in this context looks fine to me?

understanding/21/pointer-gestures.html
patrickhlauke: overflow stuff is mentioned in dragging https://deploy-preview-4843--wcag2.netlify.app/understanding/dragging-movements.

this actually makes me realise this bit is slightly inaccurate, as you don't "swipe" to scroll a page, you drag it (unless we want to say it's about "flicking" - with a quick swipe - the page)

then can include a mention of scrollable CSS containers

understanding/21/pointer-gestures.html
patrickhlauke: sure, but the context here is performing multi-pointer gestures, which you do on a touch screen. also, this is just a non-exhaustive (existing) example.

understanding/21/pointer-gestures.html
patrickhlauke:
```suggestion
				<li>A website includes a map view that supports the pinch/spread gesture to zoom into the map content. As a single-pointer alternative, the map also includes plus/minus buttons to zoom in and out.</li>
```

understanding/21/pointer-gestures.html
patrickhlauke: this risks making the example a bit less clear (also, note that scroll-snapping would have that same effect even when it's a CSS overflow scroller, so that's not a differentiator)

patrickhlauke: the difference would be more whether or not it's done via scripting on the author's side...
```suggestion
        <li>A news site has a horizontal carousel with different news item teasers. The site implements custom code that detects fast horizontal swiping/flicking motion to move to the previous / next items. It also offers forward and backward buttons on the left and right edges of the carousel, to move the carousel to the previous and next item with a simple click/tap. These buttons can be visible (for instance, as large arrow icons) or visually hidden (but still operable with a pointer).</li>
```

understanding/21/pointer-gestures.html
patrickhlauke:
```suggestion
				<li>Users who may not understand the custom gesture interaction intended by the author will be able to rely on simple, frequently used gestures, such as basic clicking or tapping, to interact. This can be especially beneficial for users with cognitive or learning disabilities.</li>
```

understanding/21/pointer-gestures.html
patrickhlauke:
```suggestion
			<p>This success criterion applies to gestures in the author-provided content, not gestures defined by the operating system, user agent, or assistive technology. Examples of operating system gestures on touch screen devices would be swiping down from the top of the screen to see system notifications, and gestures to operate built-in assistive technologies. Examples of user agent-implemented gestures on touch screen devices would be horizontal flicking motions implemented by browsers for navigating within the page history, or vertical flicking gestures to "fling" page content (or content inside natively scrollable containers) to scroll it.</p>
```

understanding/21/pointer-gestures.html
patrickhlauke: https://github.com/w3c/wcag/issues/5036

understanding/21/pointer-gestures.html
patrickhlauke: reviewing this as a whole (rather than trying to make small adjustments at sentence level), i'm struck by the fact that the entire "Challenges for people with disabilities" section doesn't actually say anything new (now that the inaccurate part about touchscreens is gone), and retreads points that really should be covered in the intent. having a more fundamental look at this now, as a last push...

patrickhlauke:
Prompted by https://github.com/w3c/wcag/pull/4843#discussion_r3026719361 made one last (hopefully) big change:

> Remove the "Challenges for people with disabilities" section - after removing the inaccurate section about the difference between touchscreen and mouse operation for range sliders, the section only contains general points that are far better made in the intent section

https://github.com/w3c/wcag/pull/4843/commits/a53c434d8685945c531f2bae42abe8bbd77bb32d

understanding/21/pointer-gestures.html
patrickhlauke: https://github.com/w3c/wcag/pull/4843/commits/a53c434d8685945c531f2bae42abe8bbd77bb32d

understanding/21/pointer-gestures.html
```diff
@@ -20,78 +20,80 @@ <h2>In brief</h2>
...
+
+      <h3>The difference between path-based Pointer Gestures and Dragging</h3>
+
+			<p>In a path-based gesture, what matters is how the pointer moves: its path or direction (and sometimes its speed). For some gestures, even the exact start- and end-points may not matter.</p>
 
-			<p>Examples of path-based gestures include swiping, sliders and carousels dependent on the direction of interaction, and other gestures which trace a prescribed path such as drawing a specific shape. Such paths may be drawn with a finger or stylus on a touchscreen, graphics tablet, or trackpad, or with a mouse, joystick, or similar pointer device.</p>
+			<figure id="figure-path-based-gesture">
+				<img src="img/2.5.1-path-based-gesture.png" width="400" alt="Illustration of a stylised finger on a hand touching an initial point 1, then moving in a straight horizontal line to a second point 2.">
+				<figcaption>A <strong>path-based gesture</strong> where pointer movement is only allowed in a straight line from the start-point to the end-point. If the user strays from the straight directional path, the gesture is not recognized, has no effect, or is aborted.</figcaption>
```
patrickhlauke:
```suggestion
				<figcaption>A <strong>path-based gesture</strong> where pointer movement is only recognized when it occurs in a straight line from the start-point to the end-point. If the user strays from the straight directional path, the gesture is not recognized, has no effect, or is cancelled.</figcaption>
```

understanding/22/dragging-movements.html
```diff
@@ -86,8 +83,12 @@ <h2>Examples of Dragging Movements with alternatives</h2>
...
          <li>A task board that allows users to drag and drop items between columns also provides an additional pop-up menu after tapping or clicking on items for moving the selected element to another column by tapping or clicking on pop-up menu entries.</li>
          <li>A radial control widget (color wheel) where the value can be set by dragging the marker for the currently selected color to another position, also allows picking another color value by tapping or clicking on another place in the color wheel.</li>
-         <li>A linear slider control widget, where the value can be set by dragging the visual indicator (thumb) showing the current value, allows tapping or clicking on any point of the slider track to change the value and set the thumb to that position. </li>  
+         <li>A linear range slider control widget, where the value can be set by dragging the visual indicator (thumb) showing the current value, allows tapping or clicking on any point of the slider track to change the value and set the thumb to that position. </li>
```
patrickhlauke:
```suggestion
         <li>A range slider control widget, where the value can be set by dragging the visual indicator (thumb) showing the current value, allows tapping or clicking on any point of the slider track to change the value and set the thumb to that position. </li>  
```

bruce-usab:
@detlevhfischer we did some editing on the TF call 4/3.  Please take another look, to see if you are comfortable with the additional changes.  (And please 👍 or remove your 👎 at top, commenting as you like.)  We had consensus with folks on the call for keeping the “in some rare instances” example.

PR #4845, opened 2026-01-02
Consistency: "user agent" rather than "user-agent"

patrickhlauke:
There's only a few straggler uses of "user-agent" (hyphenated). The preferred usage is "user agent" (no hyphen).


bruce-usab:
Reviewed on TF call 1/16 call and moved to Ready for approval.

PR #4846, opened 2026-01-02
Build: Support entire content of technique h1 being wrapped

kfranqueiro:
This fixes the issue reported in #4831 while avoiding regressions to existing link text truncation logic.

(This does cause some whitespace changes in output, but they are not noticeable in the rendered HTML.)


adampage:
Thanks, @kfranqueiro.

iadawn (approved):

PR #4847, opened 2026-01-03
Chore: code cleanup

patrickhlauke:
* Remove the now unnecessary `xml:space="preserve"` for `<pre>` blocks (since we don't use XSLT anymore for publication)
* Remove the odd straggler instances of `xmlns:wcag` (presumably used at some point, again in the old XSLT days)

### DIFF
--- a/11ty/techniques.ts
+++ b/11ty/techniques.ts
@@ -399,7 +399,7 @@ export async function getTechniquesByTechnology(guidelines: FlatGuidelinesMap) {
       technology,
       title,
       titleHtml,
-      truncatedTitle: title.replace(/\s*\n[\s\S]*\n\s*/, " … "),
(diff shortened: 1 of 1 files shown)


kfranqueiro:
I removed more instances of `xml:space="preserve"`. a few were likely missed due to a preceding newline, but I'm not sure how the others were missed.

patrickhlauke:
thanks @kfranqueiro ... very odd. I think perhaps I `git commit`ted too quickly after doing the find/replace in vs code, and it hadn't yet written all the files to the filesystem at that exact moment... was clearly too keen.

bruce-usab:
Discussed and approved on TF call 9 Jan 2026.  Per process, PR should be mentioned in next email to AG but, as a bug fix, expected to be merged sooner than that.

PR #4850, opened 2026-01-05
Update footer design to match the WAI website's

remibetin:
This applies the same footer design as on the WAI website.

Related to https://github.com/w3c/wai-website-theme/pull/163.

### DIFF
--- a/xslt/base.xslt
+++ b/xslt/base.xslt
@@ -360,54 +360,35 @@
 	</xsl:template>
 
 	<xsl:template name="site-footer">
-    <footer class="w3c-global-footer wai-global-footer default-grid" aria-label="Site">
-      <div class="inner">
-        <div class="wai-global-footer__content">
-          <div class="wai-global-footer__subsite"><a href="https://www.w3.org/WAI/" lang="en" dir="auto" translate="no">W3C Web Accessibility Initiative (WAI)</a></div>
-          <div class="wai-global-footer__subsite-desc">Strategies, standards, resources to make the Web accessible to people with disabilities</div>
-          <div class="wai-global-footer__cta w3c-l-cluster w3c-l-cluster--vertical-align">
-            <ul class="w3c-clean-list" role="list">
-              <li class="w3c-with-icon--before w3c-with-icon--larger">
-                <img class="w3c-icon w3c-icon--larger" src="https://www.w3.org/WAI/assets/images/email.svg" width="30" height="30" alt="" aria-hidden="true">
-                <a href="https://www.w3.org/WAI/news/subscribe/">Get News in Email</a>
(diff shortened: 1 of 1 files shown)


iadawn:
This looks good to me. If Ken is happy I can do an understanding and techniques update this Thurs.

xslt/base.xslt
```diff
@@ -360,35 +360,54 @@
...
-		</footer>
+    <footer class="w3c-global-footer wai-global-footer default-grid" aria-label="Site">
+      <div class="inner">
+        <div class="wai-global-footer__content">
+          <div class="wai-global-footer__subsite"><a href="https://www.w3.org/WAI/" lang="en" dir="auto" translate="no">W3C Web Accessibility Initiative (WAI)</a></div>
+          <div class="wai-global-footer__subsite-desc">Strategies, standards, resources to make the Web accessible to people with disabilities</div>
+          <div class="wai-global-footer__cta w3c-l-cluster w3c-l-cluster--vertical-align">
+            <ul class="w3c-clean-list" role="list">
+              <li class="w3c-with-icon--before w3c-with-icon--larger">
+                <img class="w3c-icon w3c-icon--larger" src="https://www.w3.org/WAI/assets/images/email.svg" width="30" height="30" alt="" aria-hidden="true">
```
kfranqueiro: I'm not sure we should even worry about what's in `base.xslt`, as it has probably already missed a few prior updates (it hasn't been updated since I started working here). Moreover, it looks like the code pasted here would not be valid XML, such as this non-self-closing `img` tag, which is why I suggest we leave it out.

I had previously gone through and removed parts of the XSLT that we definitely weren't using anymore; maybe I should remove more of it... cc @iadawn

iadawn: Yes, we can ignore base.xslt... or indeed any XSLT ... please, please? :)

remibetin: > it looks like the code pasted here would not be valid XML, such as this non-self-closing img tag, which is why I suggest we leave it out.

Indeed, thanks for flagging!

> I'm not sure we should even worry about what's in base.xslt, as it has probably already missed a few prior updates 

I guess the only question I have is: if we don't update this file, will some pages still display the old footer?

kfranqueiro: Nope, we have relied exclusively on the Eleventy build for roughly 18 months at this point. I partially cleaned out parts of the XSLT that I knew for sure we would never use again but had left any build tasks I'd had doubts about; I'm suspecting we don't have a use for any of what's remaining at this point, either.

remibetin: Thanks @kfranqueiro I've reverted the changes made to base.xslt at this point.

PR #4853, opened 2026-01-12
Update header design

remibetin:
This PR:
- updates the header design to match the [updated WAI website theme](https://github.com/w3c/wai-website-theme/pull/175)
- removes in `css/base.css` the blocks that could come in conflict with the theme stylesheets.

Discussed with @kfranqueiro:
- [x] There is no cache busting at the moment. This seems needed whenever we update the theme.

To do before merging:
- [x] Point to the `style.css` stylesheet in production (the current path is used for the preview only): `https://www.w3.org/WAI/assets/css/style.css`

⚠️ Important: This update needs to be synced with the release of the theme update.

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -11,9 +11,9 @@ isUnderstanding and isTechniques (each set to true in respective folder)
         <a href="{{ headerUrl }}">WCAG {{ versionDecimal }} {{ headerLabel }}</a>
       </div>
       {% if isTechniques %}
-      <div class="minimal-header-subtitle">Examples of ways to meet WCAG; not required</div>
+      <p class="minimal-header-subtitle">Examples of ways to meet WCAG; not required</p>
       {% elsif isUnderstanding %}
-      <div class="minimal-header-subtitle">Informative explanations, not required to meet WCAG</div>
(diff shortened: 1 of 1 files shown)


kfranqueiro:
RE cache-busting: I've implemented cache-busting filename fragments for CSS files output by this repo's Eleventy build system in #4856. The CSS files produced by this repo's build seem to get an Expires and max-age of 1 week in HTTP headers, so cache-busting is certainly valuable on this end.

Meanwhile, WAI's style.css seems to get a max-age of 10 minutes, yet its Expires header currently seems to give me a timestamp in the past? So I'm not sure we need to worry as much about cache-busting style.css, and I'm not sure if using a GET parameter might mess with anything in the chain server-side (e.g. potentially causing no caching at all).

Note that the (temporary?) changes to stylesheet URLs in this PR may come up as conflicting with #4856 when it merges, due to the proximity of the two style tags. Hopefully that will go away if the style tag changes in this PR are reverted, otherwise you might need to deal with conflict resolution even though the changes in these PRs are not on the same lines, only neighboring ones.

remibetin:
Thanks @kfranqueiro I'll make sure to update the stylesheet URLs.

PR #4856, opened 2026-01-13
Implement cache-busted CSS filenames

kfranqueiro:
This PR revises copying logic in the Eleventy config to append cache-bust suffixes to CSS filenames that are copied into the `techniques` and `understanding` output folders, and updates templates referencing these resources to match.

This PR also removes `techniques.css` and `understanding.css`, since as far as I could tell, they're never imported and the classnames they reference aren't used. I guess that's something I failed to pick up on when first migrating the build system to Eleventy...

Ultimately, this means this only applies to `base.css` and `a11y-light.css`.

This is intended to help support #4853 which involves synchronizing style updates between this repo and the rest of the WAI website, as otherwise these CSS assets seem to have a max-age of a week. (cc @remibetin)

### DIFF
--- a/_includes/techniques/head.html
+++ b/_includes/techniques/head.html
@@ -1,1 +1,1 @@
-<link rel="stylesheet" href="../base.css">
+<link rel="stylesheet" href="../base.{{ cacheBustValue }}.css">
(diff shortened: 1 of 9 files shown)


kfranqueiro:
I spoke with @iadawn and we want to check with the Systems Team first to see if there is a more desirable approach before resorting to publishing new asset filenames on every monthly push to the WAI website.

kfranqueiro:
Closing this, as we've instead opted to temporarily reduce cache expiration headers ahead of merging #4853 rather than adding complexity to the build system and website deployment with filename revving.

PR #4863, opened 2026-01-14
Nullify WAI blockquote styles

### LINKED ISSUE #4134: Problems with blockquotes' pseudo-element styles
Blockquotes in the Understanding docs include styling which adds an open quote on `blockquote p:first-of-type::before`, and a close quote on `blockquote p:last-of-type::after`. Unfortunately, this causes problems in many situations that involve multiple paragraphs, which often involves nesting.

- [Understanding Conformance Requirements](https://www.w3.org/WAI/WCAG22/Understanding/conformance#conformance-requirements) is the most prominent example
- This also affects [the Intent section of Parsing](https://www.w3.org/WAI/WCAG22/Understanding/parsing#intent)
- Meanwhile, there are blockquotes containing lists in [the Intent section of Focus Appearance](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance#intent) which do not get these open/close quotes, which may or may not be intentional.

Example captured by @mbgower: 
image

kfranqueiro:
Fixes #4134.

This technically stems from styles in the WAI-site-wide stylesheet, but I suspect it is written that way for reasons (rather than e.g. only applying to only-child paragraphs inside blockquotes) and will be less lift in terms of reviewing to override it on this end instead.

@netlify /understanding/conformance#conformance-requirements


patrickhlauke:
Compare with https://www.w3.org/WAI/WCAG22/Understanding/conformance#conformance-requirements - note the double-quotes on each line of the blockquote, which this PR suppresses

patrickhlauke (approved):

bruce-usab:
Reviewed on TF call 1/16 call and moved to Ready for approval.

PR #4876, opened 2026-01-16
Tweak Understanding Headings and Labels 2.4.6

bruce-usab:
Editorial suggestion for Understanding for Headings and Labels, removing the use of <q>requires</q> in a couple of places.

A [preview diff](https://services.w3.org/htmldiff?doc1=https://www.w3.org/WAI/WCAG22/Understanding/headings-and-labels&doc2=https://deploy-preview-4876--wcag2.netlify.app/understanding/headings-and-labels) is available.

### DIFF
--- a/guidelines/act-mapping.json
+++ b/guidelines/act-mapping.json
@@ -1773,6 +1773,57 @@
       }
     },
     {
+      "title": "ARIA global properties not used where prohibited",
+      "permalink": "/standards-guidelines/act/rules/kb1m8s/proposed/",
+      "successCriteria": [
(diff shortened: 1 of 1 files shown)


bruce-usab:
I was prompted to propose change by “does not _require_ headings or labels” when meeting WCAG _does_ (generally) require (good) use of headings and labels.  I will also confess to being triggered by the semicolon.

patrickhlauke:
the "does not require" wording was added (by me at the time) because far too often i was seeing auditors failing pages under 2.4.6 for not having headings or labels where they felt there should be some

PR #3985, opened 2024-07-22
Remove superseded XSLT build targets and update README

kfranqueiro:
**Note: This is intended to land immediately _after_ #3975.** I'm sending it up-front so it can get reviewed, since it doesn't require code from the other branch.

WAI site pushes and gh-pages deploys no longer depend on the XSLT build process, and #3975 includes code that will work _only_ with the Eleventy-based build system. To avoid any accidents, this PR removes the parts of the XSLT process that are no longer needed, i.e. anything exclusively related to generating techniques and understanding pages. (Anything related to JSON generation has been left intact, until such time as that is tackled under the new process as well.)

This PR also updates the main README to include a reference to the new build system README, remove references to XML and XSLT, update documentation around term definitions, and add documentation about marking obsolete techniques (which is what #3975 adds).

Additional README fixes:

- Typos
- Notation that was being swallowed during markdown parsing (e.g. `<version>`)
- Default branch references (updated from master to main)

PR #4885, opened 2026-01-23
Remove all remaining unused XML/XSLT

kfranqueiro:
This is a spiritual successor to #3985. It removes:

- The remainder of the XSLT build system, which has been unused since #3917 was merged 18 months ago
- The contents of `lib`, which were only necessary for running the XSLT build
- The entirety of the `wcag20` folder, which has not been built in any way for an even longer time than that, and has typically only tripped up folks in the WCAG 2.x Backlog TF
- A reference to one `wcag20` file in `11ty/biblio.ts`, which has been replaced with a map directly containing the output that was previously computed at build time
- References to these files in the README, `.gitignore`, and the `check-files` workflow
  - Note this means that anyone who has ever run the XSLT build system locally (probably a low-single-digit number of people) will need to remove some old build artifacts that will show up in `git status` after pulling once this is merged

This branch occupies roughly 90MB less space than `main` when checked out, largely due to removal of `lib` and Flash and Silverlight examples under `wcag20/Techniques/working-examples`. (These will still exist in git history.)

This PR produces no changes in the Eleventy build output.

### DIFF
--- a/.github/workflows/cspell.yml
+++ b/.github/workflows/cspell.yml
@@ -10,3 +10,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: streetsidesoftware/cspell-action@v7
+        with:
+          # incremental_files_only would scan every file in the PR,
+          # even those that would typically be ignored
+          incremental_files_only: false


patrickhlauke:
this is magical! thanks for diving in on this long-standing bugbear of mine...

iadawn (approved):

patrickhlauke:
So stoked to see this merged!

Issue #2462, opened 2022-05-23
"Note" CSS / padding for Understandings not same as Main document, 0px Vs. 5px


jake-abma:
Note CSS / padding for Understandings not same as Main document

Main doc: https://w3c.github.io/wcag/guidelines/22/#focus-appearance-minimum
Understanding: https://w3c.github.io/wcag/understanding/focus-appearance-minimum.html#benefits

When looking at the "Notes" the Understanding pages do not have a top padding of 5 CSS px

Main doc:
```
<div class="note"...>

.issue, .note, .example, .advisement, blockquote, .amendment, .correction, .addition {
    padding: 0.5em;
    ...
}
```

Understanding:

```
<div class="note">

div.constraint, div.issue, div.note, div.notice, div.example {
    ...
    padding-top: 0;
}
```

patrickhlauke:
I would actually love it if the notes in understanding documents matched more closely the way notes look in the main spec. It would make them stand out more, and provide some visual consistency with the spec itself.

It would also avoid visually confusing things like in https://www.w3.org/WAI/WCAG22/Understanding/images-of-text where a note follows a definition list, and the reliance purely on margin makes it difficult to work out if the note only relates to the last definition.

> ![screenshot showing the top of the images of text understanding, with the 'success criterion' boxout; in the boxout, the 'Note' is distinguished only by indentation/left margin, but the preceding definition list also uses indentation/left margin, making it visually confusing](https://github.com/user-attachments/assets/75cb73fc-0dba-44e2-b793-d417038f217e)

As a test, transplanting the styles from the spec to the understanding would give us something like this:

> ![the same boxout, but the note now uses the same green background/light green border as in the spec, making it much clearer visually](https://github.com/user-attachments/assets/62ef29bf-2258-4000-a27f-2f5f34c54109)

The same would then also apply to notes in the understanding prose elsewhere... like further down on that same understanding page, this note

> ![screenshot of a note later on in the understanding document, relying only on left margin/indent](https://github.com/user-attachments/assets/83daddaf-a127-4c6f-a83c-3fbd3e86088a)

would then look like this

> ![the same note as before, now with the green boxout styling](https://github.com/user-attachments/assets/e650e8a0-895c-4ca5-acd8-3aad95750c33)

patrickhlauke:
Happy to make a prototype PR to introduce these note styles to the understanding stylesheet, if we think that'd be useful. I for one would find it much clearer, visually, and much more visually consistent with the spec.

[...]

PR #4890, opened 2026-01-26
Transplant note styling from spec base.css to understanding base.css

patrickhlauke:
This transplants the `.note` styles from the main spec to the understanding docs.

EDIT: for an example, see the latest proposed changes https://github.com/w3c/wcag/pull/4890#issuecomment-3877976451

Closes https://github.com/w3c/wcag/issues/2462


scottaohara:
oh, very much yes please.  thanks!

iadawn:
Can this be confined to notes within the SC box? If so, 👍. If not, I might need to think a bit more. In the broader body text, the notes then grab a lot of attention.

patrickhlauke:
> In the broader body text, the notes then grab a lot of attention.

I'm actually a fan of the notes standing out a bit more, but perhaps the spec styles are just a bit heavy-handed. could conceivably tone them down a little bit (away from that eye-searing acid green)

As an example of how this change then manifests itself also in the body ... current:

> a note in the current understanding doc - only denoted by a subtle indentation

With the PR applied:

> the same example, but with the green boxout style note

It break up the "wall of text" look of some of the long understanding docs with lots of notes. Makes it much more digestible, to me anyway.

iadawn:
OK, Shawn and I had a chat about this and we are of a mind to tone down the styling for notes outside the SC block. Sure, good to break up the wall of text but using notes to do this isn't ideal. Notes should be more asides and the big green styling makes them stand out way and shout for attention way to much.

patrickhlauke:
> OK, Shawn and I had a chat about this and we are of a mind to tone down the styling for notes outside the SC block

sure, i can iterate on this some more - copy the exact style inside the SC block, tone it down for the rest. putting this back in "In progress"

patrickhlauke:
did some tweaking, and landed on two separate styles: the "acid green" note style that matches what's in the normative spec, but it's only used inside the actual success criterion box. otherwise, it uses a much more sedate (but still distinctive) light blue/gray styling

> screenshot showing a zoomed out view of an understanding page, with the green note boxes visible in the success criterion box, and a more sedate light blue/gray note later on in the prose

https://deploy-preview-4890--wcag2.netlify.app/understanding/focus-appearance

thoughts @iadawn et al?

iadawn:
I like... the in-page note style doesn't grab visual attention as much as the in-SC note.

@shawna-slh

bruce-usab:
Nice improvement!

My apologies for scope creep, but I think figures 18 and 19 should have mention of the colors (or contrast ratio) and mention that it's the middle button which is the problem.  I’ll try and make a single-line comment in a timely way to be a bit more specific wrt my suggestion.

kfranqueiro:
> My apologies for scope creep, but I think figures 18 and 19 should have mention of the colors (or contrast ratio) and mention that it's the middle button which is the problem. I’ll try and make a single-line comment in a timely way to be a bit more specific wrt my suggestion.

Please make a separate PR or issue for this. This PR concerns site-wide styles; it is not changing content.

PR #4883, opened 2026-01-22
Add missing related technique SCR39 to F95 (and vice-versa)

venkatvemuri28:
####  **Bug Summary**
• Missing related technique reference SCR39 in F95 failure technique

---

####  **Root Cause Analysis**
• Not updated in F95.html Related Techniques section
• Closes #3789

---

####  **Fix Details**
• Replaced placeholder in techniques/failures/F95.html

---

####  **Risk / Impact Areas**
• Minimal risk: single file update
• Affects only navigation to related technique
• No impact on site functionality or other techniques

---

####  **Checklist**
- [x] I tested this across supported browsers/devices
- [x] No console errors or warnings

---

#### **Reviewer Notes**
• Verify link resolves correctly in built site
• Check consistency with other technique links (e.g., .html extension)

PR #4896, opened 2026-01-28
Adds SCR39 as related technique to F95

umut-gultekin:
Adds SCR39 as related technique to F95

Preview: https://deploy-preview-4896--wcag2.netlify.app/techniques/failures/f95

### DIFF
--- a/.github/workflows/11ty-publish.yaml
+++ b/.github/workflows/11ty-publish.yaml
@@ -32,9 +32,8 @@ jobs:
           npm i
           npm run build
           cp guidelines/guidelines.css guidelines/relative-luminance.html _site/guidelines/22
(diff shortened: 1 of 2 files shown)


patrickhlauke:
for TF discussion...

question is ... do we generally link failure techniques to positive techniques? their "relation" as such is that they both relate to 1.4.13

https://www.w3.org/WAI/WCAG22/Techniques/failures/F95
https://www.w3.org/WAI/WCAG22/Techniques/client-side-script/SCR39

if yes, then likely we also want the relationship to be reciprocal, i.e. adding a link to F95 to SCR39

patrickhlauke:
answering my own question: failure techniques do indeed have "related (positive) techniques".

remaining question: do we need a back-reference from SCR39 to the failure technique?

patrickhlauke:
confusingly, both this one AND https://github.com/w3c/wcag/pull/4883 actually work the same way once you look at the previews. I believe there's some magic link rewriting that happens when the techniques actually get published. For consistency with other similar cross-references from FXX techniques to other related techniques, going to close this one and go with https://github.com/w3c/wcag/pull/4883

PR #4836, opened 2025-12-28
Clarification of 2.5.3 Label in Name understanding (case insensitivity)

patrickhlauke:
While the understanding document https://www.w3.org/WAI/WCAG22/Understanding/label-in-name.html does mention that capitalization "may" be irrelevant, the advice buries the lede. This makes it a lot more obvious and explicit, while also mentioning that as best practice capitalization/matching accName and visible label may be meaningful for SR users.

---

In addition, one aspect that always felt strange and out of place in the understanding was the use of inline "live" examples.

screenshot of the understanding document, showing the inline live examples circled

They look amateurish, and the first two don't actually "work" - the accessible names of the controls match exactly what the visible labels are. Replaced these with figures AND code samples that *do* actually show a divergence between label and accName in terms of punctuation/capitalization.

Lastly, this PR fixes some awkwardness in the prose for the mathematical section.

---

Closes https://github.com/w3c/wcag/issues/3812

Preview: https://deploy-preview-4836--wcag2.netlify.app/understanding/label-in-name
[HTML diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FUnderstanding%2Flabel-in-name&doc2=https%3A%2F%2Fdeploy-preview-4836--wcag2.netlify.app%2Funderstanding%2Flabel-in-name)

Issue #3812, opened 2024-05-02
2.5.3 Label in Name - case insensitve?


patrickhlauke:
I'm going to assume (based on the described use cases/benefits) that 2.5.3 is case insensitive. For instance, I'd assume the following passes:

```
<button aria-label="Delete">DELETE</button>
```

If others agree, suggest this should be clarified/mentioned in the understanding document (don't think it currently does, unless I missed)

https://www.w3.org/WAI/WCAG22/Understanding/label-in-name.html

giacomo-petri:
I’m curious if it makes a difference whether an acronym is used in uppercase vs lowercase. 

Let's consider:
- "ADD" as a word in the context of "add to bag":
`<button aria-label="add">ADD</button>`
No issues
- "ADD" as an acronym for "Attention Deficit Disorder":
`<a href="..." aria-label="add">ADD</a>`
Could there be any issue here?

I know that VoiceOver, for instance, tends to read uppercase acronyms as individual letters. However, I’m unsure how speech recognition technologies handle this. Do they recognize acronyms letter by letter, or do they interpret the entire word? 

In the context of "Attention Deficit Disorder," I wouldn't expect someone to pronounce "add" as a word. Instead, I would anticipate it being spelled out as "A," "D," "D."

Does it make a difference for speech recognition technology users if you use:

`<a href="..." aria-label="add">ADD</a>`
vs
`<a href="..." aria-label="ADD">ADD</a>`

?

filippo-zorzi:
Voice Control (with Chrome, for what it matters) doesn't distinguish between "add" and "ADD".

https://github.com/user-attachments/assets/dcbc1e9c-ff3b-48f9-acbd-de73348690ff

"A.D.D." has to be spelled out but this is not in the focus of this issue.

giacomo-petri:
Another relevant example is when uppercase and lowercase letters convey different meanings. For instance, in chemistry, "CO" represents carbon monoxide, while "Co" stands for cobalt. "CO" is read letter by letter as "C–O", whereas "Co" is pronounced as a single syllable, "Co" (at least in Italian).

Thus, although in most cases uppercase and lowercase make no difference, it is still important to include a note explaining that, when the specific letter case is meaningful, the accessible name must reflect the visual representation.

If we all agree, I'll take care of it.

patrickhlauke:
[...]

PR #4897, opened 2026-01-28
Understanding Label in Name editorial

bruce-usab:
Minor editorial tweaks to “punctuation and capitalization” subsection of Understanding for SC 2.5.3 Label in Name.
1. `&`times`;` -> `&times;` and `>` -> `&gt;`
1. Wrap three inline examples into `figure` tags.
1. Don’t overwrite (!) `name` in radio button example (and change to check boxes, just because).
1. Use `q` element for body text quotation marks.
1. Use lower case for what the screen reader vocalizes.
1. Use distinct values for attributes so if someone looks at source code, it’s clear which `id` is being used by `label for`.

**To Do**
* incorporate work from #3812
* incorporate work from #4836 
* Review use of quotation marks
  - original used `q` element a couple places.  The element generates noisy code.
  - replace neutral quotation marks with typographical one
  - remove quotation marks from heading
* Review treatment of terms, sometimes they are bold emphasis, sometimes in-page links, sometimes fully qualified link.
  - Make all consistently hyperlinks

Anchor to subsection of [preview](https://deploy-preview-4897--wcag2.netlify.app/understanding/label-in-name#text-labels-express-something-in-human-language) and [diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FUnderstanding%2Flabel-in-name.html&doc2=https%3A%2F%2Fdeploy-preview-4897--wcag2.netlify.app%2Funderstanding%2Flabel-in-name#text-labels-express-something-in-human-language).


understanding/21/label-in-name.html
```diff
@@ -37,38 +37,38 @@ <h3>Identifying label text for components</h3>
...
 			<section>
-				<h3>Text labels "express something in human language"</h3>
+				<h3>Text labels “express something in human language”</h3>
```
bruce-usab:
```suggestion
				<h3>Text labels <q>express something in human language</q></h3>
```

understanding/21/label-in-name.html
```diff
@@ -37,38 +37,38 @@ <h3>Identifying label text for components</h3>
...
-          <p>The use of punctuation and capitalization in labels <em>may</em> also be considered optional for the same reason. For example, the colon conventionally added at the end of input labels does not express something in human language, and capitals on the first letter of each word in a label do not normally alter the words' meaning. This is particularly relevant in the context of this SC, since it is primarily aimed at users of speech recognition; capitals and most punctuation are frequently ignored when a user speaks a label as a means of interacting with a control.</p>
+          <p>The use of punctuation and capitalization in labels <em>may</em> also be considered optional for the same reason. For example, the colon conventionally added at the end of input labels does not express something in human language, and capitals on the first letter of each word in a label do not normally alter the word’s meaning. This is particularly relevant in the context of this SC, since it is primarily aimed at users of speech recognition; capitals and most punctuation are frequently ignored when a user speaks a label as a means of interacting with a control.</p>
```
kfranqueiro: Reverting this because it is correctly "words" plural

```suggestion
          <p>The use of punctuation and capitalization in labels <em>may</em> also be considered optional for the same reason. For example, the colon conventionally added at the end of input labels does not express something in human language, and capitals on the first letter of each word in a label do not normally alter the words' meaning. This is particularly relevant in the context of this SC, since it is primarily aimed at users of speech recognition; capitals and most punctuation are frequently ignored when a user speaks a label as a means of interacting with a control.</p>
```

baldino-m:
It might be worthy replacing the examples which are currently implemented as "real" components, such as the "Next" button, "First name" input field and math checkboxes with a picture + code snippet, since cleaner.

bruce-usab:
Discussed on TF call 30 Jan, moved back to In Progress.  Bruce needs to double check use of `q` element and add `name` attribute.

We are looking for a volunteer to create screen shot snippets, since that would be better than the current use of incomplete `form` UIC elements.

understanding/21/label-in-name.html
```diff
@@ -21,11 +21,11 @@ <h2>In brief</h2>
...
 			<p>Mismatches between visible labels and programmatic names for controls are even more of an issue for speech-input and text-to-speech users who also have cognitive challenges. Mismatches create an extra cognitive load for speech-input users, who must remember to say a speech command that is different from the visible label they see on a control. It also creates extra cognitive load for a text-to-speech user to absorb and understand speech output that does not match the visible label.</p>
-			<p>Note that when a <a>user interface component</a> lacks an <a href="https://www.w3.org/TR/accname/#dfn-accessible-name">accessible name</a> — a failure of <a href="name-role-value">4.1.2 Name, Role, Value</a> — and has a visible text label, then it also fails this success criterion.</p>
+			<p>A <a>user interface component</a> with a visible label but without an <a href="https://www.w3.org/TR/accname/#dfn-accessible-name">accessible name</a> fails both this success criterion _and_ <a href="name-role-value">4.1.2 Name, Role, Value</a>.</p>
```
scottaohara: re the use of `_and_` - do these files support markdown?  should this be an `<em>and</em>`?

understanding/21/label-in-name.html
```diff
@@ -21,11 +21,11 @@ <h2>In brief</h2>
...
+			<p>Where text labels exist and are properly linked to the user interface components through established authoring practices, the label and name will normally match. When they don't match, speech-input users who attempt to use the visible text label as a means of navigation or selection will be unsuccessful. For example, the user says “move to password” (because that is the lable they see) but programmatic name of the field is not “password”. The speech-based navigation fails because the visible label spoken by the user does not match (or is not part of) the accessible name that is enabled by voice recognition software as a speech-input command. In addition, when the accessible name is different from the visible label, it may function as a hidden command that can be accidentally activated by speech-input users.</p>
```
scottaohara:
```suggestion
			<p>Where text labels exist and are properly linked to the user interface components through established authoring practices, the label and name will normally match. When they don't match, speech-input users who attempt to use the visible text label as a means of navigation or selection will be unsuccessful. For example, the user says “move to password” (because that is the label they see) but programmatic name of the field is not “password”. The speech-based navigation fails because the visible label spoken by the user does not match (or is not part of) the accessible name that is enabled by voice recognition software as a speech-input command. In addition, when the accessible name is different from the visible label, it may function as a hidden command that can be accidentally activated by speech-input users.</p>
```

understanding/21/label-in-name.html
patrickhlauke: ah, good catch... it's the sort of thing i keep accidentally doing...
```suggestion
			<p>A <a>user interface component</a> with a visible label but without an <a href="https://www.w3.org/TR/accname/#dfn-accessible-name">accessible name</a> fails both this success criterion <em>and</em> <a href="name-role-value">4.1.2 Name, Role, Value</a>.</p>
```

bruce-usab:
Made obsolete by #4836

PR #4906, opened 2026-02-02
Update affiliations

iadawn:


acknowledgements/ag-wg-participants.html
```diff
@@ -89,7 +89,7 @@ <h4>Participants active in the AG WG at the time of publication</h4>
 		<li>Lemon, Gez (The Paciello Group, LLC)</li>
```
patrickhlauke, line 89:
```suggestion
		<li>Lemon, Gez (TetraLogical)</li>
```

patrickhlauke: ah actually...at the time of publication being the original? or even the updates? if the former, then ignore me... otherwise (if the updates now count, gez and also steve faulkner are TetraLogical)

PR #4916, opened 2026-02-06
Remove allowCompoundWords from cspell config to catch more errors

kfranqueiro:
This applies a learning I gained while working on the WCAG 3 repo: it turns out that while `allowCompoundWords` can be convenient, it causes cspell to miss legitimate typos, as it essentially allows various words or parts of words to be stitched together, even if the outcome isn't a real word.

Removing `allowCompoundWords: true` results in many more failures. Some are typos or compounds that would typically be hyphenated, which I've fixed. Some are valid words or reasonable exceptions, so I've added those either to `custom-words.txt` for cases that should be universally allowed, or within `cspell.yml` as exceptions to specific files or folders.

There were a couple of errors within `guidelines/` that could be considered legitimate, but I chose to ignore them for now, and can send a separate, more focused PR for editorial fixes within the recommendation.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2908,9 +2908,10 @@
       }
     },
     "node_modules/markdown-it": {
-      "version": "14.1.0",
-      "resolved": "https://registry.npmjs.org/markdown-it/-/markdown-it-14.1.0.tgz",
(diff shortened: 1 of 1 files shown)


bruce-usab:
Discussed and reviewed on TF call 13 Feb.  Many adjustments including hyphenation (e.g. “non-verbally” instead of “non verbally”) and compound words  (e.g. “screenreader” instead of “screen reader”).  An example of prefix+suffix variants allowing a misspelling is “infoation”.  Moved to Ready for approval.

PR #4917, opened 2026-02-08
fix: use correct HTML attribute `aria-describedby`, not `aria-describedBy`

d-koppenhagen:
Some example show to use `aria-describedBy` (with a big "B"). The attribute should be `aria-describedby` instead.


patrickhlauke:
good catch, yup

patrickhlauke:
note that we don't touch the `wcag20` folder anymore (and the .xml sources will be removed soon altogether), so ... would you mind removing those from the PR, making it purely about the `techniques/aria/ARIA21.html` file alone?

d-koppenhagen:
@patrickhlauke done

PR #4920, opened 2026-02-11
Normative change to SC 2.2.2 - replace "information" with "content", rewrite note 3

### LINKED ISSUE #3478: 2.2.2 Pause, stop, hide SC text should make it clear it applies to all content
Opening this issue due to a public comment, [Issue 232](https://github.com/w3c/wcag2ict/issues/232), we received on the WCAG2ICT FPWD. The issue indicates that the normative SC text in WCAG says "information" rather than "content" and thought it was interpreted to mean that the SC doesn't apply to all content on the web page. This brought in question the WCAG2ICT notes that indicated the SC applies to all content and the reviewer questions whether WCAG2ICT is expanding the scope of the SC.  

In WCAG, the non-normative Note 2 does say, "all content on the Web page (whether it is used to meet other success criteria or not) must meet this success criterion." It would be most helpful to have a clarification in the WCAG errata that normatively, this SC applies to all content, as it seems that was the intent for this SC.  SC 2.2.2 is not just limited to content that is considered "information". Otherwise, the notes and understanding should be updated to clarify what type(s) of content are out of scope for meeting this SC.

See also the WCAG2ICT TF discussion on this, documented in the [5 October meeting minutes](https://www.w3.org/2023/10/05-wcag2ict-minutes#t03).

### LINKED ISSUE #4921: Clarification: Decorative autoplay background video and SC 2.2.2 vs SC 1.1.1
Hello WCAG Working Group,

I’m opening this issue to summarize and confirm a clarification that emerged from discussion in another thread. (Apologies for starting that discussion in the wrong place — I’m reposting here to keep the topic organized.)

The relevant discussion began in #3478 (see comment by @etherplain: https://github.com/w3c/wcag/issues/3478#issuecomment-3887152381).

My understanding from that discussion is:

- Autoplay decorative background motion may trigger SC 2.2.2 (Pause, Stop, Hide) and require a user mechanism to pause or stop motion.
- This requirement is independent of semantic classification under SC 1.1.1.
- Decorative media may still remain hidden from assistive technologies and does not require alt text or transcripts if it conveys no information.

In other words, pause controls address the impact of motion on users, not the informational status of the media.

I’m documenting this here to confirm that this interpretation accurately reflects working group intent.

Thank you for any confirmation or correction.

patrickhlauke:

---

There is confusion relating to why the SC uses "information" rather than "content". While there may have been discussions back in the day when this was first drafted (possibly to make a distinction between "decorative" and "non-decorative" content?), the notes and understanding don't bear out this interpretation. Also, logically - if one of the primary aims of the SC is to help users that are distracted/confused by moving/auto-updating content that happens in parallel to other content ... it makes no difference whether that moving/updating stuff is "decorative" or "non-decorative".

In addition, Note 3 is currently almost unreadable (and grammatically wrong at the end as well: "and in many situations could be misleading to do so"?)

This PR proposed a normative change to the SC text itself and Note 3

Closes https://github.com/w3c/wcag/issues/3478
Closes https://github.com/w3c/wcag/issues/4921


guidelines/sc/20/pause-stop-hide.html
```diff
@@ -1,57 +1,39 @@
...
-      meet other success criteria or not) must meet this success criterion. See <a href="#cc5">Conformance Requirement 5: Non-Interference</a>.
-   </p>
+      meet other success criteria or not) must meet this success criterion. See <a href="#cc5">Conformance Requirement 5: Non-Interference</a>.</p>
    
-   <p class="note">Content that is updated periodically by software or that is streamed to the user agent
-      is not required to preserve or present information that is generated or received between
-      the initiation of the pause and resuming presentation, as this may not be technically
-      possible, and in many situations could be misleading to do so.
-   </p>
+   <p class="note">When auto-updating content was paused, it is not required
```
ljoakley:
```suggestion
   <p class="note">When auto-updating content is paused, it is not required
```

guidelines/sc/20/pause-stop-hide.html
```diff
@@ -1,57 +1,39 @@
...
+      for any of the updates that happened while the content was paused to be presented
+      to the user once the content is un-paused again – as this may not be technically
+      possible, and in many situations could be confusing or misleading.</p>
    
    <p class="note">An animation that occurs as part of a preload phase or similar situation can be considered
       essential if interaction cannot occur during that phase for all users and if not indicating
-      progress could confuse users or cause them to think that content was frozen or broken.
-   </p>
-   
+      progress could confuse users or cause them to think that content was frozen or broken.</p>
```
patrickhlauke:
```suggestion
      progress could confuse users or cause them to think that content is frozen or broken.</p>
```

scottaohara (approved):

awkawk:
This can be clarified in Understanding more easily.

As an aside, is it possible to have a easier-to-read diff for changes in the normative documents? It seems like a blank line or two being deleted makes it necessary to read more deeply to find changes. In this case, the second to last note, which is substantially changed.

bruce-usab (approved):
Link to 2.2.2 in [diff preview](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FTR%2F2024%2FREC-WCAG22-20241212%2F&doc2=https%3A%2F%2Fdeploy-preview-4920--wcag2.netlify.app%2Fguidelines%2F#pause-stop-hide) for @awkawk et al.

patrickhlauke:
> This can be clarified in Understanding more easily

I'd posit that doing an erratum is clearer than having normative language that says one thing, and then having to explain in the understanding "when we say X, we actually mean Y"

awkawk:
> I'd posit that doing an erratum is clearer than having normative language that says one thing, and then having to explain in the understanding "when we say X, we actually mean Y"

And then when we make that change, every bit or training, tooling, and documentation created around the world to support 2.2.2 will need to be updated, which represents possibly a greater risk to clarity and an important consideration before changing the normative language.

patrickhlauke:
let's never make any errata or changes ever then, @awkawk ... cast WCAG in amber/carbonite

awkawk:
Thats not what I'm saying, Patrick. To quote myself: "an important consideration before changing the normative language".

In the case of this particular change, I don't think that it is worth it.

Issue #4315, opened 2025-04-03
Improve the examples given in PDF9


DuffJohnson:
[PDF9](https://www.w3.org/WAI/WCAG22/Techniques/pdf/PDF9) includes the following:

> Heading markup can be used:
> 
> - to indicate start of main content;
> - to mark up section headings within the main content area;
> - to demarcate different navigational sections, such as top or main navigation, left or secondary navigation, and footer navigation;
> - to mark up images (containing text) which have the appearance of headings visually.

However, in the PDF context (the technology to which this technique applies)...

- the 1st bullet is dubious at best (especially if H1 is used for the document's title), as many documents locate front-matter (which isn't "main content") after the title.
- the 2nd bullet would ideally be enhanced to read: "to mark up section headings (text or images) within the main content area;"
- the 3rd bullet is simply inapplicable in the PDF context, and should be removed
- the 4th bullet can be retired based on the suggested change to the 2nd bullet.

Ironically, the given examples are far more applicable to HTML content than to PDF... but the various techniques applicable to HTML don't provide these examples.

DuffJohnson:
I've [posted to this effect](https://github.com/w3c/wcag/issues/4315) (more or less) on the W3C's repo.

DuffJohnson:
Based on internal discussions I'm going to drop this; it's not a necessary or especially useful change.

PR #4922, opened 2026-02-13
Update PDF9.html

patrickhlauke:
Removed unnecessary qualifications, streamlined the text to address #4315.

(This was an orphaned stale branch - opening this PR to decide if we want this proposed, but never sent, change from @DuffJohnson)


patrickhlauke:
ah, apologies ... i just connected the dots with https://github.com/w3c/wcag/issues/4315 and @DuffJohnson decided to drop it (but the open branch was left there by accident). closing, which should also delete the stale branch

PR #4927, opened 2026-02-13
Update 1.4.3 and 1.4.6 to include references to text in video

### LINKED ISSUE #1939: Contrast and video
I'm wondering if the two contrast criteria include text in video. I understand the exceptions for Logotypes and Incidental ("… or that are part of a picture that contains significant other visual content, have no contrast requirement") parts of the criterion, but am wondering about a fairly common situation where, at the end of a video, you'd have a corporate logo, followed by a company slogan, and a URL. Would the slogan and URL fall under 1.4.3 and 1.4.6?

patrickhlauke:
Filing as a draft PR for now... based on an branch that was set up at the time but without a matching PR due to ongoing discussion in the original issue.

This, once refined, closes #1939


understanding/20/contrast-enhanced.html
patrickhlauke:
```suggestion
         understood as text, which includes text in videos.. Incidental text, such as in photographs that happen to include
```

understanding/20/contrast-enhanced.html
```diff
@@ -87,17 +87,16 @@ <h2>Intent of Contrast (Enhanced)</h2>
       
       <p>The contrast requirements for text also apply to images of text
          (text that has been rendered into pixels and then stored in an image format) - see
-         <a href="images-of-text">Success Criterion 1.4.5: Images of Text</a>.
+         <a href="images-of-text">Success Criterion 1.4.5 Images of Text</a>.
       </p>
       
       <p>This requirement applies to situations in which images of text were intended to be
-         understood as text. Incidental text, such as in photographs that happen to include
+         understood as text, which includes text in videos.. Incidental text, such as in photographs that happen to include
```
patrickhlauke:
```suggestion
         understood as text, which includes text in videos. Incidental text, such as in photographs that happen to include
```

bruce-usab:
Discussed on TF call 27 Feb and moved to Ready for Approval.

understanding/20/contrast-minimum.html
```diff
@@ -113,7 +113,7 @@ <h2>Intent of Contrast (Minimum)</h2>
       </div>
 
       <p>This success criterion applies to text in the page, including
-         placeholder text and text that is shown when a pointer is hovering over an object
+         placeholder text, text in videos, and text that is shown when a pointer is hovering over an object
```
dbjorge, line 116: :-1: I don't think this has enough detail to make this understanding doc clear about the scope. The SC applies specifically to "The visual presentation of [text](https://www.w3.org/WAI/WCAG22/Understanding/contrast-minimum.html#dfn-text) and [images of text](https://www.w3.org/WAI/WCAG22/Understanding/contrast-minimum.html#dfn-image-of-text)"; text in videos *can* qualify as image of text, but *often doesn't*. The "image of text" definition is:

> **image of text**: text that has been rendered in a non-text form (e.g., an image) in order to achieve a particular visual effect
> **Note**: This does not include text that is part of a picture that contains significant other visual content.
> **Example**: A person's name on a nametag in a photograph.

I think this definition suggests that if you had a video that was "I animated the word 'Sale!' bouncing from side to side", that would be in scope, but if you had a video where a low-contrast street sign happened to be part of a movie scene, that would not be in scope.

I think *most* text in videos in practice would *not* be in scope, so I think we should be more clear than this update suggests.

bruce-usab: It occurs to me that “A person‘s name on nametag is a photograph” could be read as being an example of images of text — when the opposite is intended!

bruce-usab: And to @dbjorge good point on call wrt captioning -- those often are actual text (not images of text).  With digital TVs and modern closed captioning, sufficient contrast is guaranteed because modern TVs let one pick foreground and background colors (and other settings, including font).  The source media lacking CC data is addressed by other SC.

TestPartners: > I think _most_ text in videos in practice would _not_ be in scope.

There are certainly plenty of videos like that, but most or all of the text would be in scope in many of the videos we encounter. This includes opening credits, title frames between scenes, and multiple end cards that can contain a large amount of information. Important text is often displayed throughout educational videos and animations. And some clients still create videos with open captions.

PR #4939, opened 2026-02-17
Don't just refer to SCs by their number, don't use colon

patrickhlauke:
For consistency:

* don't use a colon between the SC number and its title - my main reason is that it looks somehow "stilted", and when used in a sentence it can end up "fighting" with the punctuation of the actual sentence (particularly when not part of a link), and just makes it harder to parse
* don't just refer to SCs by number alone - (at least for first mentions in a document) this helps documents be more readable without assuming that the reader walks around remembering what each SC number is/what criterion it actually represents.

It would be good to get something decided/officially mentioned in the style guide (particularly for the use of `:` or not)

EDIT: normative changes split into https://github.com/w3c/wcag/pull/4947

### DIFF
--- a/understanding/20/error-identification.html
+++ b/understanding/20/error-identification.html
@@ -27,7 +27,7 @@
          failed.
          The error must be indicated in <a>text</a>.</p>
          <p>This SC requires that users be provided with information about the nature of the error, including the identity of the item in error. What the user should do to correct the item in error is covered by 
-         <a href="error-suggestion">3.3.3 Error Suggestion</a>. Often, the error description can be phrased so that it meets both SC 3.3.1 Error Identification and SC 3.3.3 Error Suggestion at the same time. For instance, "Email is not valid" would pass SC 3.3.1, but "Please provide a valid email address in the format name@domain.com" also conveys how it can be fixed and passes both.
+         <a href="error-suggestion">3.3.3 Error Suggestion</a>. Often, the error description can be phrased so that it meets both Success Criteria 3.3.1 Error Identification and 3.3.3 Error Suggestion at the same time. For instance, "Email is not valid" would pass 3.3.1, but "Please provide a valid email address in the format name@domain.com" also conveys how it can be fixed and passes both.
       </p>
       <p>An "input error" includes:</p>
 


shawna-slh:
+1 to no colon to match the punctuation in the [WCAG standard](https://www.w3.org/TR/WCAG22/#non-text-content); for example:

> Success Criterion 1.1.1 Non-text Content

kfranqueiro:
I did a pass through the PR, found and fixed additional typos, made a couple of tweaks to front-load expansions found multiple times within the same paragraph, and removed colons added within the build system.

I do think we should split this; IMO it makes sense to keep the informative changes in this PR since that applies to the majority of changes, and move the `guidelines/` changes to a separate one. (If you want, I can help to do that.)

bruce-usab:
Reviewed on TF call 20 Feb.  Consensus on Ready for approval, but needs to be split into two PRs to address TR space files separately.

patrickhlauke:
with thanks to @kfranqueiro the normative part is now split out into https://github.com/w3c/wcag/pull/4947

moving this and the other one to "Ready for approval"

understanding/20/three-flashes.html
```diff
@@ -32,8 +32,9 @@ <h2>Intent of Three Flashes</h2>
       </p>
       
       <p>Whereas 
-         <a href="three-flashes-or-below-threshold" class="guideline">Success Criterion 2.3.1</a> allows flashing if it is dim enough or has a small enough area, Success Criterion
-         2.3.2 does not allow flashing greater than 3 per second, regardless of brightness
+         <a href="three-flashes-or-below-threshold" class="guideline">Success Criterion 2.3.1 Three Flashes or Below Threshold</a>
+         allows flashing if it is dim enough or has a small enough area, Success Criterion
```
bruce-usab: Maybe replace comma with semicolon here?  It was okay before, but with the ten extra words it’s a bit harder to parse.  And the clauses could stand on their own each as a sentence.

bruce-usab:
I tagged as “substantive” to highlight that there is a bit more to go through than is typical for editorial changes.

understanding/20/three-flashes.html
kfranqueiro: Changing the comma to a semicolon would make the sentence grammatically incorrect.

Incidentally, this was one of the specific situations I raised to Patrick, where expanding the number to number + name could be argued as unnecessary because 2.3.2 is literally the page we're looking at right now. Or we could rephrase "Success Criterion 2.3.2 Three Flashes" to "this success criterion".

maryjom:
```suggestion
         allows flashing if it is dim enough or has a small enough area. Success Criterion
```

understanding/20/three-flashes.html
```diff
@@ -32,8 +32,9 @@ <h2>Intent of Three Flashes</h2>
...
+         2.3.2 Three Flashes does not allow flashing greater than 3 per second, regardless of brightness
          or size. As a result, even a single flashing pixel would violate this criterion. The
```
maryjom: Restructured sentences so clarify what "this criterion" is referring to.
```suggestion
         or size, which means even a single flashing pixel would violate this criterion. The
```

understanding/20/three-flashes.html
patrickhlauke: making this a full stop makes no sense. The sentence starts with "Whereas...". You can't just stop it there.

patrickhlauke: I amended this whole sentence. This should make it more readable now...

understanding/20/three-flashes.html
patrickhlauke: Amended the whole sentence to be more readable

understanding/20/error-identification.html
```diff
@@ -27,7 +27,7 @@ <h2>Intent of Error Identification</h2>
          failed.
          The error must be indicated in <a>text</a>.</p>
          <p>This SC requires that users be provided with information about the nature of the error, including the identity of the item in error. What the user should do to correct the item in error is covered by 
-         <a href="error-suggestion">3.3.3 Error Suggestion</a>. Often, the error description can be phrased so that it meets both SC 3.3.1 and SC 3.3.3 at the same time. For instance, "Email is not valid" would pass SC 3.3.1, but "Please provide a valid email address in the format name@domain.com" also conveys how it can be fixed and passes both.
+         <a href="error-suggestion">3.3.3 Error Suggestion</a>. Often, the error description can be phrased so that it meets both SC 3.3.1 Error Identification and SC 3.3.3 Error Suggestion at the same time. For instance, "Email is not valid" would pass SC 3.3.1, but "Please provide a valid email address in the format name@domain.com" also conveys how it can be fixed and passes both.
```
patrickhlauke:
```suggestion
         <a href="error-suggestion">3.3.3 Error Suggestion</a>. Often, the error description can be phrased so that it meets both Success Criteria 3.3.1 Error Identification and 3.3.3 Error Suggestion at the same time. For instance, "Email is not valid" would pass 3.3.1, but "Please provide a valid email address in the format name@domain.com" also conveys how it can be fixed and passes both.
```

PR #4947, opened 2026-02-20
Don't just refer to SCs by their number alone

patrickhlauke:
Normative counterpart to https://github.com/w3c/wcag/pull/4939

> For consistency:
> 
> * don't just refer to SCs by number alone - (at least for first mentions in a document) this helps documents be more readable without assuming that the reader walks around remembering what each SC number is/what criterion it actually represents.


patrickhlauke:
Discussed this in the TF meeting 20 Feb. Group agreed to put this to the wider WG.

bruce-usab (approved):

kfranqueiro:

PR #4952, opened 2026-02-24
Update normative wording of 4.1.3 and definition of "status message"

### LINKED ISSUE #4676: Confusion regarding definition of "status message"

patrickhlauke:

---

The current normative wording for [4.1.3 Status Messages](https://www.w3.org/TR/WCAG22/#status-messages) and the [definition for the term status message](https://www.w3.org/TR/WCAG22/#dfn-status-messages) have ... issues:

* when substituting the term definition in the normative SC wording, it's unclear *what* a status message is / when the SC actually applies - see https://github.com/w3c/wcag/discussions/4672. just going by the normative wording, any change of content that isn't a change of context needs to be announced (missing the fact that it only applies to discrete/small messages/notifications)
* the mention of `through role or properties` would disallow the use of upcoming/new solutions, such as [ariaNotify](https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaNotify), since they are programmatic but do *not* rely on roles/properties - there is https://github.com/w3c/wcag/pull/5079 that attempts to awkwardly patch this in understanding, but it would be better to fix this at source
* the preface of `In content implemented using markup languages` limits the scope of this SC, which results in 4.1.3 not being applicable in cases where WCAG is being applied to other technologies (WCAG2ICT) - the current https://www.w3.org/TR/wcag2ict/#applying-sc-4-1-3-status-messages-to-non-web-documents-and-software notes that the SC "applies as written", meaning that in non-markup-language scenarios, it doesn't ... but then provides a note saying it sure would be nice if it did ("there is still a user need...") - and this handwave is also present in current [draft versions of the upcoming EN 301 549 v4.1.0](https://www.etsi.org/deliver/etsi_en/301500_301599/301549/04.01.00_20/en_301549v040100ev.pdf); many auditors currently "ignore" the markup language part when evaluating native mobile/desktop apps against WCAG, and take the fundamental principle - that AT must announce status messages when these don't receive focus - regardless

I believe this change, while quite deep, is backwards-compatible with the existing/current version of the wording the way it was originally intended

EDIT: Reintroduced the limitation of scope by adding "In technologies that support it" as a preamble

---

### Current normative wording (SC + term)

Current normative wording of the SC, the definition, and "grafting" (with tweaks from singular to plural) that definition into the normative wording:

> In content implemented using markup languages, **status messages** can be _programmatically determined_ through _role_ or properties such that they can be presented to the user by _assistive technologies_ without receiving focus.

> change in content that is not a _change of context_, and that provides information to the user on the success or results of an action, on the waiting state of an application, on the progress of a _process_, or on the existence of errors

> In content implemented using markup languages, **change[s] in content that [are] not a _change of context_, and that [provide] information to the user on the success or results of an action, on the waiting state of an application, on the progress of a _process_, or on the existence of errors** can be _programmatically determined_ through _role_ or properties such that they can be presented to the user by _assistive technologies_ without receiving focus.

---

### Proposed normative rewording (SC + term)

This PR proposes the following change to the SC, the definition, and shows the resulting "grafted" version  (with tweaks from singular to plural) :

> In technologies that support it, **status messages** that appear dynamically (without a _change of context_) can be _programmatically determined_ without receiving focus, so that they can be presented to the user by _assistive technologies_.

> discrete piece of content whose only purpose is to provide information to the user on the results of an action, the waiting state of an application, the progress or completion of a _process_, or the existence of errors

>In technologies that support it, **discrete piece[s] of content whose only purpose is to provide information to the user on the results of an action, the waiting state of an application, the progress or completion of a _process_, or the existence of errors** that appear dynamically (without a _change of context_) can be _programmatically determined_ without receiving focus, so that they can be presented to the user by _assistive technologies_.

---

- [SC diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fwcag%2Fguidelines%2F22%2F&doc2=https%3A%2F%2Fdeploy-preview-4952--wcag2.netlify.app%2Fguidelines%2F#status-messages)
- [Term diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fwcag%2Fguidelines%2F22%2F&doc2=https%3A%2F%2Fdeploy-preview-4952--wcag2.netlify.app%2Fguidelines%2F#dfn-status-messages)

Closes #4676

This PR supersedes https://github.com/w3c/wcag/pull/4840 (as that one originally only tried to update the definition, but further discussion on the PR made it clear that there's more fundamental work to be done, if we're already aiming for a normative change)

See also https://github.com/w3c/wcag/pull/5079 - if *that* PR is merged, this PR will need to be updated to account for the changes made there/mesh them in better.


guidelines/terms/21/status-message.html
```diff
@@ -1,6 +1,4 @@
 <dt><dfn id="dfn-status-messages" data-lt="status messages">status message</dfn></dt>
 <dd>
-   					
-   <p>change in content that is not a <a>change of context</a>, and that provides information to the user on the success or results of an action, on the waiting state of an application, on the progress of a <a>process</a>, or on the existence of errors</p>
-   
+   <p>discrete piece of content that is added or updated dynamically, and whose only purpose is to provide information to the user on the success or results of an action, on the waiting state of an application, on the progress or completion of a <a>process</a>, or on the existence of errors</p>
```
fstrr: "success" is the result of an action, so I think we can take that out.

```suggestion
   <p>discrete piece of content that is added or updated dynamically, and whose only purpose is to provide information to the user on the results of an action, on the waiting state of an application, on the progress or completion of a <a>process</a>, or on the existence of errors</p>
```

guidelines/sc/21/status-messages.html
```diff
@@ -1,9 +1,5 @@
 <section id="status-messages" class="sc">
-   					
    <h4>Status Messages</h4>
-   					
    <p class="conformance-level">AA</p>
-   					
-   <p>In content implemented using markup languages, <a>status messages</a> can be <a>programmatically determined</a> through <a>role</a> or properties such that they can be presented to the user by <a>assistive technologies</a> without receiving focus.</p>
-   				
+   <p><a>Status messages</a> that don't cause a <a>change of context</a> can be <a>programmatically determined</a> such that they can be presented to the user by <a>assistive technologies</a> without receiving focus.</p>
```
electrickite: Small style note - contractions such as "don't" in this line are not used elsewhere in the normative text of WCAG.

patrickhlauke: good catch

patrickhlauke:
```suggestion
   <p><a>Status messages</a> that do not cause a <a>change of context</a> can be <a>programmatically determined</a> such that they can be presented to the user by <a>assistive technologies</a> without receiving focus.</p>
```

guidelines/terms/21/status-message.html
```diff
@@ -1,6 +1,4 @@
 <dt><dfn id="dfn-status-messages" data-lt="status messages">status message</dfn></dt>
 <dd>
-   					
-   <p>change in content that is not a <a>change of context</a>, and that provides information to the user on the success or results of an action, on the waiting state of an application, on the progress of a <a>process</a>, or on the existence of errors</p>
-   
+   <p>discrete piece of content that is added or updated dynamically, and whose only purpose is to provide information to the user on the results of an action, on the waiting state of an application, on the progress or completion of a <a>process</a>, or on the existence of errors</p>
```
adampage:
```suggestion
   <p>discrete piece of content that is added or updated dynamically, and whose only purpose is to provide information to the user on:</p>
   <ul>
     <li>the results of an action</li>
     <li>the waiting state of an application</li>
     <li>the progress or completion of a <a>process</a></li>
     <li>the existence of errors</li>
   </ul>
```
Only because we’re already touching this, a soft formatting suggestion for parsability.

patrickhlauke: while i'd generally agree, the fact that the idea with these "terms" is that you should be able to replace them in place with their definition, adding this sort of structure makes this concept a bit more difficult (though sure, this wouldn't be the only term that does this)

adampage (approved):

detlevhfischer:
Very nice to see the reference to _content implemented using markup languages_ removed (in the light of applying the requirement to mobile apps)

bruce-usab:
I agree that it would be good to remove “in content using markup languages” here and elsewhere.  It gives too much of a free pass to mobile and WCAG2ICT.  As I noted in previous draft (#4840) excising “changes that are not a change of context” from the defined term is tremendously helpful.

bruce-usab:
Just FYI, from [editors draft of WCAG2ICT](https://w3c.github.io/wcag2ict/#status-messages):
> 4.1.3 Status Messages — Depends upon status messages being programmatically determinable using role or properties.

As @scottaohara noted, limiting authors to _only_ (ARIA) role or properties is counter to current best practices.

bruce-usab:
@detlevhfischer provided current formulation of 4.1.3 in EN 301 549 V.4.1:

> Where ICT is, or includes, non-web software that provides a user interface, and the software includes functionality that is not closed to programmatic access of information by assistive technologies, and the functionality that is not closed supports status message notifications, status messages from the functionality that is not closed can be programmatically determined so that they can be presented to the user by assistive technologies without receiving focus.
> 
> The WCAG success criterion is restricted to content implemented using markup languages, but the WCAG2ICT Task Force has concluded that even “where status messages are not implemented using markup languages, there is still a user need to have status messages be programmatically exposed so that they can be presented to the user by assistive technologies without receiving focus.”

bruce-usab:
On TF call we noted that moving focus incorrectly should not be a loophole.  @baldino-m suggest updating Understanding to allow for change of context to the right place.

patrickhlauke:
to expand, at TF meeting concern was raised that the "without causing a change of context" part didn't explicitly say that the status message is/was the one that received focus. so the gap is: a status message is updated, but focus moves SOMEWHERE ELSE on the page, then it would still be exempted. this gap is already present in the current term/normative SC wording, but just "hidden" by being buried in the definition.

@scottaohara raised the point that ariaNotify *is* going to be the de-facto solution that ARIA-WG is pushing, and `aria-live` will be legacy, so at the very least removing the "role or property" part of the normative wording is essential, as otherwise the SC won't be applicable to reality.

@giacomo-petri suggested that now that the "In markup languages..." part is removed, it would still be prudent to have wording along the lines of "In technologies that support it..." or similar, to provide some guard rails.

Will iterate on this and bring it back to the group in future.

bruce-usab (approved):
See [preview diff of SC 4.1.3](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FTR%2F2024%2FREC-WCAG22-20241212%2F&doc2=https%3A%2F%2Fdeploy-preview-4952--wcag2.netlify.app%2Fguidelines%2F#status-messages) and revised [glossary term](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FTR%2F2024%2FREC-WCAG22-20241212%2F&doc2=https%3A%2F%2Fdeploy-preview-4952--wcag2.netlify.app%2Fguidelines%2F#dfn-status-messages).

patrickhlauke:
Made changes based on the discussion at the last TF call:

* added a "In technologies that support it" preamble
* simplified the rest of the sentence to spell out that this means "status messages that don't receive focus". any other changes of context would already be out of scope based on the new definition that talks about DYNAMIC changes of content (i.e. loading a new page or similar would not be DYNAMIC anymore)

bruce-usab (approved):

guidelines/sc/21/status-messages.html
```diff
@@ -1,9 +1,5 @@
 <section id="status-messages" class="sc">
-   					
    <h4>Status Messages</h4>
-   					
    <p class="conformance-level">AA</p>
-   					
-   <p>In content implemented using markup languages, <a>status messages</a> can be <a>programmatically determined</a> through <a>role</a> or properties such that they can be presented to the user by <a>assistive technologies</a> without receiving focus.</p>
-   				
+   <p>In technologies that support it, <a>status messages</a> that don't receive focus can be <a>programmatically determined</a> so that they can be presented to the user by <a>assistive technologies</a>.</p>
```
baldino-m: I am not sure about this: does it leave out a "Thanks for your email"  message (which is loaded on a confirmation page, so it does represent a change of context) scenario? Can be this considered a status message?

patrickhlauke:
```suggestion
   <p>In technologies that support it, <a>status messages</a> can be <a>programmatically determined</a> without receiving focus, so that they can be presented to the user by <a>assistive technologies</a>.</p>
```

bruce-usab:
On TF call 4/3, we decided to work on this more.  In particular the example of a confirmation message -- both as a new page and as a status message -- would be a clarifying example.

guidelines/sc/21/status-messages.html
```diff
@@ -1,9 +1,5 @@
 <section id="status-messages" class="sc">
-   					
    <h4>Status Messages</h4>
-   					
    <p class="conformance-level">AA</p>
-   					
-   <p>In content implemented using markup languages, <a>status messages</a> can be <a>programmatically determined</a> through <a>role</a> or properties such that they can be presented to the user by <a>assistive technologies</a> without receiving focus.</p>
-   				
+   <p>In technologies that support it, <a>status messages</a> can be <a>programmatically determined</a> without receiving focus, so that they can be presented to the user by <a>assistive technologies</a>.</p>
```
patrickhlauke:
```suggestion
   <p>In technologies that support it, <a>status messages</a> that appear dynamically (without a <a>change of context</a>) can be <a>programmatically determined</a> without receiving focus, so that they can be presented to the user by <a>assistive technologies</a>.</p>
```

guidelines/terms/21/status-message.html
```diff
@@ -1,6 +1,4 @@
 <dt><dfn id="dfn-status-messages" data-lt="status messages">status message</dfn></dt>
 <dd>
-   					
-   <p>change in content that is not a <a>change of context</a>, and that provides information to the user on the success or results of an action, on the waiting state of an application, on the progress of a <a>process</a>, or on the existence of errors</p>
-   
+   <p>discrete piece of content that is added or updated dynamically, and whose only purpose is to provide information to the user on the results of an action, the waiting state of an application, the progress or completion of a <a>process</a>, or the existence of errors</p>
```
patrickhlauke:
```suggestion
   <p>discrete piece of content whose only purpose is to provide information to the user on the results of an action, the waiting state of an application, the progress or completion of a <a>process</a>, or the existence of errors</p>
```

guidelines/sc/21/status-messages.html
patrickhlauke: i think my latest attempt covers this aspect now.

patrickhlauke:
Had a little flash of inspiration after our meeting https://github.com/w3c/wcag/pull/4952/changes/33b5a03c06842136d63b7cf82536adbf8aca34f2

bruce-usab:
@patrickhlauke I think you need either an _i.e._ or _e.g._ right after the opening parenthesis.  At present, “without a change of context” can be read either way (both seem like they work, but I think _i.e._ is what you mean) — or even as a constraint/modifier to “dynamically” (which is an ambiguity  you definitely don’t mean).

I did a quick scan of 2.2 SC and spotted only three that use parentheticals.  They are all variations on providing examples, and you’re essentially providing an synonym.
- [2.1.4](https://www.w3.org/TR/wcag/#character-key-shortcuts)
> …using only letter (including upper- and lower-case letters), punctuation, number, or symbol characters, then…
> …one or more non-printable keyboard keys (e.g., Ctrl, Alt);
- [3.3.9](https://www.w3.org/TR/wcag/#accessible-authentication-enhanced)
> A cognitive function test (such as remembering a password or solving a puzzle) is not required for any step…
- [4.1.2](https://www.w3.org/TR/wcag/#name-role-value)
> For all user interface components (including but not limited to: form elements, links and components generated by scripts), the name and role…

I also see that merely remove the parentheses is problematic.

patrickhlauke:
wondering if it should be even more specific, as "change of context" is a bit overloaded as is. maybe dynamically could just be clarified as "(i.e. without moving to a new page)" or similar...

guidelines/sc/21/status-messages.html
```diff
@@ -1,9 +1,5 @@
 <section id="status-messages" class="sc">
-   					
    <h4>Status Messages</h4>
-   					
    <p class="conformance-level">AA</p>
-   					
-   <p>In content implemented using markup languages, <a>status messages</a> can be <a>programmatically determined</a> through <a>role</a> or properties such that they can be presented to the user by <a>assistive technologies</a> without receiving focus.</p>
-   				
+   <p>In technologies that support it, <a>status messages</a> that appear dynamically (without a <a>change of context</a>) can be <a>programmatically determined</a> without receiving focus, so that they can be presented to the user by <a>assistive technologies</a>.</p>
```
bruce-usab:
```suggestion
   <p>In technologies that support it, <a>status messages</a> that appear dynamically (that is, without a <a>change of context</a>) can be <a>programmatically determined</a> without receiving focus, so that they can be presented to the user by <a>assistive technologies</a>.</p>
```

bruce-usab (changes requested):
I preferred the version without the parenthetical and avoided “change of context”.

guidelines/sc/21/status-messages.html
```diff
@@ -1,9 +1,5 @@
 <section id="status-messages" class="sc">
-   					
    <h4>Status Messages</h4>
-   					
    <p class="conformance-level">AA</p>
-   					
-   <p>In content implemented using markup languages, <a>status messages</a> can be <a>programmatically determined</a> through <a>role</a> or properties such that they can be presented to the user by <a>assistive technologies</a> without receiving focus.</p>
-   				
+   <p>In technologies that support it, <a>status messages</a> that appear dynamically (that is, without a <a>change of context</a>) can be <a>programmatically determined</a> without receiving focus, so that they can be presented to the user by <a>assistive technologies</a>.</p>
```
patrickhlauke:
```suggestion
   <p>In technologies that support it, <a>status messages</a> that appear dynamically on the current page can be <a>programmatically determined</a> without receiving focus, so that they can be presented to the user by <a>assistive technologies</a>.</p>
```

patrickhlauke:
> wondering if it should be even more specific, as "change of context" is a bit overloaded as is. maybe dynamically could just be clarified as "(i.e. without moving to a new page)" or similar...

changed to "dynamically on the current page" which i believe is exactly what we want/mean

joelamyman:
I've been looking in to an issue related to the "change of context" wording, and this update would be an improvement, particularly the use of "appear dynamically". This change makes the wording a lot easier to understand.

bruce-usab (approved):

WilcoFiers (changes requested):
I like this update, but I strongly disagree doing this through an errata. These changes adjust the scope of the success criterion. Changing both in which situations this criterion applies, and what solutions are valid. 

This is not the same success criterion. It requires a different number, and it requires a new version of WCAG so that people can understand that something changed. A new version triggers new translations, ensures people don't conflate the two criteria, encourages everyone to update the copies they have all over the place, etc.

This kind of thing is why I've advocated for an annual release of WCAG, and for a smaller incremental path to WCAG 3.  We have to be transparent about these sorts of things.

patrickhlauke:
@WilcoFiers 

> but I strongly disagree doing this through an errata

currently, within the scope of the WCAG 2 TF, that's the only mechanism we have at our disposal. however, we *are* earmarking proposed PRs that are too "radical" for a potential WCAG 2.3 release. this will be part of the pre-CfC and CfC that @alastc will circulate soon. so they won't go as actual errata, but be merged into 2.3 (*if* that ever happens)

guidelines/sc/21/status-messages.html
```diff
@@ -1,9 +1,5 @@
 <section id="status-messages" class="sc">
-   					
    <h4>Status Messages</h4>
-   					
    <p class="conformance-level">AA</p>
-   					
-   <p>In content implemented using markup languages, <a>status messages</a> can be <a>programmatically determined</a> through <a>role</a> or properties such that they can be presented to the user by <a>assistive technologies</a> without receiving focus.</p>
-   				
+   <p>In technologies that support it, <a>status messages</a> that appear dynamically on the current page can be <a>programmatically determined</a> without receiving focus, so that they can be presented to the user by <a>assistive technologies</a>.</p>
```
baldino-m: can this be re-prhased so that the fact that this also includes text that is dynamically updated, not only added? This should work: "status messages that appear (or are updated) dynamically"

patrickhlauke: probably semantics whether or not "appear" already includes the idea that text was updated (and the change now appears) ... but sure

patrickhlauke:
```suggestion
   <p>In technologies that support it, <a>status messages</a> that appear or update dynamically on the current page can be <a>programmatically determined</a> without receiving focus, so that they can be presented to the user by <a>assistive technologies</a>.</p>
```

awkawk:
This is a major change and beyond the scope of what the WG should handle in an erratum. If there is confusion that can be clarified in the understanding document, we should do that first. Agree with @WilcoFiers on this being problematic.

WilcoFiers:
> for a potential WCAG 2.3 release

I'm all for an incremental path to WCAG 3.0, where we take WCAG 2 criteria, update the ones that urgently need correcting, and release a new version every year, or every other year. Something like that. But the AG chairs have decided otherwise. There is no WCAG 2.3 in the next charter. Any time you spend on it is time not spent on things that actually can lead to results. Instead of writing this, I could have reviewed another WCAG 3 outcome too. 

As for republishing WCAG 2.2, that's just as problematic. DoJ recently used [WCAG 2's republications and changing techniques](https://www.federalregister.gov/documents/2026/04/20/2026-07663/extension-of-compliance-dates-for-nondiscrimination-on-the-basis-of-disability-accessibility-of-web#citation-57-p20906) as a reason for delaying Title II.  Not the main reason. And the techniques argument is obviously flawed, every technique in its header says it is not normative and not required for conformance. But it goes to show that to legislators, the stability of WCAG 2 matters.

The point they make about linking to a dated version is very valid. The dated version they use now shows an angry red "You are on an outdated version!" modal. They trusted W3C to keep their versioned document stable, but instead we are now shouting at anyone who wants to use the version Title II requires them to use, to use a different version. There isn't even an obvious way to see what has changed, so that people can quickly understand that the version W3C is aggressively directing you to is essentially the same as the one you wanted to opened.

patrickhlauke:
@WilcoFiers can i suggest that this meta-discussion happens elsewhere, NOT on this PR. take it up with the chairs and the wider AGWG.

bruce-usab:
@awkawk how is this a “major change” in the context of W3C [Classes of Changes](https://www.w3.org/policies/process/#correction-classes)?

Do you disagree that anything less than **4. New features** is within the scope of the charter for 2.x work?

This PR (from Class 2, emphasis added):
> …constitutes changes that do not affect conformance, i.e. changes that **reasonable implementers would not interpret as changing architectural or interoperability requirements or their implementation.**

That section continues (again, emphasis added):

> ***Changes which resolve ambiguities in the specification*** are considered to change (by clarification) the implementation requirements and do not fall into this class. …If there is any doubt or disagreement as to whether a change functionally affects interpretation, that change does not fall into this class.

So that puts this PR in **Class 3 Other changes _that do not add new features_** and this PR is more than editorial but is still erratum in the scope of what the WG could (and should) handle.

[updated my comment to correct markdown link syntax]

bruce-usab:
> The point they make about linking to a dated version is very valid.

I respectfully disagree.  They are merely doubling-down on reading comprehension errors from the rule.  DOJ deliberately cited to a version outdated at the time they published the rule, and gave a nonsensical explanation for that choice.

> The dated version they use now shows an angry red "You are on an outdated version!" modal. They trusted W3C to keep their versioned document stable, but instead we are now shouting at anyone who wants to use the version Title II requires them to use, to use a different version.

Except that we know it’s _not_ a different _version_ but merely a reprinting which incorporates errata.  What does it mean to cite to an edition that doesn’t include errata published at the time of the citation?

> There isn't even an obvious way to see what has changed, so that people can quickly understand that the version W3C is aggressively directing you to is essentially the same as the one you wanted to opened.

That's a fair concern, but it’s a problem DOJ created.  We could help fix that by including a preview diff link in the outdated version modal.  It’s not a reason to stop work on errata.

awkawk:
> @awkawk how is this a “major change” in the context of W3C [Classes of Changes(https://www.w3.org/policies/process/#correction-classes)?
> 
> Do you disagree that anything less than 4. New features is within the scope of the charter for 2.x work?

Confusing question. Please restate.
 
> This PR (from Class 2, emphasis added):
> …constitutes changes that do not affect conformance, i.e. changes that reasonable implementers would not interpret as changing architectural or interoperability requirements or their implementation.

This PR is Class 3 in my opinion. The PR states that it is unclear what a status message is and the correction would "functionally affect interpretation of the document".

> That section continues (again, emphasis added):
> 
> Changes which resolve ambiguities in the specification are considered to change (by clarification) the implementation requirements and do not fall into this class. …If there is any doubt or disagreement as to whether a change functionally affects interpretation, that change does not fall into this class.

Good, we agree!

> So that puts this PR in Class 3 Other changes that do not add new features and this PR is more than editorial but is still erratum in the scope of what the WG could (and should) handle.

Sure, there is a process for handling non-editorial errata, but it is up to the WG and the W3C as to whether to follow that path. Historically and for the purposes of stability of the specification, harmonization of references which rely on dated versions, and all of the external tools, training, and resources that exist, the WG has chosen to not take this path.

That is why I have advocated for clarifications in techniques and understanding documents for interpretations of WCAG, and I think it makes sense here also.

awkawk:
> Except that we know it’s not a different version but merely a reprinting which incorporates errata. What does it mean to cite to an edition that doesn’t include errata published at the time of the citation?

This is why we have https://www.w3.org/WAI/WCAG22/errata/. Editorial errata are generally regarded as part of the original date-based standard as they do not change the meaning or interpretation of the standard.

patrickhlauke:
> That is why I have advocated for clarifications in techniques and understanding documents for interpretations of WCAG

just on that point ... funnily enough, the current understanding for 4.1.3 does already try to explain what a status message is along the rough lines that this PR does ... just that this PR wants to make that clear in the actual normative definition and SC, rather than in the understanding. and then there's the usual tension of "at which point is an understanding document that tries to get authors to read the normative part a certain way actually just making/changing normative requirements by stealth?". and having an understanding that says "you know where normatively we say X? yeah, well, pretend we said Y, ok?" isn't... great?

WilcoFiers:
> @WilcoFiers can i suggest that this meta-discussion happens elsewhere, NOT on this PR. take it up with the chairs and the wider AGWG.

Yes please.

> That's a fair concern, but it’s a problem DOJ created. We could help fix that by including a preview diff link in the outdated version modal. It’s not a reason to stop work on errata.

I disagree. I don't like erratas. I like republishing WCAG with the errata's added onto it even less, and when it comes to class 3 changes that's an automatic -1 from me.

These documents underpin legal requirements throughout the world. We do not have the authority to change legal requirements. It worries me that we're somehow in a position to do so, because legislators linked to w3.org rather than keep a copy of it. We shouldn't do that, even if we think the change is insignificant or an improvement. It's messy at best, and we ultimately aren't the ones who need to decide whether orgs are in compliance or not. Others may disagree that those are acceptable changes to make.

I'm all in favor of regular publications of WCAG. I think annually would be fine. I personally prefer evergreen. I think that's something that if we plan it well, likely we can figure out a way to do it. There are lots of standards that are legal requirements that get updated regularly. Car safety standards for one. We can figure this out. But we have to acknowledge that /tr/wcag22/ is linked from a EN 301 549. If we want changes to it, we have to republish that version to a new URL, and with a clearly distinct title.

Issue #1779, opened 2021-05-04
example sentence for 3.2.5: Change on Request already covered by 3.2.2: On Input


jake-abma:
https://www.w3.org/WAI/WCAG22/Understanding/change-on-request

3.2.5: Change on Request mentions:

> This Success Criterion aims to eliminate potential confusion that may be caused by unexpected changes of context such as automatic launching of new windows, automatic submission of forms after selecting an item from a list, etcetera.

Now the second example given:

> automatic submission of forms after selecting an item from a list

Is already covered by 3.2.2: On Input, not?

https://www.w3.org/WAI/WCAG22/Understanding/on-input

So as an example to use as a reason for this extra SC while already covered seems to loose weight.

jake-abma:
3.2.5 is not even used as an example at the before / after demo:

Nr. 3

https://www.w3.org/WAI/demos/bad/before/annotated/template.html

> `<select onchange="...">`
> Note: Errors on the template are propagated throughout the entire website.
> 
> Success Criterion 2.4.1 - Keyboard
> Failure 54: Failure of Success Criterion 2.1.1 due to using only pointing-device-specific event handlers (including gesture) for a function
> 
> Success Criterion 3.2.2 - On Input
> Failure 36: Failure of Success Criterion 3.2.2 due to automatically submitting a form and presenting new content without prior warning when the last field in the form is given a value

mraccess77:
On a related note - 3.2.1 and 3.2.2 do not officially cover the situation where a dialog randomly pops up on page shifting focus as they are very scoped to on focus and on input.  However, all most everyone would flag this as a violation of SC 3.2.1 although technically it falls under SC 3.2.5.  This has been a long standing messy relationship between the two.  Also new windows that open on activation but without wording is crammed into SC 3.2.5 as well complicating things - some people continue to flag this as an issue under SC 3.2.1 which is not correct as indicating that links open new windows is not covered in Level A or AA.

jake-abma:
@mraccess77 yeh, and change on input does not cover inputs, but also placed under 3.2.5, I think this is not correct and must be part of 3.2.2, also from a user perspective (not being AAA) but also from a 'doable perspective'

alastc:
So the AAA criterion is more expansive than the AA one, I'm not seeing an issue.

jake-abma:
[...]

PR #4955, opened 2026-02-25
Remove G200, tweak G201 to apply to links and buttons, fix association of the technique with appropriate SCs

### LINKED ISSUE #4043: G200: why is 3.2.1 On Focus referenced?
[Technique G200](https://www.w3.org/WAI/WCAG21/Techniques/general/G200) references 3.2.1 On Focus but makes no mention or reference to causing a context change on focus. Should 3.2.1 be removed from this technique?

### LINKED ISSUE #4197: G201: referenced in 3.2.1 On focus and 3.2.2 On input but does it really make sense?
Hi,

I do not understand why [G201](https://www.w3.org/WAI/WCAG22/Techniques/general/G201.html) is referenced in [3.2.1 On focus](https://www.w3.org/WAI/WCAG22/Understanding/on-focus) and [3.2.2 On input](https://www.w3.org/WAI/WCAG22/Understanding/on-input).

- G201 is only about links opening in a new window.
- 3.2.1 is about user interface component receiving focus. (I saw #4043 where it was already discussed with G200.)
- 3.2.2 is about forms.

Does it make sense in any way?

giacomo-petri:
Closes #4043
Closes #4197

EDIT: (@patrickhlauke : having discussed this with @giacomo-petri separately, went ahead to make some more fundamental tweaks here)

* deletes G200 - this "technique" isn't really a technique as such ... it just gives examples of where opening a new window can be considered ok? seems pointless
* removes G201 as advisory technique for 3.2.1 On Focus and 3.2.2 On Input ... as it has nothing to do with either of them; instead it adds it as *advisory* technique for 3.2.5 Change on Request, and *advisory* for 2.4.4 Link Purpose (in Context), 2.4.6 Headings and Labels, and 2.4.9 Link Purpose (Link Only)
* tweaks G201 to make it applicable to both links and buttons
* tweaks 3.2.5 Change on Request understanding slightly for consistency/clarity - and renames the folder for the working example, as it's not *just* a css technique shown there

Making G201 *advisory* for 2.4.4/2.4.6/2.4.9 hopefully also signals how it's a nice to have for those SCs, but not required (hopefully?), which answers another perennial question i see bubbling up ("there's a link that opens a new window, can i fail it because it doesn't have an icon/text that says it opens in a new window?")

- [Preview](https://deploy-preview-4955--wcag2.netlify.app/techniques/general/g201)
- [Diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FTechniques%2Fgeneral%2FG201&doc2=https%3A%2F%2Fdeploy-preview-4955--wcag2.netlify.app%2Ftechniques%2Fgeneral%2Fg201)


giacomo-petri:
Hey @kfranqueiro, I'm not allowed to access the logs. Can you please double-check why the preview cannot be deployed?

Thanks

techniques/general/G200.html
patrickhlauke:
```suggestion
      <p>Pages that open new windows and tabs</p>
```

patrickhlauke:
```suggestion
      <p>The objective of this technique is to limit the use of links or buttons that open new windows or tabs within web content. In general, it is better not to open new windows and tabs since they can be disorienting for people, especially people who have difficulty perceiving visual content (and who may therefore not realise that a new window or tab was opened). However there are some situations where it is preferable from an accessibility perspective to open a new window or tab. Here are two such situations:</p>
```

patrickhlauke:
```suggestion
<!DOCTYPE html><html lang="en"><head><title>Opening new windows and tabs from a link or button only when necessary</title><link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove"></head><body><h1>Opening new windows and tabs from a link or button only when necessary</h1><section class="meta"><p class="id">ID: G200</p><p class="technology">Technology: general</p><p class="type">Type: Technique</p></section><section id="applicability"><h2>When to Use</h2>
```

patrickhlauke:
```suggestion
         <li>Opening a page containing context-sensitive information, such as help instructions, will significantly disrupt a multi-step workflow, such as filling in and submitting a form, if the page is opened in the same window or tab.</li>
```

patrickhlauke:
```suggestion
         <li>The user is logged into a secured area of a site, and following a link to a page outside of the secured area would terminate the user's login session. In this case, opening external links in an external window allows the user to access such references while keeping their login active in the original window.</li>
```

patrickhlauke: have to admit though this scenario seems fairly artificial ... maybe back in the dark ages it was the case, but these days your login/session isn't terminated if you navigate away. heck, i could even close the browser altogether and reopen it, and the session's likely still ok.

patrickhlauke:
```suggestion
            <p>An online form provides extensive context-sensitive help for each form field on a separate page because there is too much text to include within the form. The links to the context-sensitive help open in new windows or tabs to prevent the loss of any form data that has already been entered.</p>
```

patrickhlauke: just removing odd double spaces

patrickhlauke:
```suggestion
            <p>A page on a secure website includes a link to an external page that is outside of the secure session. The link opens in a new window or tab, since opening the link in the same window will end the secure session.</p>
```

patrickhlauke: again, see comment above about how this seems contrived nowadays

techniques/general/G200.html
```diff
@@ -21,9 +21,9 @@ <h3>Secure website</h3>
          
       </section>
       <section class="example">
-         <h3>Date Picker</h3>
+         <h3>Document link</h3>
          
-            <p>An online form includes a date field that allows the user to  manually type in the date or select a date from a calendar-based date  picker on a separate page. The link to the calendar-based date picker  opens in a new window or tab to prevent the loss of any form data that  has already been entered.</p>
+            <p>A page presenting substantial content includes a link to a supporting document provided in another format (PDF). The document opens in a new tab to allow users to consult the additional material while preserving their current context and progress on the originating page.</p>
```
patrickhlauke: maybe could even generalise this a bit further/make it even more real: the original page is a big list of documents available for viewing/download in different formats, including PDF. each view/download link opens in a new window or tab, so the user doesn't lose their place in the original page that lists all the documents

techniques/general/G200.html
```diff
@@ -37,4 +37,4 @@ <h3>Date Picker</h3>
               <a href="https://www.nngroup.com/articles/the-top-ten-web-design-mistakes-of-1999/">Top-10 New Mistakes of Web Design</a>
```
patrickhlauke: party like it's 1999...

techniques/general/G200.html
giacomo-petri: Should we get rid of it completely? I'm on board with that

kfranqueiro:
> Hey @kfranqueiro, I'm not allowed to access the logs. Can you please double-check why the preview cannot be deployed?
> 
> Thanks

Looks like the spec-generator invocations failed, and looks like the reason why _might_ be a new front-loaded "are you sure?" step that GitHack added. I might need to do something about that separately. This doesn't mean there's anything wrong with the PR.

FYI, I recently added the "Build PR / Run build" check which also performs the build of the informative docs, so if you see that passing, you should be safe. (Doesn't resolve that there's currently no preview though.)

techniques/general/G200.html
```diff
@@ -1,17 +1,17 @@
...
+      <p>Pages that open new windows and tabs</p>
    </section><section id="description"><h2>Description</h2>
-      <p>The objective of this technique is to limit the use of links or  buttons that open new windows or tabs within web content. In general, it is better not to open new windows and tabs since they can be disorienting for people, especially people who have difficulty  perceiving visual content. However there are some situations where it is preferable from an accessibility perspective to open a new window or tab. Here are two such situations:</p>
+      <p>The objective of this technique is to limit the use of links or buttons that open new windows or tabs within web content. In general, it is better not to open new windows and tabs since they can be disorienting for people, especially people who have difficulty perceiving visual content (and who may therefore not realise that a new window or tab was opened). However there are some situations where it is preferable from an accessibility perspective to open a new window or tab. Here are two such situations:</p>
```
patrickhlauke:
```suggestion
      <p>The objective of this technique is to limit the use of links or buttons that open new windows or tabs within web content. In general, it is better not to open new windows and tabs since they can be disorienting for people, especially people who have difficulty perceiving visual content (and who may therefore not realize that a new window or tab was opened). However there are some situations where it is preferable from an accessibility perspective to open a new window or tab. Here are two such situations:</p>
```

techniques/general/G200.html
patrickhlauke: I wouldn't miss it if it was gone...

patrickhlauke: and again... I wouldn't miss it if it was gone...

techniques/general/G201.html
```diff
@@ -17,8 +17,11 @@ <h2>When to Use</h2>
 		</section>
 		<section id="description">
 			<h2>Description</h2>
-			<p>The objective of this technique is to provide a warning before automatically
-            opening a new window or tab. Opening new windows automatically when
+			<p>The objective of this technique is to provide a warning before opening a new 
+            window or tab when a link is activated by a user's direct action (for example, 
+            a left-click or pressing Enter when focused, rather than through a browser or 
+            context menu option that allows the user to open the link in a new window).
```
patrickhlauke:
```suggestion
			<p>The objective of this technique is to provide advanced warning for links that opening a new 
            window or tab when activated by the user. This only applies to "regular" activation, such as 
            a left-click on a mouse, or pressing <kbd>Enter</kbd> when the link has focus – actions that normally
            would open the link target in the current tab or window; the technique does not apply to situations
            where the user explicitly chooses to open the link in a new tab or window – for instance, using the
            browser's context menu option, or middle-clicking on a mouse).</p>
```

techniques/general/G201.html
```diff
@@ -17,8 +17,11 @@ <h2>When to Use</h2>
...
+            Opening new windows automatically when
```
patrickhlauke:
```suggestion
            <p>Opening a new tab or window when
```

techniques/general/G201.html
```diff
@@ -17,8 +17,11 @@ <h2>When to Use</h2>
...
             a link is activated can be disorienting for people who have difficulty
             perceiving visual content, and for some people with cognitive disabilities,
             if they are not warned in advance. Providing a warning allows the user
```
patrickhlauke:
```suggestion
            if they are not warned in advance. Providing a warning or indication that a link
            will open a new tab or window allows the user
```

techniques/general/G201.html
```diff
@@ -27,6 +30,12 @@ <h2>Description</h2>
             to go to the new window. It will help them understand that the "back" button
             will not work and that they have to return to the last window they
             had open, in order to find their previous location.</p>
```
patrickhlauke:
```
to decide it they want to leave the current tab or window. It will also help
them understand that, once they follow the link, their browser's "back" button
will not work and that they have to return to the last tab or window they
had open, in order to find their previous location.</p>
```

patrickhlauke: urgh, github isn't allowing me to do this as a multiline suggestion ... will have to dive into vscode in a bit and make a proper commit

giacomo-petri:
@patrickhlauke, it looks like the following section may have been accidentally removed:

```
<div class="note">
            <p>This technique is advisory. While indicating that a link "opens in a new window"
              is considered a good practice, is not required to meet the requirements of any
              particular success criterion.</p>
          </div>
```

I think restoring it would be helpful, as it clarifies the final note in this issue description.

patrickhlauke:
as the question came up in https://github.com/w3c/wcag/issues/1779 about what the difference is between 3.2.2 On Input and 3.2.5 Change on Request ... for 3.2.2, one way of satisfying the requirements of the SC is to "warn" the user - ` the user has been advised of the behavior before using the component`.

this "if you warn the user, it's not a problem" escape clause is *not* present in 3.2.5 Change on Request, which demands that you must have a mechanism to prevent unexpected changes of context. for this reason, G201 is not *sufficient* for 3.2.5, but only advisory.

(note that the clarification about the potential overlap, but also the difference, between 3.2.5 and 3.2.2 that #1779 raises will be dealt with in a separate PR)

patrickhlauke:
> I think restoring it would be helpful

Done (after I initially thought it was sufficient for 3.2.5, but then realised it's advisory for that one too)

bruce-usab:
Discussed on TF call 27 Feb, keeping in Drafted for now.

understanding/20/change-on-request.html
```diff
@@ -40,97 +37,71 @@ <h2>Intent of Change on Request</h2>
...
       
       <div class="note">
-         
          <p>It is possible for more than one change of context to occur simultaneously. For example,
-            clicking on a link which automatically opens a new window is an example of two separate
+            clicking on a link which opens a new window or tab is an example of two separate
             changes of context related to the change in content and to the change in the viewport
-            (window). The change in the content in this case is initiated by user request when
+            (window or tab). The change in the content in this case is initiated by user request when
             they click on the link, but unless the user can be aware that the link will open in
```
adampage:
```suggestion
            they click on the link, but <em>opening a new window or tab</em> cannot be regarded as user-initiated if it happens without the user explicitly requesting it.
```

understanding/20/change-on-request.html
```diff
@@ -40,97 +37,71 @@ <h2>Intent of Change on Request</h2>
...
-            a new window then that change of context cannot be regarded as user-initiated.
+            a new window or tab, then that change of context cannot be regarded as user-initiated.
```
adampage:
```suggestion
```

Goes with the [above suggestion](https://github.com/w3c/wcag/pull/4955/changes#r2865320710).

bruce-usab:
I am still marginally in favor of fixing  G200 rather than retiring it.  The [OG G200](https://www.w3.org/TR/WCAG20-TECHS/G200.html) mentions it being relevant to a _guideline_ rather than a success criteria.
> This technique relates to: Understanding Guideline 3.2 (Predictable) [Make Web pages appear and operate in predictable ways].

Do we have other techniques which address _only_ a Guideline?  This aspect of G200 (being an outlier) could be part of the obsoleted message.

PR #4960, opened 2026-02-27
Reformat tables out of Understanding 1.4.11

### LINKED ISSUE #4641: At narrow viewports, Understanding SC 1.4.11 has undesirable horizontal scrolling overflow
At narrow viewports approaching `320px`, non-responsive tables on [Understanding Non-Text Contrast](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html#:~:text=Passing%20User%20Interface%20Component%20Examples) scroll off the right edge of the viewport.

<details>
<summary>Animated example</summary>
A looping video capture of the Understanding Non-Text Contrast in a narrow window, demonstrating how some content is wider than the viewport and requires horizontal scrolling to reveal.
</details>

adampage:
Resolves #4641 by reformatting the layout of non-text contrast examples to improve their legibility across different viewport sizes.

Also relocates two especially complicated examples — [the bar chart and pie chart](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html#:~:text=In%20order%20to%20understand,be%20required%20for%20understanding.) —  out of the grid layout so that they can be made larger.

Please compare the [current page](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html) with the [deploy preview](https://deploy-preview-4960--wcag2.netlify.app/understanding/non-text-contrast#user-interface-component-examples).

## Wide viewport

|Before|After|
|-----|-----|
|A desktop screencap of the page before this PR, in which several examples and their supporting descriptions are arranged in a table.|A desktop screencap of the page after this PR, in which several examples and their supporting descriptions are arranged in a layout table roughly roughly comparable to the before, except the images are now on the left instead of the right and there are no column headers.|

## Narrow viewport

|Before|After|
|-----|-----|
|A mobile screencap of the page before this PR, in which several examples and their supporting descriptions are arranged in a table that is cropped by the right edge of the viewport.|A mobile screencap of the page before this PR, in which several examples and their supporting descriptions are arranged in a single, uncropped column.|

## Bar and pie charts

|Before|After|
|-----|-----|
|A desktop screencap of the page before this PR, in which the bar chart and pie chart examples are inside a grid layout, small and difficult to make out.|A desktop screencap of the page after this PR, in which the bar chart and pie chart examples are in a single full-width column, larger and easier to make out.|


patrickhlauke:
i quite like the proposed grid look. much more user-friendly for a visual user

bruce-usab:
Discuss on TF call 27 Feb, looking very promising.  Feedback provided to @adampage and PR moved back to In Progress.

adampage:
Heya @patrickhlauke / @kfranqueiro / @bruce-usab, I’ve made a couple follow-up tweaks — please [have a look](https://deploy-preview-4960--wcag2.netlify.app/understanding/non-text-contrast)?

fstrr:
Instead of a heading that says "Passing User Interface Component Examples" and then one that says "Examples with Insufficient Contrast", could the second heading say "Failing User Interface Component Examples" (which is what the current page has)?

adampage:
Oof, yes, this was my mistake, thanks for catching that, @fstrr.

While restructuring the markup, I re-typed that heading from memory, or so I thought. 🤦🏻 I’ve [restored the original text](https://github.com/w3c/wcag/pull/4960/changes/ca0ad045063d9ba47226cafc434c2ebd7e346d51).

kfranqueiro:
Some comments that came up in discussion on April 24 backlog call:

- Bottom of grid may need extra margin or divider to distinguish the following content
- Grid may accidentally seem to convey relation between horizontally-arranged items, or may be harder to read due to across-then-down orientation
- May run into sub-optimal distinctions between existing WAI heading level styles, though the PR appreciably avoids overriding WAI styles
- The grid ends up seeming inconsistent with figures presented earlier in the page, which are purely linear and also have intermediate paragraphs; it might be feasible to apply similar treatment for these as well

bruce-usab:
So much better than current version!  The grid layout invites side-by-side comparison that isn’t intended.

adampage:
Heya reviewers, here’s a follow-up attempt after the [feedback from yesterday’s 2.x TF meeting](https://github.com/w3c/wcag/wiki/2026%E2%80%9305%E2%80%9308#in-progress).

Each figure collection will now be presented as a layout table in sufficiently wide viewports. A couple key differences compared to the current version in production:

- No more table semantics, and therefore no more column headers. Neither seem necessary to me for this content.
- Instead of a width-greedy 3-column table, I went with a 2-column layout by combing the heading and caption into the cells of 1 column, allowing more room for the image.
- I’m now displaying the image at the _start_ of the row rather than the _end_. I.e., lead with the visual aid.
  - However, I made sure the heading still precedes the image in DOM reading order.

Please check out the [updated before and after screencaps](https://github.com/w3c/wcag/pull/4960#issue-3998554129) and the [deploy preview](https://deploy-preview-4960--wcag2.netlify.app/understanding/non-text-contrast#user-interface-component-examples).

cc: @patrickhlauke  / @kfranqueiro / @giacomo-petri / @bruce-usab

fstrr:
Do you want to convert the text in the SVGs to paths so that they look consistent if the user doesn't have Helvetica installed?

kfranqueiro:
> Do you want to convert the text in the SVGs to paths so that they look consistent if the user doesn't have Helvetica installed?

Is there a critical reason to do that in this case? I don't think that's been done with prior SVGs in this repo (edit: apparently it had at some point, sorry I missed the memo on that), and I would suspect that would largely nullify the benefits of using SVGs over PNGs from the perspective of efficiently storing history.

fstrr:
Not a critical reason, but more of a presentational one. Patrick edited one of my SVGs recently for this reason, so I thought I'd suggest it here.

patrickhlauke:
yeah, i usually do it for my SVGs, mainly because SVGs don't handle different fonts well in most cases (things not being centered anymore, things sometimes overlapping, etc as they still deal with absolute positions on things)

the benefit is still much cleaner adaptation/resizing without things either looking pixelated, or requiring monstrously large pngs

adampage:
> Do you want to convert the text in the SVGs to paths so that they look consistent if the user doesn't have Helvetica installed?

I do usually convert `<text>`s to outlines, but made an exception here since the specimens are so simple; I erred on the side of slightly smaller payloads and more readable SVG markup.

However, I should have caught the hardcoded `font-family="Helvetica"` back in #4573 where I first introduced these SVGs. 🤦🏻  I’ve updated to a more resilient `font-family="sans-serif"` just now in c22d265.

If the group would prefer to convert those to paths, I’d be glad to take care of that in a separate PR.

kfranqueiro:
I suspect that at least in the first 3 examples, converting the text strings to paths may be warranted, as I'm noticing that the underline / focus rectangle / centering is off, respectively when I view it in Linux:

Three SVGs. In the first two, the underline runs out before the text does. In the second, the depicted focus rectangle overlaps the last 't' in 'Text'. In the third, the text is clearly off-center.

I expect this would also be a minor issue for the examples containing "Menu" and "Some link text". But I suppose if we change some of them, then we might as well change all of them, in order for them to match.

What bugs me is, will this get to a point where storing a PNG at sufficiently high resolution would actively take up less space/bandwidth?

adampage:
> I suspect that at least in the first 3 examples, converting the text strings to paths may be warranted, as I'm noticing that the underline / focus rectangle / centering is off, respectively when I view it in Linux:

@kfranqueiro — shoot, I agree. 😐  Thanks for doing Linux recon. I’ll go ahead and push up a separate PR to de-`<text>` those SVGs. cc: @patrickhlauke / @fstrr 

> What bugs me is, will this get to a point where storing a PNG at sufficiently high resolution would actively take up less space/bandwidth?

I’m 99% confident that rasterized PNG could _never_ be smaller than SVG for the simplistic artwork that we’re dealing with in these examples. One other benefit of inlining these SVGs directly in the document is that it reduces the number of individual network requests for the page.

adampage:
Just pushed up w3c/wcag#5122 to take care of [those `<text>` elements](https://github.com/w3c/wcag/pull/4960#issuecomment-4454836288).

Also, in [e0f9a72](https://github.com/w3c/wcag/pull/4960/commits/e0f9a7212a8adb058a658379623a389decff6004) just now, I reverted my small update to the SVGs in this PR to avoid merge conflicts later.

bruce-usab (approved):
In particular, thanks for the treatment of the line graph and bar chart!

giacomo-petri (approved):

hidde:
Woah, great work @adampage !

gundulaniemann:
I definitely appreciate the improvements for the narrow view and the responsiveness. Yet I miss the row headers in the wide view in the new version. As they are present in the narrow view, users who need good structure are now disadvantaged in the wide view.
They might loose orientation (at least I do).
is there a way to get the first column (the row headers) back again in the wide view?

adampage:
> Yet I miss the row headers in the wide view in the new version.

Thank you, @gundulaniemann.

We did discuss this in the task force and agreed that changing the semantics of this content from a data table to a pattern of subheading + figure + caption would be a structural improvement. I fear that reintroducing that third column would more strongly suggest that we should also return to data table semantics.

But I’m sensitive to the disorientation you described. Do you think it would be an improvement to make the styling of those subheadings a bit more prominent? For example:

|Before|After|
|-----|-----|
|A partial screencap of the desktop view in which each example’s subheading is only slightly larger than the paired body text.|A partial screencap of the desktop view in which each example’s subheading is considerably larger than the paired body text.|

Another thought would be to introduce a bit more visual separation between the example cases. The [Current Members & Testimonials page on w3.org](https://www.w3.org/membership/list/) has a a very similar content pattern pairing a logo on the left with details on the right, but the separation between each pair is more pronounced than the styles in this PR. Do you think it would be an improvement to imitate that here?

adampage:
Heya @patrickhlauke, I’ve [bumped the size on those subheadings](https://github.com/w3c/wcag/pull/4960#issuecomment-4661785779) and resolved the merge conflict.

The conflict was related to [this bit](https://github.com/w3c/wcag/pull/5064/changes#diff-365e883aab3c4ffb86e62f1de22de162e3dc48ce1c181d2f790705b56a3bec17R306-R313) in #5064, so I’d be grateful if you’d double-check [these two examples](https://deploy-preview-4960--wcag2.netlify.app/understanding/non-text-contrast#user-interface-component-examples:~:text=Text%20input%20%2D%20low,pass%20this%20criterion.) to make sure that I preserved your earlier content changes.

patrickhlauke:
@adampage thank you, i checked, looks good to me.

PR #4995, opened 2026-03-11
Clarifications to labels or instructions SC

### LINKED ISSUE #4977: Consider providing examples of when controls _don't_ need labels - 3.3.2 labels or instructions
There are very common patterns where form controls are provided, but _not_ in the context of a traditional form - which have led some people to believe that 3.3.2 is not applicable.  

i'm a bit on the fence about that pov... but i have a hard time arguing that all controls need visible labels when they're used in contexts where the intent of the form control is rather 'clear'

for instance, a rich text editor where within a toolbar controls are present to choose the font size or family.  Those are commonly presented without visible labels.  [here's an example of one such editor](https://tinymce.github.io/tinymce-demos/tinymce-basics/custom-toolbar-button-menu-and-context-toolbar.html).

The understanding doc doesn't really make the line in the sand clear, imo.  so would there be any opposition to try and help clarify some of this nuance so that people dont' have very loose or draconian takes on this SC?

scottaohara:
Closes #4977

This PR attempts to clarify expectations around when labels, instructions or both might be necessary for form fields / controls.
Key changes:
- expand the concepts introduced in the first paragraph of the current doc into multiple paragraphs in this revision
- try to clarify that the example of radios/checkboxes labels and the labels for options within a select-only combobox are similar in that they are about selecting a choice - but that there can still be a need for providing a `select` a visible label (as I've encountered those that feel that the wording of the current example is indicating that only the options of a select, and not the select itself, require labels.  Selects can require labels as well... and this update tries to emphasize that, while also acknowledging that under certain cases, it may not be necessary
- per the above, adds a new example to clarify those use cases
- adds an acknowledgement that one may pass 3.3.2, but could fail 2.5.3 if the label provided doesn't meet the requirements of label in name.

Preview: https://deploy-preview-4995--wcag2.netlify.app/understanding/labels-or-instructions#intent
Diff: https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FUnderstanding%2Flabels-or-instructions.html%23intent&doc2=https%3A%2F%2Fdeploy-preview-4995--wcag2.netlify.app%2Funderstanding%2Flabels-or-instructions%23intent#intent


understanding/20/labels-or-instructions.html
```diff
@@ -22,27 +22,55 @@ <h2>In brief</h2>
...
-         if they are out of the customary formats or if there are specific rules for correct
-         input. Content authors may also choose to make such instructions available to users
-         only when the individual control has focus especially when instructions are long and
-         verbose.
+         or labels to identify the purpose of fields and controls that expect user input, so that  
+         users will know what sort of values to enter, or adjust. Commonly, labels or instructions 
+         will be necessary to help users understand how to fill out the input fields of a form. 
+      </p>
+      <p>
+         Despite their moniker, form fields and controls are not limited to use within a form. They
```
detlevhfischer:
```suggestion
         Form fields and controls are not limited to use within a form. They
```

understanding/20/labels-or-instructions.html
```diff
@@ -22,27 +22,55 @@ <h2>In brief</h2>
...
+         understand their purpose and the expected values they will need to enter or adjust to
+         use the controls efficiently while mitigating errors.
+      </p>
+      <p>
+         In the case of radio buttons, checkboxes, comboboxes, or similar controls that provide
+         users with choices, each choice will need a label so that users know what they are actually
+         selecting. In some cases, these labels may be enough to indicate the purpose of the parent 
+         control (such as a select-only combobox), or grouping of controls (like groupings of radios
+         or checkboxes), and thus further visible labels may not be necessary. However, in cases
+         where the labels for the choices would be ambiguous without an overarching label or instructions
```
detlevhfischer:
```suggestion
         where the labels for the choices would be ambiguous without a group label or instructions 
```

understanding/20/labels-or-instructions.html
```diff
@@ -22,27 +22,55 @@ <h2>In brief</h2>
...
+         use the controls efficiently while mitigating errors.
+      </p>
+      <p>
+         In the case of radio buttons, checkboxes, comboboxes, or similar controls that provide
+         users with choices, each choice will need a label so that users know what they are actually
+         selecting. In some cases, these labels may be enough to indicate the purpose of the parent 
+         control (such as a select-only combobox), or grouping of controls (like groupings of radios
+         or checkboxes), and thus further visible labels may not be necessary. However, in cases
+         where the labels for the choices would be ambiguous without an overarching label or instructions 
+         for a control or group of controls, then authors would need to provide one or both for users.
```
detlevhfischer: It may not be clear enough for some readers what "one or both" refers back to - possibly better spell out?

Above I have replaced "moniker" and "overarching" since I expect these terms may be unfamiliar.

understanding/20/labels-or-instructions.html
```diff
@@ -22,27 +22,55 @@ <h2>In brief</h2>
...
+      </p>
+
+      <p>
+         Beyond identifying the purpose of a field or control, instructions or labels may specify the 
+         necessary format(s) for data entry, especially when if they are out of the customary formats
```
detlevhfischer:
```suggestion
         necessary format(s) for data entry, especially when they deviate from the usual formats 
```

understanding/20/labels-or-instructions.html
```diff
@@ -22,27 +22,55 @@ <h2>In brief</h2>
...
+         control (such as a select-only combobox), or grouping of controls (like groupings of radios
+         or checkboxes), and thus further visible labels may not be necessary. However, in cases
+         where the labels for the choices would be ambiguous without an overarching label or instructions 
+         for a control or group of controls, then authors would need to provide one or both for users.
+      </p>
+
+      <p>
+         Beyond identifying the purpose of a field or control, instructions or labels may specify the 
+         necessary format(s) for data entry, especially when if they are out of the customary formats 
+         or if there are specific rules for correct input. Content authors may also choose to make such
```
detlevhfischer:
```suggestion
         or when there are specific rules for correct input. Content authors may also choose to make such 
```

understanding/20/labels-or-instructions.html
```diff
@@ -22,27 +22,55 @@ <h2>In brief</h2>
...
+         where the labels for the choices would be ambiguous without an overarching label or instructions 
+         for a control or group of controls, then authors would need to provide one or both for users.
+      </p>
+
+      <p>
+         Beyond identifying the purpose of a field or control, instructions or labels may specify the 
+         necessary format(s) for data entry, especially when if they are out of the customary formats 
+         or if there are specific rules for correct input. Content authors may also choose to make such 
+         instructions available to users only when the individual control has focus, or may provide a 
+         mechanism to dynamically render or navigate a user to the instructions, especially for when
```
detlevhfischer:
```suggestion
         mechanism to dynamically display these instructions, especially when 
```

understanding/20/labels-or-instructions.html
```diff
@@ -22,27 +22,55 @@ <h2>In brief</h2>
...
+         for a control or group of controls, then authors would need to provide one or both for users.
+      </p>
+
+      <p>
+         Beyond identifying the purpose of a field or control, instructions or labels may specify the 
+         necessary format(s) for data entry, especially when if they are out of the customary formats 
+         or if there are specific rules for correct input. Content authors may also choose to make such 
+         instructions available to users only when the individual control has focus, or may provide a 
+         mechanism to dynamically render or navigate a user to the instructions, especially for when 
+         instructions are long and verbose.
```
detlevhfischer:
```suggestion
         they are long and verbose.
```

understanding/20/labels-or-instructions.html
```diff
@@ -22,27 +22,55 @@ <h2>In brief</h2>
...
       </p>
       
       <p>The intent of this success criterion is not to clutter the page with unnecessary information
          but to provide important cues and instructions that will benefit people with disabilities.
```
detlevhfischer, line 55: it feels a bit odd to focus on people with disabilities here since not having things like date format etc. spelled out really affects all users. 

One could explain that people with some disabilites may be less able to infer such information from the context (due to cognitive ( memory / retention  issues or due to reduced size of viewport of LV users) so having it in place is more important to them.

understanding/20/labels-or-instructions.html
```diff
@@ -22,27 +22,55 @@ <h2>In brief</h2>
...
-         Too much information or instruction can be just as harmful as too little.
-         The goal is to make certain that enough information is provided for the user to accomplish
-         the task without undue confusion or navigation.
+         Too much information or instruction can be just as harmful as too little. Rather, the goal is to
```
detlevhfischer, line 56: possibly drop this?

understanding/20/labels-or-instructions.html
```diff
@@ -54,14 +82,16 @@ <h2>Intent of Labels or Instructions</h2>
...
          providing an accessible name or description for form controls and inputs has been used — that aspect is
          covered separately by <a href="name-role-value">4.1.2 Name, Role, Value</a>. It is possible
-         for controls and inputs to have an appropriate accessible name or description (e.g. using <code>aria-label="..."</code>)
-         and therefore pass Success Criterion 4.1.2, but to still fail this success criterion (if the labels or instructions
-         aren't presented to all users, not just those using assistive technologies).
+         for controls and inputs to have an appropriate accessible name or description (e.g., using <code>aria-label="..."</code>)
+         and therefore pass Success Criterion 4.1.2 Name, Role, Value, but to still fail this success criterion 
+         (if the labels or instructions ren't presented to all users, not just those using assistive technologies).
+         Additionally, it can be possible to pass this success criterion by providing a visible label, but fail Success Criterion
+         2.5.3 Label in Name, if the perceivable visible text label is not included in the accessible name of the control.
```
detlevhfischer:
```suggestion
         2.5.3 Label in Name if the visible text label is not included in the accessible name of the control.
```

understanding/20/labels-or-instructions.html
```diff
@@ -54,14 +82,16 @@ <h2>Intent of Labels or Instructions</h2>
...
       </p>
       <p>This success criterion does not apply to links or other controls (such as an expand/collapse widget, or similar
          interactive components) that are not associated with data entry.
       </p>
-      <p>While this success criterion requires that controls and inputs have labels or instructions, whether or
+      <p>While this success criterion requires that form fields and controls labels or instructions, whether or
```
detlevhfischer:
```suggestion
      <p>While this success criterion requires that form fields and controls labels or instructions are present, whether or
```

understanding/20/labels-or-instructions.html
```diff
@@ -86,7 +116,11 @@ <h2>Benefits of Labels or Instructions</h2>
             fields) can prevent users from making incomplete or incorrect form submissions, which prevents
             users from having to navigate once more through a page/form in order to fix submission errors.
          </li>
-         
+
+         <li>Providing labels can help users identify the expected name of a control, so that it may be accessed
```
detlevhfischer:
```suggestion
         <li>Providing labels helps users identify the expected accessible name of a control, so that it may be accessed 
```

understanding/20/labels-or-instructions.html
```diff
@@ -120,6 +154,15 @@ <h2>Examples of Labels or Instructions</h2>
             having a single label "Name" (which would appear to leave the second text field unlabelled),
             each field is given an explicit label — "Given Name" and "Family Name".
          </li>
+
+         <li>
+            A rich text editor presents two select-only comboboxes to allow a user to choose the font family and
+            font size of the text they are editing. Each option of the comboboxes are accurately labeled. The
```
detlevhfischer:
```suggestion
            font size of the text they are editing. Each option of the comboboxes is accurately labeled. The 
```

understanding/20/labels-or-instructions.html
```diff
@@ -120,6 +154,15 @@ <h2>Examples of Labels or Instructions</h2>
             having a single label "Name" (which would appear to leave the second text field unlabelled),
             each field is given an explicit label — "Given Name" and "Family Name".
          </li>
+
+         <li>
+            A rich text editor presents two select-only comboboxes to allow a user to choose the font family and
+            font size of the text they are editing. Each option of the comboboxes are accurately labeled. The 
+            comboboxes are given accessible names ("font family" and "font size", respectively). 
+            The context of being part of the rich text editor's toolbar, as well as the clear labels of each option provide the
+            necessary information for a user to determine their purpose, without the need for persistent visible labels
```
detlevhfischer:
```suggestion
            sufficient information for a user to determine the purpose of the select-only comboboxes, without the need for persistent visible labels
```

understanding/20/labels-or-instructions.html
```diff
@@ -120,6 +154,15 @@ <h2>Examples of Labels or Instructions</h2>
             having a single label "Name" (which would appear to leave the second text field unlabelled),
             each field is given an explicit label — "Given Name" and "Family Name".
          </li>
+
+         <li>
+            A rich text editor presents two select-only comboboxes to allow a user to choose the font family and
+            font size of the text they are editing. Each option of the comboboxes are accurately labeled. The 
+            comboboxes are given accessible names ("font family" and "font size", respectively). 
+            The context of being part of the rich text editor's toolbar, as well as the clear labels of each option provide the
```
detlevhfischer:
```suggestion
            The context of being part of the rich text editor's toolbar, as well as the clear labels of each option provide
```

understanding/20/labels-or-instructions.html
```diff
@@ -120,6 +154,15 @@ <h2>Examples of Labels or Instructions</h2>
...
+            necessary information for a user to determine their purpose, without the need for persistent visible labels
+            for the individual comboboxes.
```
detlevhfischer:
```suggestion

```

detlevhfischer (approved):

understanding/20/labels-or-instructions.html
scottaohara: this change wouldn't work though, because it's a group or a control that might need the label.  not just a group.

understanding/20/labels-or-instructions.html
scottaohara: this change would negate the point of indicating someone could provide long instructions in another location (e.g., navigate to a new page).  so i'll make a different change than this suggestion...

understanding/20/labels-or-instructions.html
scottaohara: agreed, don't need to say instructions again since they were just mentioned.

understanding/20/labels-or-instructions.html
scottaohara: this is the existing text and from what i can tell, has been this text since wcag 2.0.

understanding/20/labels-or-instructions.html
scottaohara: this is again long standing text, and i don't think it makes sense to drop it in a PR that is trying to help clarify this point - that it's sometimes unnecessary to add extra text, and doing so can be detrimental to the UX of the UI

understanding/20/labels-or-instructions.html
scottaohara: i tried using those words, and imo it felt very clunky to have to repeat "labels, instructions or both".

i'll think more if this can be reworded further.

understanding/20/labels-or-instructions.html
scottaohara: thanks for this.  accidentally deleted the "have" that was supposed to come before "labels or..."  i'll get that back in there.

understanding/20/labels-or-instructions.html
patrickhlauke: I think' Scott idea here was to ease the reader into the fact that "form fields" can also be outside of a form...acknowledging that it's probably a misnomer/confusing terminology.

Maybe we can retain the idea, while ditching the "Moniker" term...

```suggestion
         Despite colloquially being referred to as "form fields", these inputs and controls are not limited to use within a form. They
```

understanding/20/labels-or-instructions.html
patrickhlauke: i don't mind "overarching", but i'm biased because I use exactly that term all the time myself when talking about legends

understanding/20/labels-or-instructions.html
```diff
@@ -22,27 +22,54 @@ <h2>In brief</h2>
...
+         use the controls efficiently while mitigating errors.
+      </p>
+      <p>
+         In the case of radio buttons, checkboxes, comboboxes, or similar controls that provide
+         users with choices, each choice will need a label so that users know what they are actually
+         selecting. In some cases, these labels may be enough to indicate the purpose of the parent 
+         control (such as a select-only combobox), or grouping of controls (like groupings of radios
+         or checkboxes), and thus further visible labels may not be necessary. However, in cases
+         where the labels for the choices would be ambiguous without an overarching label or instructions 
+         for a control or group of controls, then authors would need to provide one or both for users.
```
```suggestion
         where the labels for the choices are ambiguous on their own,
         authors need to provide an overarching label / legend or instructions for the control or group of controls.
```

understanding/20/labels-or-instructions.html
patrickhlauke: suggestion for turning the sentence around https://github.com/w3c/wcag/pull/4995/changes#r3442705508

understanding/20/labels-or-instructions.html
```diff
@@ -22,27 +22,54 @@ <h2>In brief</h2>
...
+      </p>
+
+      <p>
+         Beyond identifying the purpose of a field or control, instructions or labels may specify the 
+         necessary format(s) for data entry, especially if they are out of the customary formats 
+         or if there are specific rules for correct input. Content authors may also choose to make such 
+         instructions available to users only when the individual control has focus, or they may provide a
```
```suggestion
         or if there are specific rules for correct input. Content authors can choose to make such 
         instructions available to users only when the individual control has focus, or they can provide a 
```

would prefer "can" to "may"

patrickhlauke:
related, just came across this question on slack https://web-a11y.slack.com/archives/C7KTV5CBA/p1786523815294159 where somebody was asking if, for quantity-type controls (number and plus/minus buttons), where it's usually clear from context, but nominally requirements still say you must provide a visual label, if adding a suffix to the number would work 
screenshot of an item in an ecommerce site. the quantity field has a suffix of "st" - the site is norwegian? - to provide a visible label to indicate it's about quantity

understanding/20/labels-or-instructions.html
```diff
@@ -54,14 +81,16 @@ <h2>Intent of Labels or Instructions</h2>
       <p>Further, this success criterion does not take into consideration whether or not alternative methods of
          providing an accessible name or description for form controls and inputs has been used — that aspect is
          covered separately by <a href="name-role-value">4.1.2 Name, Role, Value</a>. It is possible
-         for controls and inputs to have an appropriate accessible name or description (e.g. using <code>aria-label="..."</code>)
-         and therefore pass Success Criterion 4.1.2, but to still fail this success criterion (if the labels or instructions
-         aren't presented to all users, not just those using assistive technologies).
+         for controls and inputs to have an appropriate accessible name or description (e.g., using <code>aria-label="..."</code>)
+         and therefore pass Success Criterion 4.1.2 Name, Role, Value, but to still fail this success criterion 
+         (if the labels or instructions ren't presented to all users, not just those using assistive technologies).
```
detlevhfischer, line 86:
```suggestion
         (because the labels or instructions are not presented to all users, but only to those using assistive technologies).
```

detlevhfischer (approved):

patrickhlauke:
another scenario that seems to come up frequently these days are chat inputs, where users type in their message.

a text input with placeholder 'type message' and a send button

in this particular case, there's at least a placeholder ... which yes, is not a permanent label ... but again, i doubt that somebody goes to a chat window, types their chat message ... then has to step away for a bit, and when they come back, they're puzzled about what the purpose of that input was. but also consider the case where there might be nothing, just an empty text field and a send button.

while not as pervasive, the paper airplane send icon does label the input. also, the actual overall context makes it clear ... you're in a chat dialog/overlay, you have an input there and a button ... whatever could that be?

maybe not as part of this PR, but ... it would be good to provide a few visual examples of cases where there's sufficient external labelling/context for a form field NOT to require an explicit label.

bruce-usab (approved):

understanding/20/labels-or-instructions.html
```diff
@@ -54,14 +81,16 @@ <h2>Intent of Labels or Instructions</h2>
       <p>Further, this success criterion does not take into consideration whether or not alternative methods of
          providing an accessible name or description for form controls and inputs has been used — that aspect is
          covered separately by <a href="name-role-value">4.1.2 Name, Role, Value</a>. It is possible
-         for controls and inputs to have an appropriate accessible name or description (e.g. using <code>aria-label="..."</code>)
-         and therefore pass Success Criterion 4.1.2, but to still fail this success criterion (if the labels or instructions
-         aren't presented to all users, not just those using assistive technologies).
+         for controls and inputs to have an appropriate accessible name or description (e.g., using <code>aria-label="..."</code>)
```
```suggestion
         for controls and inputs to have an appropriate accessible name or description (for example, using <code>aria-label="..."</code>)
```

understanding/20/labels-or-instructions.html
```diff
@@ -132,6 +174,13 @@ <h2>Examples of Labels or Instructions</h2>
             of the three fields. The value of this attribute for the three fields are, respectively,
             "Area Code", "Exchange", and "Number". 
          </li>
+
+        <li>A website presents a form as a series of individual pages (a wizard) rather than on one large page.
+          All pages in the wizard are introduced by a heading. A few pages contain only a single form field. For these pages
+          the form field is not given its own distinct label, as the heading is enough to indicate the purpose and
+          expected value the user is to enter.  E.g., a page with the heading of "Email address" is sufficient in indicating
```
```suggestion
          expected value the user is to enter. For instance, a page with the heading of "Email address" is sufficient in indicating
```

understanding/20/labels-or-instructions.html
```diff
@@ -132,6 +174,13 @@ <h2>Examples of Labels or Instructions</h2>
             of the three fields. The value of this attribute for the three fields are, respectively,
             "Area Code", "Exchange", and "Number". 
          </li>
+
+        <li>A website presents a form as a series of individual pages (a wizard) rather than on one large page.
+          All pages in the wizard are introduced by a heading. A few pages contain only a single form field. For these pages
+          the form field is not given its own distinct label, as the heading is enough to indicate the purpose and
+          expected value the user is to enter. For instance, a page with the heading of "Email address" is sufficient in indicating
+          the purpose of the single form field present on the page.
+        </li>
```
patrickhlauke, line 184: Any chance of adding these two extra common examples that bubble up all the time?

```suggestion
        <li>A shopping site presents users with a list of products; each product has a quantity input field with a plus/minus control. Even though the input itself lacks an explicit "Quantity" label, it is not normatively required, as it's clear from the context what the purpose of the input is.</li>
        <li>A chat interface has a single text input field, followed by a "paper plane in flight" icon button to send the message. It's clear from the context of being in a chat what the purpose of the input is, without requiring a visible "Message" label (though authors should consider adding a placeholder hint when the message field is empty)</li> 
```

PR #4996, opened 2026-03-12
Enable to remove old logs on serverlog.html working-example

### LINKED ISSUE #4993: There is no function to remove server logs on aria-role-log/serverlog.html working-example
I tried `serverlog.html` working-example and confirmed that the log addition are announced by AT correctly, but I was not able to confirm that the removal of the log is not announced, because there is no button to remove and also auto removal due to the log size limitation is not performed.

https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA23
https://www.w3.org/WAI/WCAG22/working-examples/aria-role-log/serverlog.html

It looks that the algorithm to remove logs is commented out, so when we uncomment it and make some fix to work correctly, we'll be able to confirm that logs are correctly removed (and I'm happy to fix it if it's OK).

https://github.com/w3c/wcag/blob/c9c2c61c4d4df8d19544b673a541b95dc9026ce2/working-examples/aria-role-log/serverlog.html

mehm8128:
close https://github.com/w3c/wcag/issues/4993

I implemented the removal of old logs when server logs reach their limit (5), and some refactoring such as removing redundant spacing and line breaks.
However, it looks some browser doesn't support `aria-relevant="removals"`, even if I add `removals` to the `aria-relevant` token list, I was not able to confirm that the removals are announced.
Anyway, I implemented log removals, so this working example now works as expected.

Preview: https://deploy-preview-4996--wcag2.netlify.app/working-examples/aria-role-log/serverlog.html (compare to https://www.w3.org/WAI/WCAG22/working-examples/aria-role-log/serverlog.html)


patrickhlauke:
From testing, it does indeed seem like there are bugs in browser implementations. In particular, in Chrome and Edge on Windows, with NVDA/JAWS, when the oldest node is deleted, the entire content of the log is read out again. Historically, `aria-related` has had poor/flaky implementation - I suspect this is the reason why the technique had the code to delete the oldest entry (and the hidden button to do it), but this was eventually commented out.

I'd suggest NOT making this change (as it's broken in practice in Chrome at least), and instead changing the explanation/prose at the start of the example (and in ARIA23) to remove mention of removals.

patrickhlauke:
Closing, superseded by https://github.com/w3c/wcag/pull/5077

PR #4841, opened 2025-12-30
Changes to 3.1.3 Unusual Words understanding

patrickhlauke:
Based on the discussion in https://github.com/w3c/wcag/issues/3606 this PR clarifies that "context" (and by extension intended target audience) does apply when deciding if a word is used in an unusual/restricted way. ( @alastc seemed to agree with this in https://github.com/w3c/wcag/issues/3606#issuecomment-1860197941 )

In addition, this PR adds a note about neologisms and made-up/invented words.

Closes https://github.com/w3c/wcag/issues/3606 and https://github.com/w3c/wcag/issues/3591

Preview: https://deploy-preview-4841--wcag2.netlify.app/understanding/unusual-words
[Diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fwcag%2Funderstanding%2Funusual-words&doc2=https%3A%2F%2Fdeploy-preview-4841--wcag2.netlify.app%2Funderstanding%2Funusual-words)

PR #5009, opened 2026-03-20
Replace "Single-A" and "Double-A" with "Level A" and "Level AA"

patrickhlauke:
follow-up to https://github.com/w3c/wcag/pull/4841 where we first spotted this sort of antiquated reference


kfranqueiro (approved):

patrickhlauke:
As we discussed this point during our review of #4841 in the WCAG 2 backlog TF (and were almost going to add this to that PR, before deciding to do a separate one for it), I'm adding this already to the "Ready for Approval" column.

Issue #2215, opened 2022-02-07
6. Glossary: ADD Definition for “Intentional Text”


BridgesHelpdesk:
Suggest adding a definition of “intentional text” (new definition in ALL CAPS):
INTENTIONAL TEXT 
1.	TEXT DISPLAYED ON A VIDEO THAT IS PROVIDED FOR THE PURPOSE OF THE VIEWER READING THE TEXT (EXAMPLES INCLUDE DOCUMENT SHARES, GRAPHICS WITH READABLE TEXT, AND HANDWRITTEN MESSAGE VIDEOS) AND
2.	TEXT THAT HAS BEEN ADDED BY THE VIDEO CREATOR AS AN OVERLAY ON THE VIDEO

Myndex:
ALL CAPS FOR LONG PASSAGES makes reading difficult. Let's not do all caps.

PR #5013, opened 2026-03-21
Rework note and example in definition of "image of text"

### LINKED ISSUE #2217: 6. Glossary: EDIT Image of Text Definition
The current definition of “image of text” is quite limiting; it allows “other visual content” to suffice for text accessibility. 
#Image of Text (current)
image of text
text that has been rendered in a non-text form (e.g., an image) in order to achieve a particular visual effect
Note: This does not include text that is part of a picture that contains significant other visual content.
Example: A person's name on a nametag in a photograph.
Instead, this definition needs to ensure that the image of text will include a full text alternative, regardless of whether additional visual content is provided:
#Image of Text (recommended amendment)
(Additions in ALL CAPS)
image of text
text that has been rendered in a non-text form (e.g., an image) in order to achieve a particular visual effect
Note: This does not include text that is part of a picture that contains significant other visual content.
Example: A person's name on a nametag in a photograph.
NOTE: A TEXT ALTERNATIVE MUST BE PROVIDED FOR ALL CONTENT COMPRISING THE IMAGE OF TEXT, REGARDLESS OF THE PRESENCE OR ABSENCE OF NON-TEXT VISUAL CONTENT.

patrickhlauke:
The note is confusing (see https://github.com/w3c/wcag/discussions/5012), as it appears to exclude important text in combination with other visual elements within the same image, which is not the case (though this point, oddly, appears to be in contention?).

The example is confusing because the context here is unclear - if the main important aspect of the photograph is to show the portrait of a person, and they happen to wear a nametag...then that nametag may well be incidental...unless the whole point of the photo was to in fact show/illustrate the person's name by including the nametag? In any case, this example - particularly coming after that note currently, adds to the confusion.

If the note was intended to set up the idea of text being incidental, this should be reworded. And if the example was in reference to the *note*, it should be part of the note.

This PR rewords the note, and grafts the example (with more explanation/context) directly into the note

Closes https://github.com/w3c/wcag/issues/2217
x-ref https://github.com/w3c/wcag/issues/2215


patrickhlauke:
going through https://github.com/w3c/wcag/discussions/5012 there is strong evidence that for the purpose of evaluating Contrast (Minimum) requirements, the vast majority of practitioners have been interpreting the vague note to mean that incidental text was excluded, but NOT text in an image that is still conveying meaning/content. changing/clarifying the note would bring the normative text in line with the reality of how auditors have been interpreting it.

as https://github.com/w3c/wcag/discussions/5012#discussioncomment-16268947 confirms: "notes are not normative" so changing it should be acceptable...and bring this in line with reality of how people have been interpreting it.

If this is not acceptable, then a change for a potential WCAG 2.3 would be to tweak the normative wording of the contrast SC to not only cover "text and images of text" but also text in images that is not incidental (a more fundamental/disruptive change, but a backwards-compatible one)

PR #5020, opened 2026-03-25
Add pointerevents alias to biblio.ts

### LINKED ISSUE #5004: Biblio reference to `[[pointervents]]` in understanding target size (minimum) and (enhanced)
In the "Key terms" sections of the two target size understanding documents

* https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum#key-terms
* https://www.w3.org/WAI/WCAG22/Understanding/target-size-enhanced.html#key-terms

the definition for "pointer input" includes a biblio reference to `[pointervents]`. However, in contrast to the term in the main spec https://www.w3.org/TR/WCAG22/#dfn-pointer-inputs the version in the understanding docs is broken/not clickable.

Trying to build the site locally, this problem is also shown in the build output:

```
./understanding/21/target-size-enhanced.html: Unresolved biblio ref: pointerevents
./understanding/22/target-size-minimum.html: Unresolved biblio ref: pointerevents
```

kfranqueiro:
Fixes #5004

We rely on a local alias map to avoid incurring multiple requests to specref on every build. Last month, pointerevents was updated in specref (https://github.com/tobie/specref/pull/906) to be an alias, which means our map requires updating as well.

### DIFF
--- a/11ty/biblio.ts
+++ b/11ty/biblio.ts
@@ -12,6 +12,7 @@ export const biblioPattern = /\[\[\??([\w-]+)\]\]/g;
 // Resolve necessary aliases locally rather than requiring an extra round trip to specref
 const aliases = {
   "css3-values": "css-values-3",
+  pointerevents: "pointerevents4",
 };
 
 /** Compiles URLs from local biblio + specref for linking in Understanding documents. */


11ty/biblio.ts
```diff
@@ -12,6 +12,7 @@ export const biblioPattern = /\[\[\??([\w-]+)\]\]/g;
 // Resolve necessary aliases locally rather than requiring an extra round trip to specref
 const aliases = {
   "css3-values": "css-values-3",
+  pointerevents: "pointerevents4",
```
patrickhlauke, line 15: should it point to Level 4 (which is still a working draft) ? Level 3 is still in CR, the latest REC is Level 2 (actually makes me wonder if https://www.w3.org/TR/pointerevents/ shouldn't always just point to the latest REC one? not sure what the policy on that is)

kfranqueiro: The aim of this PR is to match the normative document's behavior. Judging by the response in #5004, it seems like /TR/pointerevents/ pointed to Level 4 since late 2024 where it was intentionally updated.

patrickhlauke: yeah sorry, it's probably a separate question. so go ahead with this as is.

PR #5030, opened 2026-03-28
Clarify Dismissible paragraph in Understanding SC 1.4.13

### LINKED ISSUE #1409: Incomprehensible paragraph in chapter "Dismissable" (Understanding SC 1.4.13)
https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html#dismissable

> For most triggers of relatively small size, it is desirable for both methods to be implemented. 

In my opinion the size of the trigger is not relevant here, but the size of the content that is automatically displayed when the trigger is reached.

> If the trigger is large, noticing the additional content may be of concern if it appears away from the trigger. In those cases, only the second method may be appropriate. 

Again, this is probably more about the size of the displayed content and not about the size of the trigger. But more important: The content can't be displayed far away from the trigger, because otherwise "Hoverable" is not fulfilled.

venkatvemuri28:
## Summary
This PR addresses issue #1409 by clarifying one paragraph in the Dismissible section of Understanding SC 1.4.13 (Content on Hover or Focus).

The previous wording could be read as internally inconsistent because it focused on trigger size without clearly explaining the placement tradeoff.

[HTML Diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fwcag%2Funderstanding%2Fcontent-on-hover-or-focus&doc2=https%3A%2F%2Fdeploy-preview-5030--wcag2.netlify.app%2Funderstanding%2Fcontent-on-hover-or-focus#intent)

## What changed
- Updated one paragraph in `understanding/21/content-on-hover-or-focus.html`
- Clarified why the second method alone may be appropriate in some large-trigger scenarios

## Why
- Better explains the implementation tradeoff

Closes #1409


bruce-usab (commented):
Discussed on TF call 19 June.  Paragraph at line 52 needs a fuller re-write than just tweaking a few words.

The entire dismissible section above that paragraph needs reworking.

patrickhlauke:
Thank you for your effort @venkatvemuri28 but looking closer at the whole "Dismissible" section in the understanding document, the group felt that it needs a much wider/more fundamental reworking, which we'll do separately

PR #5038, opened 2026-04-02
Consistency: use "touchscreen" rather than "touch screen" (normative)

patrickhlauke:
single word is more common in the wild, and used more often in WCAG and its documentation than the two-word variant

noticed as part of work on https://github.com/w3c/wcag/pull/4843

this is the normative companion PR to https://github.com/w3c/wcag/pull/5037


bruce-usab (approved):

awkawk:
"Touch screen" is not even incorrect, so I don't know how we would consider this an editorial (or even important) change to make.

patrickhlauke:
@awkawk for consistency within WCAG itself (see non-normative companion too https://github.com/w3c/wcag/pull/5037) and other W3C specs like https://www.w3.org/TR/pointerevents4/

patrickhlauke:
to put it in raw numbers...uses in WCAG (the spec, understanding, techniques, etc):

* `touch screen` (two words) appears 18 times in 11 files
* `touchscreen` (one word) appears 54 times in 11 files

hidde:
Given the numbers @patrickhlauke presents, it makes a lot of sense to me to make this change (and try and make it in the other places too).

kfranqueiro:

PR #5060, opened 2026-04-10
Remove the swipe example from G219, tweak wording of example single pointer non-drag gestures

### LINKED ISSUE #2404: Last option "swipe gesture" needs to be deleted in G219 Ensuring that a single pointer alternative
In G219 there is an alternative added which is the core of the problem for dragging or pointer gestures... :-)
https://w3c.github.io/wcag/techniques/general/G219

Total text:

> Such dragging movements are difficult or impossible to carry out for some users with motor disabilities. The alternative to dragging movements operates the underlying function by one or several [single pointer](https://w3c.github.io/wcag/guidelines/#dfn-single-pointer) activations. A single tap or click may reveal controls (arrows) to move a target in a stepwise fashion; open a drop-down menu where the drop position can be selected; or allow moving it to an ajacent postion by a swipe gesture.

Last part is:

> or allow moving it to an ajacent postion by a swipe gesture.

The swipe gesture part is a problem for "some users with motor disabilities"...

(also ajacent should be adjacent... but no change needed as all needs to be deleted)

patrickhlauke:
Swipes would satisfy dragging SC, but then still fail pointer gesture SC, so a bad example to give.

Closes https://github.com/w3c/wcag/issues/2404


adampage (approved):

bruce-usab (approved):
[diff preview](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FTechniques%2Fgeneral%2FG219&doc2=https%3A%2F%2Fdeploy-preview-5060--wcag2.netlify.app%2Ftechniques%2Fgeneral%2Fg219)

PR #5065, opened 2026-04-12
Expand 1.4.5 Images of Text benefits

### LINKED ISSUE #4334: 1.4.5 Benefits section lists beneficiaries
The Benefits section of the understanding document for criterion 1.4.5 Images of Text lists the beneficiaries of the criterion instead of its benefits. That is odd, and not in line with most (or maybe all) other criterions, where the actual benefits are explained.

https://www.w3.org/WAI/WCAG22/Understanding/images-of-text.html#benefits

patrickhlauke:
Make the benefits not just about the users, but how these users BENEFIT from content that satisfies the requirement

Closes https://github.com/w3c/wcag/issues/4334


understanding/20/images-of-text.html
```diff
@@ -64,13 +64,13 @@ <h2>Intent of Images of Text</h2>
    <section id="benefits">
       <h2>Benefits of Images of Text</h2>
       <ul>
-         <li>People with low vision (who may have trouble reading the text with the authored font
-            family, size and/or color).
+         <li>People with low vision who may have trouble reading the text with the authored font
+            family, size, and/or color can change all these aspects.
```
bruce-usab:
```suggestion
            family, size, and/or color can change font attributes.
```

bruce-usab (changes requested):
I am not thrilled with “can change all these _aspects_” and made a suggestion.

understanding/20/images-of-text.html
adampage:
```suggestion
            family, size, and/or color can customize all these aspects.
```
Soft editorial suggestion.

bruce-usab (approved):

bruce-usab (approved):

understanding/20/images-of-text.html
```diff
@@ -64,13 +64,13 @@ <h2>Intent of Images of Text</h2>
    <section id="benefits">
       <h2>Benefits of Images of Text</h2>
```
awkawk, line 65: is the heading now reading "Benefits of Images of Text"? The bullets seem confusing when positioned in that way. I get that it is referring to the SC that the Understanding document is about, but perhaps it would be better to preface the list with "When content meets this criterion..."?

patrickhlauke: the headings always get stripped out. and note that i did not touch that in this PR, this is exactly as it's been until now

PR #5072, opened 2026-04-15
Clarify conditions for resize failure in F69

### LINKED ISSUE #5071: F69 is an insufficient Failure technique for Resize Text
[F69: Failure of Success Criterion 1.4.4 when resizing visually rendered text up to 200 percent causes the text, image or controls to be clipped, truncated or obscured](https://www.w3.org/WAI/WCAG22/Techniques/failures/F69) is a problematic failure technique.

There is already an issue open against it, objecting that it does a poor job of defining the "clipped, truncated, or obscured" criteria used in its own test (#2573 ). I'm not going to propose language to address that, although it could be worked into the same PR.

More broadly, the test does not check if there is a mechanism to reveal the truncated text. The SC stipulates resizing can happen "without loss of content or functionality". However, it is entirely possible to truncate text in a managed way that both makes it clear that text has been intentionally truncated and to provide a function to display the truncated text. 

As the Understanding document notes:
> truncation is acceptable if the component's full content is available on focus or after user activation and an indication that this information can be accessed, is provided to the user in some way besides the fact that it is truncated.

The simplest example is a string that ends in an ellipsis, where the ellipsis is a button that can be activated to reveal the entire string. It's important to emphasize that use of such a technique is not motivated only by accessibility. Ellipses are commonly used anytime regular-sized text exceeds a prescribed area (think of a long string in a table column header cell).

Personally, I find some of the wording on the check confusing, and besides adding a third check and updating the outcome, I suggest rewording 2.

1. Increase the text size of the content by 200%.
1. Check that text is clipped, truncated, or obscured as a result of the resizing.
1. Check that no mechanism exists to display text that was clipped, truncated or obscured due to text resizing.

If check #2 or #3 is true, then the failure condition applies and the content fails these Success Criterion.

mbgower:
Closes #5071


techniques/failures/F69.html
```diff
@@ -59,12 +59,13 @@ <h3>Truncated text</h3>
          </div>
          <ol>
             <li>Increase the text size of the content by 200%.</li>
-            <li>Check that no text is clipped, truncated, or obscured.</li>
+            <li>Check that text is clipped, truncated, or obscured as a result of the resizing.</li>
+            <li>Check that no mechanism exists to display text that was clipped, truncated or obscured due to text resizing.</li>
```
mbgower: adding Oxford comma to match existing check
```suggestion
            <li>Check that no mechanism exists to display text that was clipped, truncated, or obscured due to text resizing.</li>
```

bruce-usab (changes requested):
Please see [my comment](https://github.com/w3c/wcag/issues/5071#issuecomment-4297064398) in #5071.

PR #5073, opened 2026-04-16
Add  example section for visual markup explanation to G96

CharlesBelov:
Added a text-specific example section explaining visual markup with semantic meaning.


techniques/general/G96.html
```diff
@@ -93,6 +93,10 @@ <h4>The <abbr title="Cascading Style Sheets">CSS</abbr></h4>
  }</code></pre>
 				</section>
 </section>
+    <section class="example">
+         <h3>Visual markup with semantic meaning has explicit text explaining the meaning</h3>
+			<p>A redline document designates deleted text with strikethrough and added text with double underline. Deletions are preceded by the text "[begin deletion]" and followed by the text "[end deletion]. Additions are preceded by the text "[begin addition]" and followed by the text "[end addition].</p>
```
patrickhlauke, line 98: shouldn't additions/deletions use `<ins>` and `<del>` ? or are you concerned about whether or not AT currently announce/expose these?

CharlesBelov: @patrickhlauke I do have concerns about whether `<ins>` and `<del>` are read by screen readers, an area I don't have expertise in, but the main issue is with publishing the document to the web as a PDF, which my understanding is that WCAG covers.

If I use track changes in Word to track insertions and deletions, when it is saved as PDF, the deleted text is removed as is the markup that there has been an addition. There doesn't appear to be any option in Word to Save to PDF with the tracked changes intact.

If I print, the tracked changes are there. But printing to PDF does not produce accessibility tags.

In the case of my agency, the San Francisco Municipal Transportation Agency, we don't use track changes to create these PDF files. Instead, we apply strikethrough to deleted content and a particular font and underscores to inserted text.

There is the further complication that additions and deletions can - admittedly rarely - be done by both the SFMTA Board and by the Board of Supervisors of the City and County of San Francisco. In such cases, we have two different deletion styles and two different insertion styles, something which `<ins>` and `<del>` can't handle semantically. 

None of this markup makes it from Word to PDF tags when saving as PDF. Our only available workaround is to explicitly show the additions and deletions as verbiage in a conforming alternative document.

Issue #884, opened 2019-09-05
What does "support the following text style properties" mean (1.4.12)?


JAWS-test:
If I have a closed environment where the developer can set all the text style properties, but the user can't (e.g. browser in a company where the use of user styles is not possible), what then?
* does the developer have to implement a setting possibility for the user, so that she or he can adjust the text style properties? or
* is 1.4.12 not applicable because the user cannot adjust the text style properties (without the help of the developer)?

patrickhlauke:
the latter

awkawk:
@patrickhlauke I'd say the former. If a user has a locked-down version of a browser that is required for use in company X then the author needs to provide this functionality. It is just like resize text where the Understanding document says:

> The author cannot rely on the user agent to satisfy this Success Criterion for HTML content if users do not have access to a user agent with zoom support. For example, if they work in an environment that requires them to use IE 6.
> 
> If the author is using a technology whose user agents do not provide zoom support, the author is responsible to provide this type of functionality directly or to provide content that works with the type of functionality provided by the user agent. If the user agent doesn't provide zoom functionality but does let the the user change the text size, the author is responsible for ensuring that the content remains usable when the text is resized.

patrickhlauke:
> If the author is using a technology whose user agents do not provide zoom support

the technology in this case (i'm assuming) is HTML/CSS, and default user agents provide support for this. the IE6 case (for text size) was specifically of its time, and related to not being able to rely on IE6 resizing text set in pixel sizes...I think that clause is very much "of its time" and mainly due to the preponderance of IE use at the time.

if the environment that the user has to use uses some cut-down browser (like a point of sales type device for instance), is this a web content author responsibility, or a user agent responsibility?

patrickhlauke:
or are you suggesting that it's the web author's responsibility now that for a point-of-sales type environment, THEY must provide / backfill anything that was not provided by default in the user agent or THEY fail WCAG? seems a stretch to me...

awkawk:
> if the environment that the user has to use uses some cut-down browser (like a point of sales type device for instance), is this a web content author responsibility, or a user agent responsibility?

I'd say author. Otherwise the author can't make a conformance claim that has them relying on accessibility-supported technology.

patrickhlauke:
to split hairs, for 1.4.12

> In content implemented using markup languages that support the following text style properties, no loss of content or functionality occurs by setting all of the following and by changing no other style property

[...]

PR #5081, opened 2026-04-23
Add new Technique: Guarding keyboard event handlers against IME composition

### LINKED ISSUE #5075: Proposal: New Failure Technique — Failure of SC 2.1.1 and SC 3.2.2 due to using `keydown`/`keyup` event handlers without guarding against IME composition
I'd like to propose a new failure technique for **SC 2.1.1 Keyboard** and **SC 3.2.2 On Input**.

## Failure description

When a `keydown` or `keyup` handler checks for the Enter key (or other keys used as IME shortcuts) without guarding against IME composition, users who rely on an IME — such as those in CJK regions — experience broken behavior. Pressing `Enter` to confirm an IME candidate fires `keydown`, which accidentally triggers the application's action (e.g., form submission) before the user has finished composing their input.

- **SC 2.1.1**: The keyboard interface becomes unreliable for IME users, as the IME's own keyboard shortcuts collide with the application's shortcuts.
- **SC 3.2.2**: An unexpected change of context (e.g., form submission) is triggered by input that the user intended solely for the IME.

## Failure example

```js
// ❌ Failure — triggers during IME composition
inputElement.addEventListener("keydown", (event) => {
  if (event.key === "Enter") submit();
});
```

## Sufficient techniques to avoid this failure

```js
// ✅ Option 1 — guard with isComposing (+ keyCode 229 for Safari)
inputElement.addEventListener("keydown", (event) => {
  if (event.isComposing || event.keyCode === 229) return;
  if (event.key === "Enter") submit();
});

// ✅ Option 2 — use the form submit event instead
formElement.addEventListener("submit", (event) => {
  event.preventDefault();
  submit();
});
```

The `keyCode === 229` check is necessary to cover a [known Safari bug](https://bugs.webkit.org/show_bug.cgi?id=165004) where `compositionend` fires before `keydown`, causing `isComposing` to be `false` on the final confirmation keypress. MDN also [recommends this workaround](https://developer.mozilla.org/en-US/docs/Web/API/Element/keydown_event#keydown_events_with_ime).
note: This bug has been fixed in a recent Safari version, but users on older versions remain affected. Therefore, adding `|| event.keyCode === 229` as a fallback is still recommended to ensure correct behavior across the full range of Safari versions in use.

## References
- [New lint rule: require IME-safe keyboard event handling · biomejs/biome · Discussion #10041](https://github.com/biomejs/biome/discussions/10041)
- [If You Handle Enter, Check event.isComposing - DEV Community](https://dev.to/yukimi-inu/why-16-billion-east-asians-are-quietly-raging-at-your-enter-key-handler-1po0)
- [Stop implementing “Enter to submit” by listening directly to raw keydown events.](https://dninomiya.github.io/form-guide/stop-enter-submit)
- [Your Form Submits Mid-Japanese Input | A static site to link people to when their web form submits upon the user confirming their IME input.](https://kai-rin.github.io/your-form-submits-mid-japanese-input/)
- [#45371 (Add tag: Use `event.isComposing` to support input method editor (IME) converters) – WordPress Trac](https://core.trac.wordpress.org/ticket/45371)
- https://x.com/p_craft/status/2044418886175793154?s=20

mehm8128:
close https://github.com/w3c/wcag/issues/5075


mehm8128:
FYI: [Add initial draft for input method article by xfq · Pull Request #884 · w3c/i18n-drafts](https://github.com/w3c/i18n-drafts/pull/884)

mehm8128:
@patrickhlauke 
Is there any reason why it was moved back from Drafted to In progress?

By the way, it seems that this issue was also mentioned in Google's Modern Web Guidance.
https://x.com/ChromiumDev/status/2095950953895645266?s=20

patrickhlauke:
> Is there any reason why it was moved back from Drafted to In progress?

Mainly because we've not gotten around to reviewing it first before discussing it in our Friday meetings ... once that's happened, it will move back and then generally goes to the wider group within a few weeks

mehm8128:
Understood. Looking forward to moving this PR forward.

PR #5082, opened 2026-04-23
Add index page for working-examples in dev and PR previews

kfranqueiro:
This implements an index page for the working-examples folder, only exposed in dev and for PR previews.

(This was suggested by @patrickhlauke a few weeks ago.)

This has no visible impact on normal builds.

Note that the index page does not list PHP files, since they won't work in these contexts anyway; those would still require copying the working-examples to a web server that supports PHP.

## Changes

- Adds `working-examples/index.html`, populated via `working-examples/index.11tydata.js`
  - The latter makes use of `tinyglobby`; in the future I'd like to migrate our existing `glob` usage to this package, but will do that separately. Installing `tinyglobby` adds less than 200KB to `node_modules`.
- Updates `CustomLiquid` to prevent processing anything outside of `techniques` and `understanding`
  - It turns out this had been needlessly and negligibly processing the top-level `index.html`, which this change also prevents
  - There's also a bit of refactoring to shorten the many uses of `scope.page.inputPath` by using a local variable for it
- I realized that the dev server was fully restarting when I changed `index.11tydata.js`, which is unnecessary, so I added an `--exclude` argument to `tsx watch`


patrickhlauke:
amazing stuff, @kfranqueiro

patrickhlauke (approved):

iadawn (approved):

PR #5088, opened 2026-04-26
Add clarification that moving phone/tablet camera for QR/bar code is not a motion actuation concern

### LINKED ISSUE #1120: Moving phone camera over QR code a case for 2.5.4 Motion Actuation?
Just wondering whether people agree that the need to move a phone camera over a printed or on-screen QR code to instantiate an app and start using the instance is a case covered by 2.5.4 Motion Actuation? I would argue that an alternative (such as a text field for the onput of the numeric equivalent) would be required to pass 2.5.4. Thoughts? (This is an native app case, but arguably the same could apply to a web app.) Any thoughts?

patrickhlauke:
Closes https://github.com/w3c/wcag/issues/1120

Preview: see last sentence in the first large note in the intent https://deploy-preview-5088--wcag2.netlify.app/understanding/motion-actuation


patrickhlauke:
closing in favour of https://github.com/w3c/wcag/pull/5089

PR #5089, opened 2026-04-27
Add clarification that moving phone/tablet camera for QR/bar code is not a motion actuation concern

### LINKED ISSUE #1120: Moving phone camera over QR code a case for 2.5.4 Motion Actuation?

### LINKED ISSUE #1404: 2.5.4 Motion Actuation and camera access (for QR codes etc, not actual "gesturing towards the device")
I've seen questions bubbling up a few times (most recently today https://webaim.org/discussion/mail_message?id=45215) about whether or not functionality like QR code reading fails 2.5.4 because "users have to move their device to point the camera at the QR code". 

my take is that no, this does not fail, because the "motion" - having to potentially move the device to point the camera to the QR code - is ancilliary ... the motion itself is not what triggers any functionality. the app/site is not reacting to the motion per se. originally, from my memories on the MATF at the time, the intent was essentially a specific scenario of a more general "reacting to sensors, like gyros, near-field sensors, accelerometers, etc.", but pared down to just cover motion.

the understanding document does, confusingly, mention cameras (in relation to users gesturing towards the device), but perhaps it needs to add an extra note about things like QR codes?

detlevhfischer:
Adding in the intent section that pointing a device camera at an object such as QR code or image does not fall under movement actuation.

Closes #1120
Closes #1404


patrickhlauke:
duplicate of sorts of https://github.com/w3c/wcag/pull/5088 - suggest in future checking if there's already something marked as closing an issue :)

understanding/21/motion-actuation.html
```diff
@@ -24,7 +24,7 @@ <h2>In brief</h2>
...
+<p class="note">This criterion concerns content that interprets motion such as explicit eye, face, or hand gestures to a camera, or tilting and shaking a device, as an input. It does not cover the movement of users through space as registered by geolocation sensors or beacons, the movement necessary to point a device camera at a particular object (e.g. a QR code, a document, or an image for the purpose of recognizing it), or events observed by the device other than intentional input gestures and movements by the user. The "Supported Interface" exception directly exempts incidental user motion – such as the movement necessary to type on/operate a keyboard, move/click a mouse, the hand and finger movements used to operate a touchscreen, or any movement necessary to operate assistive technology.</p>
```
patrickhlauke:
```suggestion
<p class="note">This criterion concerns content that interprets motion such as explicit eye, face, or hand gestures to a camera, or tilting and shaking a device, as an input. It does not cover the movement of users through space as registered by geolocation sensors or beacons, the movement necessary to point a device camera at a particular object (such as a QR code, a document, or an image for the purpose of recognizing it), or events observed by the device other than intentional input gestures and movements by the user. The "Supported Interface" exception directly exempts incidental user motion – such as the movement necessary to type on/operate a keyboard, move/click a mouse, the hand and finger movements used to operate a touchscreen, or any movement necessary to operate assistive technology.</p>
```

trying to avoid "e,g,", "i.e.", and similar, for better clarity

patrickhlauke:
tossing a coin as these are very similar...going to go with this one over mine, just to diversify the contributor pool a bit ;)

detlevhfischer:
I guess I just checked (maybe not thoroughly enough?) whether a PR was already linked in the issue - honestly wouldn't know how to check if a "closes #"  has already been used somewhere else on a PR... But never mind.

patrickhlauke:
You can see it on the issue itself

> screenshot of 1120 with the linked PR after the issue title, and the 'development' section listing linked PRs

patrickhlauke (approved):

adampage (approved):
LGTM 🚀

bruce-usab (approved):

PR #5090, opened 2026-04-29
Add C40 (two-color focus) as sufficient for 1.4.11 Non-text contrast, make sufficient techniques for 1.4.11 situationally specific

### LINKED ISSUE #2575: Should C40 be also sufficient for 1.4.11 Non-Text Contrast?

patrickhlauke:
Based on the discussion and agreement in #2575, it seems it was agreed that C40 does indeed cover 1.4.11 as a sufficient technique, but it was never added in #3112 /cc @dbjorge

Edit: in addition, this PR reworks how sufficient techniques are identified for 1.4.11 - qualifying situations where a sufficient technique is actually sufficient

Closes https://github.com/w3c/wcag/issues/2575

Previews:

* https://deploy-preview-5090--wcag2.netlify.app/understanding/non-text-contrast#techniques
* https://deploy-preview-5090--wcag2.netlify.app/techniques/general/g174
* https://deploy-preview-5090--wcag2.netlify.app/techniques/general/g195
* https://deploy-preview-5090--wcag2.netlify.app/techniques/css/c40
* https://deploy-preview-5090--wcag2.netlify.app/techniques/general/g207
* https://deploy-preview-5090--wcag2.netlify.app/techniques/general/g209


detlevhfischer:
Provisional thumbs down, because my understanding is that a Sufficient Technique should ensure that when applied, the page passes the SC it refers to. Having contrasty focus indication says nothing about passing 1.4.11 for any other graphical content on a page, so I don't see how C40 on its own it can be considered a sufficient technique for 1.4.11.

bruce-usab (approved):

understanding/understanding.11tydata.js
```diff
@@ -535,7 +535,7 @@ export default function (data) {
           {
             title:
               "Situation A: Color is used to identify user interface components or used to identify user interface component states",
-            techniques: ["G195", "G174"],
+            techniques: ["G195", "G174","C40"],
```
kfranqueiro: Formatting consistency

```suggestion
            techniques: ["G195", "G174", "C40"],
```

patrickhlauke:
> Provisional thumbs down, because my understanding is that a Sufficient Technique should ensure that when applied, the page passes the SC it refers to. Having contrasty focus indication says nothing about passing 1.4.11 for any other graphical content on a page, so I don't see how C40 on its own it can be considered a sufficient technique for 1.4.11.

@detlev https://github.com/w3c/wcag/pull/5090/changes/920af828b10d8d1cd69021d900765f7b1a18fe6f should address this now ... can you check and, if you're happy with it, change your thumbs-down?

bruce-usab (changes requested):
G174 is, I think, missing a precondition?  Something like:

> For each page with insufficient contrast, check for the following:

This would be first [line under procedure](https://deploy-preview-5090--wcag2.netlify.app/techniques/general/g174#tests), before the numbered test steps.

Test procedures for G195, G207, and G209 all start with “For each…”

patrickhlauke:
@bruce-usab it's a bit out of scope for this, but sure...

bruce-usab (approved):

PR #5097, opened 2026-05-01
Update Bypass Blocks to match other SC 

### LINKED ISSUE #5096: site-oriented SC use inconsistent phrasing
WCAG uses slightly different phrasing when meaning the same thing:
- [2.4.1 Bypass Block](https://www.w3.org/TR/2024/REC-WCAG22-20241212/#bypass-blocks)
  * …repeated on multiple web pages
- [2.4.5 Multiple Ways](https://www.w3.org/TR/2024/REC-WCAG22-20241212/#multiple-ways)
  * …web page within a set of web pages
- [2.4.8 Location](https://www.w3.org/TR/2024/REC-WCAG22-20241212/#location) (AAA)
  * …within a set of web pages
- [3.2.3 Consistent Navigation](https://www.w3.org/TR/2024/REC-WCAG22-20241212/#consistent-navigation)
  *  …repeated on multiple web pages within a set of web pages
- [3.2.4 Consistent Identification](https://www.w3.org/TR/2024/REC-WCAG22-20241212/#consistent-identification) 
  * …within a set of web pages
- [3.2.6 Consistent Help](https://www.w3.org/TR/2024/REC-WCAG22-20241212/#consistent-help) (new with 2.2)
  * …repeated on multiple web pages within a set of web pages

I submit that either:
1. “within a set of web pages” be added to 2.4.1 (for consistency); or
2. “within a set of web pages” be deleted from 3.2.3 and 3.2.6 (as being superfluous)

bruce-usab:
Add “within a set of web pages” to end of SC.  Makes something implicit explicit, and conforms this SC to the 5 (!) other SC.

This is editorial, but touches the normative text.  Inconsistency noted on WCAG2ICT TF call 30 April 2026.

Closes #5096


patrickhlauke:
> This is editorial, but touches the normative text.

which of course means it'll cause an existential crisis to some folks...

patrickhlauke (approved):

alastc:
I'm not sure about this, I think there is a difference between bypass blocks and Consistent nav/help.

For the consistency ones, they are scoped to a set of web pages so that the navigation/help can be different between sets of web pages. (But consistent within the set.)

Bypass blocks isn't related to sets of pages, it's about whether certain content is repeated across pages that may or may not be a set. It's about having a link present, but it doesn't have to be in the same place across a set of web pages.

I think this artificially narrows the scope of the SC.

patrickhlauke:
@alastc it's probably a question of whether or not by virtue of having content that is repeated across multiple pages that needs a bypass mechanism, those multiple pages effectively become a "set" ...

bruce-usab:
I think I must agree that, linguistically, “blocks of content that are repeated on multiple web pages” is a wider scope than “blocks of content that are repeated on multiple web pages [within a [set of webpages](https://www.w3.org/TR/WCAG22/#dfn-set-of-web-pages)]” since the second **seems** to be adding a constraint on the first.

Than is in tension with the general principle of using the same phrasing when the same thing is meant.

My other concern is that if we _don’t_ make this change, then aren’t the navigation/help SC (i.e., 3.2.3, 3.2.6) weakened?  Those somehow _don't_ apply when they **are repeated on multiple** pages not on **set** of web pages?

patrickhlauke:
based on @alastc 's concern in https://github.com/w3c/wcag/pull/5097#issuecomment-5294999845 i'm actually now leaning towards not making this change. left as is, the SC still works fine, while making it explicitly about pages in a set may require contortions like my devil's advocate one above ("are they part of a set by virtue of having those repeated elements that need a bypass?")

patrickhlauke:
> My other concern is that if we don’t make this change, then aren’t the navigation/help SC (i.e., 3.2.3, 3.2.6) weakened? Those somehow don't apply when they are repeated on multiple pages not on set of web pages?

I think the narrower scoping here is important, because for a whole site, you may have different sections (different "sets") with their distinct navigation and features. and removing "within a set of web pages" from 3.2.3 and 3.2.6 would then fail as soon as you compare two pages from different sets but all still part of the same site.

Issue #4113, opened 2024-10-17
3.1.2 Language of Parts clarification


EmptyEnd6893:
### Discussed in https://github.com/w3c/wcag/discussions/4111

<div type='discussions-op-text'>

<sup>Originally posted by **EmptyEnd6893** October 17, 2024</sup>
I have been wondering about the second example for the [Language of Parts understanding page](https://www.w3.org/WAI/WCAG22/Understanding/language-of-parts.html). I have repeated it below for convenience.

'An HTML Web page includes links to versions of the page in other languages (e.g., Deutsch, Français, Nederlands, Catalan, etc.). The text of each link is the name of the language, in that language. The language of each link is indicated via a lang attribute.'

Are the names of the languages not considered proper names in this case? If so, it seems that the criterion would not apply. I realise that language names are slightly unusual in that languages frequently have their own word for other languages, but wouldn't they still count as proper names? Any help would be appreciated.</div>

EmptyEnd6893:
Hi. I am opening an issue based on the above Q&A as it could be an opportunity to make the understanding page for 3.1.2 a little clearer.

EmptyEnd6893:
The gist of it is that the understanding document does not appear to provide any guidance on how to handle proper names that are different in different languages, such as Deutsch changing to German. From what I have read, there does not appear to be a requirement to set a lang attribute (or use another technical method to set language) on proper names in a foreign language. This seems odd to me. Can anyone confirm whether this is the intended interpretation of the criterion?

patrickhlauke:
> Can anyone confirm whether this is the intended interpretation of the criterion?

Yes

EmptyEnd6893:
Does that mean that it is not required to set a lang attribute on the <a> elements shown in the second example on the understanding page. The language names would count as proper names correct?

patrickhlauke:
To me, the names of languages aren't "proper names", so the example looks correct to me, but this is likely where linguists can debate things at length.

patrickhlauke:
in german, practically every single word that's not an adjective or verb is capitalised. doesn't mean that it makes every one of those words "proper names".

EmptyEnd6893:
[...]

Issue #1174, opened 2020-06-22
3.1.2 Language of parts - can unmarked single words considered a PASS


detlevhfischer:
I apologise for returning to an issue that has already been discussed a few times here: https://github.com/w3c/wcag/issues/297 and https://github.com/w3c/wcag/pull/808 with no clear decision yet taken due to more urgent business.
Considering the loads of other stuff,  I mark this as discussion, not something requiring a working group vote.

Our test procedure for _3.1.2 Language of Parts_ currently requires words that are neither spoken in a similar way as in the source language nor have become part of the vernacular to be marked up.  

We got a change request calling out a mistake here, arguing that single words do not fall under 'passage or phrase'. The user argues that marking up single words forces screen readers to switch language, and that while some desktop screenreaders can turn off language change, this this not possible in VoiceOver and Talkback, making pages where single words are diligently marked up unpleasant to read for screen reader users.

I would like to get a quick straw poll of other people's opinion on whether you would consider the lack of markup on single words necessarily a FAIL. Would you accept the trade-off argument that the practice is so disruptive in many contexts (especially mobile) that it should be considered a PASS when individual words are not marked up (while longer passages like quotations, book citations of titles etc. would of course still require lang markup)? An exception where single word mark-up seems important appears to be language choice menus.

kerstinp:
Hi all,

I just want to add [Issue 287](https://github.com/w3c/wcag21/issues/287).

Cheers

Kerstin

kerstinp:
Hi all,

I’m a bit confused cause I opened in 2017 this Issue ([Issue 287](https://github.com/w3c/wcag21/issues/287)) for seeking clarification on what the SC says? In this SC we have exceptions and “passage or phrase”. Single words are no passages. So my question was, what is meant by “phrase” – cause there is no definition in the normative glossary. The answer was:

> WCAG 2.0 is definitely referring to the "group of words" definition.

Acccording to this answer it is a pass when single words from a foreign language don't have markup.  

Screenreader:

How AT are dealing with words from a foreign language does not just depend on the screenreader but (also) on the voice and in some cases also on the OS. Some years ago I’ve tested with a short german text: 101 words, 18 english (Disclaimer: The text includes also proper names which is according to the SC an exception. And four words that are not pronounced differently, which is not described in the SC):

This was the german text:

[...]

PR #5098, opened 2026-05-02
Add missing definition for "proper name"

### LINKED ISSUE #4094: 3.1.2 Lang of Parts exceptions need to be Key Terms
It would be very helpful in practice to get Key Terms for the "except for" part of 3.1.2 Language of Parts. 

Second, I'd like to understand this team's intent for the phrasing of the 3.1.2 Language of Parts SC, primarily the reason to use "except for":

> The [human language](https://www.w3.org/WAI/WCAG22/Understanding/language-of-parts.html#dfn-human-language) of each passage or phrase in the content can be [programmatically determined](https://www.w3.org/WAI/WCAG22/Understanding/language-of-parts.html#dfn-programmatically-determined) except for proper names, technical terms, words of indeterminate language, and words or phrases that have become part of the vernacular of the immediately surrounding text.

This SC seems to imply the second clause is clearly not able to be programmatically determinable. Is the intent that the "except for" items not be pronounced by AT in their original language even if it can be pronounced?

- "can be" implies there are situations that "can't be" ; I know the 3.1.1 SC uses the same phrasing so maybe just ignore this bullet. 

- The "except for" part of the SC some people have interpreted as meaning that any of these exceptions are **exempt** from needing (or even, "should **never** use") a language attribute.

In the case I'm thinking of, a company uses several branding product names that are purposely not translated into a second language when the surrounding text is in that second language. It is a decision; these words are trademarked. That product name is clearly in a programmatically determinable language (say, clearly in English, which has a programmatically determinable language "en"). As the name of a "thing" it could be considered a "proper name" if using a generic definition from an internet search, like a "person, place, or thing". WCAG does not define "proper name". It seems to do AT users, particularly SR users, a disservice by resulting in poor pronunciation for items that are clearly one language's version of a noun.  If it used a lang attribute, AT users would understand this product is the same product as what is advertised on commercials; if not using a lang attribute some SR butcher the words as unintelligible in the second language.

Also, most languages have alternative proper nouns for places and things, like "Twin Towers" in English is "Torres Gemelas" in Spanish; it was a place, it has proper nouns, but the proper nouns can change. 

TL; DR
I would strongly suggest formally adding each exception as its own Key Term. 
I would strongly suggest limiting the "proper name" term to specific use cases
I would strongly suggest a sentence in the Understanding document that permits using a programmatically determinable language on these exceptions (as defined in a key term) if the author determines it may serve the intended audience, but that it would not fail the SC.

patrickhlauke:
Two success criteria just casually mention "proper names" without having an actual definition of what the term means.

The exact meaning of "proper name" (and "proper noun") appears to be a hotly contested / not quite clear-cut area in linguistics. Some languages also may not have a concept of "proper name/noun" that maps neatly to the English meaning.
This PR tries to give a fairly high-level definition (inspired by wikipedia, admittedly), for clarity/consistency. In particular, it surfaces the idea that a proper name/noun "can be used across languages" (which is relevant to explain its use in 3.1.2 Language of Parts in particular).

Relates to https://github.com/w3c/wcag/issues/4113 where the question came up whether or not the names of languages, in their own language (like "Deutsch", "Italiano", etc) are not considered "proper names" because they were capitalised, and one folksy definition of proper name/noun leans on the idea that "they're words that are usually capitalised".

This PR just spells out explicitly what the interpretation has always been (?), so this does not introduce any new normative requirement by the backdoor.

Previews:

* Term https://deploy-preview-5098--wcag2.netlify.app/guidelines/#dfn-proper-name-s
* 3.1.2 SC https://deploy-preview-5098--wcag2.netlify.app/guidelines/#language-of-parts
* 3.1.5 SC https://deploy-preview-5098--wcag2.netlify.app/guidelines/#reading-level

Also relates to https://github.com/w3c/wcag/issues/1174
Closes https://github.com/w3c/wcag/issues/4094


guidelines/terms/20/proper-name.html
```diff
@@ -0,0 +1,4 @@
+<dt><dfn id="dfn-proper-name-s" data-lt="proper names|proper name(s)">proper name</dfn></dt>
+<dd>
+  <p>a name which is taken to uniquely identify its referent in the world – such as the name of a specific person, place, or product – and that can be used across languages without modification</p>
```
WilcoFiers: This definition seems kinda clunky. So people who were named after a parent / grandparent don't have a proper name because the name wasn't intended to be unique? That's odd, and hard to test. It also seems that things like book and movie titles fall under this. "Project Hail Mary" is the name of a book and movie, it's intended to be unique, and you can't translate it because you're then talking about the translated book. Books and movies wouldn't traditionally be considered a proper name, but your definition doesn't seem to rule them out. I would expect a language attribute on those to be required.

"Can be used across languages without modification" is a confusing phrase to me. I had to look up the issue to figure it out. I think what you want to say here is that "the name cannot be translated to the language of the text without changing its meaning". That way "Deutsch" in an English text requires a language, as does "Deutschland". But "Shell Global" in a German text is fine, even if you can translate those words, since the English name of that company is also its name used in Germany.

detlevhfischer:
Looks good to me, can't see why that should be an issue, @WilcoFiers

WilcoFiers (changes requested):
I'm opposed to making substantive changes to WCAG's existing documents. See https://github.com/w3c/wcag/pull/4952#issuecomment-4359200925

This change can be done in an understanding document.

guidelines/terms/20/proper-name.html
patrickhlauke: i encourage you to come up with a better definition, @WilcoFiers

patrickhlauke: (and the fact we're arguing/disagreeing is exactly WHY this should be a normative definition in spec ... since this determines whether something passes or fails the requirements in question that use that undefined term ... and understanding documents should not be the ones that influence pass or fail per se, no?)

bruce-usab (changes requested):
+1 to Wilco’s concern for the proposed definition.  I suggest that the change should be for the SC to use “proper nouns” instead of “proper names” and then we _really_ don’t have to add a term.

Can anyone articulate a difference between the two?  I think this is essentially a long-standing typo!  Compare Google (or dictionary) search for [proper name](https://www.google.com/search?q=define%3A%22proper+name%22) to [proper noun](https://www.google.com/search?q=define%3A%22proper+noun%22).

patrickhlauke:
>  I suggest that the change should be for the SC to use “proper nouns” instead of “proper names” and then we really don’t have to add a term

do you think "proper nouns" is unambiguous enough that it doesn't warrant a term/definition? I disagree, to me it's just as vague a term as "proper names". keeping in mind that the audience includes a large number of authors etc who don't have English as their first language either.

and just noting that even saying "we should change names to nouns" would still be an erratum that I believe Wilco would be opposed to

guidelines/terms/20/proper-name.html
gundulaniemann: I suggest to drop the term "uniquely" from the definition.

adampage:
```suggestion
  <p>a name which is taken to distinctly identify its referent in the world – such as the name of a specific person, place, or product – and that can be used across languages without translation</p>
```
“distinctly” instead of “uniquely” to accommodate proper names that are not truly unique?

“translation” instead of “modification” to allow for transliterating a proper name into another writing system?  (e.g., “Project Hail Mary” → “プロジェクト・ヘイル・メアリー”)

patrickhlauke: > I suggest to drop the term "uniquely" from the definition.

I think @adampage 's change softens this https://github.com/w3c/wcag/pull/5098/changes/6fc74571403538fe6e25bdb4ec606ae7ebdf5d99

patrickhlauke:
Discussed in today's TF call. There were a few smaller wordsmithing concerns, and the question of whether or not this would actually directly address the original issue that spawned this (which I believe it does, as the actual term definition would make it clearer by inference that "Deutsch", "Italiano", etc are not "proper names" when used in English prose), but I think generally the feeling was positive.

bruce-usab:
I think the current proposed definition is an improvement from earlier:
> a name which is taken to distinctly identify its referent in the world – such as the name of a specific person, place, or product – and that can be used across languages without translation

But is it not a bit circular?  The “used across languages without translation” is the basis for exception in the SC, so it’s off-loading some of the subjectivity to the definition.  I am not sure this is an improvement.  I still favor using proper noun (instead of proper name), a term found in dictionaries.  I do think “proper nouns” is unambiguous enough that it doesn't warrant a glossary entry.

Yes, that is still an erratum.  It’s correcting a typo, so I think that is more defensible (as compared to adding a term).

bruce-usab:
I agree that “distinctly” is better than “uniquely”  and “without translation” is better than ”without translation” but, assuming we add a definition, can we avoid “referent” and “is taken to?  I am also not sure about “in the world” as I think fairly abstract things can have proper names.  How about:
> a name which is used to distinctly identify something – such as the name of a specific person, place, or product – and that can be used across languages without translation

bruce-usab:
Discussed on TF call 15 May and we have consensus with moving forward to AG.  A recommendation was made to reinforce in PR that main point of SC is for pronunciation by screen reading software.

guidelines/terms/20/proper-name.html
```diff
@@ -0,0 +1,4 @@
+<dt><dfn id="dfn-proper-name-s" data-lt="proper names|proper name(s)">proper name</dfn></dt>
+<dd>
+  <p>a name which is taken to distinctly identify its referent in the world – such as the name of a specific person, place, or product – and that can be used across languages without translation</p>
```
iadawn:
```suggestion
  <p>a name which is taken to distinctly identify its referent in the world – including, but not limited to, the name of a specific person, place, or product – and that can be used across languages without translation</p>
```

guidelines/terms/20/proper-name.html
iadawn: > "the name cannot be translated to the language of the text without changing its meaning"

Not sure that it is a 'cannot' be translated. My name, for example, is a proper noun but 'White' can be translated into other languages.

Having said that if the definition of 'proper name'/'proper noun' "appears to be a hotly contested / not quite clear-cut area in linguistics", I am wondering if this is something that we can or should wade in with a definition for.

patrickhlauke: > My name, for example, is a proper noun but 'White' can be translated into other languages.

but then, it changes its meaning, because you are not "Kevin Bianco", you are "Kevin White", which is your proper name

patrickhlauke: > I am wondering if this is something that we can or should wade in with a definition for

and leave the normative wording ambiguous? and therefore interpretation and auditing results ambiguous/down to people's interpretation?

chaals:
> do you think "proper nouns" is unambiguous enough that it doesn't warrant a term/definition? 

Yes, absolutely - and given that the WCAG group are not lexicographers (although some of us, taking a broad definition of "group", might meet the definition given by Johnson a few centuries ago) I think it is better to look for terms that we can avoid defining.

patrickhlauke:
retracting this for now - too much pushback to actually defining normatively what the term means. will switch to a note in Language of Parts understanding instead then...

patrickhlauke:
Superseded by https://github.com/w3c/wcag/pull/5135 which tries to solve this in the non-normative understanding document instead (though this seems suboptimal compared to actually providing a normative definition)

PR #5102, opened 2026-05-06
Implemented the F13 as a 1.1.1 and 1.3.3 failure

### LINKED ISSUE #4784: F13 is ambiguous as for what causes the SC to fail

baldino-m:
[Diff preview](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FTechniques%2Ffailures%2FF13&doc2=https%3A%2F%2Fdeploy-preview-5102--wcag2.netlify.app%2Ftechniques%2Ffailures%2Ff13)

Preview: https://deploy-preview-5102--wcag2.netlify.app/techniques/failures/f13

Closes #4784


bruce-usab (changes requested):
I like the prose of the rewrite — but I disagree with swapping 1.4.1 out for 1.3.3 (because 1.3.3 is limited to _instructions_ only).

baldino-m:
> I like the prose of the rewrite — but I disagree with swapping 1.4.1 out for 1.3.3 (because 1.3.3 is limited to _instructions_ only).

Hi @bruce-usab, 

I disagree: in my opinion, the alternative text is providing effective instructions about how to "read" (so to speak) the bar chart ("The red bars indicate sales that were below the yearly quota"). The SC states: "Instructions provided for understanding and operating content [...]" and I think that this applies. The instructions just don't make sense as an alternative text is aimed (mainly) at non-sighted users. I am more than happy to discuss with the group and get everyone's opinion about this.

bruce-usab (approved):
I have changed my mind upon another re-reading, and I agree 1.3.3 is a fit and 1.4.1 is not.

techniques/failures/F13.html
```diff
@@ -18,37 +20,49 @@ <h2>When to Use</h2>
...
-      who are blind, have low vision, or have color vision deficiency, because they will not be able to perceive the
-      information conveyed by the color differences.</p>
+         the text alternative for an image relies on color differences only. This can cause problems for
+         people
+         who are blind or have low vision, because they will not be able to perceive the visual cues the alternative
+         text refers to.</p>
+
+      <div class="note">
+         <div>
+            <p>Depending on its visual appearance, such an image might or might not fail success criterion <a href="../../Understanding/use-of-color">1.4.1 Use of Color</a>. For example, bar charts, such as the one described in the following section, can provide information based on color alone (which represents a failure) or using a combination of color and pattern for their bars (which is a pass). Please remember that 1.4.1 Use of Color addresses the needs of sighted users, who are able to perceive images and don't usually rely on text alternatives.</p>
```
kfranqueiro:
```suggestion
            <p>Depending on its visual appearance, such an image might or might not fail Success Criterion <a href="../../Understanding/use-of-color">1.4.1 Use of Color</a>. For example, bar charts, such as the one described in the following section, can provide information based on color alone (which represents a failure) or using a combination of color and pattern for their bars (which is a pass). Please remember that 1.4.1 Use of Color addresses the needs of sighted users, who are able to perceive images and don't usually rely on text alternatives.</p>
```

techniques/failures/F13.html
```diff
@@ -18,37 +20,49 @@ <h2>When to Use</h2>
...
-         is conveyed by the color red in the image. The alternative should
-         indicate which people did not meet the sales quota rather than
-         relying on color.</li>
+            yearly sales figures for four employees in the Sales Department. The
+            text alternative for the image says: "Bar chart
+            showing the yearly sales figures for the Sales Department. Mary
+            sold 3.1 Million; Fred, 2.6 Million; Bob, 2.2 Million; and Andrew,
+            3.4 Million. The red bars indicate sales that were below the yearly
+            quota". Information about the employees not meeting the yearly quota is not provided anywhere else on the
+            page.<!-- making the bar chart image the only source of information. This accounts for a scenario in which a full-data table other kind of content (in text form) is provided below the bar chart, in which case the alternative text for the image would not rely on colour differences only --> This alternative fails to provide all the relevant information,
```
kfranqueiro, line 44:
```suggestion
            page. This alternative fails to provide all the relevant information,
```

patrickhlauke:
rereading this, it's such a convoluted multi-failure example. to me, at first reading, and in order of "priority"

- the image of the chart itself as described fails 1.4.1 Use of Color
- i can then see how you could argue that the text alternative, while describing things correctly visually, fails to fully convey the meaning/content of the image, so failure of 1.1.1 Non-text Content
- if you then insist, yes i can understand why you'd then ALSO say this fails 1.3.3 Sensory characteristics. though I am still not convinced altogether than the text alternative is an instruction - the alternative is intended to act as the thing itself...once somebody gets the alternative, you'd assume they don't see the actual image. so it's not "instructing" so much as it is "replacing" (though yes, i'm aware that there are scenarios where low-vision users may both be able to at least partially perceive the image itself and then still use the alt as a description). I more naturally count instructions as being somewhere else on the page

(leaving the interpretation of 1.3.3 aside, one way of making it more clearly just a 1.1.1/1.3.3 failure would be if the actual image of the chart used something other that the red colour - like a pattern, or bold text, or whatever that doesn't count as just colour, and the alt text then decided to ONLY mention the red colour. feels a bit of a stretch perhaps?)

maybe a much less ambiguous fix here would be to make this a failure of 1.4.1 Use of Color and 1.3.3 Sensory Characteristics (and leave 1.1.1 aside). Say there's the image of the chart that relies on colour, AND there's a visible caption underneath the chart referring to the red colour. there, unambiguous failure of both, without the awkward "is the alternative to be considered something that users would also consume at the same time as seeing the original?".

For the scenario of the chart just using red and the alt reflecting that, it's enough that that would fail 1.4.1 I'd say (without trying to then crowbar some failure of 1.1.1 into it).

If it was a chart that used red and some other clue, so not failing 1.4.1, but the alt text only talked about the red colour, then just fail it under 1.1.1 and be done with it for not being appropriate alternative for the non-text content....no need to try and drag in an extra 1.3.3 failure out of it

PR #5107, opened 2026-05-09
Update On Input to address different button types

### LINKED ISSUE #5059: On Input understanding doc makes broad strokes about buttons
The rational that buttons are not applicable under On Input is summed up in this part of the opening intent for the SC:

>Changing the setting of any user interface component is changing some aspect in the control that will persist when the user is no longer interacting with it. So checking a checkbox, entering text into a text field, or changing the selected option in a list control changes its setting, but activating a link or a button does not.

But this reasoning seems flawed.  A button can behave exactly like a "checkbox" which _is reviewed_ under On Input, if the button's purpose is to convey a selected state (aria-pressed).  Quite often a button can be interacted with and something about the button changes, where a change of context would not be expected.

For instance, the aria-pressed example, or activating a button to indicate it's the "current" button in a set of buttons.  Expand/collapse buttons.  A button that performs an action that results in a loading state, where in performing that action the button becoming disabled and thus loses focus - which commonly results in screen reader virtual cursors being kicked back up to the body element (change in context), when sighed mouse / keyboard users attention can remain on the now inactive button, and even pressing of tab key will continue navigating a user to the next focusable element in the document's focus order.

If I understand the intent of this callout, it was because actions like submitting a form are reasonably expected to navigate to a new page, or activating a standard button is reasonably expected to move focus to a new location in the page (dialog, first focusable element of a listbox or menu popup, etc.) but broadly excepting buttons without nuance seems like a weird informative doc normative change to "changing the setting of ANY user interface component..."

fstrr:
Closes #5059 

Clarifying different types of buttons and whether they're in or out of scope

Preview: https://deploy-preview-5107--wcag2.netlify.app/understanding/on-input

[Diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fwcag%2Funderstanding%2Fon-input&doc2=https%3A%2F%2Fdeploy-preview-5107--wcag2.netlify.app%2Funderstanding%2Fon-input)


understanding/20/on-input.html
```diff
@@ -36,11 +36,16 @@ <h2>Intent of On Input</h2>
...
       <div class="note">
          
-         <p>This success criterion covers changes in context due to changing the setting of a
-            control. Clicking on links or buttons is activating a control, not
-            changing the setting of that control.
-         </p>
-         
+         <p>This success criterion covers changes in context due to changing the setting of a control. Clicking on a link is activating it rather than changing its setting. Buttons differ depending on their action:</p>
+        <dl>
+          <dt>Buttons that change their state</dt>
```
fstrr:
```suggestion
          <dt>Buttons that change state</dt>
```

scottaohara (approved):
thanks for working on this @fstrr 

i think this helps / fixes the over simplification that the understanding doc currently has.

going to mark as approved because i think this is the right direction to take

understanding/20/on-input.html
```diff
@@ -36,11 +36,16 @@ <h2>Intent of On Input</h2>
       
       <div class="note">
          
-         <p>This success criterion covers changes in context due to changing the setting of a
-            control. Clicking on links or buttons is activating a control, not
-            changing the setting of that control.
-         </p>
-         
+         <p>This success criterion covers changes in context due to changing the setting of a control. Clicking on a link is activating it rather than changing its setting. Buttons differ depending on their action:</p>
```
patrickhlauke:
```suggestion
         <p>This success criterion covers changes in context due to changing the setting or value of a control.</p>
         <p>In contrast, clicking on a link is activating it, rather than changing its setting. Buttons differ depending on their action:</p>
```

understanding/20/on-input.html
```diff
@@ -36,11 +36,16 @@ <h2>Intent of On Input</h2>
...
+        <dl>
+          <dt>Buttons that change state</dt>
+          <dd>A button that has its state toggled, for example: by using <code>aria-pressed</code> or <code>aria-expanded</code> attributes, is changing its setting.</dd>
```
patrickhlauke: i think with this an subsequent ones, it would make more sense to remove the `:`

understanding/20/on-input.html
```diff
@@ -36,11 +36,16 @@ <h2>Intent of On Input</h2>
...
+          <dd>A button that becomes disabled when activated is changing its setting.</dd>
+          <dt>Buttons that submit data</dt>
+          <dd>A button that submits a form and navigates the user to a new page is activating a control, not changing a setting.</dd>
+          <dt>Buttons that move focus</dt>
+          <dd>A button that moves focus, for example: to a newly opened modal dialog, a slide-out navigation panel, or the first item in a <code>listbox</code> is activating a control, not changing its setting.</dd>
+</dl>
```
patrickhlauke, line 49: i wonder if this list could be reordered/reorganised somehow into two groups: buttons that, when clicked/pressed, counts as changing a setting; and then buttons that trigger an action.

maybe we want to also give a similar list of controls that count as changing settings? radio buttons, checkboxes, select groups, text inputs?

bruce-usab (approved):
Discussed on TF call 15 May and consensus that exposition on different button types is a helpful clarification.  See [preview diff at Intent](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fwcag%2Funderstanding%2Fon-input&doc2=https%3A%2F%2Fdeploy-preview-5107--wcag2.netlify.app%2Funderstanding%2Fon-input#intent).

understanding/20/on-input.html
patrickhlauke:
```suggestion
          <dd>A button that has its state toggled, for example by using <code>aria-pressed</code> or <code>aria-expanded</code> attributes, is changing its setting.</dd>
```

understanding/20/on-input.html
```diff
@@ -36,11 +36,17 @@ <h2>Intent of On Input</h2>
...
+         <p>This success criterion covers changes in context due to changing the setting or value of a control.</p>
+         <p>In contrast, clicking on a link is activating it, rather than changing its setting. Buttons differ depending on their action:</p>
+        <dl>
+          <dt>Buttons that change state</dt>
+          <dd>A button that has its state toggled, for example by using <code>aria-pressed</code> or <code>aria-expanded</code> attributes, is changing its setting.</dd>
+          <dd>A button that becomes disabled when activated is changing its setting.</dd>
+          <dt>Buttons that submit data</dt>
+          <dd>A button that submits a form and navigates the user to a new page is activating a control, not changing a setting.</dd>
+          <dt>Buttons that move focus</dt>
+          <dd>A button that moves focus, for example: to a newly opened modal dialog, a slide-out navigation panel, or the first item in a <code>listbox</code> is activating a control, not changing its setting.</dd>
```
patrickhlauke:
```suggestion
          <dd>A button that moves focus, for example to a newly opened modal dialog, a slide-out navigation panel, or the first item in a <code>listbox</code> is activating a control, not changing its setting.</dd>
```

understanding/20/on-input.html
```diff
@@ -36,11 +36,17 @@ <h2>Intent of On Input</h2>
...
-         <p>This success criterion covers changes in context due to changing the setting of a
-            control. Clicking on links or buttons is activating a control, not
-            changing the setting of that control.
-         </p>
-         
+         <p>This success criterion covers changes in context due to changing the setting or value of a control.</p>
+         <p>In contrast, clicking on a link is activating it, rather than changing its setting. Buttons differ depending on their action:</p>
+        <dl>
+          <dt>Buttons that change state</dt>
+          <dd>A button that has its state toggled, for example by using <code>aria-pressed</code> or <code>aria-expanded</code> attributes, is changing its setting.</dd>
```
patrickhlauke:
```suggestion
          <dd>A button that has its state toggled, for example by using <code>aria-pressed</code> attributes, is changing its setting.</dd>
```

understanding/20/on-input.html
patrickhlauke: Deferring this to a later, more fundamental expansion/rewrite of On Input

awkawk:
@fstrr I'm not sure if this is necessary - we are clear in Understanding that with a button we are activating the control not changing a setting. If I click a "pay now" button and it becomes inactive to prevent paying again, I have still activated the button and that might send me to a dialog to complete the transaction, and that seems appropriate. 

The use case that drove this was the wide-spread use of "jump menus" and radio groups with sub-items - in both cases a keyboard user would be sent off to a different location as they navigated through the options presented. 

In the APG they differentiate between "widgets" and "input widgets" and I think that reflects what I'm thinking about these here. Spinbuttons and switches may use buttons also, but are not buttons when the correct role is applied, for example...

scottaohara:
>we are clear in Understanding that with a button we are activating the control not changing a setting.

but per to the current note in the understanding doc that changes "any user interface component" from the normative text to essentially say "any user interface component except for links and buttons" - that then over steps and makes it read that a toggle button may allow for a change in context even though the control is effectively the same as a checkbox which wouldn't allow that.

the clarification from this PR maintains what the goal of the SC was, per your description of the intent, but corrects for the lack of nuance in the different types of buttons / that not all buttons will always submit a form or open a dialog,...

awkawk:
@scottaohara can you provide some other examples of buttons that are used as input and the result is a changed setting for the control? Maybe I'm just not thinking about the same use cases you are.

scottaohara:
beyond the toggle button example that is a great example of a clear contradiction that shows the current informative language is too broad in its effort to modify the normative text... per your example of a pay now button that becomes disabled - as you said, focus "might" go to a dialog upon activation.  so, imagine the situations where that doesn't happen. i've seen a lot.  for instance, a dynamic message appears instead to say added to cart, or that now the item is sold out.  The button becomes disabled, resulting in focus being removed from the button and now is programmatically reset to the body.

In React apps, developers might entirely re-render a toggle button, expand/collapse button resulting in a loss of focus / resetting context to the body.  this is also unfortunately not uncommon from my experiences.

PR #5119, opened 2026-05-14
Change 'serialize' to 'linearize' in Note 2 of Consistent Help

bruce-usab:
Change [serialize](https://www.google.com/search?q=define%3Aserailized) to [linearize](https://www.google.com/search?q=define%3Alinearize) in Note 2 of SC 3.2.6 Consistent Help.

Use of the incorrect word caused some confusion on WCAG2ICT call today, see [minutes](https://www.w3.org/2026/05/14-wcag2ict-minutes.html#a866).

(FWIW, since I looked it up, this sort of error is a [catachresis](https://en.wiktionary.org/wiki/catachresis).  [Edited to use Wiktionary rather than Google.])


bruce-usab:
> For this success criterion, ”the same order relative to other page content” can be thought of as how the content is ordered when the page is <del>serialized</del> <ins>linearized</ins>.

patrickhlauke:
i believe in computing the terms are often used interchangeably though - serialisation in computing is about breaking down complex data structures into simpler forms that can be transmitted (like converting a complex structure into JSON is referred to as "serialisation" to convert the complex structure into a string)

bruce-usab:
> i believe in computing the terms are often used interchangeably…

Fair point, but for a lay person, [serialized](https://en.wiktionary.org/wiki/serialized) raises the idea that the page is meant to be broken into a series of pages.  Which is almost a coherent idea; enough so to trip someone up.  Regardless, [linearized](https://en.wiktionary.org/wiki/linearized) is closer to plain language.

awkawk:
I stand by serialize as the correct term here. In the Understanding document we wrote: "If the help item is visually in a different location, but in the same serial order, that is not helpful from a user's point of view, but it would not fail this criterion."

If the point is to make this clearer to lay people it should be clarified in Understanding, it is correct in the main document.

patrickhlauke:
@dbjorge mentioned in our call that "serialize" has a very specific and different meaning to what's intended here, even for technical people (indeed, serialization is about converting into a format that can be transmitted and then later reconstructed)

GreggVan:
Having read all this -- I  think linearize is the best term.   Serial has both the serial data  AND the serial as in broken up into pieces/episodes/installments  aspects.   Linearize does not

PR #5130, opened 2026-05-26
Add "also" to Note 1 for 2.2.2 Pause, Stop, Hide

patrickhlauke:
> Note 1: For requirements related to flickering or flashing content, refer to [Guideline 2.3](https://www.w3.org/TR/WCAG22/#seizures-and-physical-reactions).
https://www.w3.org/TR/WCAG22/#pause-stop-hide

The note currently gives the impression that 2.2.2 doesn't apply to flickering or flashing content, and that auditors/authors should *instead* refer to Guideline 2.3. This is incorrect ... it's not either/or. Flickering or flashing content that moves, blinks, scrolls, or auto-updates automatically is still subject to 2.2.2. Then, *in addition*, that content must be checked against 2.3.1/2.3.2

See also https://github.com/w3c/wcag/pull/5131 (the non-normative part of this change)


patrickhlauke:
Decided that this may be unnecessary (and would otherwise require a normative change along the same lines for the note in sensory characteristics). The non-normative note is sufficient to clarify (if it wasn't clear from context) that it's not an either/or relationship.

PR #5131, opened 2026-05-26
Add "also" to note about flashing content in 2.2.2 Pause, Stop, Hide understanding

patrickhlauke:
> "Flashing" content that starts automatically will need to be evaluated against 2.3.2 Three Flashes and 2.3.1 Three Flashes or Below Threshold

Currently the note in https://www.w3.org/WAI/WCAG22/Understanding/pause-stop-hide.html seems to suggest that flashing content should be evaluated against 2.3.1/2.3.2 *instead of* 2.2.2. This is incorrect ... it's not either/or. Flickering or flashing content that moves, blinks, scrolls, or auto-updates automatically is still subject to 2.2.2. Then, in addition, that content must be checked against 2.3.1/2.3.2

In addition, this PR reorders mention of the two SCs to be in numerical order

See also https://github.com/w3c/wcag/pull/5130 (the normative part of this change)


bruce-usab (approved):

HaTheo (approved):
Agree with this change.

PR #5133, opened 2026-05-27
Add clarification to 1.4.11 Non-text contrast understanding about logos acting as user interface components - UIC "inherits" the exemption for logos with insufficient contrast

### LINKED ISSUE #1275: No exception regarding contrasts for logos if they are interactive (SC 1.4.3, 1.4.11)

### LINKED ISSUE #1739: Logo excepted from 1.4.3 Contrast (Minimum), but does it fail as a link for 1.4.11 Non-text Contrast?

### LINKED ISSUE #4376: SC 1.4.11 - Solid colour logo as link

patrickhlauke:
**Counter-proposal to https://github.com/w3c/wcag/pull/5113**

The normative wording for 1.4.11 https://www.w3.org/TR/WCAG22/#non-text-contrast

> The visual presentation of the following have a contrast ratio of at least 3:1 against adjacent color(s):
> 
> **Graphical Objects**

has two distinct aspects/subjects - user interface components (UICs) and graphical objects.

Logos that have insufficient non-text contrast are exempted when they act as graphical objects, as their (low contrast) colours fall under the "except when a particular presentation of graphics is essential" clause.

However, when logos are the only content inside a UIC, the logo itself may be exempt, but the UIC around it has no explicit exemption. A button/control only containing a logo with insufficient contrast, and no other sufficiently contrasting visual indicator, normatively doesn't appear to be exempt.

This PR handwaves this and says that the exemption for the logo carries over/is inherited by the UIC

---

Follow-up to https://github.com/w3c/wcag/pull/4402


detlevhfischer:
scratches head - why do you thumb-down your own proposal?

Personally, I still see the normative wording can be read in two ways depending on how you parse it, and the conclusion that logos once linked must meet contrast requirements is not quite convincing. I think @awkawk has a point when noting that logos frequently act as links to the root page. The normative text simply does not address either way the conflict between the essential nature of logos and the requirement for UICs, and is in my view therefore an insufficent basis for the proposed change.

patrickhlauke:
> scratches head - why do you thumb-down your own proposal?

because personally, i don't like it/agree with it...

> the conclusion that logos once linked must meet contrast requirements is not quite convincing

the conclusion (from the other PR) is that the logo is exempt - that doesn't change. but the UIC itself needs something else that has sufficient contrast to make it actually perceivable - whether it's using a variant of the logo, a border, an extra line, something adjacent...

understanding/21/non-text-contrast.html
```diff
@@ -412,8 +412,10 @@ <h4>Logos</h4>
 
 				<p>Logos are exempted from contrast requirements, under the assumption that they must comply with stricter color choices mandated
 				by corporate identity or brand guidelines. However, this can be problematic when logos act as <a>user interface components</a>
-				(such as a link or other interactive control). In these cases, as a best practice, authors should consider choosing a variant of
-				the logo that has sufficient contrast, if allowed by the corporate identity or brand guidelines.
+				(such as a link or other interactive control). While the exemption only explicitly applies to logos as <em>graphical objects</em>,
+				<em>user interface components</em> whose only visible aspect is a logo are <strong>also exempted</strong>.
```
detlevhfischer, line 662: > While the exemption only explicitly applies to logos as graphical objects, user interface components whose only visible aspect is a logo are also exempted.

I find this wording unclear and the construction confusing. 

The wording also leaves a gap considering the many cases where a logo has a textual byline that is part of the same graphic, and is exempt via the 1.4.3. exception "Text that is part of a logo or brand name has no contrast requirement." Given the suggested wording, would the logo, used as UIC, **not** fall under the exception when there is more than mere graphics in the logo image?

I would suggest something simpler: "Logos fall under the essential exception. They are exempt even when used as user interface element." (Followed, of course, by the strong recommendation to use a contrasty version whenever possible.)

detlevhfischer (changes requested):

understanding/21/non-text-contrast.html
awkawk: You beat me to it @detlevhfischer. The phrasing seems like it was written to convey the author's disagreement with the words being written. :)

I suggest:
`<p>Logos are exempted from contrast requirements, both as graphics and when used as a <a>user interface component</a>, under the assumption that they must comply with stricter color choices mandated by corporate identity or brand guidelines. As a best practice, authors should consider choosing a variant of the logo that has sufficient contrast, if allowed by the corporate identity or brand guidelines.</p>`

This is an overreach and should be removed:
`<p>If logos are presented with an insufficient contrast, but their presentation was an author choice rather than	being mandated by corporate identity or brand guidelines, then that particular low contrast presentation is <em>not</em> "essential", and the logo is <em>not</em> exempt from the contrast requirements.</p>`
				
`<figure id="figure-low-contrast-logos-author-choice">
					A heading: 'ACME tool is trusted by the world's most innovative companies.'; below the heading, a series of nine company logos, all presented as very dark grey logos on the page's black background, with very low contrast; one logo is hovered with the mouse, and is displayed in white, with high contrast against the page background.
					<figcaption>An author chooses to present company logos with low contrast by default, until they are hovered; the fact that these are logos doesn't exempt this scenario from failing the requirements of this success criterion, as the initial low contrast presentation is not "essential"</figcaption>`

I suggest also adding:
`<p>While logos are exempted, it is worth noting that <a>user interface components</a> containing logos (such as a link to a company home page) are not exempted from other aspects of this success criterion, such as the requirement to have sufficient contrast for the focus indicator for the control.</p>`

patrickhlauke: > The phrasing seems like it was written to convey the author's disagreement with the words being written. :)

The words were trying hard to provide an explanation for a handwavy rationale that is not in fact present in the normative wording...

> This is an overreach and should be removed:

That was part of a previous PR, not part of this PR. And that PR already passed the review period with support from other AGWG members https://github.com/w3c/wcag/pull/4402

> I suggest also adding:

Good suggestion.

awkawk: > That was part of a previous PR, not part of this PR. And that PR already passed the review period with support from other AGWG members https://github.com/w3c/wcag/pull/4402

Still problematic.

bruce-usab (approved):

gundulaniemann:
If a user cannot see the graphical element and there is no indication that there is something interactive, they are disadvantaged and miss a functionality.
It is the same kind of disadvantage as for a keyboard user if the element is not focusable and the functionality is not available for them (like via an alternative), or for a screen reader user if the element is not described via screen reader.

understanding/21/non-text-contrast.html
```diff
@@ -412,8 +412,10 @@ <h4>Logos</h4>
 
 				<p>Logos are exempted from contrast requirements, under the assumption that they must comply with stricter color choices mandated
 				by corporate identity or brand guidelines. However, this can be problematic when logos act as <a>user interface components</a>
-				(such as a link or other interactive control). In these cases, as a best practice, authors should consider choosing a variant of
-				the logo that has sufficient contrast, if allowed by the corporate identity or brand guidelines.
+				(such as a link or other interactive control). While the exemption only explicitly applies to logos as <em>graphical objects</em>,
```
hidde, line 661: I find 'exemption only explicitly applies to' confusing word choice; the exemption is 'when a particular presentation of graphics is [essential](https://www.w3.org/TR/WCAG22/#dfn-essential) to the information being conveyed'.

As I understand it 'essential' implies logos; this is implicit, not explicit.

patrickhlauke:
Superseded by https://github.com/w3c/wcag/pull/5250

PR #5135, opened 2026-05-27
Add a non-normative explanation of "proper name" to 3.1.2 Language of Parts understanding

### LINKED ISSUE #4094: 3.1.2 Lang of Parts exceptions need to be Key Terms

patrickhlauke:
Supersedes https://github.com/w3c/wcag/pull/5098 - adding a generalised explanation of what "proper names" are in the 3.1.2 understanding document, rather than attempting to make a normative term definition (allowing for some more handwaving).

Preview: https://deploy-preview-5135--wcag2.netlify.app/understanding/language-of-parts#intent (just after the big Note)

Relates to https://github.com/w3c/wcag/issues/4113 and https://github.com/w3c/wcag/issues/1174
Closes https://github.com/w3c/wcag/issues/4094


giacomo-petri:
Easy for you! your name is almost always pronounced correctly... Are you asking for my review because my name keeps getting misspelled as “Ghiacomo” or “Gaicomo”? 😄 

Jokes aside, and setting aside the fact that that normative wording doesn't perfectly reflect how things work in practice (anyway, this is what the normative wording says), I think your change does a good job of clarifying the concept of "proper name"!

giacomo-petri (approved):

chaals:
Gi' a comma, I think this is a net improvement so we should accept it and refine if we have the energy.

That said my "inner pedant" (almost as big as the "outer me") wants to refine this a little further. It isn't so much the concept and definition of proper names that changes across languages, as how they are treated. Which is pretty variable even in English.

Beginning with pronunciation, because that is what was behind the original proposal during the development of WCAG 1 in the 90s, I would *like* the statement to be clearer that proper names are treated according to somewhat unpredictable customs:

Ian Jacobs (sometime WCAG editor and participant in this group) and Ian Fleming (who wrote the original James Bond novels) share the same first name, but pronounced differently in english (the relevant native language in both cases).

"Albert Camus" in common english usage is typically spoken with the english pronunciation of "Albert" and the (approximate*) French pronunciation of "Camus" - WIkipedia follows (or substantiates) that, giving the French pronunciation of his surname only in its English language entry. And notably, one of his most famous works, "L'Étranger" is known by two different names in English - "The Stranger" or "The Outsider".

In English "Ypres" is often pronounced "wipers", but something like "iprhe" in French (the original language) while Cannes is mixed between "can" and "cans", and Calais is generally "kalay".

Although as noted there are also a lot of unpredictable rules for proper names in translation, I think we don't need to get into that aspect. But if we want to do so then it matters to make sure what we say is correct. Again, I'm inclined to go with something like "there are a lot of unpredictable conventions, rather than simple rules".

Proper names can *often* be used across languages without translation. The simplest counter-examples are probably well-known places. London and København are often translated in nearby languages, Roma and Paris less so, Buenos Aires and Oslo rarely if at all. At the same time, being well-known proper names these are widely recognised by people and software in practice. (In particular, English forms are generally recognised in non-english languages even if speakers of those languages would use a different form by default).

In Spanish "Carolus III Rex" - King Charles of various commonwealth countries - is generally known as "Rey Carlos", (son of "la Reina Isabel") but the french musician Charles Aznavour is commonly known as Charles (and so am I in official dealings, and the US musician Ray Charles goes by the same name in Spanish and English).

* the French "y" sound (similar to German "ü" is one of the few that I don't know of in English - a language that includes one of the widest range of sounds, so we typically approximate it with something like the "oo" in "too".

patrickhlauke:
@chaals if either your inner or outer pedant can propose changes to the PR itself (without turning it into a treatise), by all means. my main concern though here is trying to address the question of "would a site fail (be 'illegal'/'break the law' in places that require 'WCAG conformance') for not putting a `lang="..."` around this thing that we'd consider a proper name", not necessarily the trickier aspect of "how to guarantee a consistent and language-appropriate vocalisation/announcement by SRs"

chaals:
I didn't have time (yet) to trim my treatise into a PR I liked. Which is 
how I got to the conclusion (see the first line: We should adopt this).

I'm on board with you - the main concern is explaining what you need to do 
to conform to WCAG (and why that is what you need to do).

understanding/20/language-of-parts.html
```diff
@@ -50,6 +50,14 @@ <h2>Intent of Language of Parts</h2>
          <p>As these problems are caused by limitations of user agents and assistive technologies, and are outside of an author's control, they <em>don't fail</em> the requirements of this success criterion. When evaluating content against this success criterion, auditors can therefore not solely rely on the spoken output from assistive technologies, but must verify whether or not changes in natural language have been identified correctly in the underlying code or markup.</p>
       </div>
 
+      <p>While the concept and definition of <em>proper names</em> varies across languages,
+      it generally refers to a name which is taken to distinctly identify its referent – including,
+      but not limited to, the name of a specific person, place, or product. Proper names can be used
+      across languages without translation and without the need to change writing systems. Proper names
```
adampage:
```suggestion
      across languages without translation. Proper names
```
Proper names _often_ need to change writing systems to be understood though, right?

E.g., an English speaker can better understand “桐生 一馬” when it’s transliterated to “Kazuma Kiryu”.

adampage (approved):

understanding/20/language-of-parts.html
patrickhlauke: that bit was a last second addition ... but i probably didn't think it through. i WAS thinking about the transliteration case, but probably got myself confused about what i'd want to pass or not without a `lang`. probably closer to what i had in mind was "without translation (though it may be transliterated from its original writing system)"...

patrickhlauke:
```suggestion
      across languages without translation (though it may be transliterated from its original writing system). Proper names
```

patrickhlauke: i think with that in place, the logic would be that as kiryu-chan's name is a proper name, you can have an english sentence that includes 桐生 一馬 or its transliteration Kazuma Kiryu *without* needing to use `lang="jp"`

adampage: Yes, got it, and agree. 🚀 

Author hat on, I could not abide _omitting_ the lang tag from `<span lang="ja-Latn">Kazuma Kiryu</span>`, but I accept that WCAG shouldn’t necessarily require it. 😅

patrickhlauke: yeah same, it's a slippery slope though, which is the problem (as with many WCAG SCs and their binary nature)

understanding/20/language-of-parts.html
adampage, line 74:
```suggestion

      <div class="note">
         <p>
            Pronunciation quality and accuracy can vary significantly across
            screen readers and speech synthesizers.
         </p>
         <p>
            Although a borrowed French word like “rendezvous” may be pronounced
            intelligibly in an English context, others may not. This can be
            especially true of <a
            href="#:~:text=While%20the%20concept,being%20in%20French.">proper
            names</a>.
         </p>
         <p>
            Though <strong>not normatively required</strong> to satisfy this
            success criterion, it’s a best practice to programmatically indicate
            the source language of these words to encourage accurate
            pronunciation and improve understandability. For example, using the
            <code>lang</code> attribute in HTML:
         </p>
         <pre><code class="language-html">It can be a &lt;span lang="fr"&gt;faux pas&lt;/span&gt; to omit the &lt;code&gt;lang&lt;/code&gt; attribute.</code></pre>
         <p>
            When the language of the part uses a different writing system (or a transliteration)
            than the language of the page, explicitly specifying the source language is
            especially helpful for accurate pronunciation:
         </p>
         <pre><code class="language-html">Good &lt;span lang="zh-Latn-pinyin"&gt;feng shui&lt;/span&gt; begins with good markup.</code></pre>
      </div>

```

In addition to your clarifications, how about a “best practice” note to nudge authors toward good hygiene?

A screencap showing how the suggested code and text above renders in a web browser, with nothing essential communicated by style.

☝🏻  Here’s a screencap of how the suggestion renders.

(Edit from Patrick: sorry, as I tweaked - with Adam's permission - the above code suggestion a bit more, the screencap is now not accurate)

bruce-usab (approved):
Nice job squaring the circle!

understanding/20/language-of-parts.html
```diff
@@ -50,6 +50,14 @@ <h2>Intent of Language of Parts</h2>
          <p>As these problems are caused by limitations of user agents and assistive technologies, and are outside of an author's control, they <em>don't fail</em> the requirements of this success criterion. When evaluating content against this success criterion, auditors can therefore not solely rely on the spoken output from assistive technologies, but must verify whether or not changes in natural language have been identified correctly in the underlying code or markup.</p>
       </div>
 
+      <p>While the concept and definition of <em>proper names</em> varies across languages,
+      it generally refers to a name which is taken to distinctly identify its referent – including,
+      but not limited to, the name of a specific person, place, or product. Proper names can be used
```
patrickhlauke:
```suggestion
      but not limited to, a product's brand name or the name of a specific person or place. Proper names can be used
```

bruce-usab (approved):

PR #5175, opened 2026-06-20
Clarify 1.4.13 Content on Hover or Focus understanding explanation for "Dismissible"

### LINKED ISSUE #1409: Incomprehensible paragraph in chapter "Dismissable" (Understanding SC 1.4.13)

patrickhlauke:
The explanation provided in https://www.w3.org/WAI/WCAG22/Understanding/content-on-hover-or-focus.html#dismissible is slightly confusing - the problem is arguably not stated clearly enough, and in the second half it starts to muddy the waters with discussions on when different approaches are valid or not.

This PR attempts to make it clearer, and removes the confusing talk about small and large sized targets.

Closes https://github.com/w3c/wcag/issues/1409
Supersedes https://github.com/w3c/wcag/pull/5030

Preview: https://deploy-preview-5175--wcag2.netlify.app/understanding/content-on-hover-or-focus#dismissible

[Diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fwcag%2Funderstanding%2Fcontent-on-hover-or-focus&doc2=https%3A%2F%2Fdeploy-preview-5175--wcag2.netlify.app%2Funderstanding%2Fcontent-on-hover-or-focus)


bruce-usab (approved):
Much better!

bruce-usab:
Reviewed (and approved) on Backlog call 26 June).

understanding/21/content-on-hover-or-focus.html
```diff
@@ -42,15 +42,12 @@ <h2>Intent</h2>
...
+			<p>The condition includes two exemptions where being able to dismiss the additional content without moving pointer hover or keyboard focus is not required:</p>
+			<ul>
+				<li>The additional content communicates an input error message: these can persist without being dismissible, as they generally require the user's attention, explicit confirmation or remedial action.</li>
+				<li>The additional content does not replace or obscure other content (including the trigger element itself): if the additional content only appears over white space or purely decorative content, such as a background graphic which provides no information, it won't interfere with a user's ability to view or operate the page.</li>
```
fstrr: Should there be a qualifier for "obscure other content"? As currently written, that's going to include purely decorative content, which is then "disqualified" later in the sentence. Would linking to [the definition of content work](https://www.w3.org/TR/WCAG22/#dfn-content)?

patrickhlauke: don't think the definition of content would help here, as that definition leaves it open to decorative content as well. possibly "meaningful content"?

patrickhlauke:
```suggestion
				<li>The additional content does not replace or obscure other meaningful content (including the trigger element itself): if the additional content only appears over white space or purely decorative content, such as a background graphic which provides no information, it won't interfere with a user's ability to view or operate the page.</li>
```

PR #5178, opened 2026-06-24
Add note to 3.1.2 Language of Parts to clarify single words can be phrases/passages

### LINKED ISSUE #1174: 3.1.2 Language of parts - can unmarked single words considered a PASS

patrickhlauke:
Closes https://github.com/w3c/wcag/issues/1174

Preview: https://deploy-preview-5178--wcag2.netlify.app/understanding/language-of-parts#intent

### DIFF
--- a/understanding/20/language-of-parts.html
+++ b/understanding/20/language-of-parts.html
@@ -22,11 +22,17 @@
       <h2>Intent of Language of Parts</h2>
 
       <p>The intent of this success criterion is to ensure that user agents can correctly present
-         phrases, passages, and in some cases words written in multiple languages. This makes it possible for user agents and
+         phrases or passages written in multiple languages. This makes it possible for user agents and
          assistive technologies to present content according to the  presentation and pronunciation
          rules for that language. This applies to graphical browsers as well as screen readers,
          braille displays, and other voice browsers.
       </p>
+
+      <div class="note">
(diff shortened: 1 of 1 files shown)


bruce-usab (approved):

bruce-usab:
Reviewed and discussed on Backlog call 26 June.  Approved without further edits.

Issue #3950, opened 2024-07-08
Unclear if focus indicator counts when determining if the focus is obscured


rachaelbradley:
2.4.11 and 2.4.12 reference the "user interface component" but the definition of this does not explicitly include or exclude the focus indicator as part of the component. Assuming the intent is that the focus indicator is not part of the component, some of the language in the understanding documents makes this unclear. The understanding documents also do not explicitly state that an external focus indicator should not count towards focus not obscured. 

This causes an issue both in passing 2.4.11 and in failing 2.4.12.

detlevhfischer:
I think it is pretty clear that the the SC normative text, "When a [user interface component](https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-minimum.html#dfn-user-interface-component) receives keyboard focus...",  the UIE is the element that the SC focuses on, not whatever focus indication is used on it. 

Whenever the focus is some outline around the UIE, the focus will then also not be fully obscured. The SC misses cases where the focus indicator may be an underline or run inside the perimeter of the UIE. Taking the normative text narrowly, content would then seem to pass if the UIE - say, a link - is partly visible but not its focus (like the underline). I have read through the 2.4.11 Understanding text and see nothing that contradicts this interpretation, and nothing that suggests the focus indicator should be conasidered part of the UIE. 

@rachaelbradley Which bits of understanding text do you think are unclear?

One could certainly strengthen the understanding by advising that neither the UIE not its focus indcation should be fully obscured, but that is not a "must" - that would need a normative change.

patrickhlauke:
the main point of confusion ... is in the name of the SC itself. "focus not obscured" on its face implies that it means the focus indicator, particularly since the spec uses "focus" to mean "focus indicator" in other SCs like "focus visible" and "focus appearance". too late now, of course, but "focused *element* not obscured" would have avoided any misunderstanding...

patrickhlauke:
The understanding for both Focus Not Obscured (Minimum) and Focus Not Obscured (Enhanced) now contain a note that tries to clarify the situation explicitly:

> Note
>
[...]

PR #5183, opened 2026-06-27
Rename Focus Not Obscured SCs to Focused Component Not Obscured

patrickhlauke:

---

The name of these two SCs is confusing, particularly because other SCs (Focus Visible, Focus Appearance) refer to the focus *indicator* with the "Focus" shortform, whereas these SCs refer to the focused *component*.

x-ref https://github.com/w3c/wcag/issues/3950 https://github.com/w3c/wcag/discussions/5194


giacomo-petri (approved):

adampage (approved):

yatil:
YES, PLEASE!

bruce-usab (commented):
This would be a helpful [class 2](https://www.w3.org/policies/process/#correction-classes) editorial change.  The current short name is misleading because the SC is about the _component_, and **not** the _focus_ (border) applied to the component.

patrickhlauke:
@bruce-usab if you think this could be turned into an erratum for 2.2 rather than something set aside for a potential 2.3 ... happy to move it. was just anticipating lots of (usual) pushback.... (i see there's already a thumbs-down, no surprises)

ljkoakley:
I do not agree with changing the name to refer to a component. Web pages contain components but focus can be on a specific area within the component. As an example, the component "grid" with many lines of content with the edit icon at the end of the line. The grid can be partially behind a pop-up or a drawer, if the focus is on the edit icon which is behind the pop-up or drawer, the component does not have focus, the edit icon has the focus.

fstrr:
@ljkoakley The only thing that is changing is the name. The SC has always been about the focused component and that content isn't changing.

bruce-usab:
I remember once hearing that the names of SC are not themselves normative.  I looked for, but couldn’t find, anything about that — or even that “name” is the best way to refer to these handles/labels.  Can anyone hit me with a clue bat?

yatil:
The shortnames (that's what I have been calling them) are in the normative part of the spec, but their goal or task is to summarize the requirement or give it a name. They have usually no bearing on how to interpret the success criterion itself (which are the paragraphs following the shortnames).

So changing them changes normative text, but should never have bearing on the normative interpretation. 

Case in point: despite this SC's shortname referring to the focus not being obscured, the SC text makes it clear that the focused component cannot be obscured. 

NB: If it was on me, I would even do a more drastic change to the shortname: _Focused Component not entirely hidden_

That would put it in line with the SC text, where obscured isn't used.

patrickhlauke:
to @yatil 's point ... I think this may require further discussion when (if) we get to implementing this in (a potential) 2.3. I believe there *is* an intended difference here between "obscured" (meaning covered, overlapped) and the more general "hidden" (which could also include things like messing with the `visibility` or `opacity` of a component, positioning it off-screen, or similar). I believe the original intention was always to just cover the former (overlapped/covered) and not the latter. this is further reinforced by the current wording in the understanding document.

if this is indeed the case, i'd actually flip this suggestion around and say the normative SC wording should be changed to say "obscured".

yatil:
I think (and I maybe remembering that wrong) that the reason for the SC wording came from an inability to properly define what "obscured" meant. Would it only apply to content (web example) that has a higher z-index, for example. I (might falsely) remember that using hidden in general was much easier to define and also more universally useable across technology. 

(That said there is a certain overlap with "Focus visible" currently, which is not necessarily bad.)

PR #5184, opened 2026-06-28
Avoid focus outline "hugging" container

patrickhlauke:
When navigating with the keyboard, jumping to a term (in the main spec) or following a section link (in understanding) shows the focus outline uncomfortably hugging the container, making it look amateurish/like a mistake. This PR adds a bit of breathing space, making it look intentional.

Before (in Chrome/Win):

a focused glossary term, with the hugging outline

a focused understanding section, with the hugging outline

---

After (in Chrome/Win):

a focused glossary term, with some offset to provide breathing room

a focused understanding section, with some offset to provide breathing room

### DIFF
--- a/css/base.css
+++ b/css/base.css
@@ -537,6 +537,6 @@ pre {
 dfn:focus {
     outline-offset: 0.25em;
 }
-main:focus, aside:focus, section:focus, div:focus {
+:is(main,aside,section,div):focus {
     outline-offset: 0.75em;
 }
--- a/css/base.css
+++ b/css/base.css
@@ -540,3 +540,13 @@ dfn:focus {
 :is(main,aside,section,div):focus {
     outline-offset: 0.75em;
 }
+
+[id][tabindex="-1"]:is(main,aside,section,div) {
+  > :first-child {
+    padding-block-start: 0.5em;
+  }
+
+  > :last-child {
+    padding-block-end: 0.5em;
+  }
+}
--- a/css/base.css
+++ b/css/base.css
@@ -532,3 +532,11 @@ pre {
     padding: 0 0.25em;
     display: inline-block;
 }
+
+/* avoid focus outline "hugging" container */
+dfn:focus {
+    outline-offset: 0.25em;
+}
+main:focus, aside:focus, section:focus, div:focus {
+    outline-offset: 0.75em;
+}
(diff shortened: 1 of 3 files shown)


giacomo-petri:
On Chrome and Safari, the focus indicator looks a bit odd when the container receives focus. The top border follows the heading first and then the rest of the container:

screenshot visualising the focus indicator behaviour mentioned above

This behavior already existed before this PR, so the changes here aren't making it worse. However, since we're already working on the focus presentation, it might be worth addressing it as well.
I believe this is caused by margin collapsing.
A possible fix would be to add 10px of top padding to the heading and remove the corresponding 10px top margin from the container (applied consistently throughout the whole CSS).

bruce-usab:
> On Chrome and Safari, the focus indicator looks a bit odd when the container receives focus…
> This behavior already existed before this PR, so the changes here aren't making it worse.

Just noting that Safari already doesn’t have the too-tight focus outline.  Is the current behavior a bug with Edge?

css/base.css
```diff
@@ -532,3 +532,11 @@ pre {
...
     display: inline-block;
 }
+
+/* avoid focus outline "hugging" container */
+dfn:focus {
+    outline-offset: 0.25em;
+}
+main:focus, aside:focus, section:focus, div:focus {
+    outline-offset: 0.75em;
+}
```
adampage, line 552:
```suggestion
}

[id][tabindex="-1"]:is(main,aside,section,div) {
  > :first-child {
    padding-block-start: 0.5em;
  }

  > :last-child {
    padding-block-end: 0.5em;
  }
}
```

For your consideration, @patrickhlauke  / @giacomo-petri / @kfranqueiro. 🥴 

Before and after:

Screencaps comparing before and after this code suggestion. In the before, two red annotation arrows point to locations where the focus outline poke out and cause the otherwise rectangular shape to be irregular. In the after, the outline is a plain rectangle.

css/base.css
```diff
@@ -532,3 +532,11 @@ pre {
     padding: 0 0.25em;
     display: inline-block;
 }
+
+/* avoid focus outline "hugging" container */
+dfn:focus {
+    outline-offset: 0.25em;
+}
+main:focus, aside:focus, section:focus, div:focus {
```
adampage:
```suggestion
:is(main,aside,section,div):focus {
```
...and just because [I was in the neighborhood](https://github.com/w3c/wcag/pull/5184/changes#r3521260967).

adampage (approved):

PR #5187, opened 2026-06-29
Reinstate G200 only as advisory for 3.2.5 Change on Request, strip "secure session" inaccuracy

patrickhlauke:
G200 was completely deleted, rather than being obsoleted. After some discussion, decided to reinstate it, slightly slimmed down, but only as advisory for 3.2.5

Follow-up to https://github.com/w3c/wcag/pull/4955

Preview: https://deploy-preview-5187--wcag2.netlify.app/techniques/general/g200


techniques/general/G200.html
```diff
@@ -0,0 +1,49 @@
...
+    <p class="type">Type: Technique</p>
+  </section>
+  <section id="applicability">
+    <h2>When to Use</h2>
+    <p>Pages that open new windows or tabs</p>
+  </section>
+  <section id="description">
+    <h2>Description</h2>
+    <p>The objective of this technique is to limit the use of links or buttons that open new windows or tabs within web content. In general, it is better not to open new windows and tabs since they can be disorienting for people, especially people who have difficulty perceiving visual content. However there are some situations where it is preferable from an accessibility perspective to open a new window or tab, particularly in cases where it's important for a user not to accidentally lose track of where they are in the current page – for instance, while completing a form.</p>
+    <p>It is recommended that when links will open a new window or tab, they should provide users with an advance warning or hint, such as a "new window" icon.</p>
```
detlevhfischer, line 21: Should there something that states that this information should be available visually **and** programmatically?

detlevhfischer (approved):

techniques/general/G200.html
patrickhlauke: i think it's pretty much implied that all other SCs also still apply. otherwise we'd need to list everything else too (that if they go for an icon for instance, it must have text alternative, must have sufficient non-text contrast, etc)

detlevhfischer: yeah, forget it.

giacomo-petri (approved):

bruce-usab (approved):
[preview diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FTechniques%2Fgeneral%2FG200&doc2=https%3A%2F%2Fdeploy-preview-5187--wcag2.netlify.app%2Ftechniques%2Fgeneral%2Fg200)

baldino-m:
I ilke the re-writing, but example 2 does not really convince me, since there is no real technical reason to open these links in a new window. Although the original example 1 and 2 are very similar, I like these more than the edited example 2

Issue #3661, opened 2024-01-29
"Clearly labeled" in 1.2.1 / 1.2.2 / 1.2.3 is not ... clear


patrickhlauke:
1.2.1 Audio-only and Video-only (Prerecorded), 1.2.2 Captions (Prerecorded), 1.2.3 Audio Description or Media Alternative (Prerecorded), and the definition for "synchronized media" all include the clause that A/V used as a media alternative for text be "clearly labeled as such". Ironically, what actually constitutes a "clear label" is left up to the reader to infer - none of the understanding documents make any mention of this or provide examples.

There is one failure technique, [Failure of Success Criterion 1.2.2 and 1.2.8 due to not labeling a synchronized media alternative to text as an alternative](https://www.w3.org/WAI/WCAG22/Techniques/failures/F74), linked from 1.2.2 Captions (Prerecorded)'s understanding, that initially seems to provide some information, but it's just a tautology - the failure is not having a clear label, and the test process is "check that it's clearly labeled". Not...very informative or useful.

While I doubt this can be normatively nailed down fully, it would be nice to at least provide some indication of what "clearly labeled" actually means, in practice.

PR #5192, opened 2026-07-02
Update F74.html

detlevhfischer:
* Tried to give an idea of what is meant by a clear label (= one that clearly references the corresponding text section)
* Changed the tax form example to a registration form instruction as part of a larger set of service instructions (so that the difficulty of associating the synchronized media alternative for that one part is made more obvious when the video is at the bottom of the entire set of instructions)

### DIFF
--- a/techniques/failures/F74.html
+++ b/techniques/failures/F74.html
@@ -1,19 +1,19 @@
 <!DOCTYPE html><html lang="en"><head><title>Failure of Success Criterion 1.2.2 and 1.2.8 due to not labeling a synchronized media alternative to text as an alternative</title><link rel="stylesheet" type="text/css" href="../../css/sources.css" class="remove"></head><body><h1>Failure of Success Criterion 1.2.2 and 1.2.8 due to not labeling a synchronized media alternative to text as an alternative</h1><section class="meta"><p class="id">ID: F74</p><p class="technology">Technology: failures</p><p class="type">Type: Failure</p></section><section id="applicability"><h2>When to Use</h2>
       <p>Pages that provide synchronized media alternatives to text.</p>
    </section><section id="description"><h2>Description</h2>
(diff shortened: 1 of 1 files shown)


patrickhlauke:
Related to https://github.com/w3c/wcag/issues/3661 ?

detlevhfischer:
Yes, I now added closes... above.

patrickhlauke:
might be worth extending this to add something in the understanding documents for all relevant SCs, not just relying on shoving something in a technique (where nobody bothers to look for interpretation)

patrickhlauke:
While this PR makes a few nice little edits, i don't think it fully closes #3661 (partly because the concept is still a bit vague, and partly because this shouldn't just be hidden in a technique, but surfaced in the understanding documents at least - particularly if the concept is rather handwavy/vague).

techniques/failures/F74.html
```diff
@@ -1,24 +1,24 @@
...
          
       </section>
    </section><section id="tests"><h2>Tests</h2>
       <section class="procedure"><h3>Procedure</h3>
          <ol>
             <li>Check pages that provide synchronized media alternatives to text.</li>
-            <li>Check that synchronized media is clearly labeled with the text for which it is an alternative.</li>
+            <li>Check that synchronized media has a text label that clearly indicates the part of the text for which it is an alternative.</li>
```
patrickhlauke: does it necessarily need to be a "text label"? what does a "text label" for a video even look like/what is it? is it text that precedes or follows a video? the video's title that is shown on the initial paused/stopped video thumbnail? just the text alternative behind the video element?

detlevhfischer: I guess the video label would need to be text to reference a particular section for which it is an alternative, yes -- it might just be something like "Part 3 - video" or similar. Leaving open whether the label is above or below the video does not preclude either position, we don't want to be too prescriptive here I guess. Adjaceny would be critical though. I think the label would need to be available for all users next to the video before playing, so sighted users not good at reading longer text (for language skills or cognitive reasons) can make the connection.

PR #5229, opened 2026-07-24
Chore: run all `.png` images through `pngcrush`

patrickhlauke:
No image data is lost, just IDAT information is reduced/compressed


patrickhlauke:
admittedly not an enormous saving, but ... as i was already compressing some `.png`s from a PR recently, thought i might just do it wholesale (using `find -name "*.png" -exec pngcrush -ow -brute {} \;`)

kfranqueiro (approved):
Ran this by Kevin and we are both fine with running this. While it will add to `.git` folder filesize since git can't diff binary files, some of that is gained back when considering the reduction in filesize in the current working branch.

patrickhlauke:
Lovely, thanks @kfranqueiro and @iadawn

Issue #2750, opened 2022-10-25
Are there circumstances in which an overlay tool could be seen as providing a conforming alternate version for a non-conforming website?


patrickhlauke:
### Discussed in https://github.com/w3c/wcag/discussions/2749
[**Edit** to add [TF response comment](https://github.com/w3c/wcag/issues/2750#issuecomment-4927755569) below.]

<div type='discussions-op-text'>

<sup>Originally posted by **detlevhfischer** October 25, 2022</sup>
Public bodies increasingly fall for promises of accessibility overlay providers that putting an overlay on their sites will make them WCAG 2.1 conformant in a matter of hours. In all cases we know of this is not true, and the promise grossly misleading. Moreover, most overlay tools we have investigated have themselves some or many accessibility barriers in the settings dialog offered to users for customisation of the site. But this is not the topic I would like the WG to clarify here.

This question is this: 
**In the (unlikely) case that an overlay tool is actually able to successfully remedy all WCAG fails of a site via a specific set of user settings, can this state be considered a conforming alternate version (CAV)?**

Looking at the Conformance requirements and the definition of CAV, the situation seems a bit fuzzy.

Section [5.2. Conformance Requirements](https://www.w3.org/TR/WCAG22/#conformance-reqs) talks about one specific version ("…or a conforming alternate version is provided")  and [the definition of conforming alternate versions](https://www.w3.org/TR/WCAG22/#dfn-conforming-alternate-versions) also talks about "the conforming alternate version" (singular) . One note explicitly qualifiies "One version would need to be fully conformant in order to meet conformance requirement 1."

Where it gets fuzzy is with [the last note of the CAV definition](https://www.w3.org/TR/WCAG22/#h-note-72), which says:
> "Setting user preferences within the content to produce a conforming version is an acceptable mechanism for reaching another version as long as the method used to set the preferences is accessibility supported."

There is no stipulation that just **one** switch (say, a style switcher typically used to load a more contrasty version of a page) or **one** link to a CAV is mandated. However, I would argue that without an explicit label or an explicit description in the overlay interface what settings users would need to change to produce the CAV, the mechanism cannot be considered accessibility supported (arguably failing [2.4.6 Headings and Labels](https://www.w3.org/TR/WCAG22/#headings-and-labels)).

[...]

PR #5236, opened 2026-07-27
Add extra (non-normative) note to the definition of "conforming alternate version"

patrickhlauke:

The point of this new note is already implicit in note 4 and note 7, but this makes it even more explicit.

Relates to https://github.com/w3c/wcag/issues/2750

Preview: Note 8 in https://deploy-preview-5236--wcag2.netlify.app/guidelines/#dfn-conforming-alternate-versions


bruce-usab:
Editorial suggestion (as I found “if the content” opener awkward); current:
> If the content provides a number of settings or user preferences that, individually, only address one particular aspect of the content (for instance,

Proposed:
> Where a number of settings or user preferences are provided that, individually, only address one particular aspect of the content (for example,

bruce-usab:
Approved by TF on backlog call 7/31.  Bruce flagged that [Understanding Conformance](https://www.w3.org/WAI/WCAG22/Understanding/conformance#techsaltver) should probably be lightly refreshed to say something about author provided settings/preferences.

awkawk:
-1 as this can easily be handled in Understanding.

patrickhlauke:
so we'd have 7 existing notes that clarify the concept, but the 8th one would be better migrated somewhere completely different where readers might not even look? hmmm... that seems a very weak rationale...

awkawk:
So we should move anything that anyone thinks is easier to find into the normative document? The understanding document can be changed today. Even if this is accepted it won't be seen by the people who need it until, and if, there is an update to WCAG 2.x.

patrickhlauke:
we could do both of course :) (though it would still feel extremely weird to have 7 notes, and then one that is similar but new offshored somewhere completely separately...)

unless the MO is of course "let's never change the face of the spec ever again"

patrickhlauke:
Noting that this may need to be delayed following the reopening of https://github.com/w3c/wcag/issues/2750

PR #5219, opened 2026-07-13
Fix grammar and punctuation issues across documentation

Gabriel-Dalton:
## Summary
This PR corrects various grammar and punctuation errors throughout the WCAG documentation, including articles (a/an), comma placement, and spelling mistakes.

## Key Changes
- **Article corrections**: Fixed incorrect use of "an" to "a" (and vice versa) in multiple files:
  - "an star rating pattern" → "a star rating pattern" (ARIA10.html)
  - "an graphical arrow" → "a graphical arrow" (F31.html)
  - "an non-color way" → "a non-color way" (F81.html)
  - "an native onclick event" → "a native onclick event" (G212.html)
  - "an short text alternative" → "a short text alternative" (G94.html, G95.html)
  - "a item number" → "an item number" (PDF21.html)
  - "an non-drag alternative" → "a non-drag alternative" (dragging-movements.html)
  - "a insertion point" → "an insertion point" (focus-appearance.html)

- **Comma additions**: Added missing commas after introductory clauses in G15.html (2 instances)

- **Word choice corrections**:
  - "an corporate" → "a corporate" (G184.html)
  - "an corporate human resources information" → "a corporate human resources information system" (G184.html)
  - "the a Flash content" → "the Flash content" (FLASH17.html)
  - "origin an destination" → "origin and destination" (G177.html)
  - "have be previously been" → "have previously been" (G100.html)
  - "no affect" → "no effect" (SL27.html)
  - "is use" → "is to use" (SL4.html)
  - "by them selves" → "by themselves" (link-purpose-link-only.html)

## Implementation Details
All changes are documentation-only corrections to improve readability and grammatical accuracy. No functional code changes are included.

PR #5238, opened 2026-07-29
Fix repeated word, run-together sentences and punctuation slips

Gabriel-Dalton:
Eight editorial fixes in the Techniques and Understanding pages. All informative documents, nothing under `guidelines/` is touched, so no normative change and no erratum needed.

This deliberately does not overlap with #5219. Everything here is in a different file or a different sentence.

| File | Was | Now |
| --- | --- | --- |
| `techniques/aria/ARIA17.html` | see the the ARIA Authoring Practices | see the ARIA Authoring Practices |
| `techniques/failures/F36.html` | This is a  example | This is an example |
| `techniques/general/G56.html` | how it should be done.. | how it should be done. |
| `techniques/general/G99.html` | applications.The retrievable | applications. The retrievable |
| `techniques/general/G131.html` | the control.The label | the control. The label |
| `techniques/general/G174.html` | both Success Criterion 1.4.3 ... and 1.4.6 | both Success Criteria 1.4.3 ... and 1.4.6 |
| `understanding/20/captions-prerecorded.html` | woodsmen.." | woodsmen." |
| `understanding/20/timing-adjustable.html` | a users ability | a user's ability |

On the last one: the same sentence already writes "the user's ability" a few words earlier, so the missing apostrophe reads as a slip rather than a house style.

G174 is the only one that is arguably not a typo. The sentence says the technique "would satisfy both Success Criterion 1.4.3 Contrast (Minimum) and 1.4.6 Contrast (Enhanced)", and since two criteria are named, the plural fits.


kfranqueiro:
As with #5219, as this appears to have been LLM-generated, I have performed a manual pass, and found/fixed these cases and additional ones like them in #5241.

Note I have intentionally omitted the "Success Criterion/a" change from that PR, as that points to another class of revisions I will pursue in a separate PR.

PR #5241, opened 2026-07-30
Fix various grammatical errors

kfranqueiro:
This supersedes #5219 and #5238, treating it as a bug report warranting a full human-performed pass in which I caught more of the same types of issues.

- Incorrect a/an pairings (incl. cases interrupted by markup)
- Affect vs. effect
- Correct "* self/selves" to one word
- Correct plurality of do not / does not, don't / doesn't
- Correct duplicated articles (e.g. "the a(n)" / "a(n) the")
- Add missing "to" and/or lowercase "technique" in "The {objective,purpose,etc.} of this technique is (verb)"
- Fix doubled punctuation or missing space between punctuation and start of next sentence
- Fix missing apostrophes in possessives
- A few other small wording tweaks I found as I was making these changes


bruce-usab:
Reviewed and approved by TF on backlog call 7/31.

PR #5249, opened 2026-08-03
Add clarification to 1.4.11 Non-text contrast understanding about exempt graphical objects acting as user interface components - UIC needs to have *some* sufficiently contrasting aspect

### LINKED ISSUE #1275: No exception regarding contrasts for logos if they are interactive (SC 1.4.3, 1.4.11)

### LINKED ISSUE #1739: Logo excepted from 1.4.3 Contrast (Minimum), but does it fail as a link for 1.4.11 Non-text Contrast?

### LINKED ISSUE #4376: SC 1.4.11 - Solid colour logo as link

patrickhlauke:
**For a counter-proposal, see https://github.com/w3c/wcag/pull/5250**

**This is a redo of the previous speculative PRs - cleaned up, and expanded to not only include "logos" but more generally any graphical objects that are exempted, like flags and such**

Supersedes https://github.com/w3c/wcag/pull/5113 (which only focused on logos, but the same rationale needs to apply to any other graphical objects - like flags etc - whose contrast/colours can't be changed, but that also act as UICs)

In addition, this PR moves the "Logos" section directly under/into the "Exception examples", where it more logically belongs, fixes some markup indentation, and lowercases some of the headings.

Copying over the context/rationale from https://github.com/w3c/wcag/pull/5113

> The normative wording for 1.4.11 https://www.w3.org/TR/WCAG22/#non-text-contrast
> 
> > The visual presentation of the following have a contrast ratio of at least 3:1 against adjacent color(s):
> > 
> > **User Interface Components**
> > Visual information required to identify user interface components and states, except for inactive components or where the appearance of the component is determined by the user agent and not modified by the author;
> > **Graphical Objects**
> > Parts of graphics required to understand the content, except when a particular presentation of graphics is essential to the information being conveyed.
> 
> has two distinct aspects/subjects - user interface components (UICs) and graphical objects.
> 
> Logos that have insufficient non-text contrast are exempted when they act as graphical objects, as their (low contrast) colours fall under the "except when a particular presentation of graphics is essential" clause.
> 
> However, when logos are the only content inside a UIC, the logo itself may be exempt, but the UIC around it has no explicit exemption. A button/control only containing a logo with insufficient contrast, and no other sufficiently contrasting visual indicator, would fail the UIC clause of 1.4.11.
> 
> This means that authors would need to either choose a different version of the logo (if it exists) that has sufficient non-text contrast, or make sure there is some other visual indicator on the UIC so that people who can't perceive the low-contrast logo can still understand/see that there is a UIC to begin with.
> 
> The PR also adds a clarification/reminder that if there's an accessible alternative (such as a link/control that serves the same purpose) somewhere on the page that DOES have sufficient contrast, that can also be used to pass this criterion.
> 
> ---
> 
> Follow-up to https://github.com/w3c/wcag/pull/4402 - this PR contains the more controversial aspect that clarifies if a logo is used as a user interface component, then the UIC *does* need to have a sufficiently contrasting element so that users can perceive it.
> 
> > For 1.4.11 there is a wrinkle because the SC distinguishes between user interface controls and graphical objects. Logos on their own count as graphical objects, and if the CI/brand guidelines stipulated non-contrasting colours, then that is "essential". However, the "user interface control" part of a link/button with a logo is still subject to the requirements of the UIC needing to be identifiable.

Preview: https://deploy-preview-5249--wcag2.netlify.app/understanding/non-text-contrast

Diff: https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FUnderstanding%2Fnon-text-contrast.html&doc2=https%3A%2F%2Fdeploy-preview-5249--wcag2.netlify.app%2Funderstanding%2Fnon-text-contrast

### DIFF
--- a/understanding/21/non-text-contrast.html
+++ b/understanding/21/non-text-contrast.html
@@ -180,7 +180,7 @@
 					</figcaption>
 				</figure>
 
-		        <p>Note that this success criterion does not directly compare the focused and unfocused states of a control - if the focus state relies on a change of color (e.g., changing <em>only</em> the background color of a button), this success criterion does not define any requirement for the difference in contrast between the two states.</p>
+				<p>Note that this success criterion does not directly compare the focused and unfocused states of a control - if the focus state relies on a change of color (e.g., changing <em>only</em> the background color of a button), this success criterion does not define any requirement for the difference in contrast between the two states.</p>
 
 				<figure id="figure-focus-background">
 					{% include "./img/1.4.11-ntc-focus-background.svg" %}
@@ -197,7 +197,7 @@
 				</section>
 
 				<section>
-					<h3>User Interface Component Examples</h3>
+					<h3>User Interface Component examples</h3>
 					<p>For designing focus indicators, selection indicators and user interface components that need to be perceived clearly, the following are examples that have sufficient contrast.</p>
 
 					<h4>Passing User Interface Component Examples</h4>
@@ -410,7 +410,7 @@
 
 					</div>
 
-					<h4>Failing User Interface Component Examples</h4>
+					<h4>Failing User Interface Component examples</h4>
 
 					<p>The following are examples that have insufficient contrast.</p>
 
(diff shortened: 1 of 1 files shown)


patrickhlauke:
While my original preference was for this particular interpretation, as it follows exactly from the normative wording (if you think it through), there are some additional wrinkles that this brings up:

* nominally, the "text" part of a logo falls under the regular (text) Contrast SC, which does not have this dual "text and UIC" aspect; admittedly, the more fundamental problem is the actual thinslicing of a logo into its text and non-text parts - arguably, a logo should be considered as a whole single "thing" (particularly for logos where the "text" part is heavily stylised, to the point where it's almost a non-text representation); nonetheless, the current separation would cause an additional headache - what if the non-text part of a logo has low contrast, and is exempt, and the text part of a logo is exempt from the Contrast SC, but the logo (non-text + text) is the only content of a UIC?
* as @giacomo-petri pointed out at the time, the Non-text Contrast SC has an additional clause/exemption for styling that is down to the user agent/not changed by the author; in current browsers (not sure when this change happened in the default browser stylesheets) if you have an image wrapped in a link, the image has no default border...so having a `<a href="..."></a>` would, by default, not show any outline/border around it...

patrickhlauke:
A further aspect I'd like to address (both here and in the other PR) is the weird heading structure of the current understanding document...

understanding/21/non-text-contrast.html
```diff
@@ -736,10 +716,32 @@ <h4>Essential Exception</h4>
...
+
+					<p>If exempted graphical objects (such as logos) are presented with an insufficient contrast, but their low contrast presentation
+					was an author choice, then that particular low contrast	presentation is	<em>not</em> "essential", and the graphical object
+					is <em>not</em> exempt from the contrast requirements.</p>
+
+					<figure id="figure-low-contrast-logos-author-choice">
+						<img src="img/1.4.11-ntc-author-choice-logos.png" alt="A heading: 'ACME tool is trusted by the world's most innovative companies.'; below the heading, a series of nine company logos, all presented as very dark grey logos on the page's black background, with very low contrast; one logo is hovered with the mouse, and is displayed in white, with high contrast against the page background.">
+						<figcaption>An author chooses to present company logos with low contrast by default, until they are hovered; the fact that these are logos doesn't exempt this scenario from failing the requirements of this success criterion, as the initial low contrast presentation is not "essential" (since it's not mandated by corporate identity or brand guidelines)</figcaption>
```
bruce-usab, line 739:
```suggestion
						<figcaption>An author chooses to present company logos with low contrast by default, until they are hovered; the fact that these are logos doesn't exempt this scenario from failing the requirements of this success criterion, as the initial low contrast presentation is not "essential" (since it's not mandated by corporate identity or brand guidelines).</figcaption>
```

understanding/21/non-text-contrast.html
```diff
@@ -771,14 +773,14 @@ <h2>Benefits</h2>
 		</section>
 
 		<section id="examples">
-			<h2>Graphical Object Examples</h2>
+			<h2>Graphical Object examples</h2>
```
bruce-usab, line 776:
```suggestion
			<h2>Graphical object examples</h2>
```

bruce-usab:
Preview Diff between [5250 and 5249](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fdeploy-preview-5250--wcag2.netlify.app%2Funderstanding%2Fnon-text-contrast&doc2=https%3A%2F%2Fdeploy-preview-5249--wcag2.netlify.app%2Funderstanding%2Fnon-text-contrast#essential-exception) at Essential Exception.

patrickhlauke:
As discussed at last week's meeting, I am going to close this PR, as unfortunately the fact that there is no matching UIC logic for (text) contrast, this would cause inconsistencies for logos that are purely text / logotypes.

see https://github.com/w3c/wcag/pull/5249#issuecomment-5215099188

Issue #5248, opened 2026-08-03
Tests procedures inaccuracies


giacomo-petri:
Looking at the SC 4.1.2 failure techniques, there is at least one example ([F68](https://www.w3.org/WAI/WCAG22/Techniques/failures/F68#tests)) where the testing procedure provides step-by-step instructions for verifying whether a user interface component has an accessible name. The problem is that this guidance is no longer up to date and duplicates requirements that are already defined in other specifications, such as the HTML-AAM accessible name computation.
For example, while using the placeholder attribute is not the recommended approach, it is still part of the accessible name computation defined by HTML-AAM. However, placeholder is not mentioned anywhere in the F68 testing procedure, making the guidance incomplete and potentially misleading. My suggestion would be to review at least all of the failure techniques and determine whether their testing procedures can instead reference the authoritative specifications where appropriate, or otherwise be rewritten to be more future-proof and avoid duplicating normative guidance.

giacomo-petri:
Linking the PRs that, together, will resolve this issue. I'll update this comment as I create the remaining PRs:
- https://www.w3.org/WAI/WCAG22/Techniques/failures/F68#tests -> https://github.com/w3c/wcag/pull/5253
- https://www.w3.org/WAI/WCAG22/Techniques/failures/F38 - [issue link](https://github.com/w3c/wcag/issues/5265) -> PR TODO
- https://www.w3.org/WAI/WCAG22/Techniques/failures/F39 - [issue link](https://github.com/w3c/wcag/issues/5265) -> PR TODO
- https://www.w3.org/WAI/WCAG22/Techniques/failures/F46 - [issue link](https://github.com/w3c/wcag/issues/5267) -> PR TODO
- https://www.w3.org/WAI/WCAG22/Techniques/failures/F65 - [issue link](https://github.com/w3c/wcag/issues/5272) -> PR TODO

PR #5253, opened 2026-08-03
Update F68 test procedure to rely on computed accessible names

giacomo-petri:
Related to: #5248

This PR updates the "Tests" section of F68 to make the testing procedure more robust and future-proof.

The current procedure attempts to enumerate the possible sources of an accessible name. As a result, it has become outdated and no longer reflects the current Acc name computation algorithm and HTML-AAM. For example, the current procedure does not account for accessible names derived from the placeholder attribute. Instead of duplicating the accessible name computation algorithm, this PR updates the procedure to verify the computed accessible name exposed in the accessibility tree. This aligns the technique with the authoritative specs while reducing the maintenance burden and the risk of future inconsistencies.

The updated procedure also broadens the scope beyond form controls by referring to user interface components exposed in the accessibility tree, making it consistent with the applicability of SC 4.1.2 and the intent of F68.

This change is editorial only and does not modify the intent of F68.

[Preview diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FTechniques%2Ffailures%2FF68&doc2=https%3A%2F%2Fdeploy-preview-5253--wcag2.netlify.app%2Ftechniques%2Ffailures%2Ff68#tests)


patrickhlauke:
not sure this fully closes https://github.com/w3c/wcag/issues/5248 - maybe just related to it (unless we don't find any other failures that include this sort of thing)

giacomo-petri:
description updated

patrickhlauke (approved):

PR #5293, opened 2026-08-08
1.3.5: Mention heuristics, turning off autocomplete for forms, and tweak examples

### LINKED ISSUE #1081: Exception needed for 1.3.5 Identify Input Purpose
This SC1.3.5 has given rise to a conflict of interest regarding login ids and passwords because, in HTML web pages, the method recommended in the Understanding document is to use the autocomplete attribute.

Financial applications, and other sensitive and important systems, do not necessarily want to allow the autocompletion of passwords and ids to their system, for security reasons. It allows people to log on to the system for fraudulent purposes, perhaps at a time when a user has left their system logged out but unattended. In occasional cases designers and security technicians might also want to apply the same reasoning to other data items. 

Of course, browsers ignore you anyway if you set autocomplete to off on a password field (they apparently consider security unimportant in this context), but banks and others have various workarounds to overcome this security risk. The point is that the WCAG should not require this solution that involves autocomplete on passwords, regardless of what browsers might do to the contrary.

It is too late now to change or add an exception to this SC which is part of the WCAG 2.1, to preserve the backwards compatibility of versions. But I do suggest a green Note should be added, immediately following the SC, to except the password and user id fields and, on occasion, any other field as well if security requires it. This will resolve the dilemma posed by the SC for developers trying to achieve both secure logins and WCAG compliance. I suggest a wording of:

> Note: This criterion is usually achieved in HTML using the autocomplete attribute. However, financial and other important systems have an overriding requirement for security of password and id fields used in user authentication, that may take precedence over this accessibility requirement and require that autocompletion not be used. Not providing autocompletion will not be considered a WCAG failure where security requires not doing so for, but not limited to, password and id fields.

### LINKED ISSUE #5213: Add clarification to 1.3.5 Identify Input Purpose the heuristics-based autocomplete is not enough
A common misunderstanding of 1.3.5 is, in broad terms: "my browser does/doesn't do autocomplete ... do i pass/fail based on that?"

Autocompletion in browsers is a *side effect* of 1.3.5, not its purpose in itself.

For instance, browsers will generally use heuristics (in absence of an explicit `autocomplete` value) to make a best guess about what to autofill when something generic like `autocomplete="on"` was set - often taking hints from additional things like `type="password"` or even using language heuristics on the accessible name of a field. These cases *fail* the requirement of the SC, as it wants an explicit, programmatically determinable identification of purpose. While browser autocomplete may work, other uses/benefits of 1.3.5 (such as the - slightly mythical/theoretical - "users can define their own styles for specific fields, with nice big icons, so they're more obvious") won't work.

This should be very prominently clarified in https://www.w3.org/WAI/WCAG22/Understanding/identify-input-purpose.html as it's a misconception we encounter often (latest one here https://web-a11y.slack.com/archives/CJB3KU4F4/p1783587595156429)

In addition, would be good to explore if there's a solid way to suppress autocomplete behaviour on a `<form>` while still having correct `autocomplete="…"` values on inputs, and provide a technique for it, to satisfy the "my client doesn't want autocomplete behaviour, how can i still pass 1.3.5?"

This last part would solve https://github.com/w3c/wcag/issues/1081

(and in general, the "user left their logged-in computer unattended" case would not be any less secure if autocomplete wasn't used ... if the user has an integrated password manager, or a third party password manager, and they left that one unlocked as well, then a malicious user would be able to trigger the autocomplete/autofill from those as well, regardless of whether autocomplete was present in the HTML... the security problem happened at the point where the user left their machine logged-in but unattended, and that's not something we can work around/exempt somehow)

patrickhlauke:
* whether or not a user agent actually autofills is not relevant - what counts is the identification
* browsers autofilling using heuristics doesn't count
* rewrite examples to make it clear that the identification is what matters, not whether or not browsers autofill
* the last example is already covered by the first example.

Closes #5213

Preview: https://deploy-preview-5293--wcag2.netlify.app/understanding/identify-input-purpose
Diff: https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FUnderstanding%2Fidentify-input-purpose.html&doc2=https%3A%2F%2Fdeploy-preview-5293--wcag2.netlify.app%2Funderstanding%2Fidentify-input-purpose


understanding/21/identify-input-purpose.html
```diff
@@ -44,19 +47,18 @@ <h3>Specific Benefits of Success Criterion 1.3.5</h3>
 <section id="examples">
   <h2>Examples of Success Criterion 1.3.5</h2>
   <dl>
-    <dt>A contact form using autofill</dt>
-    <dd>A contact form auto-fills in the fields for name, street, post code, city, telephone number and email address from autofill values stored in the user's browser. Assistive technology can offer a customized way of identifying particular input fields, for example drawing on a set of symbols / icons that is familiar to the user, to communicate the purpose of the fields visually.</dd>
+    <dt>A contact form</dt>
+    <dd>A contact form explicitly identifies the purpose for its various fields: name, street, post code, city, telephone number and email address. This makes it possible for user's browser to then autofill the fields with previously stored values. Assistive technology can offer a customized way of identifying particular input fields, for example drawing on a set of symbols / icons that is familiar to the user, to communicate the purpose of the fields visually.</dd>
```
baldino-m:
```suggestion
    <dd>A contact form explicitly identifies the purpose for its various fields: name, street, post code, city, telephone number and email address. This makes it possible for the user's browser to then autofill the fields with previously stored values. Assistive technologies can offer a customized way of identifying particular input fields, for example showing a set of symbols / icons that is familiar to the user, in order to communicate the purpose of the fields visually.</dd>
```

patrickhlauke: on that first edit, i *think* we use singular ("assistive technology") in most cases, rather than plural ("assistive technologies"). personally, i also prefer the latter....

understanding/21/identify-input-purpose.html
```diff
@@ -23,13 +23,16 @@ <h2>Intent of this Success Criterion</h2>
...
+  <p>In addition to repurposing this taxonomy, when the <code>autocomplete</code> attribute technique is used to meet this Success Criterion for HTML content, browsers and other user agents can suggest and 'autofill' the right content by autocompleting these fields based on past user input stored in the browser. By defining more granular definitions of common input purposes, for example “Birthday” (<code>autocomplete="bday"</code>), browsers can store personalized values for each of these fields (the user's birthday date). The user is relieved of having to type the information and can instead confirm or, if needed, change the value of the field, a significant benefit for users with memory issues, dyslexia, and other disabilities. Because the <code>autocomplete</code> values are independent of language, users that may not be familiar with the text used to visually identify user input fields (the label) can still have that purpose consistently identified to them due to the fixed taxonomy of terms.</p>
```
bruce-usab:
```suggestion
  <p>In addition to repurposing this taxonomy, when the <code>autocomplete</code> attribute technique is used to meet this success criterion for HTML content, browsers and other user agents can suggest and 'autofill' the right content by autocompleting these fields based on past user input stored in the browser. By defining more granular definitions of common input purposes, for example “Birthday” (<code>autocomplete="bday"</code>), browsers can store personalized values for each of these fields (the user's birthday date). The user is relieved of having to type the information and can instead confirm or, if needed, change the value of the field, a significant benefit for users with memory issues, dyslexia, and other disabilities. Because the <code>autocomplete</code> values are independent of language, users that may not be familiar with the text used to visually identify user input fields (the label) can still have that purpose consistently identified to them due to the fixed taxonomy of terms.</p>
```

understanding/21/identify-input-purpose.html
baldino-m: Should not this be then "The assistive technology"...?

bruce-usab (approved):
Looks great!

understanding/21/identify-input-purpose.html
patrickhlauke: no this works without the "the"

phrchaychayphanthu33-create:
https://github.com/w3c/wcag/issues/5286{
  "id": 9554194432293,
  "title": "T-Shirt",
  "status": "active",
  "vendor": "My Store",
  "product_type": "Shirts",
  "updated_at": "2025-04-22T14:30:00-05:00",
  "variants": [
    {
      "id": 123456789,
      "title": "Default Title",
      "price": "29.99",
      "sku": "TSHIRT-001",
      "taxable": true,
      "updated_at": "2025-04-22T14:30:00-05:00"
    }
  ],
  "tags": "cotton, comfortable"
}

scottaohara (approved):

PR #5299, opened 2026-08-09
Override `blockquote p:last-of-type` bottom margin

patrickhlauke:
Currently the upstream styles remove bottom margin for the last paragraph in a blockquote. This can leads to a few outlier cases that look broken/wrong, such as the missing margin between the paragraph and the unordered list in https://www.w3.org/WAI/WCAG21/Understanding/conformance#conf-req1

Generally, this also leads to a slightly unbalanced look to notes nested inside blockquotes, which have more inside margin at the top compared to the bottom.

> screenshot of the conformance requirement blockquote, with red arrows highlighting the lack of margins

Compare to these after that specific style is overridden.

> screenshot of the same section, with margin/spacing normalised

Preview: https://deploy-preview-5299--wcag2.netlify.app/understanding/conformance#conf-req1


css/base.css
```diff
@@ -395,6 +395,18 @@ blockquote p:last-of-type::after {
...
+blockquote p:last-of-type {
+	margin-bottom: inherit;
+}
+
+blockquote > p:last-of-type {
+	margin-bottom: 0;
+}
+
+blockquote > p:last-of-type + * {
+    margin-top: 1em;
```
kfranqueiro: Consistent whitespace with the majority of the file (including the rest of this PR)

```suggestion
	margin-top: 1em;
```

css/base.css
```diff
@@ -395,6 +395,18 @@ blockquote p:last-of-type::after {
...
+}
```
adampage, line 408:
```suggestion
blockquote {
	p:last-of-type {
		margin-block-end: inherit;
	}

	> p:last-of-type {
		margin-block-end: 0;

		+ * {
			margin-block-start: 1em;
		}
	}
}
```

How about [logical properties](https://caniuse.com/css-logical-props) and [native CSS nesting](https://caniuse.com/css-nesting), now that both are baseline?

Also took care to use tabs for the white space, per @kfranqueiro’s comment.

adampage (approved):
Very nice improvement.

css/base.css
patrickhlauke: for the logical properties ... this would be the first bit of them in the entire css here, so would look oddly out of place. as for native nesting... yes it makes it more readable, but *shrug*

kfranqueiro (approved):

PR #5303, opened 2026-08-12
Expand "Understanding Levels of Conformance"

patrickhlauke:
Based on the lengthy discussions in https://github.com/w3c/wcag/issues/3889 - which kept stressing the idea that no, Level A/AA/AAA does not try to imply any kind of hierarchy of importance/severity (though this may perhaps be more wishful thinking) - this PR rewrites and expands this section, to hopefully provide some more clarification.

* removes the use of the word "essential" - this currently very clearly *does* imply that some SCs are "more essential" (thus more important/severe) than others
* moves the bullet about workarounds right after the first one, as it's related conceptually
* rewords the "freedom of expression" point (see https://github.com/w3c/wcag/pull/5309)
* adds a bullet about "interpretation" (and later mentions "subjectivity" in the example ... yes, it's a dirty word that WCAG doesn't want to acknowledge most of the time...but it's true)
* gives an example after the bullet about what these bullet points actually mean - this is primarily what was missing in the "WCAG doesn't explain itself about why levels were chosen the way they were"
* adds a call-back to WCAG 1.0 - where "priority" was used - and explains how this concept is NOT used in WCAG 2. further hammers home the idea that all SCs are important, and that deciding how to prioritise or even what level to aim for are decisions for developers/policy makers in the end

Closes https://github.com/w3c/wcag/issues/3889

Preview: https://deploy-preview-5303--wcag2.netlify.app/understanding/conformance#levels
Diff: https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FUnderstanding%2Fconformance%23levels&doc2=https%3A%2F%2Fdeploy-preview-5303--wcag2.netlify.app%2Funderstanding%2Fconformance%23levels#levels

PR #5309, opened 2026-08-13
language tweak - limits on function / look and feel

chaals:
- I don't think "freedom of expression" is relevant here.

As commonly understood I believe it means "whether you can make a particular statement". As far as I know we have never proposed to limit that. Instead, we make it clear that using some ways of presenting a given statement (such as putting tiny very light grey text on a white background) mean that many people will have difficulties in reading/understanding the statement.

- Make a direct concrete statement by removing the "would"(thus matching the other items in the list)
- Success criteria impose limits on content, not on authors

### DIFF
--- a/understanding/conformance.html
+++ b/understanding/conformance.html
@@ -428,9 +428,8 @@
                     <strong>reasonably be achieved by the content creators</strong> (that is, the knowledge and skill to meet the success criterion could be acquired
                     in a week's training or less)
                 </li>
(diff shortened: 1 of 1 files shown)


patrickhlauke:
took some of the ideas here directly to https://github.com/w3c/wcag/pull/5303/changes/c473ab3750e222558559df667518ebcb8ea47a38

chaals:
https://github.com/w3c/wcag/commit/c473ab3750e222558559df667518ebcb8ea47a38 makes this PR redundant, closing. Thanks!

PR #5315, opened 2026-08-16
Ensure img/video/svg inside figure doesn't grow wider than the figure parent container

patrickhlauke:
Currently, due to the additional `1px` border, wide images/videos/svgs inside a figure can end up 2 pixels wider than the parent `<figure>` container. This is not normally a problem, but results in an unsightly "bump" in focus outline when navigating with a keyboard and jumping to a section that contains one of these wide figures.

Example: https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast#examples (make sure to navigate to it using the keyboard, to enable the focus outline)

> example of a wide image - focus outline around the section has a "bump"

> closeup showing the bump and the margin/padding for the figure - the image itself juts out by two pixels to the right

Forcing the box-sizing to take into account the border fixes this.

Preview: https://deploy-preview-5315--wcag2.netlify.app/understanding/non-text-contrast#examples

> the same example of a wide image - focus outline around the section now doesn't have a bump

> closeup showing that the image is now only as wide as the figure parent

### DIFF
--- a/css/base.css
+++ b/css/base.css
@@ -407,6 +407,7 @@ figure img,
 figure video,
 figure svg {
 	border: 1px solid var(--line-grey);
+	box-sizing: border-box;
 }
 
 main table svg[role*="image"],


adampage (approved):
Slam dunk. 🚀

PR #5336, opened 2026-08-27
Rework 2.1.4 Character Key Shortcut understanding

### LINKED ISSUE #953: Examples for SC 2.1.4 too faint - difficult to discern
Refer to examples for SC 2.1.4 under Related Resources at
https://www.w3.org/WAI/WCAG21/Understanding/character-key-shortcuts.html

The examples are very very faint and extremely difficult to discern.
1. Single character key shortcuts affecting speech input – example 1
2. Single character key shortcuts affecting speech input – example 2

Thanks,
Sailesh

patrickhlauke:
* expand the focus to not be primarily about speech users
* expand the speech input explainer to match reality a bit better
* new section about keyboard use with screen readers
* remove the outdated resources section - in particular, the links to the two context-less youtube videos (with low audio and video quality) that demonstrate the issue with Talon on sites that have since been changed
* add working example - key event logger

Stretch goals (to discuss with TF):

* record video demonstrating voice control/speech use on working example
* record video demonstating SR+keyboard use on working example

Closes https://github.com/w3c/wcag/issues/953

- [Preview](https://deploy-preview-5336--wcag2.netlify.app/understanding/character-key-shortcuts)
- [Diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FWAI%2FWCAG22%2FUnderstanding%2Fcharacter-key-shortcuts&doc2=https%3A%2F%2Fdeploy-preview-5336--wcag2.netlify.app%2Funderstanding%2Fcharacter-key-shortcuts)


patrickhlauke:
Link-dumping some resources I looked at while working on this:

* https://talonvoice.com/
* https://talonvoice.com/docs/
* https://github.com/talonhub/community/blob/main/core/keys/letter.talon-list
* https://chromewebstore.google.com/detail/dragon-web-extension/ddaloccgjfibfpkalenodgehlhkgoahe
* https://support.apple.com/en-us/111778
* https://support.google.com/accessibility/android/answer/6151848?hl=en
* https://support.microsoft.com/en-us/accessibility/windows/voice-access/use-voice-access-to-control-your-pc-author-text-with-your-voice
* Keyboard event logger [codepen](https://codepen.io/editor/patrickhlauke/pen/01a00a17-8269-7e91-b0f2-aad0cb1bd835) / [live preview](https://es-d-49515393020260904-01a00a17-8269-7e91-b0f2-aad0cb1bd835.codepen.dev/)

understanding/21/character-key-shortcuts.html
```diff
@@ -13,33 +13,38 @@ <h2>In brief</h2>
...
+			<p>However, there are some voice control/speech recognition programs (such as <a href="https://talonvoice.com/">Talon Voice</a>) that are not "context-aware". All speech is interpreted as either a command or dictation, regardless of where the current focus is. In addition, saying certain keywords is interpreted by some programs as a shorthand for the "Press … key" command, leading to potential unintentional single key presses. Critically, programs that are not "context-aware" may also turn dictation into synthetic key presses, meaning that an unintentionally spoken word can become a sequence of single-key commands if the cursor focus happens to be in the wrong place. Character key shortcuts will be problematic for users of these non-context-aware programs.</p>
```
detlevhfischer: Reference to Talon Voice:  How long will that be valid? Leave out brand name?

understanding/21/character-key-shortcuts.html
```diff
@@ -13,33 +13,38 @@ <h2>In brief</h2>
...
+
+			<h3>Keyboard use with screen readers</h3>
+			<p>The behavior or screen readers when it comes to keyboard interactions, and whether or not key presses are passed on to web content, varies depending on the specific platform and screen reader used.</p>
+			<p>On Windows, screen readers such as Jaws, NVDA, and Narrator generally intercept all key presses, as many of the shortcuts to control screen readers use single keys. Key presses are then only passed on to web content depending on the context. For instance, if focus is on a <code>button</code> element, only <kbd>Space</kbd> and <kbd>Enter</kbd> key presses (and a few generic keys, such as <kbd>Tab</kbd>) will be passed to the web content – any other key presses will be "swallowed" by the screen reader. This behavior is overridden in HTML when elements have a <code>role="application"</code>, or are children of an element with that <code>role</code>. In general, character key shortcuts won't be problematic for Windows screen reader users, as it's unlikely that they'd accidentally trigger a shortcut.</p>
```
detlevhfischer: Good stuff - but the general concern is that all this is fairly specific and might change any time, thereafter being either wrong or needing to be updated. (Will we? How soon?)

I wonder whether acceptance of single keystroks should be linked to the SR mode (browse of forms mode) rather than highlighting role application - not sure. You have done all the research...

understanding/21/character-key-shortcuts.html
patrickhlauke: the issue here is that it seems Talon is pretty much the only tool out there that behaves this way it seems...this is a way to salvage the whole voice angle, in my view, as none of the other voice control/voice access tools commonly used now have the problem this SC purports to solve...

understanding/21/character-key-shortcuts.html
patrickhlauke: we either get specific, or we generalise and give the wrong impression (as the current understanding does ... it makes it sound like the problem described happens in all voice access/control scenarios, but it's actually only when using niche software). so the same problem may happen here, if we just generalise it as "in some cases this is a problem").

in this case, i'd rather be more specific, than pretending that things are a general issue/problem when they're not.

understanding/21/character-key-shortcuts.html
```diff
@@ -13,33 +13,38 @@ <h2>In brief</h2>
...
-      <p>A user interacting with a web page or web app that doesn't use single-character shortcuts doesn't have this problem. Inadvertent strings of characters from the speech application are not interpreted as shortcuts if a modifier key is required. A speech user filling in a text input form may find that a phrase that is accidentally picked up by the speech microphone results in stray text being entered into the field, but that is easily seen and undone. The Resources section of this page contains links to videos demonstrating these types of issues.</p>
+      <p>Speech input users generally work in a single mode where they can use a mix of dictation and speech commands. The user knows to pause before and after commands, and commands are usually at least two words long. So, for instance, a user might say a bit of dictation, such as "the small boat", then pause, and say a command to delete that dictation, such as "Delete Line". In contrast, if the user were to say the two phrases together without a pause, the whole phrase would come out as dictation (i.e., "the small boat delete line").</p>
```
baldino-m:
```suggestion
      <p>Speech input users generally work in a single mode where they can use a mix of dictation and speech commands. The user knows to pause before and after commands, and commands are usually at least two words long. So, for instance, a user might say a bit of dictation, such as "the small boat", then pause, and say a command to delete that dictation, such as "Delete Line". In contrast, if the user were to say the two phrases together without a pause, the whole phrase would come out as dictation (<i>i.e.</i>, "the small boat delete line").</p>
```

understanding/21/character-key-shortcuts.html
```diff
@@ -13,33 +13,38 @@ <h2>In brief</h2>
...
+			<p>In most cases, voice control/speech recognition software is "context-aware" – speech is not interpreted as dictation, unless focus is on a form control or field that accepts keyboard input. Otherwise, speech is only interpreted as a command. In addition, even when focus is inside a form control or field that accepts keyboard input, dictation does not generate "faked" synthetic key presses – the spoken text is generally injected directly into the relevant form control/field. Synthetic key presses only happen when a user gives an explicit command, such as "Press X key".</p>
```
baldino-m:
```suggestion
			<p>In most cases, voice control/speech recognition software is "context-aware" – speech is not interpreted as dictation, unless focus is on a form control or field that accepts keyboard input. Otherwise, speech is only interpreted as a command. In addition, even when focus is inside a form control or field that accepts keyboard input, dictation does not generate unexpected synthetic key presses – the spoken text is generally injected directly into the relevant form control/field. Synthetic key presses only happen when a user gives an explicit command, such as "Press X key".</p>
```

understanding/21/character-key-shortcuts.html
```diff
@@ -13,33 +13,38 @@ <h2>In brief</h2>
...
+			<p>However, there are some voice control/speech recognition programs (such as <a href="https://talonvoice.com/">Talon Voice</a>) that are not "context-aware". All speech is interpreted as either a command or dictation, regardless of where the current focus is. In addition, saying certain keywords is interpreted by some programs as a shorthand for the "Press … key" command, leading to potential unintentional single key presses. Critically, programs that are not "context-aware" may also turn dictation into synthetic key presses, meaning that an unintentionally spoken word can become a sequence of single-key commands if the cursor focus happens to be in the wrong place. Character key shortcuts will be problematic for users of these non-context-aware programs.</p>
+
+			<h3>Keyboard use with screen readers</h3>
+			<p>The behavior or screen readers when it comes to keyboard interactions, and whether or not key presses are passed on to web content, varies depending on the specific platform and screen reader used.</p>
```
baldino-m:
```suggestion
			<p>The behavior or screen readers when it comes to keyboard interactions, and whether or not key presses are passed on to web content, varies depending on the specific platform and screen reader in use.</p>
```

understanding/21/character-key-shortcuts.html
baldino-m: A thorough explanation of browse mode vs. focus/forms mode seems necessary to me

understanding/21/character-key-shortcuts.html
baldino-m: I agree with Detlev, I would not get as far as naming a specific software here

understanding/21/character-key-shortcuts.html
patrickhlauke: we don't italicise `i.e.` anywhere else in WCAG's docs. though we do try to avoid it altogether, for clarity. 
```suggestion
      <p>Speech input users generally work in a single mode where they can use a mix of dictation and speech commands. The user knows to pause before and after commands, and commands are usually at least two words long. So, for instance, a user might say a bit of dictation, such as "the small boat", then pause, and say a command to delete that dictation, such as "Delete Line". In contrast, if the user were to say the two phrases together without a pause, the whole phrase would come out as dictation ("the small boat delete line").</p>
```

understanding/21/character-key-shortcuts.html
patrickhlauke: not just unexpected ... none at all/any. the "faked" is more a folsky reference to the "synthetic" part. could probably remove altogether...

```suggestion
			<p>In most cases, voice control/speech recognition software is "context-aware" – speech is not interpreted as dictation, unless focus is on a form control or field that accepts keyboard input. Otherwise, speech is only interpreted as a command. In addition, even when focus is inside a form control or field that accepts keyboard input, dictation does not generate synthetic key presses – the spoken text is generally injected directly into the relevant form control/field. Synthetic key presses only happen when a user gives an explicit command, such as "Press X key".</p>
```

understanding/21/character-key-shortcuts.html
patrickhlauke: browser mode vs focus mode does not change how it behaves with regards to passing key events to the page though, so not sure it's necessary?

detlevhfischer: As to browse mode, I assumed for a second that things like SR single character shortcuts (e.g. "h" to jump to headings) would not work in forms mode, but it is in fact only when focus is on text `input` elements or `select`s that the shortcut is not passed to the AT - with `select`s even when the character does not match the start of any  `select option`.

baldino-m: well, since the section is about screen reader I believe that this deserves at least a mention. At the same time, browse mode can be turned on manually (not a likely scenario, but it can be done) even when a specific component is currenlty focused

understanding/21/character-key-shortcuts.html
```diff
@@ -13,33 +13,38 @@ <h2>In brief</h2>
...
+			<p>However, there are some voice control/speech recognition programs (such as <a href="https://talonvoice.com/">Talon Voice</a>) that are not "context-aware". All speech is interpreted as either a command or dictation, regardless of where the current focus is. In addition, saying certain keywords is interpreted by some programs as a shorthand for the "Press … key" command, leading to potential unintentional single key presses. Critically, programs that are not "context-aware" may also turn dictation into synthetic key presses, meaning that an unintentionally spoken word can become a sequence of single-key commands if the cursor focus happens to be in the wrong place. Character key shortcuts will be problematic for users of these non-context-aware programs.</p>
+
+			<h3>Keyboard use with screen readers</h3>
+			<p>The behavior or screen readers when it comes to keyboard interactions, and whether or not key presses are passed on to web content, varies depending on the specific platform and screen reader in use.</p>
```
alastc:
```suggestion
			<p>The behavior of screen readers when it comes to keyboard interactions, and whether or not key presses are passed on to web content, varies depending on the specific platform and screen reader in use.</p>
```

understanding/21/character-key-shortcuts.html
```diff
@@ -13,33 +13,38 @@ <h2>In brief</h2>
...
 			<h2>Intent</h2>
-      <p>The intent of this success criterion is to reduce accidental activation of keyboard shortcuts. Character key shortcuts work well for many keyboard users. However, they can be inappropriate and frustrating for speech input users, whose dictation is interpreted as strings of letters, and for keyboard users who are prone to accidentally hit keys. To rectify this issue, authors need to allow users to turn off or reconfigure shortcuts that are made up of only character keys.</p>
+      <p>The intent of this success criterion is to reduce accidental activation of keyboard shortcuts. Character key shortcuts work well for many keyboard users. However, particularly if the shortcuts are mapped to single characters that are triggered without any modifier keys, keyboard users may trigger them accidentally. These shortcuts can also be problematic for certain voice control/speech recognition users, as some of the software used by this user group converts any speech into simulated key presses, which can lead to unintended activations. To rectify this issue, authors need to allow users to turn off or reconfigure shortcuts that are mapped to single character keys.</p>
```
patrickhlauke:
```suggestion
      <p>The intent of this success criterion is to reduce accidental activation of keyboard shortcuts. Character key shortcuts work well for many keyboard users. However, particularly if the shortcuts are mapped to single printable characters, keyboard users may trigger them accidentally. These shortcuts can also be problematic for certain voice control/speech recognition users, as some of the software used by this user group converts any speech into simulated key presses, which can lead to unintended activations. To rectify this issue, authors need to allow users to turn off or reconfigure shortcuts that are mapped to single character keys.</p>
```

understanding/21/character-key-shortcuts.html
alastc: Noting that role="application" is used in frequently-used apps such as gmail, so can seem common.

understanding/21/character-key-shortcuts.html
```diff
@@ -55,16 +60,9 @@ <h3>Alternate Control</h3>
 		</section>
 		<section id="resources">
 			<h2>Resources</h2>
-			<p>Web apps that use character-key shortcuts and allow users to disable and/or change these shortcuts:</p>
 			<ul>
-				<li>Gmail</li>
-				<li>WordPress</li>
+				<li><a href="../../working-examples/key-event-logger/">Key event logger</a> – logs all JavaScript key events that reach the document; useful to verify if a specific combination of platform, user agent, and assistive technology fires key events.</li>
```
kfranqueiro:
```suggestion
				<li><a href="../working-examples/key-event-logger/">Key event logger</a> – logs all JavaScript key events that reach the document; useful to verify if a specific combination of platform, user agent, and assistive technology fires key events.</li>
```

understanding/21/character-key-shortcuts.html
patrickhlauke: > the shortcut is not passed to the AT

just a quick note that it's the other way around: AT sits between the keyboard and the browser. any keypresses are either consumed by the AT, or passed on to the browser (not the other way around)

understanding/21/character-key-shortcuts.html
```diff
@@ -13,33 +13,38 @@ <h2>In brief</h2>
...
+				<p>Even though this success criterion refers to <q>character <strong>keys</strong></q>, note that it's not relevant whether a shortcut can be activated using a single physical <em>key</em> on a keyboard, or if it requires a combination of keys to be pressed. For instance, on most full-size US and UK keyboard, the <kbd>?</kbd> (question mark) symbol is accessed using <kbd><kbd>Shift</kbd>+<kbd>/</kbd></kbd> (forward slash key next to the right-hand <kbd>Shift</kbd> key). On a UK keyboard, in Windows, the <kbd>é</kbd> (lowercase "e" with an acute accent) requires the use of <kbd><kbd>AltGr</kbd>+<kbd>e</kbd></kbd>. The specific key combination required for certain characters will also vary depending on the user's keyboard layout. However, shortcuts that use these characters still fall under the requirements of this success criterion. What matters is that a shortcut relies on a printable character (letters, punctuation, numbers, symbol characters), and not the number of physical keyboard keys that users need to press to trigger it.</p>
```
bruce-usab:
```suggestion
				<p>Even though this success criterion refers to <q>character <strong>keys</strong></q>, note that it's not relevant whether a shortcut can be activated using a single physical <em>key</em> on a keyboard, or if it requires a combination of keys to be pressed. For instance, on most full-size US and UK keyboard, the <kbd>?</kbd> (question mark) symbol is accessed using <kbd>Shift</kbd>+<kbd>/</kbd> (forward slash key next to the right-hand <kbd>Shift</kbd> key). On a UK keyboard, in Windows, the <kbd>é</kbd> (lowercase "e" with an acute accent) requires the use of <kbd>AltGr</kbd>+<kbd>e</kbd>. The specific key combination required for certain characters will also vary depending on the user's keyboard layout. However, shortcuts that use these characters still fall under the requirements of this success criterion. What matters is that a shortcut relies on a printable character (letters, punctuation, numbers, symbol characters), and not the number of physical keyboard keys that users need to press to trigger it.</p>
```

bruce-usab: Are the nested `kbd` purposeful?

patrickhlauke: hmmm, not sure. they may have slipped through originally

understanding/21/character-key-shortcuts.html
patrickhlauke: > well, since the section is about screen reader I believe that this deserves at least a mention. At the same time, browse mode can be turned on manually (not a likely scenario, but it can be done) even when a specific component is currenlty focused

apologies @baldino-m ... I had not grokked this scenario when first looking at your comment. you're right (and testing confirms it) that if a user switches explicitly to focus (forms) mode, then key presses will be passed on the same way as for regular non-SR keyboard users. I've now redone the section a bit, mentioning the modes, and shortening things a bit.

patrickhlauke: @alastc i removed that last sentence that seemed to suggest it's not a common scenario

PR #5337, opened 2026-08-28
Mark up quoted note as actual note in 1.4.5 Images of Text understanding

### LINKED ISSUE #5318: 1.4.5 Move the note embedded in a paragraph to the standard styling
In understanding 1.4.5, a note is embedded in a paragraph, it should be moved out of the paragraph ino the standard note format:

The definition of [images of text](https://www.w3.org/WAI/WCAG22/Understanding/images-of-text.html#dfn-image-of-text) contains the note: This does not include text that is part of a picture that contains significant other visual content. Examples of such pictures include graphs, screenshots, and diagrams which visually convey important information through more than just text.

It should be reformatted as follows:
NOTE
In the definition of [images of text](https://www.w3.org/WAI/WCAG22/Understanding/images-of-text.html#dfn-image-of-text). This does not include text that is part of a picture that contains significant other visual content. Examples of such pictures include graphs, screenshots, and diagrams which visually convey important information through more than just text.

https://www.w3.org/WAI/WCAG22/Understanding/images-of-text.html#intent

patrickhlauke:
Closes https://github.com/w3c/wcag/issues/5318

Preview: https://deploy-preview-5337--wcag2.netlify.app/understanding/images-of-text#intent (first note there)

### DIFF
--- a/understanding/20/images-of-text.html
+++ b/understanding/20/images-of-text.html
@@ -44,10 +44,8 @@
          of text to match their requirements.
       </p>
 
-      <p>The definition of <a>images of text</a> contains the note: This does not include text that is part of a picture that contains significant
-         other visual content. Examples of such pictures include graphs, screenshots, and diagrams which visually
-         convey important information through more than just text.
-      </p>
+      <p>The definition of <a>images of text</a> contains the note:</p>
+      <p class="note">This does not include text that is part of a picture that contains significant other visual content. Examples of such pictures include graphs, screenshots, and diagrams which visually convey important information through more than just text.</p>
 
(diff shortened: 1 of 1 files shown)


baldino-m:
As per my understanding, the "notes" component (so to speak) is relative to the content of the understanding document, while the re-formatted note relates to the definition (https://www.w3.org/WAI/WCAG22/Understanding/images-of-text#dfn-image-of-text), so it makes sense to me keep it as it currenlty is. Please also note how the "Examples of such pictures include graphs, screenshots, and diagrams which visually convey important information through more than just text" text is not part of the note in the "original" definition (https://www.w3.org/WAI/WCAG22/Understanding/images-of-text#dfn-image-of-text).

patrickhlauke:
the main thing for me is...does it matter? it's a note, something to take note of. yes, it came from the definition, but it's pertinent here when talking about the intent of the SC...

patrickhlauke:
Discussed during TF call. Decided to close and potentially do more work to make it clearer why that note is being added, and where.

PR #5345, opened 2026-09-03
Replace "best practice" with alternative wordings (informative)

patrickhlauke:
"'best' practice ... according to whom?"

the term "best" implies some value judgement which is often quite subjective, and much more likely to change over time.

Related to https://github.com/w3c/wcag3/pull/839 / https://github.com/w3c/wcag3/pull/834#discussion_r3875944371

Normative counterpart to this PR: https://github.com/w3c/wcag/pull/5348

PR #5348, opened 2026-09-04
Reword "best practice" in Label in Name note and guidelines introduction text (normative)

patrickhlauke:
Normative counterpart to https://github.com/w3c/wcag/pull/5345

Related to https://github.com/w3c/wcag3/pull/839 / https://github.com/w3c/wcag3/pull/834#discussion_r3875944371

Note: Both changes are non-normative (one is in the introduction, the other is in a note)


mraccess77:
I don't mind the change here - but I would not want to change other places like advisory techniques until we have a broader discussion.

patrickhlauke:
> I don't mind the change here - but I would not want to change other places like advisory techniques until we have a broader discussion.

@mraccess77 so your concern is with https://github.com/w3c/wcag/pull/5345 (the non-normative counterpart to this)

WilcoFiers (changes requested):
The word "recommendation" is used by the W3C to communicate something has gone through its rec track process. The term "recommended practice" is far too similar, and could easily be confused with W3C recommendation. You're putting this on notes and informative documents, these are NOT a W3C recommendation.

There are lots of other terms that could be used that would work:
- good practice
- sound practice
- advisory
- preferred
- etc.


Those would all work fine in the informative content. I don't think this kind of nit picky stuff is worth updating a normative document over.

guidelines/index.html
```diff
@@ -60,7 +60,7 @@ <h3>WCAG 2 Layers of Guidance</h3>
...
+				<p>Note that even content that conforms at the highest level (AAA) will not be accessible to individuals with all types, degrees, or combinations of disability, particularly in the cognitive, language, and learning areas. Authors are encouraged to consider the full range of techniques, including the advisory techniques, <a href="https://www.w3.org/TR/coga-usable/">Making Content Usable for People with Cognitive and Learning Disabilities</a>, as well as to seek relevant advice about current recommended practice to ensure that web content is accessible, as far as possible, to this community. <a href="https://www.w3.org/WAI/WCAG22/Understanding/understanding-metadata">Metadata</a> may assist users in finding content most suitable for their needs. </p>
```
alastc, line 63: We aren't saying that "this is a best practice", it is saying "go find the best practice", so I think this one is fine.

guidelines/sc/21/label-in-name.html
```diff
@@ -1,9 +1,8 @@
...
-   <h4>Label in Name</h4>
-   <p class="conformance-level">A</p>
+  <h4>Label in Name</h4>
+  <p class="conformance-level">A</p>
 	
   <p>For <a>user interface components</a> with <a>labels</a> that include <a>text</a> or <a>images of text</a>, the <a>name</a> contains the text that is presented visually.</p>
 
-  <p class="note">A best practice is to have  the text of the label at the start of the name.</p>
-	
+  <p class="note">A recommended practice is to have the text of the label at the start of the name.</p>
```
patrickhlauke, line 7:
```suggestion
  <p class="note">A suggestion is to have the text of the label at the start of the name.</p>
```

guidelines/index.html
patrickhlauke: i'd still like to eradicate that term (even though this refers to *other people's* 'best' practice, rather than AGWG claiming something is 'best'). so for the same rationale, calling it 'recommended' here is fine as it doesn't say the AGWG recommends it (bringing up the confusion about REC) #beBest

patrickhlauke:
@WilcoFiers I've changed the first instance to "suggestion". for the second case, i would say "recommended" works, as it's not talking about AGWG/WCAG recommending things, but refers to recommendations by third parties

guidelines/sc/21/label-in-name.html
```diff
@@ -1,9 +1,8 @@
...
-   <h4>Label in Name</h4>
-   <p class="conformance-level">A</p>
+  <h4>Label in Name</h4>
+  <p class="conformance-level">A</p>
 	
   <p>For <a>user interface components</a> with <a>labels</a> that include <a>text</a> or <a>images of text</a>, the <a>name</a> contains the text that is presented visually.</p>
 
-  <p class="note">A best practice is to have  the text of the label at the start of the name.</p>
-	
+  <p class="note">A suggestion is to have the text of the label at the start of the name.</p>
```
patrickhlauke:
```suggestion
  <p class="note">One way of achieving this is to have the text of the label at the start of the name.</p>
```

bruce-usab (approved):

WilcoFiers:
@patrickhlauke I'd much prefer avoiding that word all together, in the understanding doc too. But I can live with this if the group believes this is a necessary change.

patrickhlauke:
i mean...we could drop that one "recommended" altogether and leave it implied ("current practice")

guidelines/index.html
patrickhlauke, line 63:
```suggestion
				<p>Note that even content that conforms at the highest level (AAA) will not be accessible to individuals with all types, degrees, or combinations of disability, particularly in the cognitive, language, and learning areas. Authors are encouraged to consider the full range of techniques, including the advisory techniques, <a href="https://www.w3.org/TR/coga-usable/">Making Content Usable for People with Cognitive and Learning Disabilities</a>, as well as to seek relevant advice about current practice to ensure that web content is accessible, as far as possible, to this community. <a href="https://www.w3.org/WAI/WCAG22/Understanding/understanding-metadata">Metadata</a> may assist users in finding content most suitable for their needs. </p>
```

PR #5374, opened 2026-09-23
Update Intent of Accessible Authentication (Enhanced)

### LINKED ISSUE #3264: 3.3.8/3.3.9: How is "authentication process" defined?

bruce-usab:
Closes #3264 

Adds paragraph to end of Intent for 3.3.9 Accessible Authentication (Enhanced):
> This Success Criterion does not address creation of a username or initiating an account.  For many web sites, establishing an initial username is not significantly different than logging in with that username.  However, for many other web sites, the processes are quite different.  Moreover, the means used for creating an account vary more than the means used for re-authentication.  This is because initial authentication may require more due diligence with assuring the identity of a new user.  Another substantive difference between first authentication and re-authentication is that a third party, human or automated, is often required.

There is already a similar paragraph in Understanding 3.3.8 Accessible Authentication (Minimum).


patrickhlauke:
don't think this fully closes the related issue, because that one also touched on other types of "proving your identity" / "proving your humanity", and we need to make clear for both SCs that those scenarios are not covered either

